PR #2, opened 2020-09-30
Project Setup

celinecamacho:


.gitignore
```diff
@@ -0,0 +1,9 @@
+*.iml
+.gradle
+/local.properties
+/.idea/workspace.xml
```
lhmerino, line 4: Please remove "/" in front since this is not needed. .gitignore understands that if its just a name that its in the root repo directory.

.idea/.name
```diff
@@ -0,0 +1 @@
+student20_pop_fe2
```
lhmerino, line 1: Should we be including anything from ".idea" directory? Maybe others want to setup different settings in their IDE (code Styles) or use a different IDE altogether.

app/src/main/java/com/github/dedis/student20_pop/MainActivity.java
```diff
@@ -0,0 +1,14 @@
+package com.github.dedis.student20_pop;
+
+import androidx.appcompat.app.AppCompatActivity;
+
+import android.os.Bundle;
+
+public class MainActivity extends AppCompatActivity {
```
lhmerino, line 7: Please add comment, what is MainActivity?

lhmerino (approved):
Looks good, just left a few comments. Keep it up! :)

lhmerino (approved):
Awesome, thank you!

imaximumGit (approved):
All good

PR #6, opened 2020-10-05
Home UI

### LINKED ISSUE #8: Implement basic Home UI
Basic Home UI, doesn't consider the case where the user was already connected to a LAO.

celinecamacho:
Closes #8 
Basic Home UI, with the welcome message, tabs and empty Connect UI and Launch UI.


mrlasagne (approved):
Everything seems to be OK

app/src/main/java/com/github/dedis/student20_pop/MainActivity.java
```diff
@@ -1,17 +1,58 @@
...
+ * Activity used to display the different UIs
+**/
+public final class MainActivity extends FragmentActivity {
+
+    private static final String TAG = MainActivity.class.getSimpleName();
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         setContentView(R.layout.activity_main);
```
lhmerino, line 23: Where is this variable R defined? Seems to be appear in a lot of places

lhmerino (changes requested):
Great work! My only concern is R since its not very descriptive of what it represents.

app/src/main/java/com/github/dedis/student20_pop/MainActivity.java
AlexisTabin: https://stackoverflow.com/questions/4953077/what-is-the-class-r-in-android

Maybe this thread can help you :)

celinecamacho: R is an auto-generated class containing the resource IDs for all the resources of res/directory

lhmerino: Ah I see, thank you @AlexisTabin and @celinecamacho! Can you please put that thread link in the project Readme so that someone else looking at the code and is not familiar with Android development can determine what this “R” variable is about. Thanks!

lhmerino (approved):
Looks good!

PR #10, opened 2020-10-12
Launch UI

celinecamacho:
Launch UI and basic LAO entity, only needing a name to be created.

### DIFF
--- a/app/src/main/java/com/github/dedis/student20_pop/model/Lao.java
+++ b/app/src/main/java/com/github/dedis/student20_pop/model/Lao.java
@@ -1,0 +1,37 @@
+package com.github.dedis.student20_pop.model;
+
+/**
+ * Class modeling a Local Autonomous Organization (LAO)
+ */
+public class Lao {
+
+    // When launching a LAO the following values are empty (they are not needed for now):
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/github/dedis/student20_pop/model/Lao.java
```diff
@@ -0,0 +1,37 @@
...
+    public Lao(String name) {
+        this.name = name;
+    }
+
+    public String getName() {
+        return name;
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
lhmerino, line 26: We may want a unique identifier (id uint32) to be the "element" that identifies a particular LAO in case two LAOs would like to have the same name. If this is however, already planned for later, please disregard.

lhmerino (approved):
Great job! Just a single comment regarding identifying a unique LAO but I believe that may be for later?

app/src/main/java/com/github/dedis/student20_pop/model/Lao.java
celinecamacho: Yes, I will work on that later!

jbmichel14 (approved):

Issue #17, opened 2020-10-17
ScrollView in the web app


imaximumGit:
The text go outside the bound of the Scrollview (it have to stay in the black rectangle and add a scroll bar to be able to see all the text) in the ConnectConfirm component on the web version

RaulinN:
Cannot reproduce this bug since clicking on the blue camera icon in `connect` makes the navigation bar "selected tab" line indicator stuck between two tabs (`home` and `connect`). Clicking again on the blue camera icon does nothing more

<img width="1402" alt="Screenshot 2020-10-17 at 17 27 48" src="https://user-images.githubusercontent.com/44499915/96346843-3ed5ac80-109e-11eb-8165-6e140d72adfc.png">

iOS version behaves normally

imaximumGit:
Explain more in the top comment, the bug appear only in the web app, the IOS and android work

RaulinN:
Solved both issues! See commit above

Issue #16, opened 2020-10-17
TouchableOpacity on web


imaximumGit:
Utilization on this in the touchableOpacity component in the CameraButton component in the web app return an error: `Uncaught TypeError: _this$ref$props is undefined`.

RaulinN:
Solved! See commit above

PR #18, opened 2020-10-17
Basic web app

RaulinN:
Basic web app project includes : 

- Correction of the "infinite-width" bug on web only
- Correction of #16 
- Correction of #17


imaximumGit (changes requested):
The STRINGS object is import twice in the file _Organization.js_ and _Attendee.js_

RaulinN:
Nice one @imaximumGit! I removed all unwanted imports in the above commit

imaximumGit (approved):

PopApp/Components/Attendee.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react'
-import {StyleSheet, View, Text} from 'react-native'
+import { StyleSheet, View, Text } from 'react-native'
```
lhmerino, line 2: Seems there are import style discrepancies (spaces vs no spaces in imports)? (Discrepancy 1)

PopApp/Components/CameraButton.js
```diff
@@ -1,21 +1,21 @@
 import React from 'react'
-import {StyleSheet, Image} from 'react-native'
-import { TouchableOpacity } from 'react-native-gesture-handler'
+import {StyleSheet, Image, View, TouchableOpacity} from 'react-native'
```
lhmerino, line 2: Discrepancy 1

lhmerino (changes requested):
Looks great! Just a subtle style discrepancy with imports.

RaulinN:
Fixed! Well spoted @lhmerino! We unified all our imports to reflect our [coding guidelines](https://thoughtbot.com/blog/structure-for-styling-in-react-native).

PopApp/App.js
```diff
@@ -1,15 +1,15 @@
-import 'react-native-gesture-handler'
```
imaximumGit, line 1: The `import 'react-native-gesture-handler'` is compulsory to use react-navigation see [here](https://reactnavigation.org/docs/getting-started)

imaximumGit (changes requested):
Just need a little modification to follow the instructions of react navigation

lhmerino (approved):

PopApp/App.js
RaulinN: Indeed my bad! Added it back.

I will remember next time to double check before deleting an import that IntelliJ deems useless! ^^'

imaximumGit (approved):

PR #44, opened 2020-10-26
Implement keys model

jbmichel14:
Implement the generation of public and private key pair with google's Tink library

### DIFF
--- a/app/src/main/java/com/github/dedis/student20_pop/model/Keys.java
+++ b/app/src/main/java/com/github/dedis/student20_pop/model/Keys.java
@@ -1,28 +1,39 @@
 package com.github.dedis.student20_pop.model;
 
+import java.security.GeneralSecurityException;
 import java.util.Objects;
+import com.google.crypto.tink.subtle.Ed25519Sign;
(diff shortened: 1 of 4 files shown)


lhmerino (approved):
Looks good! it's great that you have been able to create identities (private-public keys)

app/src/main/java/com/github/dedis/student20_pop/model/Keys.java
```diff
@@ -1,28 +1,39 @@
...
+    private String publicKey;
+    private String privateKey;
 
     /**
      * Constructor for the Keys, generates a set of public and private keys
      */
     public Keys() {
-        // Generate the public and private keys
-        this.publicKey = "";
-        this.privateKey = "";
```
celinecamacho, line 19: You should modify PersonTest because the class Person uses getPublicKey() and getPrivateKey()

celinecamacho (changes requested):
Great, just need to modify some tests in PersonTest

celinecamacho (approved):

PR #56, opened 2020-11-09
(outdated) Low-level protocol JSON schema

pierluca:
The $id for now is a bit meaningless, but this should provide a basis to move forward.

Protocol specifications : https://docs.google.com/document/d/1fyNWSPzLhM6W9V0VTFf2waMLiJGcscy7wa4bQlLkySM/edit?ts=5fa967b8#

JSON Schema reference : https://json-schema.org

### DIFF
--- a/low-level/catchup.json
+++ b/low-level/catchup.json
@@ -1,0 +1,20 @@
+{
+    "$schema": "http://json-schema.org/draft/2019-09/schema#",
+    "$id": "http://pop2020.dedis.ch/low-level/catchup.json",
+    "properties": {
+        "action": {
+            "description": "Operation to be performed",
+            "type": "string",
+            "enum": ["catchup"]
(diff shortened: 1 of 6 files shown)


pierluca:
The following should be added to most JSONs, for strictness `"additionalProperties": false`

SFulpius:
The schemas looks good overall.
However, I think the schema used by the server to send to clients published messages is missing.

pierluca:
@SFulpius This is the low-level communication across the websocket connection, as such the "message" can be sent from either client or server to the other.

SFulpius:
@pierluca So the action would be "message"? And what should we do with the req_id field, as it is not a request from the client? Should we omit it?

SFulpius:
Closing because we have a new PR for the protocol.

PR #58, opened 2020-11-10
be2: Verify signature and compute ID

### LINKED ISSUE #53: Verify digital signature on messages

SFulpius:
Verify the signature on a message and compute the message ID.
Closes #53 and #59.


RaulinN (changes requested):
Nice! I have a question and a suggestion for you! 

- Question: What is the format you expect as input for `Hex` values (keys (organizer and witnesses), hashes and signatures). I see you are manually adding the `0x` prefix in front of the `Hex` value (`Hash.scala:18`) to conform to the JsonProtocol. In the parser I am writing, I parse `Hex` values by removing the `0x` and then transforming the resulting `String` into an `Int` value. Do you prefer something else?

- Suggestion: add the Scorex dependency in `readme.md`

SFulpius:
@RaulinN On `Hash.scala:18`, the "%02x" string is used to format binary data in hexadecimal, but it does not add 0x before it.
Also, you should not transform an hexadecimal into an Int. Public keys and signature are 256 bits long, thus it won't fit in a 32 bits Int.
In the parser, just removing the 0x should work with the code I wrote.

SFulpius:
@RaulinN I updated the README.

lhmerino (approved):
Looks great!

RaulinN (approved):

PR #80, opened 2020-11-18
Init proto structure + example files per level

### LINKED ISSUE #74: Define Json schema for the communication protocol

Huntsekkern:
Set-up the directory architecture with folders for each level of detail. 
Complete generic, return, error, publish, subscribe, message (mid-level), createLao.
Suggest to rename message (low-level) to broadcast in order to differentiate it from message (mid-level)

closes #74


method/broadcast.json
```diff
@@ -0,0 +1,10 @@
+{
+    "$comment": "Previously called message.json, I'd suggest to rename it to broadcast as 1) it's what it does 2) it disambiguates it from the mid-level field message",
+    "$comment": "If you agree, we would also need to change the enum in generic from message to broadcast",
+    "$schema": "http://json-schema.org/draft/2019-09/schema#",
+    "$id": "http://pop2020.dedis.ch/low-level/broadcast.json",
```
haoqianzhang, line 5: How is this id generated?

method/message/data/createLao.json
```diff
@@ -0,0 +1,110 @@
...
+    "type": "object",
+    "properties" : {
+        "jsonrpc": {
+            "const": "2.0"
+        },
+        "method": {
+            "description": "Operation to be performed",
+            "type": "string",
+            "enum": ["publish"]
+        },
```
haoqianzhang, line 16: Do we need to repeat this common part every time?

haoqianzhang (commented):
Great Job! I would recommend reducing the repeating. You are using a multilevel folder structure, so I feel it does not make sense to include the generic JSON-RPC part everywhere.

generic.json
```diff
@@ -0,0 +1,33 @@
...
+            "type": "integer"
+        },
+        "error": {
+            "description": "For responses, absent on success, present on error",
+            "type": "object"
+        },
+        "id": {
+            "description": "Unique identifier for the request",
+            "type": "integer"
+        }
```
pierluca, line 29: you're combining request and response fields.

An object which has both method and error fields will "comply" with this schema, which is undesirable.

You might want to "cheat" and have a look at this:
https://github.com/fge/sample-json-schemas/tree/master/jsonrpc2.0

README.md
```diff
@@ -1,2 +1,29 @@
...
+## Representation of complex data types in the protocol
+Please note that || denotes a concatenation.
+
+Hex: hexadecimal (base 16) in string format. Example: 0x5932
+Public Key: Hex
+Signature: Hex (Point 1) || Hex (Point 2). Example: 0x59...0x78...
+Take the fields specified that the signature is supposed to be on, hash it and then call Sign API on the hash. Equivalent to Sign(Hash(content))
+Hash: Hex
+Encrypted Vote: ElGamal encryption of a message. Hex (Point 1) || Hex (Point 2)
+Timestamp: uint64 representation of the Unix timestamp (seconds since January 1st, 1970)
```
pierluca, line 29: Bryan made a comment in the document, that it might be best to store base64 encoded data rather than hexadecimal strings. I would tend to agree.

method/message/data/createLao.json
```diff
@@ -0,0 +1,110 @@
...
+                                "object": {
+                                    "description": "object on which to perform the action",
+                                    "type": "string",
+                                    "enum": ["lao"]
+                                }, 
+                                "action": {
+                                    "description": "action to perform on the object",
+                                    "type": "string",
+                                    "enum": ["create"]
+                                },
```
pierluca, line 43: If it's base64 encoded, JSON Schema can't validate it. It'll just have to be a type string.

method/message/message.json
```diff
@@ -0,0 +1,67 @@
...
+                            "$comment": "BFord: Not sure that this actually needs to be transmitted, since it can simply be computed from the data and signature?",
+                            "description": "hash(data||signature)",
+                            "type": "string"
+                        },
+                        "witness_signatures": {
+                            "description": "Signature by witnesses (sender||data)",
+                            "type": "array",
+                            "items": {
+                                "$comment": "slightly unsure about the type here",
+                                "$comment": "initially empty array, will be filled on further message. Not 100% sure whether it should be in the required list (but allowed to be empty) or not",
```
pierluca, line 50: can't have two fields with the same name :) 

type should probably be string, you're right. (possibly base64 encoded signatures?)

pierluca (commented):
On a very general level, I have a few remarks:

1. Look at your JSON Schemas like your code, you want it factored, clean, etc. The KISS and DRY rules apply :-) As such, while you can do per-message JSON Schemas, you don't want to repeat the entire structure every time. You might want to consider using `$ref` to external files and schemas ( https://stackoverflow.com/a/22833129 ), and using the operator `oneOf` (or `anyOf` if relevant) to say what is expected.

2. JSON objects are like dictionaries/maps in most languages: you can't use multiple `$comment` fields, the last one will overwrite the previous ones.

3. There's necessarily going to be a barrier where you do base64 encoding. Base64 encoded data is not JSON anymore (not until you decode it), and as such you have to validate separately the outer "wrapping" of the communication (low/mid-level communication) and inner, base64-encoded communication (high-level communication)

method/broadcast.json
RaulinN: This is the `$id` "template" that @pierluca used in the outdated draft

According to the [JsonSchema doc](https://json-schema.org/understanding-json-schema/basics.html#declaring-a-unique-identifier), the `$id` should be set to 
>a URL at a domain you control

Could we maybe use the github URL? Maybe consider that all schemas are already on the `proto-specs` branch (thus `"$id": "https://github.com/dedis/student20_pop/tree/proto-specs/<path-to-file>"`)

method/message/data/createLao.json
RaulinN: I am not sure to understand how this would be possible. 

- If we want the schema to be utilizable by a JsonSchema validator, I don't understand how to do it different (because referencing a higher-level schema (e.g. `generic.json` from the top-level folder) from a lower-level schema (e.g. `createLao.json`) is to my knowledge not possible if we want to then replace the `$ref` in the imported top-level schema). 
- If we want the schema to be human-readable (practical to read), then this would be possible. In this case, should the `message.json` file resemble something like this?

```json
{
	"type": "object",
	"properties": {
		"data": {
			"type": "object",
			"$ref": ???
		},
		"sender": {
			"type": "string", 
			"pattern": "^0x[a-fA-F0-9]+$",
			"$comment": "[HexString] public key of the sender"
		},
		"signature": {
			"type": "string", 
			"pattern": "^0x[a-fA-F0-9]+$",
			"$comment": "[HexString] organizer's signature on data : Sign(Hash(data))"
		},
		// ... message_id and witness_signatures
	}
}
```

I didn't include `title`, `description`, `additionalProperties` and `required` for clarity, but the idea is there. The `???` is a reference to the data object (defined in the `data` folder).

generic.json
RaulinN: Ok I think I understand now! The idea for the "top-level schema" is to match ANY valid client request (or server answer in a different file) not matter if it is a "createLao message" or a "propagate a message message" for example?

I think we were both thinking of JsonSchema the wrong way (bottom to top instead of top to bottom)

method/message/data/createLao.json
RaulinN: @Huntsekkern defined the folder structure and a few examples. I will create the files themselves. For example, here's the validator for a "message content"

```json
{
	"title": "Match message content of a LAO creation message",
	"description": "Match message content for Witness a message",
	"type": "object",
	"properties": {
		"data": {
			"type": "string",
			"contentEncoding": "base64"
		},
		"sender": {
			"type": "string", 
			"pattern": "^0x[a-fA-F0-9]+$",
			"$comment": "[HexString] public key of the sender"
		},
		"signature": {
			"type": "string", 
			"pattern": "^0x[a-fA-F0-9]+$",
			"$comment": "[HexString] organizer's signature on data : Sign(Hash(data))"
		},
		"message_id": {
			"type": "string", 
			"pattern": "^0x[a-fA-F0-9]+$",
			"$comment": "[HexString] message id : Hash(data||signature)"
		},
		"witness_signatures": {
			"type": "array", 
			"maxItems": 0,
			"$comment": "list of public keys of witnesses : always empty in this case"
		}
	}, 
	"required": ["data", "sender", "signature", "message_id", "witness_signatures"]
}
```

method/message/message.json
Huntsekkern: In this particular case, it's easy to solve, but generally, is the two fields rule an issue even with $comment? 

Because my understanding of $comment in json-schema is that those fields are not considered as belonging to the actual json.
And multi-line comments are also impossible, so that was my alternative to make them more human readable.

method/broadcast.json
haoqianzhang: I guess it is fine to keep using the dedis.ch. We control this domain name.

generic.json
Huntsekkern: I didn't find (looked hard enough...) how to $ref between different files, both your early files and the link you pasted here only use $ref within a single file, which is not really useful to avoid code duplication, so I skipped it.

Now that I'm up to date with $ref between files, I'll update the files.

Also I like the fact of having a generic structure which can catch both request and response. But fully agree that we don't want to catch it a hybrid of both, but that could be fixed with the oneOf/not, etc. rather than making two arborescences.

RaulinN: You don't need two arborescences. I just finished writing the schema for a server answer. Everything fits in a single file (since all types are very simple). Take a look at it:

```json
{
	"$schema": ???,
	"$id": ???,
	"title": "Match a JsonRpc 2.0 server answer",
	"description": "Match a positive or negative server answer",
	"type": "object",
	"properties": {
		"jsonrpc": { "description": "[String] JsonRpc version", "const": "2.0", "$comment": "should always be \"2.0\"" },
		"result": { 
			"description": "In case of positive answer, result of the client query",
			"anyOf": [
				{ "type": "integer", "minimum": 0, "maximum": 0 },
				{ "type": "array", "items": { "type": "string" } }
			],
			"$comment": "Note: this field is absent if the answer is negative"
		},
		"error": {
			"description": "In case of negative answer, error generated by the client query",
			"type": "object",
			"properties": {
				"code": { "type": "integer", "minimum": -5, "maximum": -1 },
				"description": { "type": "string" }
			},
			"required": ["code", "description"],
			"$comment": "Note: this field is absent if the answer is positive"
		},
		"id": { "type": "integer" }
	},
	"allOf": [
		{ "required": ["jsonrpc", "id"] },
		{ 
			"oneOf": [
				{ "required": ["result"] },
				{ "required": ["error"] }
			] 
		}
	],
	"$comment": "Note: required fields: (\"jsonrpc\" \\and (\"result\" \\or \"error\") \\and \"id\")"
}
```

If you prefer, I can still put both schemas in one file, but it might now be necessary. What do you think?

RaulinN: Don't worry updating the `$ref`, I will do it when defining the other empty files

RaulinN (approved):

generic.json
Huntsekkern: Ok, it's not two "arborescence" because the return part is short, but it's still two different structures with no common parent. But your code works, but it's a matter of preference. I like having the directory structure matching the json structure for consistency and ease of reaching the file needed, so that each inner layer (object in json) is in another subfolder.

What I was going for was this at the top-layer: 
```
{
    "$schema": "http://json-schema.org/draft/2019-09/schema#",
    "$id": "http://pop2020.dedis.ch/generic.json",
    "type": "object",
    "properties" : {
        "jsonrpc": {
            "const": "2.0"
        },
        "oneOf": [
            {
                "anyOf": [
                    { "$ref": "method/subscribe.json#/properties" },
                    { "$ref": "method/unsubscribe.json#/properties" },
                    { "$ref": "method/publish.json#/properties" },
                    { "$ref": "method/broadcast.json#/properties" },
                    { "$ref": "method/catchup.json#/properties" }
                ]
            },
            {
                "anyOf": [
                    { "$ref": "returns/return.json#/properties" },
                    { "$ref": "returns/returnCatchup.json#/properties" }
                ]
            }
        ],
        "id": {
            "description": "Unique identifier for the request",
            "type": "integer"
        }

    },
    "required": ["jsonrpc", "oneOf"]
}
```

and this in the return layer:
```
{
    "$comment": "all requests except for broadcast and return expect a return as an ack",
    "$schema": "http://json-schema.org/draft/2019-09/schema#",
    "$id": "http://pop2020.dedis.ch/returns/return.json",
    "type": "object",
    "properties" : {
        "oneOf": [
            {
                "result": {
                    "description": "For responses, present on success, absent on error",
                    "type": "integer"
                }
            },
            {
                "error": {
                    "description": "For responses, absent on success, present on error",
                    "type": "object",
                    "$ref": "errors/error.json#/properties"
                }
            }
        ]

    },
    "required": ["oneOf"]
}
```

maybe @pierluca can chime in and state which is cleaner

Huntsekkern:
I'm not sure how git notif works, but just in case, refactor to avoid code duplication submitted for approval

generic.json
```diff
@@ -35,5 +29,5 @@
         }
 
     },
-    "required": ["jsonrpc"]
+    "required": ["jsonrpc", "oneOf"]
```
RaulinN, line 32: I am not sure you can do that! If I understand correctly, the `oneOf` construct (and any similar ones) will successively REPLACE the `oneOf` by one of the option inside of the construct. Thus the `"required": ["jsonrpc", "oneOf"]` will never be fulfilled

generic.json
```diff
@@ -8,25 +8,19 @@
         },
         "oneOf": [
```
RaulinN, line 9: Actually I think the problem comes from here. I didn't find a way to include a `oneOf` construct *directly*  inside the `properties` field. This result in an invalid schema

Try this simple schema on [an online validator](https://jsonschemalint.com/#!/version/draft-07/markup/json)
```json
{
  "type": "object",
  "properties": {
    "oneOf": [
      {
        "jsonrpc": {
          "type": "string"
        }
      },
      {
        "m": {
          "type": "integer"
        }
      }
    ]
  }
}
```

generic.json
RaulinN, line 32: I am not sure you can do that! If I understand correctly, the `oneOf` construct (and any similar ones) will successively REPLACE the `oneOf` by one of the option inside of the construct. Thus the `"required": ["jsonrpc", "oneOf"]` will never be fulfilled

generic.json
haoqianzhang: I think a directory structure is better, since our protocol is still subject to change and it is pretty complex already.

method/message/data/createLao.json
haoqianzhang: As Bryan said, Baes64 is the standard method of store the crypto related things. Could you update it?

generic.json
Huntsekkern: I don't know, the code below passes the validator and the only difference is adding "definitions". I think it has to do with type: object and oneOf not being interpreted as an object (when properties is the object and its type is free, seems to work better).

Otherwise, what you did with only using the xxxOf on the required fields might be the alternative?

```
{
  "type": "object",
      "definitions": {
           "properties": {
              "oneOf": [
             {
              "jsonrpc": {
               "type": "string"
             }
         },
       {
        "m": {
          "type": "integer"
        }
      }
    ]
  }
}
```

method/message/data/createLao.json
RaulinN: Yes sure! I already did it. But I was waiting for the pull request to be accepted. Should I also push changes on this branch (work-proto-raoul) @haoqianzhang ?

haoqianzhang: You can merge it. Or push the changes to this branch. Both are fine I think.

haoqianzhang (approved):
Looks fine to me.

generic.json
RaulinN: Well, it *is* a valid JsonSchema (if you add a missing `}` at the end), but it doesn't check what we want. Your schema matchs an `object` (line 2) but doesn't know what to do with the `definitions`. Thus, this schema will match ANY JsonObject (since by default `"required": []` and `additionalProperties: true`). To convince you, try adding the `additionalProperties: false` field at the end like this:

```json
{
	"type": "object",
	"definitions": {
		"properties": {
			"oneOf": [
				{ "jsonrpc": { "type": "string" } },
				{ "m": { "type": "integer" } }
			]
		}
	},
	"additionalProperties": false
}
```

Now, this code accepts ONLY the empty JsonObject (since again, it doesnt know what to do with definitions and is now restricted to the properties specified (namely, none))

RaulinN:
Files have been filled!

*Note*: I assumed we didn't change from `HexString` to `Base64` for the fields **inside** the `data` field (already encoded in base64) as it would require a lot of processing (encoding and decoding) at server side. Thus (for now) the only fields encoded in base 64 are the `Hex` types inside a `message` object. What do you think?

method/message/message.json
pierluca: JSON is bad for humans :-P

This being said, I understand your concern. It's not too much of an issue with `$comment` fields, but be careful about this with other fields as the final value will only be that of the last defined field.

generic.json
pierluca: `oneOf` is already a requirement. I would suggest you have a look at this:
https://json-schema.org/understanding-json-schema/reference/combining.html

generic.json
pierluca: As I read this code, what I understand (conceptually) is that you're willing to accept either multiple methods or multiple return. 

I can give you a helping hand this afternoon, but please clarify for me what you're trying to achieve and what's blocking.  Let's start with an English description, and we'll transform that to JSON Schema in a second step :) 

In the mean time, have a look at this:
https://json-schema.org/understanding-json-schema/reference/combining.html

If need be (not sure you need to know this), there are also conditionals:
https://json-schema.org/understanding-json-schema/reference/conditionals.html

Huntsekkern: Thanks! In English, the generic file should accept either a request or a return. When it accepts a request it should accept either of the 5 methods. When it accepts a return, it should accept either the basic return or the catchup return (which has more content than the basic return).

Lower in the structure, but this is not handled in the generic file, a basic return should accept either a successful return or an error return, and some methods like publish should accept either of the many options for the data fields (createLAO, etc.)

Either can be represented by either (^^) oneOf or anyOf. When differentiating between request and return, the fields are so different than using oneOf is safe. When differentiating between the various data of publish, as some might have similar fields, anyOf is the better choice.

answer/negativeAnswer/error.json
```diff
@@ -0,0 +1,28 @@
...
+				"-3 resource already exists (e.g. lao already exists, channel already exists, etc.)",
+				"-4 request data is invalid (e.g. message is invalid), use the data field object to provide extra information",
+				"-5 access denied (e.g. subscribing to a “restricted” channel)"
+			]
+		},
+		"description": { "description": "[String] error description", "type": "string" }
+	},
+	"additionalProperties": false,
+	"required": ["code", "description"],
+	"$comment": "Note: this field is absent if the answer is positive"
```
haoqianzhang, line 27: what does "this" refer to?

answer/positiveAnswerCatchup.json
```diff
@@ -0,0 +1,10 @@
+{
+	"$schema": "http://json-schema.org/draft/2019-09/schema#",
+	"$id": "http://pop2020.dedis.ch/answer/positiveAnswerCatchup.json",
+	"title": "All requests (except catchup (list of message) and propagate message (nothing)) expect an ACK (0) as answer",
+	"description": "[Array[String]] all high-level message objects so far, sorted in time (ascending)",
+
+	"type": "array", 
+	"items": { "type": "string", "$comment": "TODO: Message objects ?" },
```
haoqianzhang, line 8: Each item in the array is a message object.

answer/positiveAnswerGeneral.json
```diff
@@ -0,0 +1,10 @@
+{
+	"$schema": "http://json-schema.org/draft/2019-09/schema#",
+	"$id": "http://pop2020.dedis.ch/answer/positiveAnswerGeneral.json",
+	"title": "All requests (except catchup (list of message) and propagate message (nothing)) expect an ACK (0) as answer",
+	"description": "[Int] success result message",
+
+	"type": "integer",
+	"minimum": 0, "maximum": 0,
```
haoqianzhang, line 8: Using constant value?
https://json-schema.org/understanding-json-schema/reference/generic.html#constant-values

method/channel/subChannel.json
```diff
@@ -0,0 +1,6 @@
+{
+	"description": "Match a sub-channel (not root channel)",
+	"type": "string",
+	"pattern": "^/root/[^/]+$",
```
haoqianzhang, line 4: Great to have this!

method/message/utility/base64String.json
```diff
@@ -0,0 +1,6 @@
+{
+	"description": "Match string in Base64",
+	"type": "string",
+	"contentEncoding": "base64",
+	"$comment": "Note: JsonSchema doesn't check that the string is some valid Base64 string"
```
haoqianzhang, line 5: It seems possible?
https://stackoverflow.com/questions/475074/regex-to-parse-or-validate-base64-data

haoqianzhang:
> Files have been filled!
> 
> _Note_: I assumed we didn't change from `HexString` to `Base64` for the fields **inside** the `data` field (already encoded in base64) as it would require a lot of processing (encoding and decoding) at server side. Thus (for now) the only fields encoded in base 64 are the `Hex` types inside a `message` object. What do you think?

What do you mean by a lot of processing? @RaulinN

method/message/utility/base64String.json
haoqianzhang: It looks still hard. Worse case, we can check it after the JSON schema. We have to verify signatures and hashs anyway.

generic.json
pierluca: I'll come back on this tomorrow morning first thing, did not have time this afternoon :-( Sorry.

method/message/utility/base64String.json
pierluca: Agreed, you can use an ugly regex ( e.g. `^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$` ) and parse the content once you decode it :)

pierluca: Forget my last message... and read the docs!
https://json-schema.org/understanding-json-schema/reference/non_json_data.html

method/message/messageWitnessMessage.json
```diff
@@ -0,0 +1,38 @@
...
+		"signature": {
+			"description": "[String] witness' signature on data : Sign(Hash(data))",
+			"$ref": "utility/base64String.json",
+			"$comment": "Note: the string is encoded in Base64"
+		},
+		"message_id": {
+			"description": "[String] message id : Hash(data||signature)",
+			"$ref": "utility/base64String.json",
+			"$comment": "Note: the string is encoded in Base64"
+		},
```
pierluca, line 28: No need for an external ref here:
https://json-schema.org/understanding-json-schema/reference/non_json_data.html

Other than that, LGTM 👍

method/message/messageGeneral.json
```diff
@@ -0,0 +1,41 @@
...
+		"signature": {
+			"description": "[String] organizer's signature on data : Sign(Hash(data))",
+			"$ref": "utility/base64String.json",
+			"$comment": "Note: the string is encoded in Base64"
+		},
+		"message_id": {
+			"description": "[String] message id : Hash(data||signature)",
+			"$ref": "utility/base64String.json",
+			"$comment": "Note: the string is encoded in Base64"
+		},
```
pierluca, line 28: You can avoid the external ref just by using:
https://json-schema.org/understanding-json-schema/reference/non_json_data.html

generic.json
```diff
@@ -0,0 +1,65 @@
+{
```
pierluca, line 1: Functionality wise, this seems pretty good ! Good job! 

It's also probably the most not-repeating you can make it. The question then becomes: is this very readable? :)

You might want to sacrifice the repetition of jsonrpc: 2.0 across a few files in favour of being able to write:

```
 	"oneOf": [
 						{ "$ref": "method/subscribe.json" },
 						{ "$ref": "method/unsubscribe.json" },
 						{ "$ref": "method/catchup.json" },
 						{ "$ref": "method/broadcast.json" },
 						{ "$ref": "method/publish.json" },
 						{ "$ref": "answer/generic.json" }
 			]
```

This implies moving the answer to a separate file, sure, but this way, you'll have a good balance of simplicity, expressiveness and non-repetition.

Side note: you might want to validate across both back-end and both front-end teams the support for relative paths in `$ref` fields in your JSON-Schema libraries. It should work seamlessly, it's more of a precaution: it would be a pity to have to go all over them again :-)

answer/negativeAnswer/error.json
```diff
@@ -0,0 +1,28 @@
+{
+	"$schema": "http://json-schema.org/draft/2019-09/schema#",
+	"$id": "http://pop2020.dedis.ch/answer/negativeAnswer/error.json",
```
pierluca, line 3: There was a thread on $id. Just came across this bit in the documentation which I had missed:
https://json-schema.org/understanding-json-schema/structuring.html#the-id-property
it refers to differences in library implementations and how id is used as a base path for relative references.

As such, @RaulinN had a point in suggesting:

> Could we maybe use the github URL? Maybe consider that all schemas are already on the proto-specs branch (thus "$id": "https://github.com/dedis/student20_pop/tree/proto-specs/<path-to-file>")

This would make most sense ! 
Sorry for coming back to you late on that one :)

method/publish.json
```diff
@@ -0,0 +1,56 @@
...
+					"properties": {
+						"channel": { 
+							"description": "[String] name of the channel", 
+							"$ref": "channel/subChannel.json"
+						},
+						"message": { 
+							"description": "[Message] message to be published",
+							"oneOf": [
+								{ "$ref": "message/messageGeneral.json" },
+								{ "$ref": "message/messageWitnessMessage.json" }
```
pierluca, line 43: I may be missing something, but isn't the witness message a sort of sub-type of messageGeneral? 
If so, there's no need for the one-of.

method/publish.json
```diff
@@ -0,0 +1,56 @@
...
+					"required": ["channel", "message"]
+				}, 
+				{
+					"description": "Match a publish query destined for a sub-channel",
+					"type": "object",
+					"properties": {
+						"channel": { 
+							"description": "[String] name of the channel", 
+							"$ref": "channel/subChannel.json"
+						},
```
pierluca, line 38: The only difference here is in the channel.
You might want to match messageGeneral and just have a one-of for the channel specification.
This can save you a lot of repetition.

...then the question is: do you really want to verify the channel, or just match any channel? (at this level)

Keep in mind, JSON-Schema is a specification, but you can (and absolutely should) do application-level verification, beyond what JSON-Schema could provide anyways.

pierluca (commented):
This is shaping up nicely !

I haven't had the time to review every single message/data/*.json but overalls this is looking better and better!

In a general sense, be careful with the non-human-readable data: it is sometimes marked as base64 encoding, sometimes as "hex string". You should settle on one and use it across the board. (Though be careful because base64 uses the full alphabet with `+` and `/` too)

answer/negativeAnswer/error.json
RaulinN: it refers to the whole "error field" which is a field in a generic message (the entire file). I will change the comment to make it more clear!

answer/positiveAnswerGeneral.json
RaulinN: Well spotted! I thought `const` only applied to strings!

method/message/utility/base64String.json
RaulinN: @pierluca I read this page, this is what I did (set `contentEncoding` to `base64`). But I found that all strings were accepted (even non-base64 strings) through testing.

I also tried specifying `binaryEncoding` encoding instead of `contentEncoding` (keyword used by JSON Hyper-Schema Draft #4) but the result is the same

method/message/data/createLao.json
```diff
@@ -0,0 +1,56 @@
+{
+    "$comment": "By sending the lao/create message to the organizer’s server’s default channel (“/root”), the main channel of the LAO will be created with the identifier id. At that point, any of the clients, including the organizer, can subscribe to the channel id and send messages over it.",
+    "$comment": "The server is expected to verify the data, such that the last modified timestamp is equal to the creation timestamp, that the timestamp is reasonably recent with respect to the server’s clock, that the attestation is valid, etc.",
+    "$comment": "Upon successful creation, the server/organizer is expected to broadcast the LAO state to all witnesses and clients (see “LAO state broadcast”).",
+    "$schema": "http://json-schema.org/draft/2019-09/schema#",
+    "$id": "http://pop2020.dedis.ch/createLao.json",
+    "type": "object",
+    "definition" : {
+        "data": {
```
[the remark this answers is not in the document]
RaulinN: Yes! This was done a few days ago! I see you reviewed the file `method/message/data/createLao.json`, but this is one of the artifacts of the project setup I have removed. I don't see it anywhere in git, I don't know what happened 🤔.

The new file corresponding to this one is `method/message/data/dataCreateLao.json` and indeed has `properties` as a top-level field.

method/message/message.json
```diff
@@ -0,0 +1,43 @@
+{
+    "$comment": "By executing a publish action, an attendee communicates its intention to publish a specific message on a channel.",
+    "$schema": "http://json-schema.org/draft/2019-09/schema#",
+    "$id": "http://pop2020.dedis.ch/message.json",
+    "type": "object",
+    "definition" : {
```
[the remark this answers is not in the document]
RaulinN: The keyword `definition` is used to define "local `$ref`". In a JsonSchema file, we can use `definition` as a top-level field and define other schemas in it. Those schemas can then be reused using

```json
{ "$ref": "#/definitions/address" }
```

`#` is a copy if the `id` mentioned at the beginning of each file

Here's a link to the mention of `definition` in the doc : https://json-schema.org/understanding-json-schema/structuring.html#id1

method/message/messageGeneral.json
RaulinN: Indeed, it seems overkill! I will change it now!

RaulinN:
> > Files have been filled!
> > _Note_: I assumed we didn't change from `HexString` to `Base64` for the fields **inside** the `data` field (already encoded in base64) as it would require a lot of processing (encoding and decoding) at server side. Thus (for now) the only fields encoded in base 64 are the `Hex` types inside a `message` object. What do you think?
> 
> What do you mean by a lot of processing? @RaulinN

Now that I think about it, there's not much of a difference between parsing a base64 string and parsing a base16 string. But then the question is, should *all* strings be encoded in base64 or only sensitive data?

For example, should the location in an event, or the error message description in an server answer be encoded as well?

generic.json
RaulinN: That's a good point! I think I have a way of doing it so that the repetition of jsonrpc: 2.0 will only be in two distinct files while still keeping the directory based-structure we were talking about a few days ago (@haoqianzhang).

I will keep you updated!

haoqianzhang:
> > > Files have been filled!
> > > _Note_: I assumed we didn't change from `HexString` to `Base64` for the fields **inside** the `data` field (already encoded in base64) as it would require a lot of processing (encoding and decoding) at server side. Thus (for now) the only fields encoded in base 64 are the `Hex` types inside a `message` object. What do you think?
> > 
> > 
> > What do you mean by a lot of processing? @RaulinN
> 
> Now that I think about it, there's not much of a difference between parsing a base64 string and parsing a base16 string. But then the question is, should _all_ strings be encoded in base64 or only sensitive data?
> 
> For example, should the location in an event, or the error message description in an server answer be encoded as well?

Only cryptographic data is encoded in Base64. I think the reason is that they are not human-readable anyway. For the general data, we do can encode it, but humans will not able to understand it directly.

answer/negativeAnswer/error.json
RaulinN: done!

method/publish.json
RaulinN: In this piece of code, I think we have to separate both parts because there are *two* differences:

- channel name (`/root` or `/root/<channel>`);
- message type (a "witness a message" message cannot be published on the main channel)

I think there's a way to do it using if-then-else

method/publish.json
RaulinN: Yes! Now I remember why I did it! This is tied to your next comment (just below). If we post on the root channel, then the message must be a `messageGeneral` (and thus not a `messageWitnessMessage`). The difference between them it that the witness signature list is always empty for a `messageWitnessMessage`. 

Since messages (`messageGeneral` & `messageWitnessMessage`) and methods (`publish`, `catchup`, ...) are at different levels in the tree, there's no way for me to check that a message sent on the root channel is indeed *not* a `messageWitnessMessage`

RaulinN (approved):
Last thing to do : swap `HexString`s to base64 `String`s. Will discuss it during meeting since both approaches are possible (cf last. @haoqianzhang's message).

method/message/message.json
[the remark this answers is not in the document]
pierluca: Exactly, I've seen `definitions` being used (with an `s`), as in the doc, and not `definition` (without `s`), but maybe that field name is arbitrary and doesn't carry special meaning. I'm unclear in that regard, I don't have time to look much into the docs today.

pierluca (approved):
I only gave it a quick look, but this looks much better overalls.
Once a decision is made on base64 vs hex strings, I believe we have our v0.1 :)

RaulinN:
Alright! All strings have been updated following the rule: 

"if a string is designed to be human-readable => string stays as plain text; otherwise, encode it as base64".

RaulinN:
I will thus merge the pull request and open a thread (issue) to discuss potential changes/improvements

PR #85, opened 2020-11-24
Work be1 group

wouewouel:
### This version is from last week 

We can handle the message createLao, createMeeting, subscribe and Unsubscribe
Currently, the messages should be formated without implementing the base64 encoding for the 'data' object of the messages but just plug them as json string.
Next pull request, we will be able to handle most messages and a big reformating of the database and of the publish subscibe model.


haoqianzhang:
If all the code is reviewed and approved previously in your branch, free feel to merge it.

wouewouel:
New pull request up to date

Issue #91, opened 2020-11-27
Error missing id


SFulpius:
The answer message [requires an id](https://github.com/dedis/student20_pop/blob/proto-specs/answer/answer.json). However, in a specific case, the backend may not be able to always determine the id. By example, if the id field is missing, or if the parser is not able to parse the low-level json.
In this case, I propose to use a special value for the id field, by example -1.

What do you think about this proposition?

@dedis/student20_pop @haoqianzhang  @lhmerino @pierluca

pierluca:
If I rephrase, your concern is the handling of malformed JSON-RPC messages (no id, not parsable, etc.) on the server side.

I can see two ways:

1.  Ignoring them. The backend logs them (for debugging & operational purposes), and that's it. This is easiest to implement, but it makes it a little harder to notice a problem client-side.

2. Informing the client (your suggestion) in some way. As you point out, there can't be a valid id field to refer to (because it's either missing or the JSON itself is not parseable). So what you want is for the server to send a message saying "this bunch of bytes you sent me makes no sense, I'm going to ignore them" (the equivalent of a 400 Bad Request), so that the client can take notice and log it. Is my understanding correct?
Now it's my time to ask you back a couple of things: what's the purpose of the ID field in JSON-RPC ? Does it make sense in this context? If not, how could you do without it? You may find out that you are already using a JSON-RPC 2.0 feature that solves this problem somewhere in the protocol :) 
Also, keep in mind that there might be multiple RPC requests in parallel, so the server should probably echo back the invalid message (maybe truncating it if it's too long). I wouldn't necessarily do it in a production system, but this can help you a lot when debugging.

SFulpius:
I'm not sure if it is already used in the protocol, but from the [JSON-RPC specs](https://www.jsonrpc.org/specification) we have:

> If there was an error in detecting the id in the Request object (e.g. Parse error/Invalid Request), it MUST be Null.

So I think we can modify the protocol to integrate it.

pierluca:
Correct, that is indeed a solution if the id could not be detected !
I was hinting at JSON-RPC notification - I had not re-read the JSON-RPC specs - but your solution is more appropriate 👍

SFulpius:
I'll make a pull request to update the protocol then.

PR #96, opened 2020-11-30
Add the case where we can't dectect an id

SFulpius:
I made a minor change to the protocol, as discussed in #91.


pierluca (approved):

answer/answer.json
```diff
@@ -19,7 +19,10 @@
 			"description": "In case of negative answer, error generated by the client query",
 			"$ref": "negativeAnswer/error.json"
 		},
-		"id": { "type": "integer" }
+		"id": { 
+			"type": "integer",
+			"$comment": "The id match the request id. If there was an error in detecting the id, it must be Null"
+		      }
```
RaulinN, line 25: The idea proposed in #91 is very good! It will probably simplify some steps in error handling and doesn't require a pre-agreed dummy value (e.g. `-1` for missing id). The only problem with what you wrote, it that if someone tries to use the JsonSchema, a NULL value won't be accepted. I would suggest modiying your Json-Schema to look like this:

```json
"id": {
    "oneOf": [
        { "type" : "integer" },
        { "type": null  }
    ],
    "$comment": "... // your comment"
}
```
See this link for more information on [Json-Schema null](https://json-schema.org/understanding-json-schema/reference/null.html)

RaulinN (changes requested):

answer/answer.json
pierluca: Well spotted. I completely missed this 😅 Thanks @RaulinN

RaulinN (approved):

Issue #97, opened 2020-11-30
Proposition of changes to the protocol


SFulpius:
By implementing the protocol, I found some minor mistakes. Also I have some propositions of changes.

1. The [README](https://github.com/dedis/student20_pop/blob/proto-specs/README.md) still mentions the Hex format, although now we are using base64 everywhere.
2. The [dataCreateLao.json](https://github.com/dedis/student20_pop/blob/proto-specs/query/method/message/data/dataCreateLao.json)  and [dataUpdateLao.json](https://github.com/dedis/student20_pop/blob/proto-specs/query/method/message/data/dataUpdateLao.json) files still mentions that the server should broadcast the change, although it is the organizer. The [dataStateLao.json](https://github.com/dedis/student20_pop/blob/proto-specs/query/method/message/data/dataStateLao.json) is correct on this point.
3. Same as the previous point, but in the [dataCreateMeeting.json](https://github.com/dedis/student20_pop/blob/proto-specs/query/method/message/data/dataCreateMeeting.json) and [dataStateMeeting.json](https://github.com/dedis/student20_pop/blob/proto-specs/query/method/message/data/dataStateMeeting.json) files.
4. In [dataCreateLao.json](https://github.com/dedis/student20_pop/blob/proto-specs/query/method/message/data/dataCreateLao.json), the field `last_modified` seems unnecessary to me, as it is always the same as the `creation` field.
5. [dataUpdateLao.json](https://github.com/dedis/student20_pop/blob/proto-specs/query/method/message/data/dataUpdateLao.json) does not have the LAO id field, thus we don't know which LAO the organizer want to update. I propose to include the LAO id in the update request.
6.  [dataStateMeeting.json](https://github.com/dedis/student20_pop/blob/proto-specs/query/method/message/data/dataStateMeeting.json) does not mention which message created or modified the meeting. I propose to add the `modification_id` and `modification_signatures` fields, which will work as in [dataStateLao.json](https://github.com/dedis/student20_pop/blob/proto-specs/query/method/message/data/dataStateLao.json).

Feel free to comment one or more of these points. Once everyone agree, I'll make a PR to update the protocol.
@dedis/student20_pop @pierluca @haoqianzhang @lhmerino

haoqianzhang:
Looks fine to me.

SFulpius:
[...]

PR #99, opened 2020-12-02
Fixes to the protocol

SFulpius:
Solves issues discussed in #97.
Also modified hashes/signatures so they correspond to what was discussed [here](https://docs.google.com/document/d/125nDMc2rc4L6pQgC92rC4X42-BOG8YCFcyzHeY-zY9A/edit) (see picture).

### DIFF
--- a/query/method/message/messageGeneral.json
+++ b/query/method/message/messageGeneral.json
@@ -19,7 +19,7 @@
 			"$comment": "Note: the string is encoded in Base64"
 		},
 		"signature": {
-			"description": "[Base64String] organizer's signature on data : Sign(Hash(data))",
+			"description": "[Base64String] organizer's signature on data : Sign(data)",
 			"type": "string",
 			"contentEncoding": "base64",
 			"$comment": "Note: the string is encoded in Base64"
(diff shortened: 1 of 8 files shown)


RaulinN:
Hey @SFulpius! Thanks for doing these changes! :) 

I took the liberty to add a commit to your PR to correct "JsonSchema syntax". The two main errors I corrected among all files were:

1. you need to specify which properties are required in the `require` field. Otherwise, a JsonObject without the specific property is optional (I added that for the signature objects and removed the last_modified which don't exist anymore)
2. In an object definition, always add the "additionalProperties: false", otherwise, a JsonObject with $n$ additional unrelated properties would also pass the test

RaulinN (approved):

SFulpius:
Thank you for the chanes @RaulinN.
I've also updated the signatures in the messageGeneral.json file.

query/method/message/messageGeneral.json
```diff
@@ -31,16 +31,29 @@
...
+					},
+					"signature": {
+						"description": "[Base64String] witness' signature : Sign(message_id)",
+						"type": "string",
+						"contentEncoding": "base64",
+						"$comment": "Note: the strings are encoded in Base64"
+					}
+				},
+				"additionalProperties": false,
+				"required": ["witness", "signature"]
```
wouewouel, line 53: Looks really fine to me, it resolves the problem of linking

wouewouel (approved):
The proposed solution looks fine to me as we no longer have the problem of remembering which witness signed the data.

PR #104, opened 2020-12-07
Fix set LAO's title

AlexisTabin:
- [x] Creating a LAO with empty name throws an exception
- [x] Trying to Launch organisation with empty name shows toast
- [x] Calling setName on a LAO with empty name now throws an exception
- [x] When editing properties, trying to set empty string to LAO name shows a toast
- [x] There was a bug in AttendeeFragment, which is solved now
- [x] Test setName 
- [x] Test Toasts
- [x] Test LAO creation

### DIFF
--- a/app/src/main/java/com/github/dedis/student20_pop/model/Lao.java
+++ b/app/src/main/java/com/github/dedis/student20_pop/model/Lao.java
@@ -133,6 +133,8 @@ public final class Lao {
     public Lao setName(String name) {
         if (name == null) {
             throw new IllegalArgumentException("Trying to set null as the name of the LAO");
+        } else if (name.isEmpty()){
+            throw new IllegalArgumentException("Trying to set an empty name for the LAO");
         }
         return new Lao(name, time, organizer, witnesses, members, events);
     }
--- a/app/src/main/java/com/github/dedis/student20_pop/model/Lao.java
+++ b/app/src/main/java/com/github/dedis/student20_pop/model/Lao.java
@@ -33,10 +33,10 @@ public final class Lao {
     public Lao(String name, Date time, String organizer) {
         if (name == null || time == null || organizer == null) {
             throw new IllegalArgumentException("Trying to  create a LAO with a null value");
-        } else if (name.isEmpty()) {
+        } else if (name.trim().isEmpty()) {
             throw new IllegalArgumentException("Trying to set an empty name for the LAO");
         }
-        this.name = name;
+        this.name = name.trim();
         this.time = time.getTime() / 1000L;
         this.id = Hash.hash(name + time);
         this.organizer = organizer;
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/github/dedis/student20_pop/model/Lao.java
```diff
@@ -133,6 +133,8 @@ public String getAttestation() {
     public Lao setName(String name) {
         if (name == null) {
             throw new IllegalArgumentException("Trying to set null as the name of the LAO");
+        } else if (name.isEmpty()) {
```
jbmichel14, line 138: Maybe we should use isBlank(), to avoid having a name with only whitespaces

app/src/main/java/com/github/dedis/student20_pop/ui/OrganizerFragment.java
```diff
@@ -162,11 +163,16 @@ public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup c
 
         confirmButton.setOnClickListener(
                 clicked -> {
-                    viewSwitcher.showNext();
-                    lao = lao.setName(laoNameEditText.getText().toString());
-                    laoNameTextView.setText(laoNameEditText.getText());
-                    editPropertiesButton.setVisibility(View.VISIBLE);
-                    addWitnessButton.setVisibility(View.GONE);
+                        String title = laoNameEditText.getText().toString();
+                        if( title != null && !title.isEmpty()){
```
jbmichel14, line 167: same here, isBlank() more appropriate?

app/src/main/java/com/github/dedis/student20_pop/model/Lao.java
AlexisTabin: I can't find any isBlank() declaration, but what we could do is call trim() on the name to remove leading and trailling whitespace and then checking for emptiness

jbmichel14: perfect

AlexisTabin: done

app/src/test/java/com/github/dedis/student20_pop/model/LaoTest.java
```diff
@@ -33,6 +33,11 @@ public void createLaoNullParametersTest() {
         assertThrows(IllegalArgumentException.class, () -> new Lao(lao1_name, time, null));
     }
 
+    @Test
+    public void createLaoEmptyNameTest() {
+        assertThrows(IllegalArgumentException.class, () -> new Lao("", time, organizer));
```
celinecamacho, line 38: You can also test a string with only spaces

celinecamacho (approved):
It's good, I would suggest just one more test though

jbmichel14 (approved):

PR #105, opened 2020-12-07
Create Application Context

### LINKED ISSUE #100: Create Application Context
Use Application Context to set and get global variables.

celinecamacho:
Closes #100 

Use the new PoPApplication class to set and get global variables. This way we can pass Person and LAO information between activities.


jbmichel14 (approved):

app/src/main/java/com/github/dedis/student20_pop/MainActivity.java
```diff
@@ -66,17 +66,19 @@ public void onClick(View view) {
                 break;
             case R.id.button_launch:
                 String name = ((EditText) findViewById(R.id.entry_box_launch)).getText().toString();
-                // For later: send LAO and organizer information
                 Person organizer = new Person("name");
```
GabrielFleischer, line 69: `organizer =  ((PoPApplication) getApplication()).getPerson()`
We should extract organizer as us here.

GabrielFleischer (approved):
It is ok, just one change but it comes with all the changes needed now that there is a context

app/src/main/java/com/github/dedis/student20_pop/MainActivity.java
celinecamacho: Indeed, good catch

app/src/androidTest/java/com/github/dedis/student20_pop/MainActivityTest.java
```diff
@@ -47,6 +53,8 @@ public void onClickLaunchTest() {
     @Test
     public void onClickLaunchLAOTest() {
         onView(withId(R.id.tab_launch)).perform(click());
+        onView(withId(R.id.entry_box_launch)).perform(typeText("LAO"));
+        onView(withId(R.id.entry_box_launch)).perform(closeSoftKeyboard());
```
AlexisTabin, line 57: you can write this on one line : `.perform(typeText("LAO"), closeSoftKeyboard());`

AlexisTabin (approved):

app/src/androidTest/java/com/github/dedis/student20_pop/MainActivityTest.java
celinecamacho: Good tip, thanks

PR #116, opened 2020-12-09
Go backend : new code structure, more functionnalities

romain46:
### New structure :
The project is now splitted in 7 packages/folders :
1. actors : defines the actors and how they behave
2. db : helper functions to work with databases
3. define : all datastructures as well as functions to convert json messages to structures
5. JsonRPC which is just a git submodule, branching the proto-spec branch
6. test : self-explainatory. Not complete yet
7. WebSocket : handles all the network connections as well as the publish/subscribe paradigm.

### more functionnalities :
currently implemented features :
#### Organizer :
* publish something on a channel
* Subscribe / unsubscribe to a channel
* Create LAO
* UpdateProperties for LAO
* handling the witness messages (not fully implemented)
* CreateRollCall (not fully implemented)

#### Witness :
The witness currently just stores the messages it gets in his database. Does not handle Publish method yet.


define/jsonhelper.go
```diff
@@ -124,14 +134,20 @@ type DataCreatePoll struct {
...
 	Start int64  /* Timestamp */
 	End   int64  /* Timestamp, optional */
 	Extra string /* arbitrary object, optional */
 }
+type DataWitnessMessage struct {
+	Object     string
+	Action     string
+	Message_id string
+	Signature  string
+}
```
pierluca, line 151: below is a lot of logic that goes way beyond what the package name refers to `define`.
You might consider splitting this package into different packages, one dealing with just the JSON-RPC layer, another dealing with messaging, and the final one dealing with the actual message types. In that case, the definitions would be in each package.

define/securityHelpers.go
```diff
@@ -34,46 +25,48 @@ func LAOCreatedIsValid(data DataCreateLAO, message Message) error {
...
-	if data.Creation != data.Last_modified {
-		return ErrInvalidResource
-	}
+	/*if data.Creation != data.Last_modified {
+		return false
+	}*/
 	//the timestamp is reasonably recent with respect to the server’s clock,
-	if data.Creation > time.Now().Unix() || data.Creation-time.Now().Unix() > MaxTimeBetweenLAOCreationAndPublish {
-		return ErrInvalidResource
+	if data.Creation > time.Now().Unix() || data.Creation-time.Now().Unix() > MaxPropagationDelay {
```
pierluca, line 46: this logic seems wrong. Let me rewrite an equivalent statement for you:

`if data.Creation > time.Now().Unix() || data.Creation > time.Now().Unix() + MaxPropagationDelay`

there's also no margin on the 'lower bound' on the data.Creation. 
Ideally you want both an upper and a lower bound with some margin, to take into account for the hosts having slightly different clocks.

define/securityHelpers.go
```diff
@@ -84,13 +77,26 @@ func MessageIsValid(msg Message) error {
 	}
 
 	// the signature is valid
-	err := VerifySignature(msg.Sender, msg.Data, msg.Signature)
+	err := VerifySignature(msg.Sender, (msg.Data), msg.Signature)
 	if err != nil {
-		return err
+		return err //err
```
pierluca, line 82: I assume you'll want to remove that :)

define/securityHelpers.go
```diff
@@ -84,13 +77,26 @@ func MessageIsValid(msg Message) error {
 	}
 
 	// the signature is valid
-	err := VerifySignature(msg.Sender, msg.Data, msg.Signature)
+	err := VerifySignature(msg.Sender, (msg.Data), msg.Signature)
```
pierluca, line 80: needless parenthesis ?

define/securityHelpers.go
```diff
@@ -84,13 +77,26 @@ func MessageIsValid(msg Message) error {
...
 	if err != nil {
-		return err
+		return err //err
 	}
 
 	// the witness signatures are valid (check on every message??)
-	return VerifyWitnessSignatures()
+	publicKeys, err := AnalyseData(string(msg.Data))
+	if publicKeys["object"] == "lao" && publicKeys["action"] == "create" {
```
pierluca, line 87: `publicKeys["action"]` makes me think we're not dealing with publicKeys at all :)

define/securityHelpers.go
```diff
@@ -101,48 +107,48 @@ func VerifySignature(publicKey string, data []byte, signature string) error {
...
 	}
-	dataDecoded, err := Decode(data)
-	if err != nil {
-		return ErrEncodingFault
-	}
-	toCheck := append(senderDecoded, dataDecoded...)
-	for i := 0; i < len(signatures); i++ {
-		err := VerifySignature(string(publicKeys[i]), toCheck, string(signatures[i]))
+	//TODO do we only check the pairs in witnessSignaturesEnc (1) or do need to verify that the publicKey
+	// of the pair is in the publicKeys before (2)?
```
pierluca, line 144: TODO unclear for me :)

main.go
```diff
@@ -2,24 +2,40 @@
...
 	"student20_pop/WebSocket"
 	"text/template"
 
-	"flag"
 	"log"
 	"net/http"
 )
 
 // this function basically makes the webserver run
 func main() {
```
pierluca, line 17: careful, you're not validating any of the args.
You could make your life easy by using a cli dependency that allows you to define flags, acceptable formats, cli documentation, and what not.

test/db_test.go
```diff
@@ -0,0 +1,22 @@
+package main
+
+/*
```
pierluca, line 3: something tells me you did not intend to commit this

test/json_test.go
```diff
@@ -0,0 +1,43 @@
+// json.go
+package main
+/*
+import (
```
pierluca, line 4: or this

test/securityHelper_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+import (
+	ed "crypto/ed25519"
+	"crypto/sha256"
+	b64 "encoding/base64"
+	"encoding/json"
+	"fmt"
+	"student20_pop/define"
+	"testing"
+	"math/rand"
```
pierluca, line 12: Maybe you're aware, maybe not:
> For random numbers suitable for security-sensitive work, see the crypto/rand package.
> https://golang.org/pkg/math/rand/

test/securityHelper_test.go
```diff
@@ -0,0 +1,74 @@
+// define/securityHelpers
```
pierluca, line 1: This file has not been `gofmt` , I can't review it as-is, it's unreadable.

test/websockets_test.go
```diff
@@ -0,0 +1,43 @@
+// websockets.go
+package main
```
pierluca, line 2: or this

pierluca (changes requested):
I've looked through the PR and there are a lot of good things. 
Allow me to focus, on a general level, on the few problems which I've identified and I'm going to try to summarize below:

- naming (variables, packages, functions, etc.) does not express your intention. it's hindering comprehension rather than helping it. Good code should tell a story. 
The best reference for this is Code Complete, 2nd edition, Microsoft Press
Here are a few other takes on the subject:
https://dmitripavlutin.com/coding-like-shakespeare-practical-function-naming-conventions/
https://stackoverflow.com/questions/1866794/naming-classes-how-to-avoid-calling-everything-a-whatevermanager

- you tend to have a few massive packages, which will grow (in time) harder to split. You should think of packages as a way to help with the "Separation of Concerns" design principle. The protocol allows for a natural split between different concerns, take advantage of it ! You could have a low-level package (name to be defined), dealing just with the JSON-RPC structures, parsing them, creating them, etc. 
Another mid-level package (name to be defined), dealing just with the mid-level communication, and interfacing with an abstract "low-level" communication interface (should we decide to replace JSON-RPC for whatever reason)
And finally an application package (name to be defined), dealing with just the actual message processing and stuff.
The beauty of it, if this is done right, is that one could entirely replace a part of the system, and the rest would still work nicely. 

- the database choice and design seems very odd for the task at hand. You're essentially storing a sequence of items (order somewhat matters) and you need to send them out based on some filtering (per-channel), why use a key-value storage and roll out your own listing logic. What is your motivation behind this choice? (I'd rather ask you know than in january ;-) )

Things being said, you don't have to fix everything, it's up to you to judge if it's realistic to do so and whether you can defend your choices :)

test/securityHelper_test.go
wouewouel: Thanks, I've made the update but for this test I don't know if it has a significant difference

test/securityHelper_test.go
wouewouel: that's fixed

WebSocket/hub.go
```diff
@@ -13,7 +14,7 @@ import (
 	"time"
 )
 
-const SIG_TRESHOLD = 4
+const SIG_THRESHOLD = 4
```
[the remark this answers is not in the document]
wouewouel: This one stays in hub.go so we applied mixedCase

WebSocket/homePage.go
```diff
@@ -10,6 +9,6 @@ import (
 // Serves the http homepage
 func HomeHandler(tpl *template.Template) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		tpl.Execute(w, r)
+		_ = tpl.Execute(w, r)
```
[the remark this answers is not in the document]
wouewouel: You right, now we log it

WebSocket/homePage.go
```diff
@@ -1,5 +1,4 @@
-/* file that serves a webpage. Comes from the chat example of github.com/websocket with minor changes */
-
+/* file to serve a webserver. Comes from the chat example of github.com/gorilla */
 package WebSocket
```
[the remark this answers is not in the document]
wouewouel: Thanks for the convention, we applied changes

README.md
```diff
@@ -1,22 +1,48 @@
...
+Back-end for the Proof-of-personhood application. To run the backend simply : `go run main.go <mode> <address> <publicKey>` where
+`<mode>` can be either 'o' or 'w' (for organizer and witness respectively), `<address>` is the address and port on which 
+the server should run (e.g. localhost:8080)  and `<publicKey>` is the actor's public key.
+
 # Packages
 
-## db 
-This package contains helper functions to interact with databases. A big code refactor will be pushed soon and this package will absorb the current channel package.
+## Database 
+The idea is to store a big string which contains all the information (in json format)
```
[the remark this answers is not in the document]
romain46: I see just now that I PR the wrong README.md version. I'll request another review once I've fixed this

WebSocket/hub.go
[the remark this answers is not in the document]
romain46: well to be precise this one is currently not used and it should probably be in another package (actors) as it references the number of signatures needed to consider a "update properties" request validated and send a "state update" to every subscribers.

WebSocket/hub.go
```diff
@@ -55,22 +56,35 @@ type hub struct {
...
 		connIndex:       0,
 		idOfSender:      -1,
-		organizer:       actors.NewOrganizer("", "orgDatabase.db"),
-		witness:         actors.NewWitness("", "witDatabase.db"),
+	}
+
+	if mode == "o" {
+		h.actor = actors.NewOrganizer(pkey, "orgdatabase.db")
+	} else if mode == "w" {
+		h.actor = actors.NewWitness(pkey, "witdatabase.db")
```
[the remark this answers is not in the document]
romain46: yes, you're right, we should modify this.
Though this never overwrites anything, it's just going to open the files. This will cause both server to run slower (because of lock requests on the file when it is accessed), but as the protocol and our database is defined, there should be no data loss. The real problem is that a witness could access info "it should not" access, like channel he's not subscribed to for example

actors/organizer.go
```diff
@@ -22,22 +22,18 @@ func NewOrganizer(pkey string, db string) *Organizer {
...
-
 /** processes what is received from the WebSocket
- * msg : receivedMessage
+ * msg : received message
  * returns, in order :
  * message to send on channel, or nil
  * channel for the message, or nil
  * response to the sender, or nil
  */
 func (o *Organizer) HandleWholeMessage(msg []byte, userId int) ([]byte, []byte, []byte) {
```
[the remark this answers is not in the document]
romain46: Agreed for the function name, we will think about it. in the mean time I'll change to named return values, this will make the code way more readable

actors/witness.go
```diff
@@ -113,35 +113,33 @@ func (w *Witness) handlePublish(generic define.Generic) ([]byte, []byte, error)
...
-
+			return nil, nil, define.ErrInvalidAction
 		default:
 			return nil, nil, define.ErrInvalidAction
 		}
 	case "meeting":
 		switch data["action"] {
 		case "create":
-			//return w.handleCreateMeeting(message, params.Channel, generic)
+			return nil, nil, define.ErrInvalidAction
```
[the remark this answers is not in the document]
romain46: well we don't have any error in the protocol for not yet implemented. Hence we say the Action is invalid, as the message totally valid is.

But true, we can implement another "intern" error type notImplemented and return this one instead (would make the code more readable). Do you prefer we do it in this PR or in the next one? (as it will change many parts of the code)

actors/witness.go
```diff
@@ -185,40 +179,116 @@ func (w *Witness) handleUpdateProperties(message define.Message, channel string,
...
 	//stores received message in DB
-	canalLAO := channel + data.ID
-	err = db.CreateMessage(message, canalLAO, w.database)
+	err = db.CreateMessage(message, channel, w.database)
 	if err != nil {
 		return nil, nil, err
 	}
 
-	//toSign := message.Sender + string(message.Data)
+	msg := db.GetMessage([]byte(channel), []byte(data.Message_id), w.database)
```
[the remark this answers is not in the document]
romain46: mhhh yes, but it is defined this way in the protocol definition, and the json parser (Marshall) is case-insensitive but not "underscore-insensitive". So it's either we leave it this way, or we change the protocol. We could also create a structure exactly similar to Message of course, and initiate it from the values we got from the unmarshalling, but to me this is just an unuseful and not elegant code replication... What do you think ?

test/db_test.go
romain46: yep, it's in there because we still need to think about how we want to test it, but it is currently not implemented you're right

WebSocket/hub.go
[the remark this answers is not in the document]
romain46: so I deleted it, we'll add it again when needed

README.md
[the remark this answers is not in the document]
romain46: Okay I pushed a new README. Sorry you read the wrong one. Could you just tell us quickly (slack or git) if you're still confused by how we manage the storage, and if you're question here above still holds ?

README.md
```diff
@@ -1,56 +1,72 @@
...
-1. for the hub, containing a list of channels and their subscribers
-2. one for the witness (described below)
-3. one for the organizer (same structure as witness's one)
+### Database Structure
+Currently, there are 2 databases created/opened when we run the backend: orgDatabase.db/witDatabase.db and sub.db.
+
+#### Sub.db
+This database exists only to handle publisher and subscribers of channels. Currently, we only manage subscribers. There is a bucket called
+"sub", in which we store the (key, value) pairs as following : each key is the ID of one channel, and the value is an array of connections
+that subscribed to this channel. When someone subscribes or unsubscribes, we simply edit this array.
```
pierluca, line 52: does this have to be persisted to disk? the connections are lost if someone disconnects, right?

README.md
```diff
@@ -1,56 +1,72 @@
...
+
+#### witDatabase.db / OrgDatabase.db
+This file has the same structure, whether it is witness or organizer. It has 2 types of bucket :
+* The first one is called `channels` and for each channel it contains a (key, value) pair, where the key is the channel's ID
+and the value is a string containing the "current state" of the channel. There is no history of messages stored here, only the channel's name, organizer, list of witnesses, etc...
+* The second type of bucket is the bucket containing the history of the channel. There is one such bucket per channel, and 
+it has the following properties :
+    1. The bucket name is the ID of the channel
+    2. The (key, values) pairs are (ID, content) of messages sent on the channel.
```
pierluca, line 61: Is this a list of (key, value) pairs or a dictionary, indexed by key ? 
How can you (efficiently) respond to a simple query such as "give me all the messages between time X and time Y"?

test/securityHelper_test.go
pierluca: Not in the tests. I'm just pointing that out in case you use that elsewhere :)

README.md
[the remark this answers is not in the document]
pierluca: It's more understandable, I'm still not convinced that a key-value store has any advantage (in this application) over a simple relational database, but at least I understand what you did, but if there are advantages I'm failing to see, feel free to ignore me and/or (even better) prove me wrong :-)

README.md
romain46: Well, you're right. I did not think of that as I first implemented this believing we would know the connection's public keys, and then I just tweaked my implementation to make it work without the Pkeys. I'll edit this soon.

README.md
romain46: If I am not mistaken, it is indexed by key (with a B+Tree index). I also just found out there is a fork of boltDB, kept up to date and actively maintained, so I might consider switching to it.

Anyway, we currently have no efficient way to execute this query. We could build an index ourselves, by having a second bucket, containing key-values pairs with keys being timestamps and values the message sent/received at this precise time. This would double the storage required, but we could execute the query in O(logN + M) time, instead of O(N), where N is the number of entries and M the number of items corresponding to the search. We can do this quite easily, but we need to be sure it is "worth it".

To answer your first question, we use a simple key-value storage, and not a more complex relational storage, because @bford told us so in the beginning of the project. It's true we have not tried to doubt this, though as the overall goal is to create a mobile App, I think the biggest advantage is that it's lighter and so needs less ressource.

Though, if you (and Brian) think it's worth it to reconsider using a relational database, I'll gladly do it. Maybe we should discuss that in the next meeting ?

define/securityHelpers.go
romain46: yep, didn't see it, thanks !

define/securityHelpers.go
romain46: indeed

define/jsonhelper.go
romain46: so we splitted the package define in 5 packages :

* security
* events (old dataDefinitions.go)
* lib (general functions like Decode and Find)
* parser (+composer to read / write messages)
* message struct representing JSON sent/received messages

We'd be glad if you could go and take a look at it :)

define/jsonhelper.go
```diff
@@ -233,6 +273,21 @@ func AnalyseDataCreateLAO(data json.RawMessage) (DataCreateLAO, error) {
 	return m, err
 }
 
+func AnalyseDataWitnessMessage(data json.RawMessage) (DataWitnessMessage, error) {
```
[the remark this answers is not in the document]
romain46: yep, we refactored that

define/jsonhelper.go
```diff
@@ -372,12 +425,18 @@ func selectDescriptionError(err error) []byte {
 			Description: "access denied",
 		}
 		//(e.g. subscribing to a “restricted” channel)
+	case ErrIdNotDecoded:
+		errResp = ErrorResponse{
+			Code:        -2,
```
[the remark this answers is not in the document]
romain46: To me the closest is -4,  request data invalid. Anyway this does not satisfy me. It's not the request's Data that is invalid, it's the whole message, and We have no error for that yet

define/helper.go
```diff
@@ -1,4 +1,5 @@
-/*useful functions, "missing in the standard library". Currently only finding an elem in a slice */
+/* GENERAL USEFUL FUNCTIONS
```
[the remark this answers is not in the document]
romain46: we created a package lib and moved it there

README.md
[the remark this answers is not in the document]
romain46: this conversation continues in your latest comment on the README, so I am closing it

actors/actor.go
```diff
@@ -0,0 +1,48 @@
...
+	handlePublish(generic define.Generic) ([]byte, []byte, error)
+	handleCreateLAO(message define.Message, channel string, generic define.Generic) ([]byte, []byte, error)
+	handleUpdateProperties(message define.Message, channel string, generic define.Generic) ([]byte, []byte, error)
+	handleWitnessMessage(message define.Message, channel string, generic define.Generic) ([]byte, []byte, error)
+	handleLAOState(message define.Message, channel string, generic define.Generic) ([]byte, []byte, error)
+	handleCreateRollCall(message define.Message, channel string, generic define.Generic) ([]byte, []byte, error)
+}
+
+//general actors functions, act only in the "Sub" database
+func handleSubscribe(generic define.Generic, userId int) error {
```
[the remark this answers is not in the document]
Huntsekkern: The name generic came from the protocol (https://github.com/dedis/student20_pop/blob/proto-specs/genericMessage.json), however it was not up to date anymore, so we rename it to query (to match the nomenclature https://github.com/dedis/student20_pop/blob/proto-specs/query/query.json), and added one more level above, called genericMessage which englobes now both queries and answers

actors/actor.go
```diff
@@ -0,0 +1,48 @@
...
+	params, err := define.AnalyseParamsLight(generic.Params)
```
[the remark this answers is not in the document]
Huntsekkern: Is this clearer? params, err := parser.ParseParams(query.Params)
It still has the issue of parsing params to return params, but the idea is that the field query.Params is in json, while the returned params is a structure

actors/organizer.go
```diff
@@ -51,35 +47,59 @@ func (o *Organizer) HandleWholeMessage(msg []byte, userId int) ([]byte, []byte,
...
 }
 
-func handleSubscribe(generic define.Generic, userId int) error {
-	params, err := define.AnalyseParamsLight(generic.Params)
+func (o *Organizer) handleMessage(generic define.Generic) ([]byte, []byte, error) {
+	params, err := define.AnalyseParamsFull(generic.Params)
 	if err != nil {
-		fmt.Printf("3")
-		return define.ErrRequestDataInvalid
+		fmt.Printf("unable to analyse paramsLight in handleMessage()")
```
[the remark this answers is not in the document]
Huntsekkern: Changed to Params and ParamsIncludingMessage.
It can still be argued that it's not great, but unfortunately, if we want to use clean and nice structure to parse, we have to use two different structures depending on whether params only contains the channel id, vs the channel id + the message content

actors/organizer.go
```diff
@@ -262,19 +288,22 @@ func (o *Organizer) handleCreateMeeting(message define.Message, canal string, ge
...
 	err = db.CreateChannel(event, o.database)
 	if err != nil {
 		return nil, nil, err
 	}
-	channel, msg := finalizeHandling(message, canal, generic)
+	err = db.CreateMessage(message, canal, o.database)
+	if err != nil {
+		return nil, nil, err
+	}
+	channel, msg := finalizeHandling(canal, generic)
```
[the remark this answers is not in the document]
Huntsekkern: Currently, it's an absolutely useless function which only calls another function. It used to do more, so I'm not deleting the function yet, but replacing its calls by the direct parser.ComposeBroadcastMessage(query), []byte(canal)

actors/organizer.go
```diff
@@ -150,16 +177,15 @@ func (o *Organizer) handlePublish(generic define.Generic) ([]byte, []byte, error
...
 		case "state":
-
+			// TODO: waiting on protocol definition
+			return nil, nil, define.ErrInvalidAction
 		default:
 			return nil, nil, define.ErrInvalidAction
 		}
 	default:
-		fmt.Printf("8")
+		fmt.Printf("data[action] not recognized in handlepublish, generating default response ")
```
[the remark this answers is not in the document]
romain46: good point, we changed that

define/jsonhelper.go
```diff
@@ -292,23 +331,37 @@ func CreateBroadcastMessage(generic Generic) []byte {
...
+				Jsonrpc:       "2.0",
+				ErrorResponse: selectDescriptionError(err),
+				Id:            generic.Id,
+			}
+
+			b, err := json.Marshal(resp)
+			if err != nil {
+				fmt.Println("couldn't Marshal the response")
+			}
+			return b
```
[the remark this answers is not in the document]
romain46: yep, I refactorized that, thank you :)

define/securityHelpers.go
romain46: oh, yeah you're right. So we changed the logic to that :
`data.Creation < time.Now().Unix() - MaxClockDifference || data.Creation > time.Now().Unix()+MaxPropagationDelay`
This takes into account clock "dissynchronisation" for min time and for the propagation delay for max time. Maybe to be clearer we should add the MaxClockDifference to the max time as well, so we don't expect the propagation to account for propagation and Clock Dissynchronisation. What do you think ?

README.md
pierluca: Yes, let's discuss this. Sqlite (for instance) could be just as lightweight while offering more thorough functionality. 
Changing storage backend might not be the highest priority task either, and we could consider it out of scope, while taking note that it might be worth doing later on :)

actors/actor.go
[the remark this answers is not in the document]
pierluca: it might make sense to rename it like you suggest.

In a general sense, compare the following things.
```
params, err := define.AnalyseParamsLight(generic.Params)
``` 
to that:
```
params, err := jsonrpc.ParseParams(rpc.Params)
```
I think the latter fully expresses what you're doing, 

Obviously, I write the latter freely, unencumbered by the constraints of your codebase, just to express what I mean.
In a general sense, however, you should aim to achieve that kind of readability.
Code is written a couple of times and read a few dozens or hundreds of time. You're doing your future self a favor :)

actors/actor.go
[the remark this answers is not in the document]
pierluca: Kudos for matching up the code with the spec. That's an excellent decision.
And you're right that this would be perfectly clear to someone reading the code, if the package name reflected that.
Say it were called `jsonrpc.Generic`, I probably wouldn't even have asked you the question, but `define` makes it very hard to guess :)

actors/witness.go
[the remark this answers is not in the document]
pierluca: I would suggest adding a `ErrNotImplemented` in the working spec to facilitate cross-team collaboration. That won't lead to much code change for anyone.

As for how you want to handle that (this PR or a next one), that's up to you :)

actors/witness.go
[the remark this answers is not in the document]
pierluca: Not true I believe.
In the struct you can define how to do the mapping between json and golang struct.
https://golang.org/pkg/encoding/json/#Marshal

define/jsonhelper.go
pierluca: Happy to, where is that?

define/securityHelpers.go
pierluca: Up to you, both options are reasonable.
Adding MaxClockDifference on both sides makes it easier to 'formally' define the specifications :)

pierluca (commented):
First batch of re-review

pierluca (commented):
Second batch of review - I won't have time to add more tonight.

README.md
romain46: Even if I have no idea of the performance impact, I don't think it makes sense to change this on week 14. We'll wirte it down as an improvement suggestion for later, but it's most probably not going to be implemented before the end of the project.

main.go
romain46: yep, I didn't know we could do this, thank you. It's implemented now :)

define/securityHelpers.go
romain46: I did clarify it a bit but the TODO concerns mostly the developpers :)

define/jsonhelper.go
```diff
@@ -292,23 +331,37 @@ func CreateBroadcastMessage(generic Generic) []byte {
 }
 
 /*
-* Function that converts a Lao to a Json byte array
-* we suppose error is in the good range
+ * Function that create the response to the sender
+ * we suppose error is in the good range
  */
-
 func CreateResponse(err error, messages []byte, generic Generic) []byte {
```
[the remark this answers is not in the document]
romain46: good idea !

security/validateMessage.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	//we start after the creation and we end after the start
+	if data.Start < data.Creation || data.End < data.Start {
+		return false
+	}
+	//need to meet some	where
+	if data.Location == "" {
+		return false
+	}
```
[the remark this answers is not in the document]
romain46: yep, good suggestion, we've changed that.

README.md
romain46: Well I think I'm going to implement this for the next PR, as it will then make sense to have the hub managing the Publish-Subscribe level. But this would imply that the Hub make a first level of Unmarshalling the message. This is not a problem, but as the code works in it's current state, I prefer to postpone this.

romain46:
@pierluca We've implemented most of your required changes. We'd be glad if you could re-review our code. If you can't no problem, you've already given us an amazing input.

main.go
```diff
@@ -2,24 +2,48 @@
...
+		log.Fatal(http.ListenAndServe(*address+":"+strconv.Itoa(*port), router)) //here to change the srv address
+
+	case "w":
+		h := network.NewWitnessHub(*pkey, *file)
+		router := http.NewServeMux()
+		router.Handle("/", network.HomeHandler(tpl))
+		router.Handle("/ws", network.NewWSHandler(h))
+		log.Printf("serving witness on adress " + *address + ":" + strconv.Itoa(*port))
+		log.Fatal(http.ListenAndServe(*address+":"+strconv.Itoa(*port), router)) //here to change the srv address
+
```
pierluca, line 46: you might want to output an error message if it's neither "o" nor "w" :)

romain46: well, it's the lines 27 to 29 :)

README.md
```diff
@@ -1,56 +1,75 @@
...
-* the `hub.go` maintains a list of opened connections and "routes" the messages. We will change this so there is no more routing to do in the next update.
-* the `connection.go` is the file that properly handles the websocket connection. When it receives a message from the client, it passes 
-it to the hub.
-* the `index.html` acts like the client that prints messages received from the server and that can send messages to the server.
-* the `homePage.go` just serves the `index.html` file on http requests
-### tweaking the server's properties 
+### network testing 
 If you want to test the websocket protocol the server running on a different machine than the client, you have to set the
 server's IP address in `index.html`  at line 36 (instead of `localhost`). You'll have to put your computer's IP address
 before the ":8080" of `main.go` at line 29.
```
haoqianzhang, line 47: It is better to have a separate config file rather than directly editing the code.

haoqianzhang (approved):

main.go
pierluca: sorry, I missed it, I just expect to see a `default` case 😅

README.md
romain46: well this file is not "part of the project". It stands for a "minimalist frontend" that just opens a ws, sends messages through it and prints the received messages

test/db_test.go
lhmerino: Sounds good, then please add a comment to these files about their status.

lhmerino (approved):
Great job! Very clean and readable code.

pierluca (approved):
Unblocking the merge - haven't re-reviewed.

PR #118, opened 2020-12-09
Person creation and App restart

### LINKED ISSUE #112: Person creation and Application restart
Look for the user information when the Application starts. If it exists then restore the information and look for the pair of keys of the user, else create the user's information.

celinecamacho:
Closes #112
Verify user existence when starting the app:

- If exists then get information from SharedPreferences and private key from persistent storage
- Otherwise create new information 

Save the information every time the Application terminates.


app/src/main/java/com/github/dedis/student20_pop/PoPApplication.java
```diff
@@ -2,18 +2,26 @@
...
 import com.github.dedis.student20_pop.model.Person;
+import com.github.dedis.student20_pop.utility.security.PrivateInfoStorage;
+import com.google.gson.Gson;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
+import java.util.Objects;
 
 public class PoPApplication extends Application {
```
jbmichel14, line 18: Just to be clear, I also made changes to this class but I'll make sure we have everything, and will modify the tests accordingly

jbmichel14 (approved):
Nice work

AlexisTabin (approved):

PR #119, opened 2020-12-09
Implement Identity UI

AlexisTabin:
Very simple UI to show a QR code

### DIFF
--- a/app/src/main/java/com/github/dedis/student20_pop/OrganizerActivity.java
+++ b/app/src/main/java/com/github/dedis/student20_pop/OrganizerActivity.java
@@ -10,12 +10,17 @@ import androidx.core.app.ActivityCompat;
 import androidx.fragment.app.Fragment;
 import androidx.fragment.app.FragmentActivity;
 
+import com.github.dedis.student20_pop.model.Keys;
+import com.github.dedis.student20_pop.model.Lao;
 import com.github.dedis.student20_pop.ui.CameraPermissionFragment;
 import com.github.dedis.student20_pop.ui.ConnectFragment;
 import com.github.dedis.student20_pop.ui.HomeFragment;
+import com.github.dedis.student20_pop.ui.IdentityFragment;
 import com.github.dedis.student20_pop.ui.OrganizerFragment;
 import com.github.dedis.student20_pop.utility.ui.OnAddWitnessListener;
 import com.github.dedis.student20_pop.utility.ui.OnEventTypeSelectedListener;
+
+import java.util.Date;
 
 /**
  * Activity used to display the different UIs for organizers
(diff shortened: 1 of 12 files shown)


app/src/main/java/com/github/dedis/student20_pop/OrganizerActivity.java
```diff
@@ -55,6 +63,22 @@ public void onClick(View view) {
...
+                bundle.putString(PRIVATE_KEY_TAG, app.getPerson().getAuthentication());
+
+                //TODO : Retrieve this LAO from the Intent
+                Lao lao = new Lao("LAO I just joined", new Date(), new Keys().getPublicKey());
+                bundle.putString(LAO_ID_TAG, lao.getId());
+
+                // set Fragmentclass Arguments
+                IdentityFragment identityFragment = new IdentityFragment();
+                identityFragment.setArguments(bundle);
+                showFragment(identityFragment, IdentityFragment.TAG);
```
jbmichel14, line 79: Is the identity UI the same if we're attendee or organizer ?

app/src/main/java/com/github/dedis/student20_pop/ui/IdentityFragment.java
```diff
@@ -0,0 +1,97 @@
...
+                    }
+                });
+
+        //TODO : get User personal key
+        //NOTE : I was not sure which unique id to use to generate the QR code
+        //User identity is composed of :
+        //User's public key
+        //Organization's ID
+        String key = this.getArguments().getString(OrganizerActivity.PRIVATE_KEY_TAG);
+        String lao = this.getArguments().getString(OrganizerActivity.LAO_ID_TAG);;
```
jbmichel14, line 76: There are 2 semicolons at the end of the line

app/src/main/java/com/github/dedis/student20_pop/ui/IdentityFragment.java
```diff
@@ -0,0 +1,97 @@
...
+
+        // TODO :   The QR code does not appear at all unless the Name field is non-empty
+        //  and not all whitespace.
+        CheckBox anonymousCheckBox = view.findViewById(R.id.checkbox_anonymous);
+        qrCode = view.findViewById(R.id.qr_code);
+        identityEmailEditText = view.findViewById(R.id.identity_email);
+        identityNameEditText = view.findViewById(R.id.identity_name);
+        identityOrganizationEditText = view.findViewById(R.id.identity_organization);
+        identityPhoneEditText = view.findViewById(R.id.identity_phone);
+        identityTitleEditText = view.findViewById(R.id.identity_title);
```
jbmichel14, line 52: Who (other than the user) can see these when the user is not anonymous?

app/src/main/java/com/github/dedis/student20_pop/OrganizerActivity.java
AlexisTabin: For the moment, in ordre to keep it simple, I assumed that it was thes cas. There is a comment about that suject In the specifications, it is mentionned the following : 
" The paradigm might be that you’re always anonymous if you’re an ordinary attendee and always non-anonymous if you’re an organizer or witness. "
But I think we can still change that in the future

app/src/main/java/com/github/dedis/student20_pop/ui/IdentityFragment.java
AlexisTabin: For the moment, we do not use theses fields at all

app/src/main/java/com/github/dedis/student20_pop/ui/IdentityFragment.java
AlexisTabin: Removed

jbmichel14:
OK Good job!

jbmichel14 (approved):

GabrielFleischer (approved):
Everything seems fine to me

PR #120, opened 2020-12-09
Fix error in regex matching for the Protocol tests

GabrielFleischer:


MockServer/src/main/java/com/github/dedis/student20_pop/mockserver/WebSocketServer.java
```diff
@@ -55,13 +55,12 @@ public void onOpen(Session session) {
...
             Matcher matcher = pattern.matcher(message);
-            String id = matcher.group(0);
-            if(id != null) {
+            if(matcher.find()) {
                 try {
                     session.getBasicRemote().sendText(
                             String.format(Locale.US,
                                     JSONRPC_FORMAT,
-                                    Integer.parseInt(id)));
+                                    Integer.parseInt(matcher.group(1))));
```
AlexisTabin, line 63: What does the 1 represents ? Would it be clearer if it was in a variable ?

AlexisTabin (approved):

MockServer/src/main/java/com/github/dedis/student20_pop/mockserver/WebSocketServer.java
GabrielFleischer: It's the first match. It's how to use the matcher, like a 0 in an array to get the first element.

celinecamacho (approved):

PR #122, opened 2020-12-10
Store witness signatures in the db

### LINKED ISSUE #111: Update messages stored in the database when receiving witness signatures

SFulpius:
Store the witness signatures with messages they are related to in the database.
Also fix the catchup test.
Closes #111.

### DIFF
--- a/src/main/scala/ch/epfl/pop/json/json.scala
+++ b/src/main/scala/ch/epfl/pop/json/json.scala
@@ -50,7 +50,12 @@ package object json {
                                    signature: Signature,
                                    message_id: Hash,
                                    witness_signatures: List[Signature]
-                                 )
+                                 ) {
+
+    def updateWitnesses(s: Signature): MessageContent =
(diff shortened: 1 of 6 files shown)


src/main/scala/ch/epfl/pop/json/json.scala
```diff
@@ -50,7 +50,12 @@ package object json {
                                    signature: Signature,
                                    message_id: Hash,
                                    witness_signatures: List[Signature]
-                                 )
+                                 ) {
+
+    def updateWitnesses(s: Signature): MessageContent =
+      MessageContent(encodedData, data, sender, signature, message_id, s :: witness_signatures)
```
RaulinN, line 56: Once this PR will be accepted, I'll change this function and update the code to reflect changes done in commit 5a14e14 (witness_signatures is now a `List[KeySignPair]` which is a pair of public key and corresponding signature)

RaulinN (approved):

haoqianzhang (approved):

PR #123, opened 2020-12-11
Implement generic QR code functionality

AlexisTabin:
The goal of this PR is to make the QR code scanning fragment generic, that way, the same fragment can be use to different applications :
- [ ] During a Roll-call event creation, attendees should show a QR code that the organizer can scan  (Not implemented yet)
- [x] To connect to a backend/LAO, need to scan organizer's QR code
- [x] To add witness

I also add 2 UI improvements in the Connecting Fragment : 

- [x] Add Cancel button when user is in "connecting . . ." state
- [x] Add Loading animation when in connecting state

### DIFF
--- a/app/src/main/java/com/github/dedis/student20_pop/OrganizerActivity.java
+++ b/app/src/main/java/com/github/dedis/student20_pop/OrganizerActivity.java
@@ -13,20 +13,27 @@ import androidx.fragment.app.FragmentActivity;
 import com.github.dedis.student20_pop.model.Keys;
 import com.github.dedis.student20_pop.model.Lao;
 import com.github.dedis.student20_pop.ui.CameraPermissionFragment;
-import com.github.dedis.student20_pop.ui.ConnectFragment;
+import com.github.dedis.student20_pop.ui.ConnectingFragment;
 import com.github.dedis.student20_pop.ui.HomeFragment;
 import com.github.dedis.student20_pop.ui.IdentityFragment;
 import com.github.dedis.student20_pop.ui.OrganizerFragment;
-import com.github.dedis.student20_pop.utility.security.PrivateInfoStorage;
+import com.github.dedis.student20_pop.ui.QRCodeScanningFragment;
+import com.github.dedis.student20_pop.ui.QRCodeScanningFragment.QRCodeScanningType;
+import com.github.dedis.student20_pop.utility.qrcode.OnCameraAllowedListener;
+import com.github.dedis.student20_pop.utility.qrcode.OnCameraNotAllowedListener;
+import com.github.dedis.student20_pop.utility.qrcode.QRCodeListener;
 import com.github.dedis.student20_pop.utility.ui.OnAddWitnessListener;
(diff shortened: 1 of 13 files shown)


AlexisTabin:
For the moment, the Roll-Call event creation part is not implemented, so I didn't do this part (I let it unchecked on purpose)

app/src/main/java/com/github/dedis/student20_pop/OrganizerActivity.java
```diff
@@ -118,10 +125,39 @@ public void OnEventTypeSelectedListener(EventType eventType) {
...
+                break;
+            default:
+                break;
+        }
+    }
+
+    @Override
+    public void onCameraAllowedListener(QRCodeScanningType qrCodeScanningType) {
+        showFragment(new QRCodeScanningFragment(qrCodeScanningType), QRCodeScanningFragment.TAG);
+    }
```
jbmichel14, line 162: Very similar code in MainActivity and OrganizerActivity, wouldn't it be possible to have these methods only in one place?

jbmichel14 (commented):
Seems all good to me

app/src/main/java/com/github/dedis/student20_pop/OrganizerActivity.java
AlexisTabin: Yes I totally agree. For onQRCodeDetected, as I understand it, the connect fragment should never be attached to OrganizerActivity but only to MainActivity. And the add Witness functionality is only reachable when you are an organizer so inversely it should be only in OrganizerActivity and not in MainActivity. But for the future, I thought maybe we want to do some different UI when an organizer tries to connect to an event, or I don't know. 

For camera interfaces, I could avoid that by creating abstract classes instead of interfaces, but I think it is better to keep the code managing the fragments inside the Activity.

jbmichel14: Ok, it's all good for now as we just prioritize on making it work

jbmichel14 (approved):

app/src/main/java/com/github/dedis/student20_pop/utility/qrcode/QRCodeListener.java
```diff
@@ -1,9 +1,11 @@
...
+import com.github.dedis.student20_pop.ui.QRCodeScanningFragment.QRCodeScanningType;
+
 /**
  * A listener that will receive detected url of a QR code
  */
 public interface QRCodeListener {
 
-    void onQRCodeDetected(String url);
+    void onQRCodeDetected(String url, QRCodeScanningType qrCodeScanningType);
 }
```
GabrielFleischer, line 11: We might want to change this in the future the give back a more general data than an url.

GabrielFleischer (approved):
Only something to change in the future

app/src/main/java/com/github/dedis/student20_pop/utility/qrcode/QRCodeListener.java
AlexisTabin: I can change that to make it more generic

Issue #101, opened 2020-12-04
Proposition of Roll-Call Event


celinecamacho:
Here is a description of the roll-call event associated messages. We will create the JSON-Schema once everyone agrees on the proposed messages.

### Create Roll-Call
Create a roll-call event. The event can be started in open mode, in which case a start time is specified, or in future mode, where the start time is not specified. If it is in future mode, the organizer needs to open the roll-call by using the “Open Roll-Call” message.
- object: “roll_call”
- action: “create”
- name: String
- id: base64 string SHA256("roll_call"||lao_id||creation||name) 
- creation: integer, [Timestamp]
- start: integer, [Timestamp] (optional)
- location: String
- roll_call_description: String (optional)
- scheduled_time: integer, [Timestamp] (optional)

### Open Roll-Call
Open a roll-call event. The event can be opened multiple times, the start time corresponds to the new open time. 
- object: “roll_call”
- action: “open” or "reopen"
- id: base64 string SHA256("roll_call"||lao_id||creation||name)
- start: integer, [Timestamp]

### Close Roll-Call
Close a roll-call event. If the event was reopened, then it will need to be closed again. Because of this the event can be closed multiple times, and the end time and attendees are overwritten. If the event was reopened, the start time we specify when closing the event should be the time when the event was reopened, not the time when the event was opened the first time.
- object: “roll_call”
- action: “close”
- id: base64 string SHA256("roll_call"||lao_id||creation||name)
- start: integer, [Timestamp]
- end: integer, [Timestamp]
- attendees: array base64 string 

@dedis/student20_pop @haoqianzhang @lhmerino @pierluca

imaximumGit:
> ### Create Roll-Call
> 
> Create a roll-call event. The event can be started in open mode, in which case a start time is specified, or in future mode, where the start time is not specified. If it is in future mode, the organizer needs to open the roll-call by using the “Open Roll-Call” message.

I think it will be good that a future roll-call has a start time or an announce time. This will allow the front-end to show the roll-call in the future events list at the correct place

SFulpius:
It is a good idea.

RaulinN:
Good initiative! I have a few suggestions/questions for you:

- should the `location` string in roll-call creation be optional like in the meetings?
- why can't we set a `end` time for the roll-call if we specify the `start` in the roll-call creation
- I do not understand what the following line means:
> The end time corresponds to the start time specified in a create/open message
[...]

PR #125, opened 2020-12-11
Roll-call specs

SFulpius:
The JSON-Schema for the roll-call.
See #101.

### DIFF
--- a/query/method/message/data/dataCloseRollCall.json
+++ b/query/method/message/data/dataCloseRollCall.json
@@ -10,7 +10,7 @@
 		"id": {
 			"type": "string",
 			"contentEncoding": "base64",
-			"$comment": "Hash : SHA256('roll_call'||lao_id||creation||name)"
+			"$comment": "Hash : SHA256('R'||lao_id||creation||name)"
 		},
 		"start": { "description": "[Timestamp] start time", "type": "integer", "minimum": 0 },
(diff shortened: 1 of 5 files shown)


RaulinN (approved):
Looks good to me! Good job! 👍

query/method/message/data/dataCloseRollCall.json
```diff
@@ -0,0 +1,36 @@
...
+	"description": "Match a close roll-call query",
+	"type": "object",
+
+	"properties": {
+		"object": { "const": "roll_call" },
+		"action": { "const": "close" },
+		"id": {
+			"type": "string",
+			"contentEncoding": "base64",
+			"$comment": "Hash : SHA256('R'||lao_id||creation||name)"
```
romain46, line 13: Why include the LAO's ID in the Hash ?

romain46: (in every files of course)

pierluca: to prevent conflict between multiple identical events from different LAOs, on the same server.
This supports the case of a single server shared across multiple LAOs.

GabrielFleischer (approved):

imaximumGit (approved):

romain46 (approved):

PR #132, opened 2020-12-16
Implementing add witness functionality

AlexisTabin:
It is now possible to add a witness by scanning a QR code

### DIFF
--- a/app/src/main/java/com/github/dedis/student20_pop/PoPApplication.java
+++ b/app/src/main/java/com/github/dedis/student20_pop/PoPApplication.java
@@ -30,6 +30,7 @@ public class PoPApplication extends Application {
     private static Context appContext;
     private Person person;
     private Map<Lao, List<Event>> laoEventsMap;
+    private Map<Lao, List<String>> laoWitnessMap;
 
     //represents the Lao which we are connected to, can be null
     private Lao currentLao;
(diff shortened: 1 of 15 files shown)


jbmichel14 (approved):

app/src/main/java/com/github/dedis/student20_pop/PoPApplication.java
```diff
@@ -24,12 +24,17 @@
  * Class modelling the application : a unique person associated with LAOs
  */
 public class PoPApplication extends Application {
+    public static final String TAG = PoPApplication.class.getSimpleName();
 
     private static final String LOCAL_BACKEND_URI = "ws://10.0.2.2:2000";
+    public static final int ADD_WITNESS_SUCCESSFUL = 0;
+    public static final int ADD_WITNESS_ALREADY_EXISTS = 1;
```
GabrielFleischer, line 31: I would rather have a boolean return value or a dedicated enum with the possible return values.

I don't really like int return values in java

GabrielFleischer (changes requested):

lhmerino (approved):
Very nice, lots of comments and tests! I didn't see any changes to the README but that's fine if there shouldn't be.

app/src/main/java/com/github/dedis/student20_pop/PoPApplication.java
AlexisTabin: Done :+1:

GabrielFleischer (approved):

PR #133, opened 2020-12-17
Specify how concatenation works

SFulpius:
Based on Tuesday's discussion, I've added to the protocol a definition of concatenation of strings.
One thing I'm not sure is if space (or any separator) between strings is needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,5 +26,5 @@ Hash: base64
 Timestamp: uint64 representation of the Unix timestamp (seconds since January 1st, 1970)  
 
 ## Concatenation for hashing
-When concatenating strings for hashing, `"` and `\` characters must be escaped by adding a `\` characters before them. `"` becomes `\"` and `\` becomes `\\`. Then each string is put between two `"` and strings are separated by a space.
-For example, <code>a<sub>1</sub>||a<sub>2</sub>||...||a<sub>n</sub></code> = <code>"esc(a<sub>1</sub>)" "esc(a<sub>2</sub>)" ... "esc(a<sub>n</sub>)"</code> where <code>a<sub>1</sub>, ..., a<sub>n</sub></code> are strings and `esc` is an escape function.
+When concatenating strings for hashing, `"` and `\` characters must be escaped by adding a `\` characters before them. `"` becomes `\"` and `\` becomes `\\`. Then each string is put between two `"` and added in a JSON array using compact representation.  
+For example, <code>a<sub>1</sub>||a<sub>2</sub>||...||a<sub>n</sub></code> = <code>["esc(a<sub>1</sub>)","esc(a<sub>2</sub>)",...,"esc(a<sub>n</sub>)"]</code> where <code>a<sub>1</sub>, ..., a<sub>n</sub></code> are strings and `esc` is an escape function.
--- a/README.md
+++ b/README.md
@@ -24,3 +24,7 @@ Public Key: base64
 Signature: base64  
 Hash: base64  
 Timestamp: uint64 representation of the Unix timestamp (seconds since January 1st, 1970)  
+
+## Concatenation for hashing
+When concatenating strings for hashing, `"` and `\` characters must be escaped by adding a `\` characters before them. `"` becomes `\"` and `\` becomes `\\`. Then each string is put between two `"` and strings are separated by a space.
+For example, <code>a<sub>1</sub>||a<sub>2</sub>||...||a<sub>n</sub></code> = <code>"esc(a<sub>1</sub>)" "esc(a<sub>2</sub>)" ... "esc(a<sub>n</sub>)"</code> where <code>a<sub>1</sub>, ..., a<sub>n</sub></code> are strings and `esc` is an escape function.


ineiti:
I don't get it why you need this. Usually all I do is to convert the string to its utf-8 representation, and then send that array of bytes to the hashing method. 

Can you give an example of where this fails?

pierluca:
@ineiti this follows a long discussion with Bryan, basically designed to avoid the `hash("ab" + "c") == hash("a" + "bc")` problem.

@SFulpius My understanding was that Bryan's recommendation was to follow a compact JSON array representation, i.e. `["ab",123,"c"]` with no spaces, to have something both standard-looking and simple. Did you diverge from it because you identified some issue?

ineiti:
Probably what I pointed out in the document in the beginning ;) 

So out of curiosity, what's the advantage of this versus some protobuf-encoding or `hash(int(strlen1) + str1 + int(strlen2) + str2)`?

pierluca:
@ineiti yes, it's what you pointed out. It wasn't a problem with the design choices at the time, but those have since changed, causing the need for adaptations.

There are probably a thousand different ways to deal with this, we can talk about those over coffee (whenever those are allowed again :-)). The proposed solution works, that's good enough ;-)

SFulpius:
@pierluca There is no special reason to use the format I specified. I did not remember in details what Bryan proposed, so I went for something similar. Do you think it would be better to use the solution using JSON array?

ineiti:
Problem with JSON is that its specifications allow different representations to be correct. But this will give of course different results in hashing. So you have to do a fully specified JSON.

For CrowdNotifier (https://notify-me.ch), we're using protobufs. Which also allow for multiple encodings to output the same data, but it's less bad. To be most consistent, you need to decode the binary to the structure, and then re-encode it.

There are other encodings, like CBOR, messagepack, which are better suited. See also https://crypto.stackexchange.com/questions/10058/how-to-hash-a-list-of-multiple-items for some more info.

pierluca:
@ineiti you really might want to consider joining the meetings, it's easier to discuss all at once :-) 

The limitations of JSON (as a format without canonical representation) were discussed at length throughout this project.
In this particular case, we settled (on Bryan's recommendation) on something that 'looks' like JSON but is stricter: an array of UTF-8 encoded strings (or numbers), with some characters escaped (`"`, `\`) and without any whitespace between syntactic elements. Essentially it doesn't require any external dependencies (as opposed to all of the above suggestions) and can be easily implemented 'by hand', by escaping a couple of characters and adding delimiters.

If we relax the constraints on importing new libraries in the project, we could also mention bencode, BSON array, UBJSON, Amazon Ion (binary encoding-variant), ... why stop at CBOR and messagepack? ;-) 

@ineiti you do raise valid points, and I would strongly support your suggestion, if it did not imply additional dependencies that we would have to integrate across 4 different platforms at a time where there's a lot of polishing still to do on the project.

I'll be off for the next few weeks and the project is drawing to a close, so I would just encourage the students to stick to what's been discussed and make sure the project is properly tested, documented and the code polished :)

SFulpius:
I updated the specs to use compact JSON array representation.

lhmerino (approved):
Awesome! Personally, I prefer strlen(str1) + str + strlen(str2) + str2 but this works as well!

imaximumGit (approved):

README.md
```diff
@@ -24,3 +24,7 @@ Public Key: base64
 Signature: base64  
 Hash: base64  
 Timestamp: uint64 representation of the Unix timestamp (seconds since January 1st, 1970)  
+
+## Concatenation for hashing
+When concatenating strings for hashing, `"` and `\` characters must be escaped by adding a `\` characters before them. `"` becomes `\"` and `\` becomes `\\`. Then each string is put between two `"` and added in a JSON array using compact representation.  
+For example, <code>a<sub>1</sub>||a<sub>2</sub>||...||a<sub>n</sub></code> = <code>["esc(a<sub>1</sub>)","esc(a<sub>2</sub>)",...,"esc(a<sub>n</sub>)"]</code> where <code>a<sub>1</sub>, ..., a<sub>n</sub></code> are strings and `esc` is an escape function.
```
romain46, line 30: I think the goal of this PR is to have something as clear as possible for hashing data. Here we have "esc is an escape function", if we don't know what characters to escape and by what we won't have the same results for the input string. I think it would be useful to specify what escape function we want to use

wouewouel: The escape function replace `"` and `\`  by `\"` and `\\` respectively.
But I think it could be better to simply use a Json for the "array of byte" to hash...
Indeed, the Json format is standard whereas what we are doing here is prone to errors as we create the Json array by hand.

```suggestion
For example, `{1:"firstString",2:"secondString",...}` 
```

wouewouel (changes requested):

README.md
SFulpius: Why do you want to use a JSON object instead of an array. What is the advantage?
Also, in your example you still need to escape strings, otherwise you could have problems. By example, `Hash(string1||string2",3:"string3) = Hash({1:"string1",2:"string2",3:"string3"}) = Hash(string1||string2||string3)`

wouewouel: You're right, it's not a byte array so we definitely need an escape function.
I'm OK with this now, thanks

wouewouel (approved):

celinecamacho (approved):

README.md
```diff
@@ -24,3 +24,7 @@ Public Key: base64
 Signature: base64  
 Hash: base64  
 Timestamp: uint64 representation of the Unix timestamp (seconds since January 1st, 1970)  
+
+## Concatenation for hashing
+When concatenating strings for hashing, `"` and `\` characters must be escaped by adding a `\` characters before them. `"` becomes `\"` and `\` becomes `\\`. Then each string is put between two `"` and added in a JSON array using compact representation.
```
celinecamacho, line 29: > Then each string is put between two `"`

Just to be sure, do we use only `"` or its escaped character `\"`?

README.md
romain46: okay, so I just don't understand why we want to escape characters that are not present in b64. I mean the strings that are going to be in this are base64-encoded right ?

GabrielFleischer (approved):

README.md
romain46: okay I thought we used the base64 encoded version but obviously not, so it's alright for me

romain46 (approved):

Issue #98, opened 2020-12-02
TODO fe-1 milestones


RaulinN:
# TODO fe-1 milestones

## => Added 03-12-20

- [x] finish requests for non-"`Create<...>`" requests
- [x] store last connected LAO information in local storage 
- [x] update requests to store information in local storage
- [ ] "handle callbacks" for each type of messages answers (~LAO messages~, ~Witness messages~, Meeting messages) `(1)` _(waiting on local storage for meetings)_
- [x] "handle callbacks" for `PropagateMessage` `(1)`
- [x] store public key in a local storage
- [ ] store secret key... somewhere? _(stored with public key for now)_ `(2)`

---
## => Added 09-12-20

- [x] `generateId` doit générer des ids increasingly et pas randomly `(0)`
- [x] implement protocol changes from #96 `(0)`
- [ ] implement protocol changes from #99 `(0)` _(waiting on local storage for `dataStateLao` and `dataStateMeeting` (+ `messageGeneral`)_
- [x] add roll-call messages (once #101 is resolved) `(1)`
- [x] more unit tests/documentation (websocket folder targeted) `(2)`

---
## => Added 20-12-20

- [ ] update readme according to PR #121 `(2)`
- [ ] add JsonSchema validation `(1)` _(half done, need to merge next PR to wire it all up)_

---
## => Additional work if we have time

- ~`tweetnacl` seem to not work on native: investigate~ `(3)`
- ~`localStorage` seem to not work on native: investigate~ `(3)`

pierluca:
Closing this issue as it belonged to the Fall 2020 project.

PR #147, opened 2020-12-21
fe-1 milestone advancements issue 98

RaulinN:
Overview of the addition (see #98 for more information):

- improved jsonrpc id generation (increasingly instead of pseudo-randomly)
- fixed a bug where the client would sent an infinite amount of queries if the server answered negatively
- client can now handle receiving a propagation message from the server
- added json schema validation stub
- added roll-call message handling
- unit tests stub


imaximumGit (approved):
Good job

romain46 (approved):

PR #151, opened 2020-12-21
Occurence of lastModified Removed

wouewouel:
This parameter "lastModified" was obsolet but still in some querries of the schema


SFulpius:
Can you explain why you think last_modified is not used anymore?

wouewouel:
Oups, my bad :)
It was removed for the creation querries but indeed we still may want to have it for the state querries.

PR #152, opened 2020-12-21
Move connexion and reconnexion process into the proxy

GabrielFleischer:
The connexion was handled manually in PoPApplication. 
It is moved inside LowLevelClientProxy. 

This makes the sessions easier to handle as they are only one object with a mutable session.
It also generalise the reconnexion process to any session.


AlexisTabin (approved):

app/src/main/java/com/github/dedis/student20_pop/utility/network/LowLevelClientProxy.java
```diff
@@ -73,11 +82,32 @@ public LowLevelClientProxy(Session session) {
...
 
         return entry.requests.thenApply(elem -> gson.fromJson(elem, responseType));
     }
 
+    private void refreshSession() {
+        synchronized (SESSION_LOCK) {
+            if(sessionFuture.isDone()) {
+                Optional<Session> session = getSessionNow();
+                if (!session.isPresent() || !session.get().isOpen())
+                    //There was an error during competition, retry
```
celinecamacho, line 105: Maybe adding a Log.d with that message could be helpful for debugging

celinecamacho (approved):
Great, just one suggestion

AlexisTabin (approved):

PR #153, opened 2020-12-21
RollCall + bug fixes

RaulinN:


src/main/scala/ch/epfl/pop/json/JsonMessageParser.scala
```diff
@@ -51,26 +51,49 @@ object JsonMessageParser {
           obj.getFields("method") match {
             case Seq(JsString(m)) => m match {
               /* Subscribing to a channel */
-              case Methods.Subscribe() => Left(obj.convertTo[SubscribeMessageClient])
+              case Methods.Subscribe() => return Left(obj.convertTo[SubscribeMessageClient])
```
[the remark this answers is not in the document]
RaulinN: Good question! Since we added the ability for the server to parse server messages in e1dc5c1 following the little backend meeting we did, the server should now be able able to parse two types of messages:
- client queries including a `method` field in the message (see line 51 you highlighted)
- answer messages (positive or negative)

The second bullet point corresponds to lines 74-96 of commit e1dc5c1 and is situated *just after* the part you highlighted. Thus, if you parse a query message without the `return`s, then whatever happens, the line will be interpreted as an expression instead of a return value. Finally, it would throw an Exception (l.95).

Did I answer your question @SFulpius ?

src/test/scala/ch/epfl/pop/tests/json/JsonMessageParserTest.scala
```diff
@@ -27,6 +30,19 @@ class JsonMessageParserTest extends FunSuite with Matchers {
...
 
+      @scala.annotation.tailrec
+      def checkListOfKeySigPair(l1: List[KeySignPair], l2: List[KeySignPair]): Unit = {
+        l1.length should equal (l2.length)
+
+        (l1, l2) match {
+          case _ if l1.isEmpty =>
+          case (h1 :: tail1, h2 :: tail2) =>
+            h1.witness should equal (h2.witness)
+            h1.signature should equal (h2.signature)
```
haoqianzhang, line 41: Interesting. Really human-readable testing.

src/test/scala/ch/epfl/pop/tests/json/JsonMessageParserTest.scala
```diff
@@ -27,6 +30,19 @@ class JsonMessageParserTest extends FunSuite with Matchers {
...
+            checkListOfKeySigPair(tail1, tail2)
+        }
+      }
+
 
       val o1 = this.m
       val o2 = o
```
haoqianzhang, line 48: What is o1 and o2?

haoqianzhang (approved):
I am not sure that using so much replaceAll in testing is a good practice. But, yeah, it is comprehensive testing.

src/main/scala/ch/epfl/pop/json/JsonMessageParser.scala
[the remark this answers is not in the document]
SFulpius: Yes, I see why you use it. But I'm not so fan of using return in this case as it makes the control flow of the program hard to understand. You could do something like that, which I think is more readable.
```scala
val fields = obj.fields
if (fields.contains("method")) {
    ...
}
else if (fields.contains("error")) {
    ...
}
...
else {
    ...
}
```

SFulpius (approved):
Except for the remark I made, LGTM.

src/main/scala/ch/epfl/pop/json/JsonMessageParser.scala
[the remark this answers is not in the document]
RaulinN: I hesitated a long time between what I wrote and what you just proposed. The problem I had with the if-else version is that is requires us to check basically twice for "method".

However, looking at it once more, I agree that the `if-else` version is more readable at a glance! I will make the change now

src/test/scala/ch/epfl/pop/tests/json/JsonMessageParserTest.scala
RaulinN: Yes! Very good question! The important parts around those two lines resemble this:

```scala
implicit class RichJsonMessage(m: JsonMessage) {
    def shouldBeEqualUntilMessageContent(o: JsonMessage): Unit = {

        def checkListOfByteArray(...) { /* ... */ }
        def checkListOfKeySigPair(...) { /* ... */ }

        val o1 = this.m
        val o2 = o

        o1 match {
            // actual tests
        }
    }
}
```

The whole point of the `RichJsonMessage` implicit class is to implicitly transform a `JsonMessage` into a `RichJsonMessage`. This way, I can use the method `shouldBeEqualUntilMessageContent` on a `JsonMessage` without adding this method in the real code (not test cases).

Thus, when later I use

```scala
m1 shouldBeEqualUntilMessageContent m2
// m1 and m2 are both standard JsonMessage
```

then `m1` gets implicitly extended into its Rich version (argument `m` of `RichJsonMessage`) and the method is called on it using `m2` as argument (thus `o` is `m2`)

To sum it all up, we have

```scala
val o1 = this.m    // corresponds to m1 due to implicit class conversion (parameter of the class)
val o2 = o         // corresponds to m2 (parameter of the method)
```

haoqianzhang: You may have better names for them.

RaulinN: Alrigth! I will merge this PR and do this in a little PR addressing bug fixes tonight or tomorrow. If I add more code, the action that we pushed will now be run on this code and it will fail

PR #156, opened 2020-12-22
Change concatenation as described in #133

SFulpius:


src/test/scala/ch/epfl/pop/tests/MessageCreationUtils.scala
```diff
@@ -6,20 +6,22 @@ import spray.json.enrichAny
...
+  def b64EncodeToString(b: Array[Byte]): String = Base64.getEncoder.encodeToString(b)
+
   def getMessageParams(data: MessageContentData, pk: PublicKey, sk: PrivateKey, channel: ChannelName): MessageParameters = {
     val encodedData = b64Encode(data.toJson.compactPrint.getBytes).map(_.toChar).mkString
     val signature = Curve25519.sign(sk, encodedData.getBytes())
     val md = MessageDigest.getInstance("SHA-256")
-    md.update(encodedData.getBytes())
-    md.update(signature)
-    val messageId = md.digest()
+    val id = "[\"" + encodedData + "\",\"" + b64EncodeToString(signature) + "\"]"
```
pierluca, line 23: since you already have hashing logic in crypto/Hash.scala, supporting a variable number of parameters and what not, why repeat yourself?

The concatenation and hashing is somewhat error prone, you might just want to have a generic function with a variable number of arguments, returning the hash for just 1 parameter, or the proper concatenation for 2+ parameters.

This way, you only have 1 function to test and it'll be correct everywhere. Plus, if it's tested and known to be correct, you could even rely on it in the tests such as PubSubTest.scala. 

Furthermore, if the hashing function or the concatenation logic needs to change tomorrow, you'll only have to change 1 place in the code.

pierluca (changes requested):
This could work, but the repetition looks like a code smell.

pierluca:
Just ping me if/when you need me to approve this :)

SFulpius:
@pierluca 
Thank you for your feedback. I updated it.

src/main/scala/ch/epfl/pop/crypto/Hash.scala
```diff
@@ -15,10 +16,8 @@ object Hash {
    * @return the id of the message
    */
   def computeMessageId(msg: Base64String, signature: Signature): Hash = {
-    val md = getMessageDigest()
-    md.update(msg.getBytes(StandardCharsets.UTF_8))
-    md.update(signature)
-    val id = md.digest()
+    val data = arrayRepr(msg, Base64.getEncoder.encodeToString(signature))
```
RaulinN, line 19: You could also store the reference of the encoder/decoder created using `Base64.getEncoder` and `Base64.getDecoder`. 
As you can see in the [Java doc](https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html), the creation of such object is a static method, but any encoding/decoding using `encodeToString(_)` for example is not static. This means that another instance of the object is created everytime

src/main/scala/ch/epfl/pop/crypto/Hash.scala
```diff
@@ -43,15 +42,16 @@ object Hash {
 
 
   private def computeGenericId(a: Array[Byte], l: Long, s: String): Hash = {
-    val md = getMessageDigest()
-    md.update(a)
-    md.update(l)
-    md.update(s.getBytes)
-    val id = md.digest()
+    val data = arrayRepr(Base64.getEncoder.encodeToString(a), l.toString, s)
```
RaulinN, line 45: same issue here

RaulinN (changes requested):

src/main/scala/ch/epfl/pop/crypto/Hash.scala
SFulpius: So the concern is that we are creating a new object each time, right? Actually, if you look at source code, you can see that the same object is returned by every call to getEncoder. However, I'm not sure if it is a good idea to rely on this as it is internal detail and is subject to change.

RaulinN (approved):

romain46 (approved):

pierluca (approved):
Not re-reviewed, but approving to unblock

PR #157, opened 2020-12-22
Update the protocol implementation to fit with the latest specifications

GabrielFleischer:
- Refactor class names to fit with the schema
- Change Message query to Broadcast query
- Update Hash concatenation


app/src/main/java/com/github/dedis/student20_pop/utility/security/Hash.java
```diff
@@ -14,26 +15,53 @@
...
+    private static final char DELIMITER = '\"';
+
+    /**
+     * Hash some objects using SHA256.
+     * Concatenate the object's string representation following the protocol's directive.
+     * Then hash the obtained string
+     *
+     * @param data : the objects to hash
+     * @return the hashed data or null if failed to hash
+     *
```
celinecamacho, line 27: You don't need the extra space between return and throws

app/src/main/java/com/github/dedis/student20_pop/utility/security/Hash.java
```diff
@@ -14,26 +15,53 @@
...
      * Hash a data using SHA256.
      *
      * @param data to hash
      * @return the hashed data or null if failed to hash
      * @throws IllegalArgumentException if the data is null
      */
-    public static String hash(String data) {
-        if (data == null) {
+    protected static String hash(String data) {
+        if (data == null)
```
celinecamacho, line 54: In our coding conventions we always put the opening and closing brackets, even if it's just one line of code

app/src/main/java/com/github/dedis/student20_pop/utility/security/Hash.java
```diff
@@ -14,26 +15,53 @@
...
+            throw new IllegalArgumentException("Can't hash a null data");
+
+        StringJoiner joiner = new StringJoiner(",", "[","]");
+        for(Object elem : data)
+            joiner.add(DELIMITER + sanitize(elem.toString()) + DELIMITER);
+
+        return hash(joiner.toString());
+    }
+
+    private static String sanitize(String input) {
```
celinecamacho, line 41: It could be good to rename the method `esc` to follow the specification (#133)

app/src/test/java/com/github/dedis/student20_pop/utility/security/HashTest.java
```diff
@@ -2,18 +2,27 @@
...
     @Test
     public void hashTest() {
         assertNotNull(Hash.hash("Data to hash"));
     }
+
+    // Hashing : test 0 \fwa"fwa-fwa
+    // Expected : ["test","0","\\fwa\"fwa-fwa"]
+
+    @Test
+    public void objectsHashingWorks() {
```
celinecamacho, line 25: I would suggest to rename the method `hashObjectsTest` to follow the conventions. Also the comments would be better inside the method.

celinecamacho (changes requested):

app/src/main/java/com/github/dedis/student20_pop/utility/security/Hash.java
GabrielFleischer: Done

app/src/test/java/com/github/dedis/student20_pop/utility/security/HashTest.java
GabrielFleischer: Done !

celinecamacho (approved):
Good

jbmichel14 (approved):

PR #159, opened 2020-12-24
Origin/work protospec ouriel

wouewouel:
Following the last meeting, we said that the Lao creation was only between the front end and the back end of the organiser.
It means that asking witnesses to sign will come after that on update Lao. 
So we don't need anymore to have a list of witness public keys on the dataCreateLao and that's what I just deleted.

### DIFF
--- a/query/method/message/data/data.json
+++ b/query/method/message/data/data.json
@@ -14,7 +14,6 @@
 		},
 		"name": { "type": "string" },
 		"creation": { "description": "[Timestamp] creation time", "type": "integer", "minimum": 0 },
-		"last_modified": { "description": "[Timestamp] last modification's time", "type": "integer", "minimum": 0 },
 		"organizer": { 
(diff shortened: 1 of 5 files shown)


pierluca:
I believe there's a misunderstanding here. 
The list of witnesses is part of the configuration of the LAO, why would you want to remove it?

If on the other hand you meant to remove the list of witnesses from the message layer, then I'd suggest leaving it empty: that way the message layer is still standard across all messages and you don't have to complicate your application logic.

romain46:
mhhhh I agree with @pierluca. It's an initialisation parameter, we don't need to use it from the beginning does not mean we want to remove it. It's better to let them in the implementation.

wouewouel:
Oh yeah, I confused the two layers my bad.
So same trick used with witnessMessage and just ignore the witnessSignatures.
Thank you for correcting me !

PR #165, opened 2020-12-24
Fix subscribe to LAO after QR scanning

celinecamacho:
## Problem
During the Roll-Call the organizer has to scan the QR code of the soon to be attendees. Once the QR code has been scanned, the soon to be attendee has no way to know the channel's id he wants to subscribe to, nor the back end's server address to connect to.

schema

## Solution
 The following image shows in black the existing connection and in orange the needed connection. We suggest that the organizer's front end sends through a new connection, shown in green, the information the attendee needs to send a subscribe request over the orange connection (and eventually to open it in case it did not existed before). 

The message the organizer front end sends through that socket the message defined in `answer/scanQRAck.json`. This message contains the LAO channel's id, server's address, and the standard JsonRPC fields (jsonrpc and id). The websocket should be closed by the attendee after receiving this message.

Of course for this solution to be possible, the attendee has to provide, in his QR, his IP address and port to open the green connection.

This is just a quick fix and is meant to be improved and worked on in the future. In order to implement the Roll-Call as soon as possible, we need a quick solution.

### DIFF
--- a/answer/scanQRAck.json
+++ b/answer/scanQRAck.json
@@ -1,0 +1,25 @@
+{
+	"$schema": "http://json-schema.org/draft/2019-09/schema#",
+	"$id": "https://github.com/dedis/student20_pop/tree/proto-specs/answer/scanQRAck.json",
+	"title": "Match a custom JsonRpc 2.0 message answer",
+	"description": "Match a scan QR acknowledgement",
+	"type": "object",
+
+	"properties": {
+		"jsonrpc": { "description": "[String] JsonRpc version", "const": "2.0", "$comment": "should always be \"2.0\"" },
+		"address": { "description": "[String] address of server", "const": "2.0", "$comment": "Organizer's back end server address" },
+		"channel_id": { "description": "[String] channel id", "const": "2.0", "$comment": "Channel id to be subscribed to" },
+		"id": {
+			"oneOf": [
+				{ "type": "integer" },
+				{ "type": "null" }
+			],
+			"$comment": "The id match the request id. If there was an error in detecting the id, it must be null"
+		      }
+	},
+	"additionalProperties": false,
+	"allOf": [
+		{ "required": ["jsonrpc", "address", "channel_id", "id"] }
+	],
+	"$comment": "Note: required fields: (\"jsonrpc\" \\and \"address\" \\and \"channel_id\" \\and \"id\")"
+}
--- a/query/method/message/data/dataCloseRollCall.json
+++ b/query/method/message/data/dataCloseRollCall.json
@@ -10,7 +10,7 @@
 		"id": {
 			"type": "string",
 			"contentEncoding": "base64",
-			"$comment": "Hash : SHA256('roll_call'||lao_id||creation||name)"
+			"$comment": "Hash : SHA256('R'||lao_id||creation||name)"
 		},
 		"start": { "description": "[Timestamp] start time", "type": "integer", "minimum": 0 },
 		"end": { "description": "[Timestamp] end time", "type": "integer", "minimum": 0 },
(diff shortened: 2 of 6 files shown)


answer/scanQRAck.json
```diff
@@ -0,0 +1,25 @@
...
+				{ "type": "integer" },
+				{ "type": "null" }
+			],
+			"$comment": "The id match the request id. If there was an error in detecting the id, it must be null"
+		      }
+	},
+	"additionalProperties": false,
+	"allOf": [
+		{ "required": ["jsonrpc", "address", "channel_id", "id"] }
+	],
```
pierluca, line 23: you don't need allOf, if you're just using required.

```suggestion
	"required": ["jsonrpc", "address", "channel_id", "id"],
```

answer/scanQRAck.json
```diff
@@ -0,0 +1,25 @@
+{
+	"$schema": "http://json-schema.org/draft/2019-09/schema#",
+	"$id": "https://github.com/dedis/student20_pop/tree/proto-specs/answer/scanQRAck.json",
+	"title": "Match a custom JsonRpc 2.0 message answer",
+	"description": "Match a scan QR acknowledgement",
+	"type": "object",
+
+	"properties": {
+		"jsonrpc": { "description": "[String] JsonRpc version", "const": "2.0", "$comment": "should always be \"2.0\"" },
+		"address": { "description": "[String] address of server", "const": "2.0", "$comment": "Organizer's back end server address" },
```
pierluca, line 10: I'm quite sure this is not valid JSON RPC

pierluca (changes requested):
Pointed out a few low-level mistakes in the protocol, but I don't think those are the main issue.

As was pointed out during the last meeting by a student (following a mistake I made, which I think is very much aligned with your thinking), the roll-call is not necessary to join a LAO, it is necessary to acquire the Proof-of-Personhood within the LAO. As such, you can safely assume that the attendees have already joined in the LAO before getting QR-scanned.

Specifically: roll call participants are presumed to be already part of the LAO, just not roll-call'd yet. As such, they receive messages on the LAO main channel.
When a roll call is created, it has its own ID (part of the creation message), which is visible in the LAO main channel.
Should you want to have messages specific to the roll call itself in a channel separate from the LAO main channel, you could use the roll call ID as a separate chanel, but that too would be visible to any clients willing to connect to it.

Have I missed something ?

answer/scanQRAck.json
```diff
@@ -0,0 +1,25 @@
...
+		"channel_id": { "description": "[String] channel id", "const": "2.0", "$comment": "Channel id to be subscribed to" },
```
RaulinN, line 11: The `"const": "2.0"` stands for "this field can only hold value "2.0". Thus this should only be used for `jsonrpc` field

answer/scanQRAck.json
RaulinN: See this [link](https://www.jsonrpc.org/specification) guys. The problem is that (for example) `address` should be located inside the `result` object

RaulinN: If you change the file to conform to json rpc, you should be able to add one `$ref` in the file `answer/answer.json`  and define this object (`scanQRAck`) inside the `positiveAnswer` folder

If, for some reason, this cannot be done and you have to keep this file directly in the `answer` folder, then you will also need to modify the top-level file to include a `$ref` to this file

answer/scanQRAck.json
```diff
@@ -0,0 +1,25 @@
+{
+	"$schema": "http://json-schema.org/draft/2019-09/schema#",
+	"$id": "https://github.com/dedis/student20_pop/tree/proto-specs/answer/scanQRAck.json",
+	"title": "Match a custom JsonRpc 2.0 message answer",
```
RaulinN, line 4: The title is the same as the file `answer/answer.json`. Some JsonValidators will display the title when debugging. You should probably correct it so that we can distinguish the two files :)

RaulinN (changes requested):

PR #166, opened 2020-12-25
Display dates/times as human readable dates/times instead of long numbers

jbmichel14:


GabrielFleischer (approved):

app/src/main/java/com/github/dedis/student20_pop/utility/ui/EventAdapter/ExpandableListViewEventAdapter.java
```diff
@@ -21,6 +23,7 @@
     protected final Context context;
     protected final List<Event.EventCategory> categories;
     protected final HashMap<Event.EventCategory, List<Event>> eventsMap;
+    protected static final SimpleDateFormat DATE_FORMAT = new SimpleDateFormat("EEE, dd MMM yyyy HH:mm", Locale.US);
```
lhmerino, line 26: - This date format is different from the previous one - could you please explain the significance for this difference? 
- Also I think adding the timezone might be important as well. 
- Could you please explain the Locale.US argument?

lhmerino (changes requested):

AlexisTabin (approved):

app/src/main/java/com/github/dedis/student20_pop/utility/ui/EventAdapter/ExpandableListViewEventAdapter.java
jbmichel14: For the format, I figured it was better to have a more precise time (with time and day of the week) in the case of displaying events, while just the date as "dd/MM/yyyy" was sufficient for the list of LAO's. But this is completely arbitrary. Should I follow a specific format to make it more consistent?

As for the Locale.US, it's to specify a language in which the date is display, and I figured that as the rest of the app is in english, I would use dates displayed in english but this is also arbitrary.

I will add the timezone.

lhmerino: Sounds good! Any format will do, was just wondering the reasoning behind it.

lhmerino (approved):

PR #167, opened 2021-01-04
fe-1 connection of UI to the network part

imaximumGit:
- Connect the creation of a meeting and on roll-call to the network part
- Manage the store of events to have the good structure for the view
- Show only the tab corresponding to the role of the user in the organization UI
- Modify the date time picker to work on the web
- Update the prop-type

Seem a big one but all the commit off @RaulinN have already be approved in the PR #147


lhmerino:
Nice checks! Is it possible to differentiate which parts are from @RaulinN and which parts are from @imaximumGit under File Changes? If @RaulinN parts are in fe1-web then those commits by @RaulinN won't be made as part of differences.

imaximumGit:
@lhmerino. As the commits are already in the fe1-web branch, there are not shown in the file diff. I don't realize this before. The PR makes about 700 new lines as 7000 lines come from a binary file.

RaulinN:
Hey @imaximumGit ! 

Before reviewing the PR, I wanted to open the project to check that everything was indeed working as intended. It seems I unfortunately get a billion warnings in the JS console. They seem to all be related to the "UI part" of the frontend. Have you encountered them as well?

- `Warning: Failed prop type: The prop 'lao.object' is marked as required in 'AppNavigation', but its value is 'undefined'`. This warning appears multiple times, always mentioning the `lao.object` prop.
- `Warning: Cannot update a component ('ForwardRef(BaseNavigationContainer)') while rendering a different component ('Attendee').`
- `Warning: Encountered two children with the same key, '.$'. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.`. This appears when loading the app AS WELL as when sending a query to the mock server (at least a requestCreateLao) which is strange... 🤔  This definitely wasn't the case before. Maybe [this link](https://sentry.io/answers/defining-proper-key-in-props/) could help.
- `measureLayoutRelativeToContainingList threw an error measureLayoutRelativeToContainingList@http://localhost:19006/static/js/bundle.js:145113:109`. This warning appears an insane amount of times.

The app is unfortunately not working on my end at least...

RaulinN:
@imaximumGit  to easily apply @lhmerino, you could use the git cherry pick command to cherry-pick [multiple commits at once](https://stackoverflow.com/questions/1670970/how-to-cherry-pick-multiple-commits). This would imply creating a new branch from fe1-web and cherry-picking commits from our branch.

Note that cherry-picking won't work for one of your commits: the one where you merge stuff. Just run two cherry-pick instead of one.

imaximumGit:
@RaulinN. I will think about it the next time. Maybe in this situation, I should have used a git rebase instead of a git merge

RaulinN:
Nice work! I still have two more questions!

- You managed to correct the `Warning: Encountered two children with the same key` bug. However a new one appeared. Last time in private message, I could tell you exactly in which file and at what line the bug occurs. This time, the stack trace is stranger. You can reproduce the bug by simply clicking on one of the LAO buttons. [url](https://www.mmbyte.com/article/2343.html)
- Do you know why the navbar (Home & Launch) is shifted when clicking on a LAO? Here's a screenshot:
Screenshot 2021-01-06 at 16 46 14

imaximumGit:
@RaulinN 

- For the first question, the bug doesn't appear on my computer. Maybe this comes from a deprecated data store in the persisted store. Make sure to update the connectLAOsReducer with the file laoData.js and currentEventsReducer with the file EventData.js
- For the second, if I have understood correctly. For some obscure reason, the text _HOME_ and _LAO 1_ in this example does not apply their rule of placement inside the tab bar or the height of their cell is not the same as the tab bar and produce this bug.

RaulinN:
> @RaulinN
> 
>     * For the first question, the bug doesn't appear on my computer. Maybe this comes from a deprecated data store in the persisted store. Make sure to update the connectLAOsReducer with the file laoData.js and currentEventsReducer with the file EventData.js
> 
>     * For the second, if I have understood correctly. For some obscure reason, the text _HOME_ and _LAO 1_ in this example does not apply their rule of placement inside the tab bar or the height of their cell is not the same as the tab bar and produce this bug.

Alright! I will try applying the suggestion once everything is merged and the last PR with the few tweaks (that I should push soon) is merged :)

RaulinN (approved):

PopApp/Components/CreateMeeting.js
```diff
@@ -75,6 +95,9 @@ const CreateMeeting = () => {
     if (event.type === 'set' && mode === 'date') {
       showTimepicker();
     }
+    if (event.type === 'set' && mode === 'time' && finishDate && finishDate <= startDate) {
```
lhmerino, line 98: Just want to double check, finishDate should be less than or equal to startDate?

lhmerino (approved):
Looks good! Just one double check comment

PopApp/Components/CreateMeeting.js
imaximumGit: Less than because I don't want the user to set a finishDate smaller than the startDate

PR #169, opened 2021-01-04
Specify that hashes are made out of base64 encoding

Huntsekkern:
**tl;dr**: proposition: hashes should now be made on arrays of base64 encoded strings which look like `["Base64{Foo}", "Base64{Bar}"]` instead of escaping the `"` and `\` within Foo and Bar

---

Hey everyone! 

While performing tests between backend-1 and frontend-1, we stumbled upon an issue that could lead to serious troubles later (possibly next semester). Here's a quick description of the issue:

When hashing the _exact same string_, Go and JavaScript do not yield the same value for very specific cases: escaped characters. We introduced this "bug" in PR #133 when we decided to escape any `"` (quote character) and `\` (backslash character) into `\"` and `\\` respectively.

If either fe1 or be1 decides to adopt the other team's solution, we don't know how fe2 and be2 would react to these changes. Moreover, if be3 is added next semester, those might encounter the exact same problem. Thus, here's our solution to the problem:

Current implementation:
```java
public String hashStrings(Array[String] strs) {
	// transform each string input in base64
	for (String str : strs) { str = escapeString(str); }
	// creates the array in string form
	String arr = ... // result: arr = ["abc", "\"time\":2"]
	// hash the array in string form
	String hashed = hash(arr);

	// return the base64 representation of the hash
	return base64Encode(hashed);
}

// transforms any `"` into `\"` and any `\` into `\\`
public String escapeString(String str) { /* ... */ }
```

Proposed implementation:
```java
public String hashStrings(Array[String] strs) {
	// transform each string input in base64
	for (String str : strs) { str = base64Encode(str); }
	// creates the array in string form
	String arr = ... // result: arr = ["abc=", "bde="]
	// hash the array in string form
	String hashed = hash(arr);

	// return the base64 representation of the hash
	return base64Encode(hashed);
}
```

Note that only the `for-loop` is modified. This way, we will never have to escape characters since every string is encoded in base64.

:information_source: with our solution, **any** string used for hashing is converted in base64 (for-loop). Another solution could be only to encode a subset of those strings. For example, the creation time (`int`) will never contain any problematic characters. The same could be said for `R` and `M` (bytes used for roll-calls and meetings when hashing).

:information_source: frontend-1 is working in a JavaScript environment while backend-1 is working in Go.

### DIFF
--- a/query/method/message/README.md
+++ b/query/method/message/README.md
@@ -30,3 +30,6 @@ As you can see, the data field is base64 encoded. This derives from the necessit
 
 As a consequence, we decide to encode the sender's preferred representation (any valid representation) in base64, and then use that representation as input to our hash or sign function. A nice side effect of this design is that the message object signature verification is entirely independent from the structure of the data field itself: no matter what message you send, the verification will always work in the same way (compare this to signing the data field-by-field!).
 Once the data field is unencoded and parsed, the receiving peer can simply validate it and process it at the application level.
+
+## Hashes
+Hashes are made on arrays of base64 encoded strings which look like that ["Base64{Foo}", "Base64{Bar}"]. This is necessary to avoid some issues with inner "
--- a/query/method/message/README.md
+++ b/query/method/message/README.md
@@ -13,20 +13,24 @@ The message data specific to the operation is encapsulated in the data field, wh
 Origin: sender field
 Authenticity: signature and witness_signatures fields
 Uniqueness: message_id field
+```json
 {
  "data": base64({}),  /* base64 representation of an object */
  "sender": "0x123a", /* Public key of sender */
  "signature": "0x123a", /* Signature by sender over "data" */
- "message_id": "0x123a", /* hash(data||signature) */
- "witness_signatures": [],/* Signature by witnesses (sender||data) */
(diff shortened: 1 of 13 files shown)


wouewouel (approved):

lhmerino:
Thank you for the detailed explanation. Is there a specific reason why you decided to escape the characters in the first place?

Huntsekkern:
> Thank you for the detailed explanation. This looks good to me. Is there a specific reason why you decided to escape the characters in the first place?

The detailed explanation is fully to be credited to @RaulinN :)

Escaping the characters was decided in PR #133 as per @pierluca quote: (and he stands by this decision in our slack thread)
> The limitations of JSON (as a format without canonical representation) were discussed at length throughout this project.
In this particular case, we settled (on Bryan's recommendation) on something that 'looks' like JSON but is stricter: an array of UTF-8 encoded strings (or numbers), with some characters escaped (", \) and without any whitespace between syntactic elements. Essentially it doesn't require any external dependencies (as opposed to all of the above suggestions) and can be easily implemented 'by hand', by escaping a couple of characters and adding delimiters.

I guess the real advantage lies in "doesn't require any external dependencies"

GabrielFleischer (approved):

SFulpius:
Closing because we choose not to change the specs

PR #170, opened 2021-01-06
Implement Roll Call Event Creation

AlexisTabin:
To implement Roll Call Event Creation, we will add 2 fragments :
 1. One to create a new Roll Call Event
 2. The second one to add attendees to the created (or to be created) roll call event
 3. We will also add a specific layout for displaying a roll-call event in organizer and attendee fragments (Not done in this PR)

Roll Call Event Creation Fragment :
- [x] Show date/time entry field 
- [x]  Below the entry field is a Description text-entry (optional), Confirm and Cancel buttons.
- [x] Open button 

The Confirm button  (when no attendee has been added) creates the roll-call event in the “Future” state, in which attendees cannot (yet) register and the organizer cannot (yet) scan them.  This is intended for roll-calls announced in advance and scheduled for some future time.  In this state, the organizer has an Open button (replacing the Confirm and Cancel buttons in the event entry state) to open the roll-call. (For the moment, you need to add attendees in order to create an Roll Call event)

Roll Call Add Attendee Fragment :
- [x] Show camera
- [x] Participant count
- [x] Result
- [x] Close Roll-Call button

Once opened (by clicking Open either initially or from the Future state event display), a camera display appears for registering participants, a participant count (so far) appears below it, and a message, “Please scan each participant’s Roll-call QR code exactly once.”  Upon successfully scanning and verifying an Attendee’s Roll-call QR code, the camera freezes for a couple seconds and it displays “Attendee scanned successfully” as the participant count increments by one.  If the app scans a roll-call QR code that has already been scanned, it displays “Attendee already recorded” and does not add to the attendee list or increment the participant count.

Finally, at the bottom is a “Close Roll-call” button.  When tapped, it first asks the organizer to confirm, then changes to the Closed state and broadcasts both the new state and the (hopefully) final roll-call.

This won't be done yet : 

In the Closed state, the roll-call display shows the start time, recorded number of participants, and a “Re-open Roll Call” button in case the organizer accidentally closed the roll call (and confirmed it) or something happened in the real world (“Wait, you forgot these people!”) to justify re-opening a closed roll-call.  For transparency, records of each re-opening and re-closing time get recorded indelibly in the underlying log that gets broadcast, witnessed, etc.


app/src/androidTest/java/com/github/dedis/student20_pop/ui/OrganizerFragmentTest.java
```diff
@@ -42,30 +42,64 @@
...
     private final String witness1 = "Alphonse";
     private final String witness2 = "Bertrand";
     private final ArrayList<String> witnesses = new ArrayList<>(Arrays.asList(witness1, witness2));
+    private final int YEAR = 2022;
+    private final int MONTH_OF_YEAR = 10;
+    private final int DAY_OF_MONTH = 10;
+    private final String DATE = "" + DAY_OF_MONTH + "/" + MONTH_OF_YEAR + "/" + YEAR;
+    private final int HOURS = 12;
+    private final int MINUTES = 15;
+    private final String TIME = "" + HOURS + ":" + MINUTES;
```
AlexisTabin, line 69: useless add, need to remove it

app/src/main/java/com/github/dedis/student20_pop/utility/network/HighLevelClientProxy.java
```diff
@@ -141,9 +139,9 @@ public LowLevelClientProxy lowLevel() {
     /**
      * Send an open roll call message
      *
-     * @param laoId id of the lao
+     * @param laoId      id of the lao
      * @param rollCallId id of the roll call
```
GabrielFleischer, line 143: Why isn't rollCallId offset like the rest ?

GabrielFleischer (approved):

app/src/main/java/com/github/dedis/student20_pop/utility/network/HighLevelClientProxy.java
AlexisTabin: What do you mean by that ?

app/src/main/res/layout/fragment_create_roll_call_event.xml
```diff
@@ -0,0 +1,106 @@
...
+        android:layout_height="wrap_content"
+        android:orientation="horizontal"
+        app:layout_constrainedHeight="true"
+        app:layout_constraintTop_toBottomOf="@id/roll_call_guideline_horizontal">
+
+        <EditText
+            android:id="@+id/roll_call_title_text"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:hint="@string/meeting_title_required" />
```
jbmichel14, line 37: The hint says "Meeting title", it would be better if it said "Roll Call Title"

jbmichel14 (changes requested):
I noticed that in the IdentityUI, the QR code is shown only if the user is not anonymous. I don't think that should be the case, as anonymous user must still be able to participate in a roll call.

AlexisTabin:
@jbmichel14 
> I noticed that in the IdentityUI, the QR code is shown only if the user is not anonymous. I don't think that should be the case, as anonymous user must still be able to participate in a roll call.

In the UI Specifications, it is written that :
> The QR code does not appear at all unless Anonymous is un-checked and the Name field is non-empty and not all whitespace.

app/src/main/res/layout/fragment_create_roll_call_event.xml
AlexisTabin: I'll change that, good catch thanks !

jbmichel14 (approved):

PR #177, opened 2021-01-10
Display Participant # for Roll-Call Event

jbmichel14:
- move some duplicated code into the common super class
- display the participant number when the event is a roll call


AlexisTabin (approved):

app/src/main/java/com/github/dedis/student20_pop/utility/ui/eventadapter/ExpandableListViewEventAdapter.java
```diff
@@ -174,7 +177,30 @@ public boolean hasStableIds() {
...
+            convertView = inflater.inflate(R.layout.layout_event, null);
+        }
+        TextView eventTitleTextView = convertView.findViewById(R.id.event_title);
+        TextView descriptionTextView = convertView.findViewById(R.id.event_description);
+        eventTitleTextView.setText(eventTitle);
+        switch (event.getType()) {
+            case ROLL_CALL:
+                eventDescription += ("\nParticipants: " + event.getAttendees().size());
+                break;
+            default:
```
celinecamacho, line 198: Why are we not taking into account the other event types?

jbmichel14: Because they are not yet supported by the protocol, and the priority was the roll call event.

celinecamacho: Ok sounds good

celinecamacho (approved):

PR #178, opened 2021-01-11
Origin/work protospec ouriel

wouewouel:
Hi everyone,

The modification is only one word but I think it's a mistake: witness should Sign(`message_id`) and not  Sign(`data`)

Indeed, the `signature` argument of [dataWitnessMessage ](https://github.com/dedis/student20_pop/blob/proto-specs/query/method/message/data/dataWitnessMessage.json) sign the `data` field, but this signature is going to be added in the list `witness_signatures` of [messageGeneral](https://github.com/dedis/student20_pop/blob/proto-specs/query/method/message/messageGeneral.json), and in that list the witnesses sign the `message_id` field.

The confusion is certainly due to the fact that we sign `data` in the `signature` field of a message.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,3 +24,7 @@ Public Key: base64
 Signature: base64  
 Hash: base64  
 Timestamp: uint64 representation of the Unix timestamp (seconds since January 1st, 1970)  
+
+## Concatenation for hashing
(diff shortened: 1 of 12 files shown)


SFulpius (approved):
As we already discussed, I agree on this issue.

pierluca (approved):
I think you're right.
To guarantee the security, though, this means that the witness server MUST verify (i.e. recompute) the message_id before sending the signature.
Please make sure you have unit tests verifying that you can't get a witness to sign a message_id that doesn't match with the data.

wouewouel:
I merge the PR and I'll post a message on slack

PR #181, opened 2021-01-11
Fix Unix Timestamp

celinecamacho:
Replace `java.util.Date` with `java.time.Instant` for the computation of Unix Timestamp in order to avoid issues with milliseconds.


GabrielFleischer (changes requested):
You should pull the schema submodule. Right now it is downgraded

app/src/main/java/com/github/dedis/student20_pop/utility/security/Hash.java
```diff
@@ -32,8 +32,12 @@ public static String hash(Object... data) {
         }
 
         StringJoiner joiner = new StringJoiner(",", "[", "]");
-        for (Object elem : data)
+        for (Object elem : data) {
+            if (elem == null) {
```
jbmichel14, line 36: What if elem is not null but empty?

jbmichel14 (approved):

app/src/main/java/com/github/dedis/student20_pop/utility/security/Hash.java
celinecamacho: But `elem` is not always a String

AlexisTabin (approved):

GabrielFleischer (approved):

PR #187, opened 2021-01-13
Fix various issues of the json schema. Nothing about the protocol

GabrielFleischer:
There are some problems with the json schema that does not allow proper validation, here are the fix I propose :

- Fix Json syntax issue in dataStateLao.json
- In broadcast and publish, messageGeneral overlaps messageWitnessMessage and the oneOf fails
- Fix path to channels schema in publish
- Add "message" as an object property such that it is not recognized as an additional property
- Change all ids such that the point to a valid json file


answer/answer.json
```diff
@@ -1,6 +1,6 @@
 {
 	"$schema": "http://json-schema.org/draft/2019-09/schema#",
-	"$id": "https://github.com/dedis/student20_pop/tree/proto-specs/answer/answer.json",
+	"$id": "https://raw.githubusercontent.com/dedis/student20_pop/proto-specs/answer/answer.json",
```
SFulpius, line 3: Can you explain why this change was needed?

query/method/broadcast.json
```diff
@@ -15,10 +15,7 @@
 				"channel": { "description": "[String] name of the channel", "$ref": "channel/subChannel.json" },
 				"message": { 
 					"description": "[Message] message to be published", 
-					"oneOf": [
-						{ "$ref": "message/messageGeneral.json" },
-						{ "$ref": "message/messageWitnessMessage.json" }
```
SFulpius, line 20: If you delete this, witnesses will not be able to send witness messages.

query/method/publish.json
```diff
@@ -1,62 +1,63 @@
...
-
-				{
-					"if": { "properties": { "channel": { "$ref": "../channel/subChannel.json" } } },
-					"then": {
-						"properties": {
-							"message": { 
-								"description": "[Message] message to be published",
-								"anyOf": [
-									{ "$ref": "message/messageGeneral.json" },
-									{ "$ref": "message/messageWitnessMessage.json" }
```
SFulpius, line 46: Same as before

answer/answer.json
pierluca: I'm not sure that this is needed, but it does make sense. That way the $id can actually be resolved if some library tries to fetch it, as it points to the file itself rather than Github interface.

query/method/broadcast.json
pierluca: I didn't look into this, but I seem to remember having the same desire to change this.
Are you sure messageWitnessMessage isn't just a subset of messageGeneral? 
You might have answered this to me before, but IIRC messageWitnessMessage is just a messageGeneral with 0 items in the witness_signatures array, which is still a valid messageGeneral

query/method/publish.json
```diff
@@ -1,62 +1,63 @@
...
+					"description": "[String] name of the channel", 
+					"type": "string",
+					"pattern": "^\/root(\/[^\/]+)?$",
+					"$comment": "Note: the regex matches a \"/root\" or a \"/root/<channel>\""
+				},
+				"message": {
+					"description": "[String] message to be published",
+					"type": "object",
+					"$comment": "Note: general property declaration to fulfill the additionalProperties requirements"
+				}
```
pierluca, line 26: how come this was not detected before? Was it unused?

pierluca (commented):
Unsure about some points raised by @SFulpius , but overalls it makes sense 👍 
I won't have time to look further into this, so I just left questions/doubts where I had some.

answer/answer.json
GabrielFleischer: The library we use in fe2 has the abitily to get the latest json directly by downloading it. So it makes more sense that this url points to a valid json file.

This is not needed

query/method/broadcast.json
GabrielFleischer: I might have missed the difference between to messageGeneral and messageWitnessMessage.

But from what I read, messageWitnessMessage is the same as messageGeneral but it forces witness_signatures to be empty.

That if a message has empty witness_signatures, it fulfill both messageGeneral and messageWitnessMessage. So the oneOf breaks

query/method/publish.json
GabrielFleischer: Yes, I was using data and I wrongly though it was the same as generic message with a oneOf [all datas]

answer/answer.json
pierluca: I'd still consider that this is a good change to implement.

HOWEVER, careful with that sort of functionality in fe2, as if the protocol evolves but not your application, you end up with something broken, and the links are not versioned. Your application shouldn't be relying on un-versioned external dependencies that may change at any point. That's why most REST API endpoints (for instance) are designed to remain stable through time and versioned, so that even if they do change, the old version remains available with that exact same specification.

You could rely on the online thing IFF you start creating folders for versions, i.e.
`"$id": "https://raw.githubusercontent.com/dedis/student20_pop/proto-specs/1.0/answer/answer.json",`
but I wouldn't go there ;-)

RaulinN (changes requested):

query/method/publish.json
```diff
@@ -1,62 +1,63 @@
...
+						"properties": {
+							"message": { 
+								"description": "[Message] message to be published",
+								"$ref": "message/messageGeneral.json"
+							}
+						},
+						"$comment": "Note: match a publish query destined for a sub-channel"
+					}
+				}
+			],
```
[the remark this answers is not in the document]
RaulinN: `message` at line 22 should thus look like that in my opinion:

```json
"message": {
	"description": "[Message] message to be published",
	"$ref": "message/messageGeneral.json"
}
```

Moreover, the whole `allOf` part should be removed.

query/method/broadcast.json
pierluca: Oh yes @RaulinN , thanks for the reminder.
That constraint is something you can enforce at the protocol specification level (as it was so far), or at the application level ( as some of the other checks are ) .
Either are valid options with a simple tradeoff: the former makes the protocol much more verbose and the application simpler, the latter makes for a neat, concise protocol but it requires additional logic in the application.

Huntsekkern: I also had the feeling that messageWitnessMessage is just a subset of messageGeneral.

If for some reason, it is needed to keep both, json schema also offers anyOf we could use instead of anyOf so that the oneOf doesn't "break"

RaulinN (approved):

PR #190, opened 2021-01-13
Add Swipe-to-Refresh

jbmichel14:
- add swipe-to-refresh in Attendee, Organizer and Home UIs so they can be updated when a LAO or event is added/removed/modified
- create modifyLao in PoPApplication
- create a method that populates the data that we can uncomment when testing without a back-end (useful for testing the UI) (the dummy data was already there, it just makes it more clear)
- it's a detail but now, there is a clear distinction between past and present events, and their category (past, present, future) depends on their start times, not creation times
- fix bugs (AddAttendeeFragmentTest was broken)

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -65,6 +65,9 @@
     <string name="end_time_before_start_time_not_allowed">End time must be after start time</string>
     <string name="end_time_optional">End Time</string>
     <string name="picker_selection">selection</string>
+    <string name="past_events">Past Events</string>
+    <string name="present_events">Present Events</string>
+    <string name="future_events">Future Events</string>
 
     <!-- Meeting Event -->
     <string name="meeting_event">Meeting Event</string>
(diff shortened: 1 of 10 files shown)


AlexisTabin (approved):

app/src/main/res/values/strings.xml
```diff
@@ -65,6 +65,9 @@
     <string name="end_time_before_start_time_not_allowed">End time must be after start time</string>
     <string name="end_time_optional">End Time</string>
     <string name="picker_selection">selection</string>
+    <string name="past_events">Past Events</string>
+    <string name="present_events">Present Events</string>
+    <string name="future_events">Future Events</string>
```
celinecamacho, line 70: Where do you use these values?

celinecamacho (commented):
Good, just don't know if the string values for past, present and future are needed

app/src/main/res/values/strings.xml
AlexisTabin: They are used in some tests

celinecamacho (approved):

PR #195, opened 2021-01-17
Improve data management of PoPApplication

GabrielFleischer:
The way data was saved was not very consistent. The aim of this pull request is to fix it.


app/src/main/java/com/github/dedis/student20_pop/PoPApplication.java
```diff
@@ -203,6 +170,15 @@ public HighLevelProxy getProxy(URI host) {
...
     }
 
+    /**
+     * Add a new LAO to the app
+     *
+     * @param lao to add
+     */
+    public void createLao(Lao lao) {
+        laos.put(lao.getId(), lao);
+    }
```
jbmichel14, line 180: Should it also add the lao id to the Person's list of LAOs ? if not, where is it done?

app/src/main/java/com/github/dedis/student20_pop/PoPApplication.java
```diff
@@ -120,29 +117,22 @@ public Person getPerson() {
...
     }
 
     /**
-     * Returns list of LAOs corresponding to the user
+     * @return the current LAO, unsafe
      */
-    public List<Lao> getLaos() {
-        return new ArrayList<>(laoEventsMap.keySet());
+    public Lao getCurrentLaoUnsafe() {
+        return currentLao;
```
jbmichel14, line 128: Is this one kept only for the tests? if so, maybe add a comment

app/src/main/java/com/github/dedis/student20_pop/model/event/Event.java
```diff
@@ -9,7 +9,7 @@
 import java.util.Objects;
 
 /**
- * Class modeling an Event
+ * Class modeling an abstract Event
```
jbmichel14, line 12: Yes, but the class itself isn't abstract (yet). It could be, as long as we change the types and use particular events in PoPApplication's dummyEvents. Doesn't have to be done in this PR but I think it should be done some time.

jbmichel14 (commented):
If it doesn't break anything, I'm fine with it. nice work

app/src/main/java/com/github/dedis/student20_pop/model/event/Event.java
GabrielFleischer: Yes I wanted to change it to abstract, changed  the doc, faced too many problems, rollback but forgot the doc

app/src/main/java/com/github/dedis/student20_pop/PoPApplication.java
GabrielFleischer: It is done at Lao creation.

We do not really use the Model Person yet, so it is not really important

app/src/main/java/com/github/dedis/student20_pop/PoPApplication.java
GabrielFleischer: No, it is used whenever we are sure the current Lao is not null. Like in the Organizer and Attendee UI

jbmichel14: right, I saw that afterwards

jbmichel14 (approved):

app/src/main/java/com/github/dedis/student20_pop/PoPApplication.java
```diff
@@ -125,6 +125,8 @@ public Person getPerson() {
      * @return the current LAO, unsafe
      */
     public Lao getCurrentLaoUnsafe() {
+        if(currentLao == null)
+            throw new IllegalStateException();
```
celinecamacho, line 129: Add a `@throws` to the comment of this method

celinecamacho (changes requested):

celinecamacho (approved):
👍

PR #203, opened 2021-01-20
Add the host of a LAO as attribute and make necessary changes

GabrielFleischer:
The LAO now knows where it is hosted and can send messages to the right backend

Improve the DataHandler to give the handler the host and the channel a data was received on


haoqianzhang (approved):

lhmerino (approved):

GabrielFleischer:
Wait ! No ! There is a compilation error that gradle build did not find locally...

PR #208, opened 2021-02-08
Initial refactoring work

pierluca:
Please browse by commit:
- 92ad4b6 is purely about identifying issues in the current code base
- 5b56aa1 adds TypeScript support (still untested) and should support the `import 'Models/Object/hash';` syntax
- f6bb281 contains the classes themselves
- 44cd5ef can be ignored, as it merely moves the folder to facilitate a later merge.


PopApp/websockets/LocalMockServer.js
```diff
@@ -3,6 +3,8 @@ const webSocketsServerPort = 8000;
 const webSocketServer = require('websocket').server;
 const http = require('http');
 
+// FIXME: Could this be re-used further ? If not, remove it entirely.
+
```
RaulinN, line 7: Probably only useful for low-level websocket testing. It was also supposed to be used in unit tests, but GitHub actions have problems running the server

PopApp/tsconfig.json
```diff
@@ -0,0 +1,76 @@
+{
+  "compilerOptions": {
+    /* Visit https://aka.ms/tsconfig.json to read more about this file */
+
+    /* Basic Options */
+    // "incremental": true,                   /* Enable incremental compilation */
+    "target": "esnext",                       /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */
+    "module": "commonjs",                     /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', 'es2020', or 'ESNext'. */
+    "lib": ["es6"],                           /* Specify library files to be included in the compilation. */
```
RaulinN: Using `btoa` and `atob` result in a compilation error. Adding the DOM library to the tsconfig file like in [this post](https://stackoverflow.com/questions/42899209/ts2304cannot-find-name-btoa) solves the issue.

(I cannot "propose changes" directly in the PR. Could it be a permissions issue?)

PopApp/Model/Network/Method/catchup.ts
```diff
@@ -0,0 +1,3 @@
+import { JsonRpcParams } from './jsonRpcParams';
+
+export class Subscribe extends JsonRpcParams {}
```
RaulinN: "Subscribe" should read "Catchup"

PopApp/Model/Objects/base64.ts
```diff
@@ -0,0 +1,14 @@
...
+        super(value);
+    }
+
+    public static encode(text: string) : Base64Data {
+        return new Base64Data(atob(text));
+    }
+
+    public decode(): string {
+        return btoa(this.valueOf());
+    }
```
RaulinN: `atob` and `btoa` are inversed

RaulinN (changes requested):

PopApp/websockets/LocalMockServer.js
pierluca: With Gaurav we were envisioning the possibility of having something like this, both to test the backends and the frontends.

We'll have to think a bit more about it though.

Are you sure you didn't have a problem like this?
https://github.community/t/test-http-server-in-github-actions/17608/6

PopApp/tsconfig.json
pierluca: Done. 
I think the proposed changes being blocked is only due to the fact that the commit is "outdated", it's not a permission issue.

PopApp/Model/Network/Method/catchup.ts
pierluca: Oops. 👍

PopApp/Model/Objects/base64.ts
pierluca: Oops (2). 👍

PR #210, opened 2021-02-10
Add LAOLocalDataSource and define entities

gnarula:
Adds `LAOLocalDataSource` for persisting entities on disk using Room

### DIFF
--- a/app/schemas/com.github.dedis.student20_pop.model.data.LAODatabase/1.json
+++ b/app/schemas/com.github.dedis.student20_pop.model.data.LAODatabase/1.json
@@ -1,0 +1,320 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 1,
+    "identityHash": "af8855922b2f6d3d9aeff68c7c58db43",
+    "entities": [
+      {
+        "tableName": "LAO",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`channel` TEXT NOT NULL, `id` TEXT, `name` TEXT, `lastModifiedAt` INTEGER, `createdAt` INTEGER, `organizer` TEXT, `modificationId` TEXT, PRIMARY KEY(`channel`))",
+        "fields": [
+          {
+            "fieldPath": "channel",
+            "columnName": "channel",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "id",
+            "columnName": "id",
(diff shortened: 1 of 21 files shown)


app/schemas/com.github.dedis.student20_pop.model.data.LAODatabase/1.json
```diff
@@ -0,0 +1,320 @@
...
+            "fieldPath": "start",
+            "columnName": "start",
+            "affinity": "INTEGER",
+            "notNull": false
+          },
+          {
+            "fieldPath": "scheduled",
+            "columnName": "scheduled",
+            "affinity": "INTEGER",
+            "notNull": false
```
celinecamacho, line 228: When a Roll Call is scheduled it doesn't have a start time yet, and vice-versa when a roll-call starts it doesn't have a scheduled time. For this reason, if the RollCall is scheduled what value would the start element have?

app/schemas/com.github.dedis.student20_pop.model.data.LAODatabase/1.json
```diff
@@ -0,0 +1,320 @@
...
+        },
+        "indices": [],
+        "foreignKeys": []
+      },
+      {
+        "tableName": "Meeting",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`originalId` TEXT NOT NULL, `id` TEXT, `laoChannel` TEXT, `name` TEXT, `creation` TEXT, `location` TEXT, `start` INTEGER, `end` INTEGER, `extra` TEXT, `modificationId` TEXT, PRIMARY KEY(`originalId`))",
+        "fields": [
+          {
+            "fieldPath": "originalId",
```
celinecamacho, line 68: What is the difference between the `originalId` and `id` field?

gnarula: The meeting id might change because of a modification of the name field since `id` depends on a hash of `name`. The `originalId` refers to the initial id the meeting was created with

celinecamacho: I don't think it's useful to save the `originalId`, I don't know if it's needed or will be used. 

If it is needed somewhere then I think it would be good to have the same for a LAO, the LAO `id` depends on the `name` too and the name can be modified (when updating a LAO)

celinecamacho (approved):
Everything looks good to me 👍

PR #211, opened 2021-02-10
Refactor Home related classes

celinecamacho:
Implement one ViewModel class to handle Home, Launch and Connect related fragments.


gnarula (changes requested):
Have a few suggestions on invoking the viewModel methods instead of simulating the click.

I think the call to `(HomeActivity) getActivity().setupZZZButton` is safe as long as we're certain that getActivity() returns a HomeActivity. While this holds true for now, I'm not sure if it would for say the `CameraPermissionFragment` which is inflated from multiple activities.

I reckon the `CameraPermissionFragment` should be refactored to have a `CameraPermissionViewModel` which is an interface that `HomeViewModel` and `LAODetailViewModel` (not sure about the naming yet) can implement.

```java
public interface CameraPermissionViewModel {
    void onPermissionGranted();

    void onPermissionDenied();
}
```

`CameraPermissionFragment` can then look like:

```java
public class CameraPermissionFragment extends Fragment {
    // ..

    private CameraPermissionViewModel mCameraPermissionViewModel;

    @Nullable
  @Override
  public View onCreateView(
          @NonNull LayoutInflater inflater,
          @Nullable ViewGroup container,
          @Nullable Bundle savedInstanceState) {

    mLaunchFragBinding = FragmentLaunchBinding.inflate(inflater, container, false);

    Activity activity = getActivity();
    if (activity instanceof HomeActivity) {
        mCameraPermissionViewModel = HomeActivity.obtainViewModel(activity);
    } else if (activity instanceof  LAODetailActivity) {
        mCameraPermissionViewModel = LAODetailActivity.obtainViewModel(activity);
    } else {
        throw new Exception("cannot obtain view model");
    }

    mLaunchFragBinding.setViewmodel(mHomeViewModel);
    mLaunchFragBinding.setLifecycleOwner(getActivity());

    return mLaunchFragBinding.getRoot();
  }

  // ..
}
```

I'm thinking about a similar pattern for `QRCodeScanningFragment`. Perhaps the calls to `setupZZZButton` can be abstracted away in an interface as well. What do you think?

app/src/main/java/com/github/dedis/student20_pop/ui/LaunchFragment.java
```diff
@@ -4,26 +4,104 @@
...
+    // Subscribe to "cancel launch" event
+    mHomeViewModel.getCancelNewLaoEvent().observe(this, booleanEvent -> {
+      Boolean action = booleanEvent.getContentIfNotHandled();
+      if (action != null) {
+        setupCancel();
+      }
+    });
+  }
+
+  public static HomeViewModel obtainViewModel(Fragment fragment) {
```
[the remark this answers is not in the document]
celinecamacho: Indeed

celinecamacho:
> Have a few suggestions on invoking the viewModel methods instead of simulating the click.
> 
> I think the call to `(HomeActivity) getActivity().setupZZZButton` is safe as long as we're certain that getActivity() returns a HomeActivity. While this holds true for now, I'm not sure if it would for say the `CameraPermissionFragment` which is inflated from multiple activities.
> 
> I reckon the `CameraPermissionFragment` should be refactored to have a `CameraPermissionViewModel` which is an interface that `HomeViewModel` and `LAODetailViewModel` (not sure about the naming yet) can implement.
> 
> ```java
> public interface CameraPermissionViewModel {
>     void onPermissionGranted();
> 
>     void onPermissionDenied();
> }
> ```
> 
> `CameraPermissionFragment` can then look like:
> 
> ```java
> public class CameraPermissionFragment extends Fragment {
>     // ..
> 
>     private CameraPermissionViewModel mCameraPermissionViewModel;
> 
>     @Nullable
>   @Override
>   public View onCreateView(
>           @NonNull LayoutInflater inflater,
>           @Nullable ViewGroup container,
>           @Nullable Bundle savedInstanceState) {
> 
>     mLaunchFragBinding = FragmentLaunchBinding.inflate(inflater, container, false);
> 
>     Activity activity = getActivity();
>     if (activity instanceof HomeActivity) {
>         mCameraPermissionViewModel = HomeActivity.obtainViewModel(activity);
>     } else if (activity instanceof  LAODetailActivity) {
>         mCameraPermissionViewModel = LAODetailActivity.obtainViewModel(activity);
>     } else {
>         throw new Exception("cannot obtain view model");
>     }
> 
>     mLaunchFragBinding.setViewmodel(mHomeViewModel);
>     mLaunchFragBinding.setLifecycleOwner(getActivity());
> 
>     return mLaunchFragBinding.getRoot();
>   }
> 
>   // ..
> }
> ```
> 
> I'm thinking about a similar pattern for `QRCodeScanningFragment`. Perhaps the calls to `setupZZZButton` can be abstracted away in an interface as well. What do you think?

If I understand correctly, there would be two interfaces`CameraPermissionViewModel` and `QRCodeScanningViewModel`?

Also, would the interface for the setup of the buttons be implemented by `HomeActivity`?

app/src/main/java/com/github/dedis/student20_pop/home/HomeViewModel.java
```diff
@@ -151,6 +162,24 @@ public void handleResult(Result result) {
 
     }
 
+    public void launchNewLao(String laoName) {
```
gnarula, line 165: I think the argument can be removed and the name can be retrieved from `mLaoName`

app/src/main/java/com/github/dedis/student20_pop/home/HomeViewModel.java
```diff
@@ -193,4 +242,16 @@ public void openConnect() {
...
+
+    public void launchNewLao() {
+        mLaunchNewLaoEvent.setValue(new Event<>(true));
+    }
+
+    public void cancelNewLao() {
+        mCancelNewLaoEvent.setValue(new Event<>(true));
+    }
+
+    public void setLaoName(Editable name) {
```
gnarula, line 254:
```suggestion
    public void setLaoName(String name) {
```

Would make unit testing easier :)

gnarula (changes requested):

gnarula (approved):

PR #209, opened 2021-02-09
Work fe1 raulinn cleanup

RaulinN:

PR #212, opened 2021-02-10
Possiblity to create a message "fromJson"

RaulinN:
- 8cdfd71 and db8825d can be ignored (file renaming)
- df929c0 adds a static `fromJson` method to protocol messages. The validation is done directly in the constructors
- d7d148d applies suggestions found in PR #209 (add a second method for hashing


fe1-web/Model/Network/Method/Message/data/witness/WitnessMessage.ts
```diff
@@ -16,14 +16,17 @@ export class WitnessMessage implements MessageData {
         this.message_id = msg.message_id;
 
         if (!msg.signature) throw new ProtocolError('Undefined \'signature\' parameter encountered during \'WitnessMessage\'');
+        // FIXME verify signature without the public key available? 0.o + uncomment 3 tests in "FromJson" test suite
```
pierluca: Technically the public key should match the sender's, in the message containing the message data.
But this points to something we might want to fix in the protocol.

fe1-web/__tests__/model/network/FromJson.ts
```diff
@@ -0,0 +1,496 @@
...
+  StateLao, StateMeeting,
+  UpdateLao, WitnessMessage
+} from '../../../Model/Network/Method/Message/data';
+import { ProtocolError } from '../../../Model/Network';
+import {Base64Data, Hash, PrivateKey, PublicKey, Signature, Timestamp} from "../../../Model/Objects";
+import { sign } from "tweetnacl";
+import { encodeBase64 } from "tweetnacl-util";
+
+const assert = require('assert');
+const assertChai = require('chai').assert;
```
pierluca, line 17: there's normally no reason to use chai if we are on Jest. Was there a specific need that required chai?

fe1-web/__tests__/model/network/FromJson.ts
```diff
@@ -0,0 +1,496 @@
...
+  const sampleWitnessMessage: Partial<WitnessMessage> = {
+    object: ObjectType.MESSAGE,
+    action: ActionType.WITNESS,
+    message_id: mockMessageId,
+    signature: mockSecretKey.sign(mockMessageId),
+  };
+
+
+
+  it('should successfully create the objects', function () {
```
pierluca, line 156: a couple of general comments. 

1. If this test fails in the continuous integration pipeline, you'll have no idea why.
Ideally, you want to make small tests that verify just one tiny bit of functionality.

Compare the following two outputs
```
"... should successfully create the objects" - FAILED
```

vs

```
"... should successfully create the objects - simple lao" - SUCCESS
"... should successfully create the objects - update lao" - SUCCESS
"... should successfully create the objects - state lao" - SUCCESS
"... should successfully create the objects - create meeting" - SUCCESS
"... should successfully create the objects - update meeting" - FAILED
"... should successfully create the objects - state meeting" - SUCCESS
...
```

2. checking `.toThrow(ProtocolError)` may not (always) be sufficient to know precisely what failed.
You might be throwing errors for the wrong reason. Could we at least check that the message contains the name of the invalid parameter?

fe1-web/__tests__/model/network/FromJson.ts
```diff
@@ -0,0 +1,496 @@
...
+  const mockMessageId = Base64Data.encode('message_id');
+
+  const org = mockPublicKey;
+  const time = STANDARD_TIMESTAMP;
+  const name = 'PoP\'s team "LAO" or filling name';
+  const mockLaoId: Hash = Hash.fromStringArray(org.toString(), time.toString(), name);
+
+  let temp: any = {};
+
+  const sampleCreateLao: Partial<CreateLao> = {
```
pierluca, line 53: these samples are extremely precious. 
We might consider putting them in JSON files to share with the entire project.

We also discussed with @gnarula this possibility, as we could then re-use samples as part of both the documentation, validating the JSON schema and validating the sub-systems :)

fe1-web/__tests__/model/network/FromJson.ts
```diff
@@ -0,0 +1,496 @@
...
+    location: 'Lausanne',
+    start: time,
+    end: FUTURE_TIMESTAMP,
+    extra: { extra: 'extra info' },
+  };
+
+  const sampleStateMeeting: Partial<StateMeeting> = {
+    object: ObjectType.MEETING,
+    action: ActionType.STATE,
+    id: Hash.fromStringArray(mockLaoId.toString(), time.toString(), name),
```
pierluca, line 100: technically, if we're looking at a JSON, id is a string, not a `Hash` object. You might want to use the `.valueOf()` method.
Then of course, we're stuck with untyped objects, which is the (sad) reality of untrusted data from the user, servers, etc.

fe1-web/Model/Network/JsonRpcRequest.ts
```diff
@@ -8,74 +8,74 @@ import { ProtocolError } from './ProtocolError';
...
-            // RPC methods, expect an ID
-            case JsonRpcMethod.PUBLISH:
-            case JsonRpcMethod.SUBSCRIBE:
-            case JsonRpcMethod.UNSUBSCRIBE:
-            case JsonRpcMethod.CATCHUP:
-                if ( this.id === undefined ) {
-                    return false;
-                }
-                break;
+  public verify(): boolean {
```
pierluca, line 38: I seem to remember you had convinced me of dropping the verify() and just verifying these properties in the constructor, right?

fe1-web/package.json
```diff
@@ -64,9 +64,7 @@
     "eslint-plugin-jsx-a11y": "^6.3.1",
     "eslint-plugin-react": "^7.21.5",
     "jest": "^26.6.3",
-    "jest-localstorage-mock": "^2.4.6",
-    "mocha": "^8.2.1",
-    "typescript": "^4.1.3"
```
pierluca, line 69: do you really want to drop typescript?

pierluca (changes requested):
A few remarks here and there.

One important thing about the parseJson(...) : be careful about the distinction between a JSON object, which will contain dictionaries, arrays, strings, numbers, booleans and null values ; and its typed representation, which might contain classes.
the `parseJson` function will have to check some obj properties' types and build objects out of them (ex Hash, Base64Data, etc.)
I do expect that most parseJson code is buggy right now and only works because the tests themselves already create objects :)

fe1-web/__tests__/model/network/FromJson.ts
RaulinN: There was before I deleted the mocha`package (as well as the typescript package it seems ^^') on the commit you reviewed. We could indeed use the expect syntax from jest.

I keep a note of this and will remove chai once the priority list becomes smaller

fe1-web/__tests__/model/network/FromJson.ts
RaulinN: 1. I agree that is it way more clear (even though, from what I saw, Jest shows exactly where the error occured)
2. done

fe1-web/package.json
RaulinN: woops ^^'

RaulinN:
>A few remarks here and there.
>One important thing about the parseJson(...) : be careful about the distinction between a JSON object, which will contain dictionaries, arrays, strings, numbers, booleans and null values ; and its typed representation, which might contain classes.
the parseJson function will have to check some obj properties' types and build objects out of them (ex Hash, Base64Data, etc.)
>I do expect that most parseJson code is buggy right now and only works because the tests themselves already create objects :)

Oh yes correct! For example `Hash`es are in fact strings when parsed from JSON! Will correct this as soon as possible

fe1-web/Model/Network/JsonRpcRequest.ts
RaulinN: I indeeed forgot the top layer of the protocol ^^'. Working on it

fe1-web/__tests__/model/network/FromJson.ts
pierluca: Re: 1, it does show where the error occurred, but that already entails "investigating" in a way.
You should aim for the *unit* tests in your CI/CD pipeline to "scream" at you what you broke, with minimal investigation.
It's different with integration tests, but making them as screaming-clear as possible is always a good idea :)

pierluca (changes requested):
I may have missed a few things, but I think I've covered the most critical parts.

fe1-web/Model/Network/Method/Message/data/lao/CreateLao.ts
```diff
@@ -21,20 +21,20 @@ export class CreateLao implements MessageData {
...
     checkWitnesses(msg.witnesses);
-    this.witnesses = [...msg.witnesses];
+    this.witnesses = msg.witnesses.map((key) => new PublicKey(key.toString()));
 
     if (!msg.id) throw new ProtocolError('Undefined \'id\' parameter encountered during \'CreateLao\'');
     const expectedHash: Hash = Hash.fromStringArray(msg.organizer.toString(), msg.creation.toString(), msg.name);
     if (!expectedHash.equals(msg.id))
       throw new ProtocolError('Invalid \'id\' parameter encountered during \'CreateLao\': unexpected id value');
-    this.id = msg.id;
+    this.id = new Hash(msg.id.toString());
```
pierluca: The string -> type logic should be in fromJson() 

If we respect the Partial<CreateLao> interface, these are already in the "right" format.

Same thing applies in all files, sorry :-/

pierluca (changes requested):
I had missed this commit...

Same comment across all files, you might also want to simplify the import paths :)

fe1-web/.eslintrc
```diff
@@ -8,13 +8,16 @@
       "mjs": "never",
       "jsx": "never",
       "ts": "never",
-      "tsx": "never"
+      "tsx": "never",
+      "vue": "never"
```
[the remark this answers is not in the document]
RaulinN: I blindly followed the stack overflow post, but it is indeed not needed ^^'

fe1-web/Model/Network/Method/Message/data/builder.ts
```diff
@@ -1,31 +1,73 @@
...
+  }
+}
+
+function buildRollCallMessage(msgData: MessageData): MessageData {
+  switch (msgData.action) {
+    case ActionType.CREATE:
+      return new CreateRollCall(msgData);
+    case ActionType.OPEN:
+    case ActionType.REOPEN:
+      return new OpenRollCall(msgData);
```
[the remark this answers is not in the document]
RaulinN: I was also thinking about it. We definitely could for clarity and in case one of those two objects changes in the future

fe1-web/Store/objects/KeyPairData.ts
```diff
@@ -0,0 +1,6 @@
+import { PublicKey, PrivateKey } from '../../Model/Objects';
+
+export interface KeyPairData {
```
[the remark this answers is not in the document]
RaulinN: Yes. But ESlint couldn't find `Partial<KeyPair>` (or `Partial<any>`) in this file. Don't know why since it was working in the `Model/` directory

fe1-web/Store/Reducers/currentLaoReducer.ts
```diff
@@ -1,4 +1,14 @@
-const initialState = { lao: {} };
+import { AnyAction } from 'redux';
+import { LaoData } from '../objects';
+
+const initialState: LaoData = {
```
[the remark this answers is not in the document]
RaulinN: I did it this way simply to be able to construct `nextState` from the previous `state` 

```javascript
    const nextState = {
      ...state,
      name: action.value.name,
      id: action.value.id,
      // ...
    };
```

If I instead write `initialState: LaoData? = null`, then I cannot do this trick. Do you prefer one of the two options?

fe1-web/Store/Reducers/currentLaoReducer.ts
```diff
@@ -13,11 +23,16 @@ const initialState = { lao: {} };
...
       ...state,
-      lao: action.value,
+      name: action.value.name,
+      id: action.value.id,
+      creation: action.value.creation,
+      last_modified: action.value.last_modified,
+      organizer: action.value.organizer,
+      witnesses: action.value.witnesses,
     };
     return nextState || state;
```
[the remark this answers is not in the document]
RaulinN: Not necessary. It was a way to noob proof the system according to the tutorial I followed

fe1-web/Store/storage/StorageCurrentLao.ts
```diff
@@ -0,0 +1,38 @@
...
+        witnesses: value.witnesses,
+      }
+    });
+  }
+
+  public getCurrentLao(): Lao {
+
+    let lao: LaoData = this.storage.getState().currentLaoReducer;
+
+    if (lao.name && lao.id && lao.creation && lao.last_modified && lao.organizer && lao.witnesses)
```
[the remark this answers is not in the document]
RaulinN: Already did it by using `Lao | null` as initial state. The ugly if statements now looks like `if (lao !== null)`

fe1-web/__tests__/model/network/FromJson.ts
```diff
@@ -8,14 +8,15 @@ import {
   StateLao, StateMeeting,
   UpdateLao, WitnessMessage
 } from '../../../Model/Network/Method/Message/data';
+import { initialise } from '../../../Store/Storage';
 import { ProtocolError } from '../../../Model/Network';
-import {Base64Data, Hash, PrivateKey, PublicKey, Signature, Timestamp} from "../../../Model/Objects";
+import { Base64Data, Hash, PrivateKey, PublicKey, Signature, Timestamp } from "../../../Model/Objects";
```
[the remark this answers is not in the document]
RaulinN: Yes! This is Intellij auto importing. I'm trying to correct it when I see these mistakes. There are probably a few in the project

fe1-web/__tests__/model/network/FromJson.ts
```diff
@@ -38,6 +39,15 @@ const _generateKeyPair = () => {
...
 describe('=== fromJson object checks ===', function() {
 
+  beforeAll(() => {
+    const promise = new Promise((resolve, reject) => {
+      initialise();
+    });
+    promise.then(
+      () => console.log('storage configured (printed from storage setup)'),
+    );
+  });
```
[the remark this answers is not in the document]
RaulinN: That is a very strange point. I also googled yesterday to learn how to handle promises in async code in jest. I did exactly what you (and Jest) proposed. However, doing so, for some reason, does the polar inverse of what we expect. Returning a promise makes it so that the tests don't wait for `beforeAll()` to complete before executing (thus resulting is fails because nothing is in the storage)

fe1-web/__tests__/websockets/WebsocketApi.ts
```diff
@@ -97,19 +89,37 @@ const checkRequests = (request, object, action) => {
...
-};
+};*/
 
 describe('=== WebsocketApi tests ===', function() {
 
-  /* NOTE: checks are done in checkRequests since wsApi.request* return void */
+  beforeAll(() => {
+    new Promise((resolve, reject) => {
+      initialise();
+    });
```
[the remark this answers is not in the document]
RaulinN: Yeah this is at least as strange as the last problem with Promises. Doing this
```javascript
    new Promise((resolve, reject) => {
      initialise();
    }).then(() => console.log("finished");
```

will *never* print finished. However, I am 100% sure that `initialise` finishes  (adding a print at the very end of `initialise` prints its content)


Moreover, from my testing (running the test suite about 100 times in total), `storeStuffInStorage` always executes once the storage is fully initialised.

```javascript
    new Promise((resolve, reject) => {
      initialise();
    });

    storeStuffInStorage();
```

fe1-web/Store/storage/StorageCurrentLao.ts
```diff
@@ -0,0 +1,38 @@
+import { dispatch } from '../Storage';
+import { Store } from 'redux';
+import { Lao } from '../../Model/Objects';
+import { LaoData } from '../objects';
+
+
+export class StorageCurrentLao {
+
+  public readonly storage: Store;
```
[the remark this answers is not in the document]
RaulinN: Definitely not ^^

fe1-web/Store/Reducers/currentLaoReducer.ts
```diff
@@ -21,20 +15,18 @@ const initialState: LaoData = {
...
+function currentLaoReducer(state: Lao | null = initialState, action: AnyAction): Lao | null {
+  if (action.type === ActionOpenedLaoReducer.SET_CURRENT_LAO) {
+    return {
       name: action.value.name,
       id: action.value.id,
       creation: action.value.creation,
       last_modified: action.value.last_modified,
       organizer: action.value.organizer,
       witnesses: action.value.witnesses,
     };
```
pierluca: I would call the constructor of Lao here

```
return new Lao({
      name: action.value.name,
      id: action.value.id,
      creation: action.value.creation,
      last_modified: action.value.last_modified,
      organizer: action.value.organizer,
      witnesses: action.value.witnesses,
    });
```

fe1-web/Store/Reducers/currentLaoReducer.ts
```diff
@@ -21,20 +15,18 @@ const initialState: LaoData = {
...
  * @param action action to be executed by the reducer
- * @returns {{lao: *}|{lao: {}}} new LAO if action is valid, old LAO otherwise
+ * @returns new LAO if action is valid, old LAO otherwise
  */
-function currentLaoReducer(state: LaoData = initialState, action: AnyAction) {
-  if (action.type === 'SET_CURRENT_LAO') {
-    const nextState = {
-      ...state,
+function currentLaoReducer(state: Lao | null = initialState, action: AnyAction): Lao | null {
+  if (action.type === ActionOpenedLaoReducer.SET_CURRENT_LAO) {
```
pierluca: state could be "null" here. In that case, return null?

fe1-web/Store/Reducers/keypairReducer.ts
```diff
@@ -14,17 +15,14 @@ const initialState: KeyPairData = { publicKey: undefined, privateKey: undefined
...
  * @param action action to be executed by the reducer
- * @returns {{secKey: ([]|*[]|string), pubKey: ([]|*[]|string)}|{secKey: [], pubKey: []}}
- * new key pair if action is valid, old key pair otherwise
+ * @returns new key pair if action is valid, old key pair otherwise
  */
-function keypairReducer(state: KeyPairData = initialState, action: AnyAction): KeyPairData {
-  if (action.type === 'SET_KEYPAIR') {
-    const nextState = {
-      ...state,
+function keypairReducer(state: KeyPair | null = initialState, action: AnyAction): KeyPair | null {
```
pierluca: state could be "null" here. In that case, return null?

fe1-web/Store/Reducers/keypairReducer.ts
```diff
@@ -14,17 +15,14 @@ const initialState: KeyPairData = { publicKey: undefined, privateKey: undefined
...
+  if (action.type === ActionKeyPairReducer.SET_KEYPAIR) {
+    return {
```
pierluca: `return new KeyPair(` ?

fe1-web/Store/Storage.ts
```diff
@@ -1,37 +1,30 @@
...
 let persistor: Persistor;
 
-let storageKeyPair: StorageKeyPair;
-let storageCurrentLao: StorageCurrentLao;
-
 export function initialise(): void {
-
   let storage: WebStorage;
+  // FIXME doesnt compile
+  //if (!window || typeof window.localStorage === 'undefined' || window.localStorage === null) {
```
pierluca: what's the error? :)

fe1-web/Store/storage/KeyPairStore.ts
```diff
@@ -0,0 +1,43 @@
...
+      const keyPair: KeyPair = new KeyPair(
+        new PublicKey(encodeBase64(pair.publicKey)),
+        new PrivateKey(encodeBase64(pair.secretKey)),
+      );
+
+      store(keyPair);
+      return keyPair;
+    }
+
+    return new KeyPair(keys.publicKey, keys.privateKey);
```
pierluca: can't we just return keys, now ?

fe1-web/Store/storage/KeyPairStore.ts
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  export function getPublicKey(): PublicKey {
+    return get().publicKey;
+  }
+
+  export function getPrivateKey(): PrivateKey {
+    return get().privateKey;
+  }
```
pierluca: do we still need this? it seems this getter isn't doing much :) 
e.g. `store.getPrivateKey()` vs `store.get().privateKey` seem basically equivalent.

fe1-web/Store/storage/OpenedLaoStore.ts
```diff
@@ -0,0 +1,17 @@
...
+import { dispatch, getStore } from '../Storage';
+
+export namespace OpenedLaoStore {
+
+  export function store(value: Lao): void {
+    dispatch({ type: 'SET_CURRENT_LAO', value });
+  }
+
+  export function get(): Lao | null {
+    const lao: Lao = getStore().getState().currentLaoReducer;
```
pierluca: can't we just return lao ?

fe1-web/__tests__/model/network/FromJson.ts
```diff
@@ -4,16 +4,17 @@ import {
...
-  ObjectType, OpenRollCall,
+  ObjectType, OpenRollCall, ReopenRollCall,
   StateLao, StateMeeting,
   UpdateLao, WitnessMessage
-} from '../../../Model/Network/Method/Message/data';
-import { getStorageCurrentLao, initialise } from '../../../Store/Storage';
-import { ProtocolError } from '../../../Model/Network';
-import { Base64Data, Hash, Lao, PrivateKey, PublicKey, Timestamp } from '../../../Model/Objects';
+} from 'Model/Network/Method/Message/data';
+import { initialise } from 'Store/Storage';
```
pierluca: it might be worth aliasing this for clarity, like `as storeInitialize`

also, we might want to rename `initialise` to `initialize`, as american english is more common in code bases :)

fe1-web/Store/Reducers/currentLaoReducer.ts
pierluca: I have doubt, is the Lao constructor accepting a Partial<Lao> as input data?

pierluca (changes requested):
Minor remarks :)

fe1-web/model/network/JsonRpcResponse.ts
```diff
@@ -1,7 +1,50 @@
...
+    this.id = resp.id || UNDEFINED_ID;
+
+    if (resp.error !== undefined && resp.result !== undefined) {
+      throw new ProtocolError('Unexpected json-rpc answer : both \'error\' and \'result\' are present');
+    }
+
+    if (resp.result !== undefined) {
+      if (resp.result) {
+        throw new ProtocolError('Unexpected json-rpc answer : non-zero result value');
+      }
```
pierluca, line 27: you've shown me that `[...messages...]` would actually throw an error here :)

fe1-web/model/network/JsonRpcResponse.ts
```diff
@@ -1,7 +1,50 @@
...
+
+      this.result = resp.result;
+    } else if (resp.error !== undefined) {
+      // FIXME hardcoded. Use enum if we keep this idea
+      if (resp.error.code < -6 || resp.error.code > -1) {
+        throw new ProtocolError(`Unexpected json-rpc answer : unexpected error code value '${resp.error.code}'`);
+      }
```
pierluca, line 34: This makes server-specific extensions harder to tolerate, I'd drop the < -6 part and just leave `>= 0`.

fe1-web/network/NetworkConnexion.ts
```diff
@@ -0,0 +1,109 @@
...
+      new Promise((resolve, reject) => {
+        this.waitWebsocketReady(resolve, reject);
+      }).then(
+        () => this.sendPayload(payload),
+      );
+    } else {
+      let query: JsonRpcRequest = payload;
+
+      // websocket ready to be used, message can be sent
+      if (payload.id === undefined || payload.id === GENERATE_ID) {
```
pierluca, line 60: this is kind of dangerous, because the application could generate an ID that we've already used in a RPC in progress.

I'm thinking that maybe we could "force" the application to use "GENERATE_ID" and return an error if they don't, but always overwrite the `id` field here.

pierluca (changes requested):
Awesome start to the connection manager, I've * of course * bombarded with you a ton of remarks, as expected from a first draft, but I think we're clearly heading in the right direction !

fe1-web/network/NetworkConnexion.ts
```diff
@@ -0,0 +1,109 @@
+import { w3cwebsocket as W3CWebSocket } from 'websocket';
```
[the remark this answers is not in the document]
RaulinN: woops! ^^'

fe1-web/model/network/JsonRpcResponse.ts
```diff
@@ -1,7 +1,50 @@
...
+
+export const UNDEFINED_ID: number = -1;
+
+interface ErrorObject {
+  code: number;
+  description: string;
+}
+
+export class JsonRpcResponse {
+  public readonly result?: number;
```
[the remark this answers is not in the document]
RaulinN: Indeed! I was a little bit more specific by defining it as `public readonly result?: number | Message;`

fe1-web/network/NetworkManager.ts
```diff
@@ -0,0 +1,43 @@
...
+  public connect(uri: string, port: number = 8000, path: string = ''): NetworkConnexion {
+    const address: string = `ws://${uri}:${port}${(path === '') ? '' : `/${path}`}`;
+    const connexion: NetworkConnexion = new NetworkConnexion(address, (m) => { console.log(m); });
+
+    this.connexions.push(connexion);
+    return connexion;
+  }
+
+  public disconnect(connexion: NetworkConnexion): void;
+  public disconnect(address: string): void;
```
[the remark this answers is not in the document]
RaulinN: As discussed during the meeting, I also added a `disconnect` method in the `NetworkConnection` class and used it while disconnecting from the maanger as well

fe1-web/network/NetworkManager.ts
```diff
@@ -0,0 +1,43 @@
...
+  public disconnect(param: NetworkConnexion | string): void {
+    if (typeof param === 'string') {
+      const connexion = this.connexions.find((nc: NetworkConnexion) => nc.address === param);
+      if (connexion !== undefined) this.disconnect(connexion);
+    } else {
+      // TODO check that this actually works once the web version is working ^^'
+      const index = this.connexions.indexOf(param);
+      if (index !== -1) this.connexions.splice(index, 1);
```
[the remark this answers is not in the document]
RaulinN: Resolved by implementing suggestions from the last comment

fe1-web/network/NetworkConnexion.ts
```diff
@@ -0,0 +1,109 @@
...
+import { JsonRpcRequest, JsonRpcResponse, UNDEFINED_ID } from 'model/network';
+
+const GENERATE_ID: number = -1;
```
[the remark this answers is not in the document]
RaulinN: That is what I did in the index.ts that I forgot to push ^^'

I also just implemented `NetworkError`s and exported it in the index.ts file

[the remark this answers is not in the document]
RaulinN: (Not sure yet if we really not to export NetworkErrors thought). We will see once the whole network part is finished

fe1-web/network/NetworkConnexion.ts
RaulinN: This was a mechanism implemented back in the time to be able to send the same query another time (in case of server fault) with the same id.
```pseudocode
let query = // get the query from pending queue
if (retry) {
     send(id = query.id, payload = query);
} else {
    send(id = GENERATE_ID, payload = query);
}
```

However if we return promises back to the user (component), then this wouldn't be useful anymore in my opinion 🤔 . What do you think about it @pierluca ?

fe1-web/model/network/JsonRpcResponse.ts
RaulinN: done!

fe1-web/model/network/JsonRpcResponse.ts
RaulinN: argh well spotted! I am starting to hate these typescript "unions" ^^'

fe1-web/network/NetworkConnexion.ts
pierluca: @RaulinN  If there was a fault during the first query, it won't hurt the system to do the same RPC (again) with a different request id.

fe1-web/network/NetworkManager.ts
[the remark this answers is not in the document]
pierluca: After thinking about it, I think there's little point (for now) in implementing two ways to go about this.
We essentially need a way to disconnect from all connections at once (i.e. quitting the LAO) or dropping one specific connection. How a specific connection should be dropped (with which parameters?) is a question that I'd let the semester students figure out :)

fe1-web/network/NetworkConnexion.ts
RaulinN: Yes indeed! That's what I was thinking if we still need to send the same query! I just explained it poorly in my last message ^^. So I agree with you, no need to let the user choose the id anymore

fe1-web/network/NetworkManager.ts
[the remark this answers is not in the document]
RaulinN: For now, I have three methods:
- `disconnect(Networkconnection nc)`
- `disconnectFrom(string uri, number port, string path)`
- `disconnectFromAll()`

We could either remove one of those (probably `disconnectFrom`) or leave them all. In my opinion, it does not hurt to have those 3 options

fe1-web/network/NetworkManager.ts
```diff
@@ -0,0 +1,43 @@
...
+  private static INSTANCE: NetworkManager;
+
+  private constructor() {
+    this.connexions = [];
+  }
+
+  public static get(): NetworkManager {
+    if (this.INSTANCE === undefined) { this.INSTANCE = new NetworkManager(); }
+    return this.INSTANCE;
+  }
```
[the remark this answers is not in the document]
RaulinN: done! :)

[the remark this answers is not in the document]
RaulinN: @pierluca Out of curiosity, is there a reason (other than testing) why the "non private constructor" version is better than this one? 🤔

fe1-web/network/NetworkConnexion.ts
```diff
@@ -0,0 +1,109 @@
...
+        });
+        this.payloadSentCount += 1;
+      }
+
+      const promise = new Promise<JsonRpcResponse>(() => {
+        // FIXME ?
+        // how to resolve/reject the promise from outside?
+        // Aka. when we receive an answer, we parse the id, take the
+        // corresponding query out of the map and resolve the promise.
+      });
```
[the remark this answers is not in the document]
RaulinN: > if you return a promise from within a promise, that promise will be "chained" to the parent promise ;-)

Was very useful to handle the "waitWebsocketReady" part 👍

fe1-web/network/NetworkConnexion.ts
```diff
@@ -0,0 +1,109 @@
...
+
+const WEBSOCKET_READYSTATE_INTERVAL_MS = 10;
+const WEBSOCKET_READYSTATE_MAX_ATTEMPTS = 100;
+
+export class NetworkConnexion {
+  private readonly ws: any;
+
+  private payloadSentCount: number = 0;
+
+  private payloadPending: Map<number, Promise<JsonRpcResponse>>;
```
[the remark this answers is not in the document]
RaulinN: Done as discussed during meeting

fe1-web/network/NetworkConnexion.ts
```diff
@@ -0,0 +1,109 @@
...
+        // FIXME ?
+        // how to resolve/reject the promise from outside?
+        // Aka. when we receive an answer, we parse the id, take the
+        // corresponding query out of the map and resolve the promise.
+      });
+
+      this.payloadPending.set(
+        query.id as number, // as this point, there will be a number as id
+        promise,
+      );
```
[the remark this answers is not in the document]
RaulinN: Done as discussed during meeting

RaulinN:
Here we go @pierluca ! All suggestions (and improvements discussed during our meeting) were taken into account with small changes/tweaks. I am happy of the result! :)

fe1-web/network/NetworkManager.ts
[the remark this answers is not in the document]
pierluca: Nope. But testable stuff is awesome 😆

fe1-web/network/NetworkConnection.ts
```diff
@@ -0,0 +1,155 @@
...
+
+  private onMessageHandler: ResponseHandler = DEFAULT_RESPONSE_HANDLER;
+
+  public readonly address: string; // FIXME get the address from ws (log ws when app is running)
+
+  constructor(address: string) {
+    const ws = new W3CWebSocket(address);
+
+    ws.onopen = () => { console.info(`Initiating web socket : ${address}`); };
+    ws.onmessage = (message: any) => {
```
pierluca, line 48: what happens if we throw `onmessage` ? who's going to handle that error?
maybe it's good enough to just log

fe1-web/network/NetworkConnection.ts
```diff
@@ -0,0 +1,155 @@
...
+
+      if (response.id !== UNDEFINED_ID) {
+        const pendingResponse = this.payloadPending.get(response.id);
+        if (pendingResponse === undefined) {
+          throw new NetworkError(`Received a response which id = ${response.id} does not match any pending requests`);
+        }
+
+        this.payloadPending.delete(response.id);
+
+        pendingResponse.resolvePromise(response);
```
pierluca, line 66: resolve or reject, based on whether the response is a result or a failure.
for convenience, it might make sense to have a getter/helper "isError" on the JsonRpcResponse, to keep the logic very clean here

fe1-web/network/NetworkConnection.ts
```diff
@@ -0,0 +1,155 @@
...
+  constructor(address: string) {
+    const ws = new W3CWebSocket(address);
+
+    ws.onopen = () => { console.info(`Initiating web socket : ${address}`); };
+    ws.onmessage = (message: any) => {
+      console.log(`Received a new message from '${address}' : `, message.data);
+
+      let response: JsonRpcResponse;
+      try {
+        response = JsonRpcResponse.fromJson(message.data);
```
pierluca, line 53: here we're assuming that ALL the messages are JsonRpcResponses, some might be JsonRpcRequests (e.g. "Broadcast") 😅

fe1-web/network/NetworkConnection.ts
```diff
@@ -0,0 +1,155 @@
...
+      } catch (e) {
+        throw new ProtocolError('Answer is not valid Json');
+      }
+
+      if (response.id !== UNDEFINED_ID) {
```
pierluca, line 58: this logic is correct for JsonRpcResponse, not for a broadcast

fe1-web/network/NetworkConnection.ts
```diff
@@ -0,0 +1,155 @@
...
+        const pendingResponse = this.payloadPending.get(response.id);
+        if (pendingResponse === undefined) {
+          throw new NetworkError(`Received a response which id = ${response.id} does not match any pending requests`);
+        }
+
+        this.payloadPending.delete(response.id);
+
+        pendingResponse.resolvePromise(response);
+        this.onMessageHandler(response);
```
pierluca, line 67: we probably only need a handler for notifications (i.e. JsonRpcRequests) , and those don't have an id

fe1-web/network/NetworkConnection.ts
```diff
@@ -0,0 +1,155 @@
...
+  public setResponseHandler(handler: ResponseHandler) {
+    this.onMessageHandler = handler;
+  }
+
+  public sendPayload(payload: JsonRpcRequest): Promise<JsonRpcResponse> {
+    // Check that the websocket connection is ready
+    if (!this.ws.readyState) {
+      return new Promise(() => {
+        this.waitWebsocketReady().then(() => this.sendPayload(payload));
+      });
```
pierluca, line 86: technically here the Promise is never resolved. You can simply do:
```
return this.waitWebsocketReady()
   .then(() => this.sendPayload(payload));
```

fe1-web/network/NetworkConnection.ts
```diff
@@ -0,0 +1,155 @@
...
+    return promise;
+  }
+
+  public disconnect(): void {
+    this.ws.close();
+  }
+
+  private waitWebsocketReady(): Promise<void> {
+    if (this.ws.readyState) {
+      return new Promise((resolve) => { resolve(); });
```
pierluca, line 132: This time I think you can really do this :) 
```suggestion
      return Promise.resolve();
```

fe1-web/network/NetworkError.ts
```diff
@@ -0,0 +1,16 @@
+/**
```
pierluca, line 1: 👍

fe1-web/network/NetworkManager.ts
```diff
@@ -1,43 +1,64 @@
...
+      this.connections.splice(index, 1);
     }
   }
 
-  public sendPayload(payload: JsonRpcRequest): void { // FIXME return promise?
-    // For now, we only have 1 connexion opened at a time: the organizer server
-    this.connexions[0].sendPayload(payload);
+  public disconnectFrom(uri: string, port: number = 8000, path: string = ''): void {
+    const address: string = NetworkManager.buildAddress(uri, port, path);
+    const connection = this.connections.find((nc: NetworkConnection) => nc.address === address);
```
pierluca, line 44: here you can use `getConnectionByAddress`

fe1-web/network/NetworkConnection.ts
```diff
@@ -0,0 +1,155 @@
...
+
+interface PendingResponse {
+  promise: Promise<JsonRpcResponse>,
+  resolvePromise: (value: JsonRpcResponse) => void,
+  rejectPromise: () => void,
+  timeoutId: NodeJS.Timeout,
+}
+
+export class NetworkConnection {
+  private readonly ws: any;
```
pierluca, line 34: if the websocket disconnects, is there any retry-mechanism to bring it back up alive? 
If not, the first network crash will render the application unusable.

fe1-web/network/NetworkConnection.ts
```diff
@@ -0,0 +1,155 @@
...
+const DEFAULT_RESPONSE_HANDLER: ResponseHandler = (message: JsonRpcResponse) => {
+  if (message.error === undefined) {
+    // default handler does nothing which is useful for some messages
+  } else {
+    console.error(
+      'A negative network error was received:\n'
+      + `\t- error code : ${message.error.code}`
+      + `\t- description : ${message.error.description}`,
+    );
+  }
```
pierluca, line 18: I don't think we need much here, to be honest, other than just logging something like "No response handler was provided to manage message: xxxx"

pierluca (changes requested):
It's very very nice indeed ! 

Still broken though 😅 we forgot altogether to handle the Broadcast messages (and other future notifications) received in the `onmessage` callback.

fe1-web/network/NetworkConnection.ts
RaulinN: In the old version, we had a try catch block around the handler. Forgot to add it there ^^'

PR #213, opened 2021-02-16
Refactor LAO detail related classes

celinecamacho:


gnarula:
Needs a few changes to align with the work in progress about state management and handling network calls with the backend. Merging it to `fe2-android-gaurav-mvvm` for now.

PR #215, opened 2021-02-18
Migrated from Chai to Jest tests

pierluca:


fe1-web/jest.config.js
```diff
@@ -1,4 +1,7 @@
 module.exports = {
   preset: 'react-native',
   moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx', 'json', 'node'],
+  setupFilesAfterEnv: ["jest-extended"],
+  testPathIgnorePatterns: ["/node_modules/"],
+  testMatch: [ "**/__tests__/**/?(*.)+(spec|test).[jt]s?(x)", "**/?(*.)+(spec|test).[jt]s?(x)" ],
```
RaulinN, line 6: Perfect! 👍

fe1-web/store/Storage.ts
```diff
@@ -45,7 +37,9 @@ export function storeInit(): PersistStoreConfig {
     // clears the local cached storage as well as the state of the storage
     let newState = state;
     if (action.type === 'CLEAR_STORAGE') {
-      storage.removeItem('persist:root'); newState = undefined;
+      // unsafe, asynchronous operation:
+      AsyncStorage.removeItem('persist:root');
```
RaulinN, line 41: Should we create an adapter for the asynchronous storage for the students? 

I don't exactly see yet how the components will update their state/view with our new network layer (aka. after handling the server answer) so it might not be useful at all.

tests/data/keypair.json
```diff
@@ -0,0 +1,4 @@
+{
+    "publicKey": "xjHAz+d0udy1XfHp5qugskWJVEGZETN/8DV3+ccOFSs=",
+    "privateKey": "vx0b2hbxwPBQzfPu9NdlCcYmuFjhUFuIUDx6doHRCM7GMcDP53S53LVd8enmq6CyRYlUQZkRM3/wNXf5xw4VKw=="
+}
```
RaulinN, line 4: 👍

fe1-web/__tests__/model/objects/Objects.test.ts
```diff
@@ -0,0 +1,92 @@
...
+  Hash, KeyPair, KeyPairState,
+  Lao, LaoState,
+  PrivateKey,
+  PublicKey,
+  WitnessSignature,
+} from 'model/objects';
+import { channelFromId, ROOT_CHANNEL } from 'model/objects/Channel';
+
+const mockPublicKey = new PublicKey('xjHAz+d0udy1XfHp5qugskWJVEGZETN/8DV3+ccOFSs=');
+const mockSecretKey = new PrivateKey('vx0b2hbxwPBQzfPu9NdlCcYmuFjhUFuIUDx6doHRCM7GMcDP53S53LVd8enmq6CyRYlUQZkRM3/wNXf5xw4VKw==');
```
RaulinN, line 12: Should we then import those keys from the `tests/data/keypair.json` file?

(Same comment for the other test files using the keys)

fe1-web/jest.config.js
pierluca: Actually, re-reading it, I feel I can just keep the second element, as the first is redundant with it.

fe1-web/store/Storage.ts
pierluca: As I see it, the redux store will take care of managing everything, so there's no need.

Furthermore, this code is unused and only available as "debug" functionality to reset the entire system. I think it's okay to leave it with a warning.

fe1-web/__tests__/model/objects/Objects.test.ts
pierluca: Yes, I just haven't gotten around to really-really cleaning up the other non-MessageApi.test.ts files :)

PR #216, opened 2021-02-19
Ed25519 key generation

gnarula:
Generates user's Ed25519 KeyPair using Tink and stores it as a SharedPreference.

Allows injecting a keyset manager in the ViewModel for cryptographic operations.

### DIFF
--- a/app/src/main/java/com/github/dedis/student20_pop/PoPApplication.java
+++ b/app/src/main/java/com/github/dedis/student20_pop/PoPApplication.java
@@ -1,12 +1,9 @@
 package com.github.dedis.student20_pop;
 
-import android.annotation.SuppressLint;
 import android.app.Application;
 import android.content.Context;
-import android.content.SharedPreferences;
 import android.os.Handler;
 import android.os.Looper;
-import android.util.Log;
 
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/github/dedis/student20_pop/PoPApplication.java
```diff
@@ -65,43 +61,17 @@ public void onCreate() {
...
-    sp.edit().putString(SP_PERSON_ID_KEY, person.getId()).commit();
-  }
+//  @SuppressLint("ApplySharedPref")
+//  @Override
+//  public void onTerminate() {
+//    super.onTerminate();
+//    SharedPreferences sp = this.getSharedPreferences(TAG, Context.MODE_PRIVATE);
+//    // Use commit instead of apply for information to be stored immediately
+//    sp.edit().putString(SP_PERSON_ID_KEY, person.getId()).commit();
+//  }
```
pierluca, line 74: if this and the above is dead code, do you want to remove it?

gnarula: Going to remove all dead code in a later commit

PR #218, opened 2021-02-19
fe1-web: implemented new Hash logic

pierluca:


fe1-web/model/objects/Hash.ts
```diff
@@ -9,10 +9,9 @@ export class Hash extends Base64Data {
      * @return resulting hash
      */
   public static fromStringArray(...data: string[]): Hash {
-    let str = '';
-    data.forEach((item) => { str += `"${Hash.escapeString(item)}",`; });
-    // remove the last comma and add square brackets around
-    str = `[${str.slice(0, -1)}]`;
+    const str = data
+      .map((item) => Hash.computeByteLength(item).toString() + item)
+      .join('');
```
RaulinN, line 14: Simple question, any reason why we agreed to put the length before the string rather than after? 🤔 

PS: You don't even need to `.toString()` with JS ^^. Won't make much of a difference thought

RaulinN (dismissed):

fe1-web/model/objects/Hash.ts
pierluca: "PS change" applied.
Ask me the question on Monday 😋

PR #221, opened 2021-02-19
Network layer improvements + Eslint warnings removal

RaulinN:
- 96bf7d9 : added most of the last websocket suggestions of PR #212 
- d56671d : remove eslint warnings in tests 
- cfb8ada : remove eslint warnings in `LocalMockServer.js` + clarified logs
- b6fa507 : applied #212 last commit suggestions
- 4f17306 : added `OperationError` when server sends an error back


fe1-web/network/NetworkConnection.ts
```diff
@@ -7,27 +7,19 @@ import { NetworkError } from './NetworkError';
...
-    );
-  }
+  console.log('No response handler was provided to manage message : ', message);
 };
 
 const WEBSOCKET_READYSTATE_INTERVAL_MS = 10;
 const WEBSOCKET_READYSTATE_MAX_ATTEMPTS = 100;
 
-const WEBSOCKET_MESSAGE_TIMEOUT_MS = 10000; // 10 seconds max round-trip time
+const WEBSOCKET_MESSAGE_TIMEOUT_MS = 1000; // 10 seconds max round-trip time
```
RaulinN, line 16: corrected. back to 10k

fe1-web/model/network/JsonRpcResponse.ts
```diff
@@ -16,7 +16,7 @@ export class JsonRpcResponse {
   public readonly id: number;
 
   constructor(resp: Partial<JsonRpcResponse>) {
-    this.id = resp.id || UNDEFINED_ID;
+    this.id = (resp.id === undefined) ? UNDEFINED_ID : resp.id;
```
pierluca, line 19: Sorry for introducing this bug :'(

fe1-web/network/NetworkConnection.ts
```diff
@@ -61,10 +53,28 @@ export class NetworkConnection {
...
+        clearTimeout(pendingResponse.timeoutId);
         this.payloadPending.delete(response.id);
 
-        pendingResponse.resolvePromise(response);
-        this.onMessageHandler(response);
+        // reject the promise if the response is negative
+        if (response.isPositiveResponse()) {
+          pendingResponse.resolvePromise(response);
+          try {
+            this.onMessageHandler(response);
```
pierluca, line 64: do we need a handler at all here if the promise gets resolved?

fe1-web/network/NetworkConnection.ts
```diff
@@ -61,10 +53,28 @@ export class NetworkConnection {
...
+          } catch (e) {
+            console.error('Exception encountered when giving response : ', response, ' to its handler.\n', e);
+          }
+        } else {
+          // Note : impossible to have an undefined error at this stage due to isPositiveResponse()
+          pendingResponse.rejectPromise(
```
pierluca, line 70: we could consider returning a specific kind of error here, which could be constructed with the error code + description (i.e. as 2 separate fields).
This way, it'll be easier to process at a higher level.
E.g. in the GUI "Could not execute operation: ${response.error.description}"

pierluca (commented):
One minor remark :)

fe1-web/network/NetworkConnection.ts
RaulinN: So. In the last version (see last commit), I left it in. In my opinion, some JsonRpcResponse could need a handler (e.g. store something in localStorage)

RaulinN:
Regarding the last lingering messages (resolved in the last commit) of #212:

> here we're assuming that ALL the messages are JsonRpcResponses, some might be JsonRpcRequests (e.g. "Broadcast") 😅

If left some blank space for JsonRpcRequests handling in the last commit ^^' (since fall's student group barely touched this subject => not implemented in any subproject). See the "note for students". What do you think about this solution?

RaulinN:
> if the websocket disconnects, is there any retry-mechanism to bring it back up alive?
If not, the first network crash will render the application unusable.

Added with the private method `establishConnection()`

fe1-web/network/NetworkConnection.ts
RaulinN: Good idea! See commit 4f173068

fe1-web/network/NetworkConnection.ts
```diff
@@ -23,65 +24,83 @@ interface PendingResponse {
...
-            + `\t- error code : ${response.error.code}`
-            // @ts-ignore
-            + `\t- description : ${response.error.description}`,
-          );
         }
+      } else {
+        // if we have a request which needs to be forwarded
+
+        // Note to students : here should be the logic regarding witness message forwarding
+        getNetworkManager().sendPayload(parsedMessage); // FIXME ignoring promise for now
```
pierluca: here we should already be parsing Broadcast messages.. 😉

fe1-web/network/NetworkConnection.ts
```diff
@@ -23,65 +24,83 @@ interface PendingResponse {
...
+        }
       }
 
-      if (response.id !== UNDEFINED_ID) {
-        const pendingResponse = this.payloadPending.get(response.id);
-        if (pendingResponse === undefined) {
-          throw new NetworkError(`Received a response which id = ${response.id} does not match any pending requests`);
-        }
+      if (parsedMessage instanceof JsonRpcResponse) {
+        // if we have a response
```
pierluca: I would refactor for readability this as 
```
try {
  let parsedMesssage: JsonRpcResponse | JsonRpcRequest = this.parseWsMessage(message.data); // new method
  if (parsedMessage instanceof JsonRpcResponse) {
    this.processResponse(parsedMessage); // new method
  } else {
    this.processRequest(parsedMessage); // new method
  } 
} catch (e) {
  console.warn('Received invalid message: data.message');
}
```

fe1-web/network/NetworkConnection.ts
```diff
@@ -23,65 +24,83 @@ interface PendingResponse {
...
-    const ws = new W3CWebSocket(address);
+    this.ws = this.establishConnection(address);
+    this.address = this.ws.url;
+  }
+
+  private establishConnection(address: string): W3CWebSocket {
+    const ws: W3CWebSocket = new W3CWebSocket(address);
 
     ws.onopen = () => { console.info(`Initiating web socket : ${address}`); };
     ws.onmessage = (message: any) => {
```
pierluca, line 47: I would consider assigning methods. i.e.
```
ws.onopen = this.onConnectionOpen; // current lambda, extracted as a class method
ws.onmessage = this.onMessage; // current function, extracted as a class method
```

fe1-web/network/OperationError.ts
```diff
@@ -0,0 +1,20 @@
...
+ * (server sending back and error) from other JS errors.
+ */
+export class OperationError extends Error {
+  public readonly errorCode: number;
+
+  constructor(m: string, errorCode: number) {
+    super(m);
+    this.errorCode = errorCode;
+
+    // set prototype explicitly, needs to happen immediately after super() call
```
pierluca, line 12: the code violates the comment... you can put line 10 at the end of the constructor !

pierluca (changes requested):
Try to avoid eslint-disables everywhere in the code :-( 
Some warnings were entirely valid and need to be addressed (e.g. guard-for-in) 

Also, receiving a broadcast is still broken :)

RaulinN:
I guess you are talking about `LocalMockServer` eslint disables. Here's the ones I added:
- // eslint-disable-next-line new-cap (line 18) : webSocketServer defined in a library :( can't change that. I'll look if I can "rename it" using imports after my lecture ^^
- // eslint-disable-next-line prefer-const (line 82) : it was simply for simplicity ^^' (being able to comment/uncomment lines quickly and relaunch server). However it's not required I agree
- // eslint-disable-next-line guard-for-in,no-restricted-syntax : will fix it along with unsafe id parsing

fe1-web/network/OperationError.ts
RaulinN: woops!

fe1-web/network/NetworkConnection.ts
RaulinN: Fixed it.

Note : added 2 eslint-disable next to it for good reasons this time ^^

RaulinN:
@pierluca I implemented parseIncomingData the same way as before:
```javascript
    let parsedMessage: JsonRpcResponse | JsonRpcRequest;
    try {
      parsedMessage = JsonRpcResponse.fromJson(data);
    } catch (e) {
      try {
        parsedMessage = JsonRpcRequest.fromJson(data);
      } catch (ee) {
        throw new ProtocolError(`Answer is not valid Json : ${e} && ${ee}`);
      }
    }
```

I have searched online for an "optional" type in typescript (which does not exist). What I would like to do is something like this:
```javascript
let parsedMessage: Optional<JsonRpcResponse | JsonRpcRequest>;
parsedMessage = JsonRpcResponse.fromJson(data); // returns something like undefined if thrown
if (!parsedMessage) parsedMessage = JsonRpcRequest.fromJson(data);
if (!parsedMessage) throw ...

return parsedMessage;
```

Do you know of a way of doing something like this (aka. remove those two nested try catch blocks)?

fe1-web/network/NetworkConnection.ts
pierluca: @RaulinN this answers your quesoint.
If you process it in a this.parsedWsMessage you only have to deal with the try-catch in that function.

pierluca:
Re: "Do you know of a way of doing something like this (aka. remove those two nested try catch blocks)?"
The answer is no.
We can only rely on the JSON-Schema to distinguish them.
Either we use the JSON Schema validation logic into the parsing function, and pass an object to the `fromJson` static function or we have to deal with the try-catch.

RaulinN:
> @RaulinN this answers your quesoint.
If you process it in a this.parsedWsMessage you only have to deal with the try-catch in that function.

In this case the current version should meet the standards. We have 1 try-catch in onMessage (in case parseIncomingData throws) and 2 try-catch blocks in parseIncomingData (since either/both fromJson could throw (JsonSchema) and because we don't know before hand if we are dealing with a request or answer)

PR #224, opened 2021-02-20
Added Fe1-Web JSON Schema support / Downgraded JSON Schema version

pierluca:


fe1-web/model/network/method/message/data/witness/WitnessMessage.ts
```diff
@@ -16,21 +17,22 @@ export class WitnessMessage implements MessageData {
     this.message_id = msg.message_id;
 
     if (!msg.signature) throw new ProtocolError('Undefined \'signature\' parameter encountered during \'WitnessMessage\'');
-    // FIXME verify signature without the public key
-    // available? 0.o + uncomment 3 tests in "FromJson" test suite
+    // FIXME verify signature without the public key available? 0.o
```
pierluca, line 20: thought about this: need to pass the sender's signature into every fromJson() ... as second parameter.
It will enable (later on) the development of authentication / authorization

RaulinN (approved):

PR #229, opened 2021-02-23
go: remove old modules

gnarula:
Removes the modules which aren't needed after the refactor.

I've kept the `db` module since it might be useful to add persistence in future.


be1-go/README.md
```diff
@@ -2,109 +2,27 @@
...
 
-Once a message is stored, it shall not be updated, except for messages requesting witness signing. In that case, we 
-append the received witness signatures at the end of the original message's list of signatures.
 
 ## Unit-tests
 Some unit-tests exist directly in their corresponding packages, as per Go specification. They can be run with `go test -v ./...` (recursive from the top-level folder) or simply `go test` for the current package.
 
-## Basic testing
-We provide a very basic front-end in the file `test/index.html`. It does nothing but opens a websocket connection to the back-end server, and lets
-you send and receive messages. IT's accessible from the `address`:`port`/test URL. There you can the JSON-RPC requests examples given in `test/json_test_strings.txt`
```
pierluca, line 90: isn't this still relevant?

be1-go/go.mod
```diff
@@ -5,11 +5,10 @@ go 1.15
 require (
 	github.com/boltdb/bolt v1.3.1
```
pierluca, line 6: why not switch, as recommended by the students, to etcd-io/bolt ?

be1-go/test/json_test_strings.txt
```diff
@@ -1,23 +0,0 @@
-String for publishing LAO creation:
```
pierluca, line 1: why remove this ? 
The hashes are probably wrong now, but it's worth updating and could be useful as a starting point.

be1-go/test/index.html
```diff
@@ -1,92 +0,0 @@
-    <!DOCTYPE html>
```
pierluca, line 1: why remove this?

be1-go/test/README.md
```diff
@@ -1,3 +0,0 @@
-This package contains index.html which is used to run a mock server accepting strings input.
```
pierluca, line 1: and this ?

pierluca (changes requested):
I'm a bit perplexed. We're probably removing more than we should.

be1-go/README.md
```diff
@@ -2,109 +2,27 @@
...
-server configuration in `config/config_default.go`. The options in this file are : 
-* `MODE` the mode we want the server to run : either "o" for organizer or "w" for witness. Default to "o".
-* `ADDRESS` the IP address we want the server to run on. Default is "127.0.0.1" (localhost).
-* `PORT` the port we want the server to run on. Default is 8000.
-* `PKEY` the actor's public key. Default is "oui".
-* `FILE` the file we want the backend to store its database on. Default is "org.db" for an organizer and "wit.db" for a witness
 
-The server can then be accessed on address:port with the websocket protocol.
-You can generate the code documentation using the godoc command: `godoc -http:=6060`.
+You may build the `pop` CLI to interact with the server by executing `go build -o pop cli`.
```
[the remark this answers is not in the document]
gnarula: `./pop -h` to the rescue. I'll update the documentation to reflect that :)

be1-go/README.md
gnarula: I feel the effort to implement this has been replicated across all projects. Something https://hoppscotch.io would work just fine to test things out. Agreed the test_strings can be handy though. Adding them back

be1-go/go.mod
gnarula: Gotta check if the API is compatible or they've made breaking changes

gnarula:
Answering here since a lot of comments correspond to bolt/db module:

The gist is I've removed references to the DB in the README to reflect the state of the project as it stands now.

As to why persistence was removed? It was easier to keep persistence aside while working on the message structs. It shouldn't be hard to add it back now that the `message` module is stable.

There are still a few things I'd have to look - the schema may/may not change or we may chose to use etcd-io/bolt instead of bolt. As a result, I felt its best to remove that from the README to avoid ambiguity.

pierluca (approved):
I won't have time to re-review, but I'll unblock you by approving.
Whenever it gets clean-enough, make sure the doc is up to date and the sample strings work - they might not because of the hash (just thought of that)

PR #231, opened 2021-02-23
Integrated Github Action workflow across sub-systems

pierluca:


pierluca:
Consider using [step.continue-on-error](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) to make sure that we don't get (too) stuck due to failing steps.

PR #234, opened 2021-02-24
Front-end 1 - network & data ingestion

pierluca:


pierluca:
@RaulinN  still needs a bit of work, but let met know if you think this is viable.
I haven't gotten around to structuring the event reducer (I had in mind to have a list of event_ids, and an event_id->event map), but the rest should make the principles somewhat obvious.

pierluca:
1. Fix code smell detected by Sonar
2. Revert storage modification (as further refactoring is in progress)
3. Ensure logging in the incomplete functions

fe1-web/__tests__/network/MessageApi.test.ts
```diff
@@ -85,8 +85,8 @@ function checkDataCreateLao(obj: MessageData) {
   expect(data.name).toBeString();
   expect(data.name).toBe(mockEventName);
 
-  expect(data.creation).toBeNumber();
-  expect(data.creation).toBeGreaterThan(0);
+  expect(data.creation).toBeNumberObject();
+  expect(data.creation.valueOf()).toBeGreaterThan(0);
```
RaulinN, line 89: I wonder how it was working before! 🤔

RaulinN (changes requested):
Encountering a TypeError in tests though ^^'

fe1-web/__tests__/network/MessageApi.test.ts
pierluca: It was working because the mock data was in the 'wrong' type :)

pierluca:
As discussed, I'll merge this despite the TypeScript error, as it's only a side-effect of data persistence being initialized in the tests (and it probably shouldn't)

PR #236, opened 2021-02-24
Refactored LAO reducer

pierluca:
This refactors the LAO reducer to store only normalized data, while having ease of access & code cleanliness by using the redux toolkit


fe1-web/store/reducers/LaoReducer.ts
```diff
@@ -0,0 +1,83 @@
...
+  (state: LaoReducerState) => state.byId,
+  // Second input: current LAO id
+  (state: LaoReducerState) => state.currentId,
+  // Selector: returns a LaoState -- should it return a Lao object?
+  (laoMap: Record<string, LaoState>, currentId: string | undefined) => {
+    if (currentId === undefined || !(currentId in laoMap)) {
+      return undefined;
+    }
+
+    return laoMap[currentId];
```
RaulinN, line 69: Don't know about the complete list of LAOs yet, but I think it could be great idea to return a `Lao` instead of a `LaoState`. The reason for this is that we very often retrieve data from the lao when we access the storage for the current LAO. For example, I think option 1 makes more sense

```typescript
const id: Hash = useSelector(makeCurrentLao()).id; // id is a Hash
const id: Hash = new Hash(useSelector(makeCurrentLao()).id); // id is a string
```

pierluca:
Closing this PR. Comments have been integrated in the work being done on #247

PR #237, opened 2021-02-25
Removed messageData from messages sent

RaulinN:
Removes messageData from messages sent.

ℹ️  there's a strange buggy test which I commented. It doesn't make any sense to me why it does not compile

### DIFF
--- a/fe1-web/model/network/method/message/Message.ts
+++ b/fe1-web/model/network/method/message/Message.ts
@@ -71,4 +71,9 @@ export class Message {
       witness_signatures: (witnessSignatures === undefined) ? [] : witnessSignatures,
     });
   }
+
+  public toProtocolJson(): any {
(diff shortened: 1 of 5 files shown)


fe1-web/model/network/method/message/Message.ts
```diff
@@ -71,4 +71,9 @@ export class Message {
       witness_signatures: (witnessSignatures === undefined) ? [] : witnessSignatures,
     });
   }
+
+  public toProtocolJson(): any {
```
pierluca, line 75: To keep type safety , we could define an interface = Message without messageData and return it here

pierluca: This of course will require some adaptations elsewhere

pierluca (approved):

PR #238, opened 2021-02-25
Additional components/parts rework

RaulinN:
- 57f9554 : reworked Identity tab + template QR code 

ℹ️ More to come soon


fe1-web/components/QRCode.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { KeyPairStore } from 'store/stores';
+
+const QRCode = (props: IPropTypes) => {
+  const identity = KeyPairStore.getPublicKey().toString();
+  const { visibility } = props;
+
+  return (
+    // For now, we simply display a text block with the public key
+    (visibility) ? <TextBlock bold text={`ID (pk) :   ${identity}`} /> : null
+  );
```
pierluca, line 13: ternary operators are easily hard to read. Suggestion:

```suggestion
  // For now, we simply display a text block with the public key
  return (visibility)
    ? (<TextBlock bold text={`ID (pk) :   ${identity}`} />)
    : null;
```

fe1-web/components/eventList/LaoProperties.tsx
```diff
@@ -0,0 +1,58 @@
...
+import PropTypes from 'prop-types';
+import { Spacing } from 'styles';
+import TextBlock from 'components/TextBlock';
+import styleEventView from 'styles/stylesheets/eventView';
+import ListCollapsibleIcon from 'components/eventList/ListCollapsibleIcon';
+import { connect } from 'react-redux';
+import ParagraphBlock from 'components/ParagraphBlock';
+import { Lao } from 'model/objects';
+
+function buildLaoProperties(lao: Lao) {
```
pierluca, line 14: for readability, I would suggest renaming as `renderLaoProperties`
but then I wonder, do we need a separate function for this?

fe1-web/components/eventList/events/Event.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const { event } = props;
+  const { renderItemFn } = props;
+  const isOrganizer = false; // FIXME ?
+
+  const [toggleChildrenVisible, setToggleChildrenVisible] = useState(false);
+
+  const hasChildren = () => event.children && event.children.length !== 0;
+
+  const toggleChildren = () => {
+    if (hasChildren()) setToggleChildrenVisible(!toggleChildrenVisible);
```
pierluca, line 30: isn't this just:
```suggestion
    setToggleChildrenVisible(hasChildren() && !toggleChildrenVisible);
```
 ?

fe1-web/components/eventList/events/Event.tsx
```diff
@@ -0,0 +1,83 @@
...
+  };
+
+  const buildListCollapsibleIcon = () => (
+    <TouchableOpacity onPress={toggleChildren} style={{ textAlign: 'right' } as ViewStyle}>
```
pierluca, line 34: Separate style variables?

Is there maybe even something predefined for right-aligned text?

pierluca (commented):
Looks good to me overalls, but there are a few code-style-things that could use a clean-up, based on both my read-through and Sonar's complaints :)

fe1-web/components/eventList/events/Event.tsx
RaulinN: @pierluca I remember testing this yesterday without success. Tried it again just to make sure. Turns out it works if I write it like this
```javascript
setToggleChildrenVisible(!!hasChildren() && !toggleChildrenVisible); // double inversion
```

This is strange since `hasChildren` (highlighted above) has to return a boolean even with the javascript strange boolean operation standards we talked about yesterday 🤔

fe1-web/components/eventList/LaoProperties.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const { lao } = props;
+  const [toggleChildrenVisible, setToggleChildrenVisible] = useState(false);
+
+  const toggleChildren = () => (setToggleChildrenVisible(!toggleChildrenVisible));
+
+  return (
+    <>
+      <TextBlock bold text="Lao Properties" />
+      <View style={[styleEventView.default, { marginTop: Spacing.s }]}>
+        <TouchableOpacity onPress={toggleChildren} style={{ textAlign: 'right' } as ViewStyle}>
```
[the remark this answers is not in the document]
RaulinN: We could, but this is the only place where we use these styles 🤔 . Should I still put them in a variable above?

fe1-web/components/eventList/LaoProperties.tsx
RaulinN: I separated this function from the main body in prevision for the "organizer LaoProperties". We will probably have something along the lines of `renderStandardProperties` and `renderOrganizerProperties`

pierluca (approved):
Awesome !

PR #239, opened 2021-02-25
message: return the original data buffer if it exists

gnarula:
Need to return the original buffer for signature verification to succeed in client.

### DIFF
--- a/be1-go/message/message.go
+++ b/be1-go/message/message.go
@@ -52,9 +52,16 @@ func NewMessage(sender PublicKey, signature Signature, witnessSignatures []Publi
 }
 
 func (m Message) MarshalJSON() ([]byte, error) {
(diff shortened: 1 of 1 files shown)


pierluca (commented):
Seems reasonable, but is there any chance that we could modify the actual `Data` field (for some reason) and then forget to modify the `RawData` , thus leading to inconsistent output?

pierluca (approved):
Then again if we're creating, I assume the `RawData` will not be filled.
I'll approve this , feel free to merge if I answered myself correctly 😅

gnarula:
> Then again if we're creating, I assume the `RawData` will not be filled.
> I'll approve this , feel free to merge if I answered myself correctly 😅

Yes, precisely :)

PR #244, opened 2021-02-26
go: linting changes

gnarula:
- Commented exported structs and methods
- Commented out portion of db package and tests within it which aren't as relevant. To be taken up when we work on persistence
- Removed the events package since it's not relevant anymore.
- Changes the linting tool from `golint` to [staticcheck](https://staticcheck.io/) since the former is deprecated


pierluca (commented):
I'm a bit concerned by the commenting out of so much code.
Shouldn't we be fixing persistence instead? That's a core feature of the server.

gnarula:
> I'm a bit concerned by the commenting out of so much code.
> 
> Shouldn't we be fixing persistence instead? That's a core feature of the server.

Yes, but then the PR won't just be about making the linter happy.

IMHO the persistence stuff deserves a separate PR and this puts us in a position where we can start getting some feedback from the CI going forward.

pierluca (approved):
Convinced. Do make it clear in the code though (todo comment) and restore persistence ASAP

PR #253, opened 2021-03-03
db: Add Repository interface and an implementation backed by SQLite

gnarula:
This PR introduces a Repository interface that the `Hub` can use as an abstraction to add/update messages to the datastore.

Future work would be to adapt the SQLite implementation in `organizerHub`.


pierluca:
I'm not sure I will be able to review it before going on holidays. I'll see if I can tomorrow.
You might want to ask someone else in the project team (@lhmerino perhaps ?)

pierluca:
Regarding Sonar's messages:
- duplicated sections seem like a good target for refactoring
- test names are not compliant with Sonar rule: it might make sense to either fix them (based on whatever Golang recommendations are) or update the regex here https://sonarcloud.io/organizations/dedis/rules?open=go%3AS100&rule_key=go%3AS100

be1-go/db/sqlite.go
```diff
@@ -0,0 +1,199 @@
...
+)
+
+type sqlite struct {
+	conn *sql.DB
+}
+
+var (
+	DDL = `
+	CREATE TABLE messages (
+		channel_id text, message_id text, timestamp integer, message text,
```
pierluca, line 23: I would also create an index on message_id , but that might be something for the students to implement as part of the improved  catch-up logic :)

be1-go/db/sqlite_test.go
```diff
@@ -0,0 +1,183 @@
...
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestSQLite_NewRepository(t *testing.T) {
+	repo, err := createRepo("test.db")
+	require.NoError(t, err)
+
+	err = repo.Close()
+	require.NoError(t, err)
```
pierluca, line 20: this could nearly be satisfied by two functions that merely return nil as an error.
should we at least check that the test.db file gets created?

pierluca (commented):
Had a look at a few things (randomly sampled), I do have a few comments, but overalls it seems pretty neat 👍 

I'd ask the students to review this code (and others), it's a good way for them to understand what's going on and how things fit together.

be1-go/db/sqlite.go
```diff
@@ -0,0 +1,199 @@
...
+	msgBuf, err := json.Marshal(msg)
+	if err != nil {
+		return xerrors.Errorf("failed to marshal message: %v", err)
+	}
+
+	tx, err := s.conn.Begin()
+	if err != nil {
+		return xerrors.Errorf("failed to create transaction: %v", err)
+	}
+
```
flmichel, line 177: Maybe we should also do a `defer tx.Rollback()` but I'm not sure since there is only one call of `Rollback()`

be1-go/db/sqlite_test.go
```diff
@@ -0,0 +1,183 @@
...
+
+	channelID := "12345"
+
+	_, err = addMessages(repo, channelID, 100)
+	require.NoError(t, err)
+
+	messages, err := repo.GetMessages(channelID)
+	require.NoError(t, err)
+
+	require.Equal(t, 100, len(messages))
```
flmichel, line 90: I don't think it's important at all but would it be a good idea to also check if the messages are in ascending order of their timestamp?

flmichel (commented):
The code looks very good to me! I just have 2 minor comments.

be1-go/db/sqlite_test.go
gnarula: Thanks for suggesting that! Updated the tests :)

be1-go/db/sqlite.go
gnarula: Handled more gracefully using `s.WithTransaction`

flmichel (approved):

PR #254, opened 2021-03-05
go: return error on duplicate lao creation

gnarula:
Fixes a bug I found while trying to replay lao creation messages during testing.

### DIFF
--- a/be1-go/hub/organizer.go
+++ b/be1-go/hub/organizer.go
@@ -199,7 +199,7 @@ func (o *organizerHub) createLao(publish message.Publish) error {
 	encodedID := base64.StdEncoding.EncodeToString(data.ID)
 	if _, ok := o.channelByID[encodedID]; ok {
 		return &message.Error{
-			Code:        -4,
+			Code:        -3,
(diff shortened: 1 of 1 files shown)


be1-go/hub/organizer.go
```diff
@@ -197,6 +197,12 @@ func (o *organizerHub) createLao(publish message.Publish) error {
 	}
 
 	encodedID := base64.StdEncoding.EncodeToString(data.ID)
+	if _, ok := o.channelByID[encodedID]; ok {
+		return &message.Error{
+			Code:        -4,
```
flmichel, line 202: Shouldn't it be '-3' since it's a duplicate? [https://github.com/dedis/student_21_pop/blob/af8656538944c29f95dc5b1699716fa5656b2d1e/protocol/answer/negativeAnswer/error.json#L18](url)

flmichel (changes requested):

be1-go/hub/organizer.go
gnarula: Nice catch! Fixing this

flmichel (approved):

PR #257, opened 2021-03-06
added simple wallet object class

CaiMusso:
Created the class SimpleWalletObject which represents a basic wallet storing a map of LAO/RollCall identifier to generated KeyPairs for proof of presence. The basic logic behind this wallet is explained in the attached PDF. 
[SimpleWallet.pdf](https://github.com/dedis/student_21_pop/files/6094310/SimpleWallet.pdf)

Currently the functionality has a bug which causes methods to return **undefined** when retrieving the above mentioned map, which is the reason why the PR is a draft.


pierluca:
You might want to add (at least temporarily) the code needed to reproduce the bug you observe, otherwise it'll be very hard for anyone to help you.

Even better would be a Jest test case (as those in the __tests__ folder) reproducing the issue, there's value in that even if you don't end up keeping it.

CaiMusso:
**OUTDATED** - see https://github.com/dedis/student_21_pop/pull/264

PR #258, opened 2021-03-06
Setup data ingestion pipeline - network is functional

pierluca:
First of all, I apologize for the massive PR, but it was really needed to get to the desired level of functionality.

To make sense of this whole thing, I would suggest to read as follows:
1. Look at `ingestion/README.md` and the files it references, that's where most of the magic happens (and it still needs to be completed)
2. Go through the reducers to understand what they do
3. Look at Launch.tsx and the changed files in the `network` folder, because that's where the magic comes together.

With this PR, the network "sort of" works with the Be1-Go. I have been able to test:
1. Create a LAO
2. Subscribe to it, catchup
3. Create a roll call (sort-of)

Known issues to be addressed:
1. Right now, the connection "glue" code is in Launch.tsx and in proof-of-concept stage. I would make sure that we have a clean "establishLaoConnection" that does the connection + subscribe + catchup and makes sure the reducer is properly setup.
Most of the other code should be pretty clean though.

2. I had to manually "add" the LAO upon creation in the reducer: we shouldn't need to do so, as the logic above should be enough to start cleanly after a connection. If we catchup after connecting, we should get the lao/create message and instantiate everything then.


RaulinN:
As said during a meeting, it seems to be good start!

I corrected the first issue. However, the second one is strange to me. I think the watcher simply doesn't watch anything (or doesn't *see* anything that moves around in the storage). I can say that because in `Watcher.ts` line 31, `msgs` is *always* an empty array 🤔 . Moreover, for some reason, when creating a LAO, it is only visible in the local storage when the app is rebooted (which doesn't make any sense to me 🤔 ). Will try again later

RaulinN:
Update: I think I have finally found the source of the "bug"... It's not a bug ^^'

@pierluca When you create a lao, you are either connected to no lao (first time using the app => local storage empty) or connected to a previous lao (in the current lao storage) that should/will be replaced after the lao creation (so precisely do what you describe as a bug).

Why? Well let's see: if i understood correctly, you would want to wait for the catchup message to be processed in order for the client to open the lao. But that cannot work because of this (in `Watcher.ts`):
```typescript
if (!laoId) {
      // not connected to a LAO, return immediately
      return;
}

const msgState = getLaoMessagesState(laoId, state);
```

At this point (last line), the client cannot know that the stale lao *is being replaced* by a new one! Thus loads the unprocessed messages of the stale lao in the lines that follow.

Moreover, if we don't set the current lao at lao creation (thus before processing the cathcup), then a client that never connected to any lao will never be able to use the app!

RaulinN:
One thing we could do however is instead of using
```typescript
dispatch(connectToLao(lao.toState()));
```

in `MessageApi.ts`, rather move it to `CommunicationApi.ts`

PR #259, opened 2021-03-07
fe1-web: QR code showing public key of user in Identity window 

Adalsteinnjons:
The QR code is displayed in the identity window. It displays the text: 'ID (Public Key):' and displays the pk of the current user as a QR code.
image

It uses the library 'qrcode.react' to generate the QR code.
I tried to follow the styling guidelines and hope I did that correctly.


fe1-web/package-lock.json
```diff
@@ -45179,7 +45208,6 @@
           "resolved": "https://registry.npmjs.org/metro-react-native-babel-transformer/-/metro-react-native-babel-transformer-0.59.0.tgz",
           "integrity": "sha512-1O3wrnMq4NcPQ1asEcl9lRDn/t+F1Oef6S9WaYVIKEhg9m/EQRGVrrTVP+R6B5Eeaj3+zNKbzM8Dx/NWy1hUbQ==",
           "requires": {
-            "@babel/core": "^7.0.0",
```
Adalsteinnjons, line 45182: Not sure why this was removed

fe1-web/components/QRCode.tsx
```diff
@@ -1,23 +1,46 @@
...
+        </View>
+      </>
+    )
     : null;
 };
 
 const propTypes = {
   visibility: PropTypes.bool.isRequired,
+  size: PropTypes.number,
+  text: PropTypes.string,
```
RaulinN, line 35: Why did you choose to put text as optional? Can you imagine a case where you would display an "empty QRcode"?

RaulinN (commented):
Nice job!

---
Now regarding this,
> However the typecasting I do in the HTML block that gets returned doesn't seem right to me

According to a [github issue](https://github.com/microsoft/TypeScript/issues/31247) (scroll down a bit), you should define the type of component arguments  (`IPropTypes` in our case) yourself.
```javascript
type IPropTypes = {
  visibility: boolean,
  size: number,
  text: string,
};
```

fe1-web/components/QRCode.tsx
```diff
@@ -1,23 +1,46 @@
...
-    ? <TextBlock bold text={`ID (pk) :   ${identity}`} />
+    ? (
+      <>
+        <TextBlock text={String(text)} />
+        <View style={[styleContainer.topCenter, { padding: 10 }]}>
+          <QRCodeDisplay value={identity} size={Number(size)} />
+        </View>
+      </>
+    )
     : null;
```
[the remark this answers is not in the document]
Adalsteinnjons: Thank you for the info about declaring the types, this makes it much cleaner.
Also, the structure you suggest seems more sensible, I will change it. (I had the thought that usually you would want to specify what the QR code is displaying and else you could leave the `text=` value empty)

fe1-web/parts/connect/ConnectQRCode.tsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
```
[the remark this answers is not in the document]
Adalsteinnjons: Correct, this file has no purpose. It was an initial file I created for testing, forgot to delete it.

fe1-web/components/QRCode.tsx
```diff
@@ -1,23 +1,46 @@
...
-  // For now, we simply display a text block with the public key
+  // Displays a QR code with the Public key of the current user
   return (visibility)
-    ? <TextBlock bold text={`ID (pk) :   ${identity}`} />
+    ? (
+      <>
+        <TextBlock text={String(text)} />
+        <View style={[styleContainer.topCenter, { padding: 10 }]}>
+          <QRCodeDisplay value={identity} size={Number(size)} />
+        </View>
```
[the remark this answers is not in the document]
Adalsteinnjons: I really tried to use something that existed already but wasn't able to get it looking the way I wanted.
If we use the `topCenter` it looks like this:
image

If we use `anchoredCenter` then it looks like this:
image

Which I found looked better since it was near the text and not centered in the middle of the page.

But the same outcome could be made by overwritting it in the component like this:
`<View style={[styleContainer.anchoredCenter, { padding: 10, justifyContent: 'flex-start' }]}>
          <QRCodeDisplay value={identity} size={Number(size)} />
        </View>`
Would that be better?

fe1-web/components/QRCode.tsx
Adalsteinnjons: The text value was about what would be displayed above the QR code, so usually something like: "ID (Public Key):", 
I made it optional because I thought sometimes it would be nice to just display the QR code and no text.
But this is redundant if I remove the `TextBlock` out of the component as you mentioned above.

fe1-web/components/QRCode.tsx
```diff
@@ -14,33 +13,30 @@ const QRCode = (props: IPropTypes) => {
...
     ? (
-      <>
-        <TextBlock text={String(text)} />
-        <View style={[styleContainer.topCenter, { padding: 10 }]}>
-          <QRCodeDisplay value={identity} size={Number(size)} />
-        </View>
-      </>
+      <View style={[styleContainer.anchoredCenter, { padding: 10, justifyContent: 'flex-start' }]}>
+        <QRCodeDisplay value={identity} size={Number(size)} />
+      </View>
```
RaulinN, line 22: You don't need the explicit cast anymore since you defined `IPropTypes` manually :)

fe1-web/components/TextBlock.tsx
```diff
@@ -15,20 +15,24 @@ const styles = StyleSheet.create({
...
+  const { visibility } = props;
 
-  return (
-    <Text style={(bold) ? styles.textBold : styles.textStandard}>{text}</Text>
-  );
+  return (visibility)
+    ? (
+      <Text style={(bold) ? styles.textBold : styles.textStandard}>{text}</Text>
+    ) : null;
 };
```
RaulinN, line 24: It is good practice to take a look at [SonarCloud report](https://sonarcloud.io/project/issues?id=dedis_student_21_pop_fe1&pullRequest=259&resolved=false&types=CODE_SMELL). Here, it gives you a "major warning"! ^^'

RaulinN (changes requested):

RaulinN (approved):
LGTM!

CaiMusso (approved):
Looks great !

PR #260, opened 2021-03-08

CaiMusso:
The LAO properties component in the event list now offers the possibility to **edit** the name of the LAO and the creation time by using the implemented edit text buttons.
This implementation uses the EdiText library from react-editext. 

NOTE - the modifications are purely on a **visual** level, no back-end implementation is done and thus the actual **lao.name** property does not change on the onSave() call of the button. For the moment, the new name or creation are simply logged on the console.

NOTE - for the moment this modification is available to all members (organiser, witnesses, attendees) which should **not** be the case. A future commit will allow this action only to the organiser.

Capture d’écran 2021-03-08 à 16 26 32


RaulinN:
Good start! I fully realise that this is a work in progress, however, I can still give you a few pieces of advice:
- I will have trouble giving you useful info because I cannot compile the project! You omitted (at least) the `package.json` file. You should commit every file that is not ignored by the `.gitignore` for other people to be able to run your code.
- ~Moreover, I see a lot of ESlint errors. These take very little time to fix. You can see them directly in your editor or by running `npm run eslint`. Sometimes, eslint is also able to resolve them by itself (see message in `npm run eslint)~ *I see you already corrected that. Keeping it there in case other find it useful*
- There's still an issue with a missplaced `}` that eslint found

RaulinN:
Oh! We will have an issue right there! I see your editor is using the default ESlint settings. Take a look online at how to setup your editor in order to make use of the file `.eslintrc` we have. The github action is also configured with our custom eslint config, thus everything will fail.

I would suggest going back to an old commit on your branch to a point where you did not modify the whole project ^^'

CaiMusso:
> Good start! I fully realise that this is a work in progress, however, I can still give you a few pieces of advice.
> 
> * I will have trouble giving you useful info because I cannot compile the project! You omitted (at least) the `package.json` file. You should commit every file that is not ignored by the `.gitignore` for other people to be able to run your code.
> * ~Moreover, I see a lot of ESlint errors. These take very little time to fix. You can see them directly in your editor or by running `npm run eslint`. Sometimes, eslint is also able to resolve them by itself (see message in `npm run eslint)~ _I see you already corrected that. Keeping it there in case other find it useful_
> * There's still an issue with a missplaced `}` that eslint found

Hi @RaulinN I tried to install the airbnb ESlint configuration following the guide I found on the readme (https://medium.com/pvtl/linting-for-react-native-bdbb586ff694). Unfortunately still doesn't work. The problem is I always just uploaded one file and thus I did it from the browser (never from the command line) so could be I didn't have a previously functioning commit. I reinstalled all the project from the beginning and then installed the ESlint following the guide but unfortunately still doesn't work.. do you have any advice? Thank you very much

CaiMusso:
**This branch was deleted because incompatibility new branch with same pull request at** https://github.com/dedis/student_21_pop/pull/261

PR #261, opened 2021-03-08

CaiMusso:
This new **pull request is a copy** of https://github.com/dedis/student_21_pop/pull/260#issue-586867524 (which is now closed). A new PR was created in **new branch** because old branch settings were not compatible with project settings. 

The LAO properties component in the event list now offers the possibility to edit the name of the LAO and the creation time by using the implemented edit text buttons.
This implementation uses the EdiText library from react-editext.
NOTE - the modifications are purely on a visual level, no back-end implementation is done and thus the actual lao.name property does not change on the onSave() call of the button. For the moment, the new name or creation are simply logged on the console.
NOTE - for the moment this modification is available to all members (organiser, witnesses, attendees) which should not be the case. A future commit will allow this action only to the organiser.
Capture d’écran 2021-03-09 à 00 28 39


RaulinN:
Soooo mister @CaiMusso! Please try to not nuke the project when pushing! ^^' You updated (manually?) npm packages versions. Unless it is a problem on my end, it seems that the project does not compile anymore ^^. Reverting the package-lock to an ancient version corrects it.

fe1-web/components/eventList/LaoProperties.tsx
```diff
@@ -10,12 +10,38 @@ import { useSelector } from 'react-redux';
 import ParagraphBlock from 'components/ParagraphBlock';
 import { Lao } from 'model/objects';
 import { makeCurrentLao } from 'store/reducers';
+import EdiText from 'react-editext';
 
 function renderProperties(lao: Lao) {
+  const style = {
+    fontFamily: 'Helvetica Bold',
```
RaulinN, line 17: Screenshot 2021-03-10 at 01 34 55


Any reason why you decided to change the font for those two lines only? ^^'

CaiMusso: Hi no real reason actually. I wanted to change the one you posted here in this picture because it was Times New Roman instead of Helvetica. Then I changed it to bold, but I'll put it back to normal Helvetica!

CaiMusso:
> Soooo mister @CaiMusso! Please try to not nuke the project when pushing! ^^' You updated (manually?) npm packages versions. Unless it is a problem on my end, it seems that the project does not compile anymore ^^. Reverting the package-lock to an ancient version corrects it.

https://dedis.slack.com/archives/C01NL2H4L9M/p1615367027025600

PR #263, opened 2021-03-09
Implemented a base channel

flmichel:
1. Implement a generic channel that can be use for any type of channel (Lao channel, election channel...).
2. Update the actual Lao channel by using the new generic channel (base channel).

### DIFF
--- a/be1-go/hub/organizer.go
+++ b/be1-go/hub/organizer.go
@@ -198,12 +198,11 @@ func (o *organizerHub) createLao(publish message.Publish) error {
 
 	encodedID := base64.StdEncoding.EncodeToString(data.ID)
 
-	laoCh := laoChannel{
-		channel: "/root/" + encodedID,
-		clients: make(map[*Client]struct{}),
-		inbox:   make(map[string]message.Message),
-	}
+	laoChannelID := "/root/" + encodedID
 
+	laoCh := laoChannel{
+		createBaseChannel(o, laoChannelID),
+	}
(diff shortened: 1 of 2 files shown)


be1-go/hub/organizer.go
```diff
@@ -214,68 +213,26 @@ func (o *organizerHub) createLao(publish message.Publish) error {
...
 
 func (c *laoChannel) Subscribe(client *Client, msg message.Subscribe) error {
-	log.Printf("received a subscribe with id: %d", msg.ID)
-	c.clientsMu.Lock()
-	defer c.clientsMu.Unlock()
-
-	c.clients[client] = struct{}{}
-
-	return nil
+	return c.baseChannel.Subscribe(client, msg)
```
gnarula, line 220: The nice thing about embedding is the struct that embeds a parent struct as the ability to call methods on the embedded struct. Have a look at https://play.golang.org/p/iNTjT13SaIx for an example.

gnarula (changes requested):
Thanks for the PR! I think it's fine if only baseChannel implements Subscribe, Unsubscribe and Catchup. The structs embedding it will automatically pass on the call to those methods - we don't need to define them again. Please have a look at the playground link I shared for an example.

be1-go/hub/organizer.go
flmichel: Thank you, I removed the methods in laoChannel.

gnarula (approved):
LGTM!

PR #265, opened 2021-03-10
Add json schemas for elections

flmichel:
Schema for :
- ElectionSetup
- CastVote
- ElectionEnd 
- ElectionResult

### DIFF
--- a/protocol/query/method/message/data/dataElectionSetup.json
+++ b/protocol/query/method/message/data/dataElectionSetup.json
@@ -67,7 +67,7 @@
 							"type": "string",
 							"$comment": "ballot option"
 						},
-						"minItems": 1,
+						"minItems": 2,
 						"uniqueItems": true
 					},
 					"write_in": {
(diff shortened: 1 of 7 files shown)


pierluca:
@flmichel Feel free to invite some of your project team mates to review, ideally one from each tech stack :)

protocol/query/method/message/data/dataElectionSetup.json
```diff
@@ -0,0 +1,87 @@
...
+						"$comment": "supported voting method"
+					},
+					"ballot_options": {
+						"description": "[Array[String]] ballot options",
+						"type": "array",
+						"items": {
+							"type": "string",
+							"$comment": "ballot option"
+						},
+						"minItems": 1,
```
Adalsteinnjons, line 70: Perhaps minItems should be set to 2? Since a question with only 1 possible answer is not really useful, but maybe this doesn't have to be enforced at the protocol level

Adalsteinnjons (commented):
Looks good to me, seems to match the specifications. The only comment I added was about perhaps having 2 required ballot options per question, but I'm not 100% sure on that.

protocol/query/method/message/data/dataElectionSetup.json
flmichel: I agree, it makes more sense. I updated it!

lhmerino (approved):
Very nice job! It looks very thorough 👍

PR #267, opened 2021-03-11

CaiMusso:
Purpose of this PR
--------------------------
The LAO properties component in the event list now offers the possibility for the **organiser** of the LAO to edit the name of the LAO. 

----------------------------------------------------------------------------------------------------------------------------

+ This implementation uses the EdiText library from react-editext. 
+ The creation date has been formatted to be human-readable. 
+ Soon also the creation date may be modified by the organizer once an alternative date-time picker has been implemented in the typescript front-end.

**NOTE** - the modifications at the moment on a **visual level**, the implementation is not currently communicating with the back-end and thus the actual lao.name property does not change on the onSave() call of the button.

https://user-images.githubusercontent.com/47753346/113446263-beaac580-93f7-11eb-8c5a-0ffdeb253ef2.mov


Adalsteinnjons (commented):
I tried it and it looks good and works well. I have 3 small comments, but I'm not the expert here so take them with a grain of salt:
- The date should ideally be formatted and shown as a human-readable date, a date picker would be ideal to change it, currently, there is only a date picker for web implemented (see `parts/lao/organizer/eventCreation/events/CreateMeeting.tsx`), I will be looking into implementing a react native datepicker next week, so I could look into this
- Font info should probably be stored globally (We have typography.js, see also the [Guidelines](https://thoughtbot.com/blog/structure-for-styling-in-react-native))
- Another design approach would be to just have it as a `TextInput` field when the person is the organizer,  then have an edit toggle, similar how it is in `parts/lao/Identity.tsx`, not sure which is better and if this matters but this would mean one less dependency

CaiMusso:
> I tried it and it looks good and works well. I have 3 small comments, but I'm not the expert here so take them with a grain of salt:
> 
> * The date should ideally be formatted and shown as a human-readable date, a date picker would be good to change it, currently, there is only a date picker for web implemented (see `parts/lao/organizer/eventCreation/events/CreateMeeting.tsx`), I will be looking into implementing a react native datepicker next week, so I could look into this
> * Font info should probably be stored globally (We have typography.js, see also the [Guidelines](https://thoughtbot.com/blog/structure-for-styling-in-react-native))
> * Another design approach would be to just have it as a `TextInput` field when the person is the organizer,  then have an edit toggle, similar how it is in `parts/lao/Identity.tsx`, not sure which is better and if this matters but this would mean one less dependency

Thank you very much @Adalsteinnjons for the review! very helpful! Regarding the EdiText/TextInput problem I think we could keep it like this for the moment, I find it intuitive and could also be useful in other parts of the code in the future. Regarding the other remarks: 
+ I moved the style in a global context ( typography.js ) 
+ I am looking into the date time picker to make the UI look cleaner. Absolutely right on the "human-readable date". I agree it doesn't make much sense to have it in the current format, thanks for pointing this out!

 I will push a new commit once is done

RaulinN (changes requested):
From the [UI specification document](https://docs.google.com/document/d/1aVsCYj1vrTxh-V6CBrU4tRGANrC2wve47MSIXPliVbU/edit#heading=h.kpqvapi0008v): 

> [...] While in property edit mode a delete (-) button appears to the right of each listed witness. [...]
>
> In addition, while in property edit mode, an Add Witness button appears below the current witnesses if any.  Tapping Add Witness opens a camera display with a prompt, [...]

Don't forget (one of you two) to add this feature at some point to be synchronized with fe2! :)

PR #268, opened 2021-03-11
Start of be2 cleanup

RaulinN:
Here's the first step regarding be2-scala refactoring:
- 33da60a : Copied archived be2-scala project into current version. **No need to review this commit**
- d07dc34 : Added a `model` folder following all other sub-projects structure. **A few notes**:
   - I tried to take the best of both worlds (between fe2 and fe1)
   - The `data/` folder is very similar to both projects. I also defined types which are not present in fe2
   - I decided to follow fe1 path regarding the data builder since it seemed to resemble more the Scala way of doing it
- fc1531f8 : Added `Message` level model (same as both front-ends)
- 15cf438a : *Applied Céline's review comments*
- 6a005c2e : *Moved `Parsable` trait to its rightful place above in the hierarchy*
- 7deab3e7 : Added `Method`  level model (same as both front-ends but extracted the redundant "message" of fe2)
    - ⚠️ contains an error (forgot to add companion objects to all files. See cf7e9fb6)
- bad4298b : *Moved Matching trait higher in the hierarchy*
- 5eda5e0e : Added requests/responses creation model
- 90ff1961 : *Renamed ParamsSimple into Params (because `ParamsWithMessages` are (extends) `Params` and not simple params)*
- cf7e9fb6 : Added missing companion objects for method models (cannot enforce `buildFromJson` without companion object since the method is "static")

*Note*: I noticed that fe2 is *exactly* following the protocol hierarchy. Some of them were not needed anymore/simplified due to protocol changes (e.g. `MessageGeneral` becomes `Message`, `GenericMessage` becomes `JsonRpcMessage`, and changed `Error` messages a bit)

---
ℹ️ The data builder is inside an object (which is probably not the best way of doing it, but I haven't found a better way for now)

### DIFF
--- a/be2-scala/.github/workflows/be2-scala.yml
+++ b/be2-scala/.github/workflows/be2-scala.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches: [ work-be2-* ]
   pull_request:
-    branches: [ be2-scala ]
+    branches: [ master ]
 
 jobs:
   build:
--- a/be2-scala/.github/workflows/be2-scala.yml
+++ b/be2-scala/.github/workflows/be2-scala.yml
@@ -1,0 +1,29 @@
+name: PoP CI be2-scala
+
+on:
+  push:
+    branches: [ work-be2-* ]
+  pull_request:
+    branches: [ be2-scala ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up JDK 1.8
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+
+      - name: Run unit tests
+        run: sbt coverage test
+
+      - name: Coverage report generation
+        run: sbt coverageReport
+
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v1
+        with:
+          fail_ci_if_error: true
--- a/be2-scala/README.md
+++ b/be2-scala/README.md
@@ -1,0 +1,32 @@
+# student20_pop: be2-scala branch
+Proof-of-personhood, fall 2020: Scala language back-end
+
+[TOC]
+
+## Running the project
+
+There are two main possible ways of running the project :
+* import the project using IntelliJ
+* using **sbt**, execute `sbt compile` and `sbt run`
+
+---
+
+
+
+## External libraries
+
+The project relies on several sbt dependencies (external libraries) :
+
+- websockets : [**akka-http**](https://doc.akka.io/docs/akka-http/current/introduction.html) for websocket server. It uses [akka-streams](https://doc.akka.io/docs/akka/current/stream/index.html) as dependency.
+- database : [**leveldb**](https://github.com/codeborui/leveldb-scala) which relies on both [snappy](https://search.maven.org/artifact/org.xerial.snappy/snappy-java/1.1.7.3/jar) (for compression/decompression) and [akka-persistence](https://doc.akka.io/docs/akka/current/persistence.html)
+- Json parser : [**spray-json**](https://github.com/spray/spray-json) for Json encoding/decoding
+- encryption : [**tink**](https://github.com/google/tink/blob/master/docs/JAVA-HOWTO.md) to verify signatures
+- testing : [**scalatest**](https://www.scalatest.org/) for unit tests
+
+---
+
+
+
+## Coding convention
+
+Our coding guidelines can be found [here](https://docs.scala-lang.org/style/).
--- a/be2-scala/README.md
+++ b/be2-scala/README.md
@@ -1,5 +1,5 @@
-# student20_pop: be2-scala branch
-Proof-of-personhood, fall 2020: Scala language back-end
+# student21_pop: be2-scala branch
+Proof-of-personhood, spring 2021: Scala language back-end
 
 [TOC]
 
--- a/be2-scala/src/main/scala/ch/epfl/pop/crypto/Hash.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/crypto/Hash.scala
@@ -1,0 +1,70 @@
+package ch.epfl.pop.crypto
+
+import java.nio.ByteBuffer
+import java.nio.charset.StandardCharsets
+import java.security.MessageDigest
+import ch.epfl.pop.json.{Base64String, Hash, Key, Signature}
+
+import java.util.Base64
+
+object Hash {
+
+  /**
+   * Compute the id of the message, which is the Hash(msg||signature)
+   * @param msg the message in base64
+   * @param signature the signature of the message
+   * @return the id of the message
+   */
+  def computeMessageId(msg: Base64String, signature: Signature): Hash = {
+    val data = arrayRepr(msg, Base64.getEncoder.encodeToString(signature))
+    val id = getMessageDigest().digest(data.getBytes(StandardCharsets.UTF_8))
+    id
+  }
+
+  /**
+   * Compute the LAO id, which is Hash(organizer||creation||name)
+   * @param organizer the organizer of the LAO
+   * @param creation the creation timestamp of the LAO
+   * @param name the name of the LAO
+   * @return the id of the LAO
+   */
+  def computeLAOId(organizer: Key, creation: Long, name: String): Hash = {
+    val data = arrayRepr(Base64.getEncoder.encodeToString(organizer), creation.toString, name)
+    val id = getMessageDigest().digest(data.getBytes(StandardCharsets.UTF_8))
+    id
+  }
+
+
+  /**
+   * Compute the meeting id, which is Hash('M'||laoId||creation||name)
+   * @param laoID the id of the LAO
+   * @param creation the creation timestamp of the meeting
+   * @param name the name of the meeting
+   * @return the id of the meeting
+   */
+  def computeMeetingId(laoID: Hash, creation: Long, name: String): Hash = computeGenericId("M", laoID, creation, name)
+
+  /**
+   * Compute the roll-call id, which is Hash('R'||laoId||creation||name)
+   * @param laoID the id of the LAO
+   * @param creation the creation timestamp of the roll-call
+   * @param name the name of the roll-call
+   * @return the id of the roll-call
+   */
+  def computeRollCallId(laoID: Hash, creation: Long, name: String): Hash = computeGenericId("R", laoID, creation, name)
+
+
+  private def computeGenericId(typ: String, a: Array[Byte], l: Long, s: String): Hash = {
+    val data = arrayRepr(typ, Base64.getEncoder.encodeToString(a), l.toString, s)
+    val id = getMessageDigest().digest(data.getBytes(StandardCharsets.UTF_8))
+    id
+  }
+
+  private def arrayRepr(strings: String*): String =
+    "[" + strings.tail.foldLeft(escapeAndQuote(strings.head))((acc, s) => acc + "," + escapeAndQuote(s)) + "]"
+
+  private def escapeAndQuote(s: String): String = '"' + s.replace("\\", "\\\\").replace("\"", "\\\"") + '"'
+
+  private def getMessageDigest() = MessageDigest.getInstance("SHA-256")
+
+}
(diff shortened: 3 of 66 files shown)


be2-scala/.github/workflows/be2-scala.yml
```diff
@@ -0,0 +1,29 @@
+name: PoP CI be2-scala
+
+on:
+  push:
+    branches: [ work-be2-* ]
+  pull_request:
+    branches: [ be2-scala ]
```
celinecamacho, line 7: we should change this to master right?

be2-scala/README.md
```diff
@@ -0,0 +1,32 @@
+# student20_pop: be2-scala branch
+Proof-of-personhood, fall 2020: Scala language back-end
```
celinecamacho, line 2: student21_pop and spring 2021 🙂

be2-scala/src/main/scala/ch/epfl/pop/model/network/method/message/data/DataBuilder.scala
```diff
@@ -0,0 +1,42 @@
...
+  def buildMeetingData(messageData: MessageData, payload: String): Any = messageData.action match {
+    case ActionType.CREATE => CreateMeeting.buildFromJson(messageData, payload)
+    case ActionType.STATE => StateMeeting.buildFromJson(messageData, payload)
+    case _ => throw new ProtocolException(s"Unknown action '${messageData.action}' encountered while creating a Meeting Data")
+  }
+
+  def buildRollCallData(messageData: MessageData, payload: String): Any = messageData.action match {
+    case ActionType.CREATE => CreateRollCall.buildFromJson(messageData, payload)
+    case ActionType.OPEN => OpenRollCall.buildFromJson(messageData, payload)
+    case ActionType.REOPEN => ReopenRollCall.buildFromJson(messageData, payload)
```
celinecamacho, line 33: This doesn't match the new protocol, having only create/open/close messages for a roll call. Do we update that in this PR?

be2-scala/src/main/scala/ch/epfl/pop/model/network/method/message/data/lao/StateLao.scala
```diff
@@ -0,0 +1,36 @@
+package ch.epfl.pop.model.network.method.message.data.lao
+
+import ch.epfl.pop.model.network.method.message.data.{ActionType, MessageData, ObjectType, Parsable}
+import ch.epfl.pop.model.objects.{Hash, PublicKey, Timestamp, WitnessSignaturePair}
+
+case class StateLao(
```
celinecamacho, line 6: What are the conventions we are following? in `CreateLao` the parameters are all in one line but here each parameter is in one line. We should choose one convention to follow

celinecamacho (commented):
I only reviewed half of the files, I have some comments about the coding conventions and the protocol specifications. There are some implementations of the roll call following the old protocol specifications (i.e. roll call reopen)

Also, now we can add the link of be2-scala to the README :)

be2-scala/.github/workflows/be2-scala.yml
RaulinN: Yes indeed! Didn't change it yet ^^' But I think there will be more things that change. For now, I'll only add "master" in pull_requests :)

be2-scala/README.md
RaulinN: Done!

Same here. Many other changes will follow

be2-scala/src/main/scala/ch/epfl/pop/model/network/method/message/data/DataBuilder.scala
RaulinN: @celinecamacho For now, I will not change it (because I don't know how exactly) so that this base PR can be merged. However, I will keep that in mind :) Thank you!

be2-scala/src/main/scala/ch/epfl/pop/model/network/method/message/data/lao/StateLao.scala
RaulinN: So I ended up creating the new line only if the line is longer than 100 chars as stated in the coding guidelines. We can however change this if you prefer of course! :)

celinecamacho: Perfect for me

be2-scala/src/test/scala/ch/epfl/pop/tests/MessageCreationUtils.scala
```diff
@@ -0,0 +1,47 @@
...
+import ch.epfl.pop.json.JsonCommunicationProtocol.MessageContentDataFormat
+import java.nio.charset.StandardCharsets.UTF_8
+
+import java.util.Base64
+
+object MessageCreationUtils {
+
+  def b64Encode(b: Array[Byte]): Array[Byte] = Base64.getEncoder.encode(b)
+
+  def b64EncodeToString(b: Array[Byte]): String = Base64.getEncoder.encodeToString(b)
```
celinecamacho, line 18: Shouldn't this return a Base64String?

be2-scala/src/main/scala/ch/epfl/pop/model/objects/Base64Data.scala
```diff
@@ -0,0 +1,3 @@
+package ch.epfl.pop.model.objects
+
+case class Base64Data(data: String)
```
celinecamacho, line 3: What is the difference between `Base64Data` and the `Base64String` used in `MessageCreationUtils`?

be2-scala/src/test/scala/ch/epfl/pop/tests/MessageCreationUtils.scala
RaulinN: Yes but I didn't touch the tests yet since everything will change anyways

be2-scala/src/main/scala/ch/epfl/pop/model/objects/Base64Data.scala
RaulinN: Same idea. Base64String is the old Base64Data. But since we didn't touch the data part yet, it is not yet refactored :)

celinecamacho: Ok great, got a bit confused there

celinecamacho (approved):
Looks good to me

be2-scala/src/main/scala/ch/epfl/pop/model/objects/Base64Data.scala
RaulinN: @celinecamacho One more thing: the only difference is that Base64String was a type alias whereas Base64Data is a case class. That means that we can add methods such as decode directly in the base64data class now :)

be2-scala/.github/workflows/be2-scala.yml
```diff
@@ -4,7 +4,7 @@ on:
   push:
     branches: [ work-be2-* ]
   pull_request:
-    branches: [ be2-scala ]
+    branches: [ master ]
```
pierluca, line 7: You want to move this file into the root `.github/workflows/ci.yml`

be2-scala/src/main/scala/ch/epfl/pop/model/network/JsonRpcRequest.scala
```diff
@@ -0,0 +1,24 @@
+package ch.epfl.pop.model.network
+import ch.epfl.pop.model.network.method.Params
+import ch.epfl.pop.model.network.method.message.data.MessageData
+
+case class JsonRpcRequest(
+                           jsonrpc: String,
+                           method: Method.Method,
```
pierluca, line 7: `Method.Method` looks odd. I understand why it was done like this, but I'm wondering if it would be more understandable by renaming it `Method.MethodType`

be2-scala/src/main/scala/ch/epfl/pop/model/network/Method.scala
```diff
@@ -0,0 +1,18 @@
...
+  // uninitialized placeholder
+  val INVALID: Value = MatchingValue("__INVALID_METHOD__")
+
+  val BROADCAST: Value = MatchingValue("message")
+  val PUBLISH: Value = MatchingValue("publish")
+  val SUBSCRIBE: Value = MatchingValue("subscribe")
+  val UNSUBSCRIBE: Value = MatchingValue("unsubscribe")
+  val CATCHUP: Value = MatchingValue("catchup")
+
+  def MatchingValue(v: String): Value with Matching =  new Val(nextId, v) with Matching
```
pierluca, line 15: double whitespace here, I assume by mistake?
 between `=` and `new Val`

be2-scala/src/main/scala/ch/epfl/pop/model/network/Parsable.scala
```diff
@@ -0,0 +1,7 @@
+package ch.epfl.pop.model.network
+
+import ch.epfl.pop.model.network.method.message.data.MessageData
+
+trait Parsable {
+  def buildFromJson(messageData: MessageData, payload: String): Any
```
pierluca, line 6: I don't understand this. What's payload? What do we expect to use this for?

A bit of documentation here might help.

be2-scala/src/main/scala/ch/epfl/pop/model/network/method/message/data/MessageData.scala
```diff
@@ -0,0 +1,6 @@
+package ch.epfl.pop.model.network.method.message.data
+
+trait MessageData {
+  val _object: ObjectType.ObjectType
```
pierluca, line 4: why put an underscore here?

pierluca (approved):
LGTM in general, just a few remarks here and there, that could be addressed later on.

be2-scala/src/main/scala/ch/epfl/pop/crypto/Hash.scala
```diff
@@ -0,0 +1,70 @@
...
+   * @param laoID the id of the LAO
+   * @param creation the creation timestamp of the roll-call
+   * @param name the name of the roll-call
+   * @return the id of the roll-call
+   */
+  def computeRollCallId(laoID: Hash, creation: Long, name: String): Hash = computeGenericId("R", laoID, creation, name)
+
+
+  private def computeGenericId(typ: String, a: Array[Byte], l: Long, s: String): Hash = {
+    val data = arrayRepr(typ, Base64.getEncoder.encodeToString(a), l.toString, s)
```
gnarula, line 58: This has changed to `len(s1)||s1||len(s2)||s2...`

gnarula (changes requested):

gnarula (approved):
Ah, my feedback isn't relevant to the PR because it concerns the first commit. Nvm :)

be2-scala/src/main/scala/ch/epfl/pop/model/network/JsonRpcRequest.scala
RaulinN: Answering messages even though we talked about them in today's meeting just in case someone else is following this PR :) 

Renamed it! :)

be2-scala/src/main/scala/ch/epfl/pop/model/network/Method.scala
RaulinN: Corrected on all 3 enums

be2-scala/src/main/scala/ch/epfl/pop/model/network/Parsable.scala
RaulinN: Taking a note of this! Will do if we end up keeping this

be2-scala/src/main/scala/ch/epfl/pop/model/network/method/message/data/MessageData.scala
RaulinN: Because object is a reserved keyword in scala. There's no way to name a varialbe object :/

be2-scala/.github/workflows/be2-scala.yml
RaulinN: Done! Didn't merge it with the common file since probably some things will change in the scala CI

PR #270, opened 2021-03-12
Election data definition

flmichel:
Define the election data type in the go backend.

### DIFF
--- a/be1-go/message/data.go
+++ b/be1-go/message/data.go
@@ -274,32 +274,50 @@ type WitnessMessageData struct {
 	Signature Signature `json:"signature"`
 }
 
+// ElectionAction represents the action associated with an "election" data message.
 type ElectionAction DataAction
 
 var (
-	SetupElectionAction    ElectionAction = "setup"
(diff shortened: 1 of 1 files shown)


lpatel99:
type ElectionSetupData struct {
	*GenericData

	ID        []byte     `json:"id"`
	LaoID     []byte     `json:"lao"`
	Name      string     `json:"name"`
	Version   string     `json:"vesrion"`
	CreatedAt Timestamp  `json:"created_at"`
	StartTime Timestamp  `json:"start_time"`
	EndTime   Timestamp  `json:"end_time"`
	Questions []Question `json:"questions"`
}

looks good to me except "vesrion" (don't know how to show code in a comment)

Tesa27:
The different election Data looks coherent with the specification so no concerns from my side.

gnarula:
I haven't reviewed this yet but I'd imagine it depends on #265 right?

flmichel:
> I haven't reviewed this yet but I'd imagine it depends on #265 right?

Yes that is right!

PR #271, opened 2021-03-12
Ingestion bugfixes

RaulinN:
- bd79772 : the app was throwing when receiving an integer result value for catchup messages
- d755939 : high-level API used (for now only) to create LAOs
- e103a1c : small changes to the mock server. Now able to correctly answer catchup queries

---
⚠️  see #258 regarding the second bug


pierluca (approved):
LGTM

PR #274, opened 2021-03-18
Election channel structure

flmichel:


Tesa27 (commented):
This layout looks good and intuitive to me!

Tesa27 (dismissed):

Tesa27 (commented):
maybe change the type of ballotOptions from []string to []message.BallotOption but that's a detail

Tesa27 (commented):
And let me know what you think, we could maybe change the type of `questions:` have a map where the id would be the key and the defined `question` type would be the value, I think it makes it easier to query/update `validVotes`

flmichel:
> maybe change the type of ballotOptions from []string to []message.BallotOption and fhe type of method from string to message.PluralityMethod but that's a detail

Yes, good idea but I guess we have to wait that [#270](https://github.com/dedis/student_21_pop/pull/270) get accepted.

flmichel:
> now what you think, we could maybe change the type of `questions:` have a map where the id would be the key and the defined `question` type would be the value, I think it makes it easier to query/update `validVotes`

I agree with that, I'm going to do the modifications!

PR #275, opened 2021-03-19
Scala network start

RaulinN:
- 4810443 : Start of the DSL actor graph
- 55e7755 : Case classes used to differenciate message types (so that we don't have to compute `object` and `action` every time we need it)
- 59d2dea : Will be used by Validator actor in the flow
- fd52d53 : Applied review comments
---


be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/MessageDecoder.scala
```diff
@@ -0,0 +1,19 @@
...
+import ch.epfl.pop.json.JsonMessageParser
+
+object MessageDecoder {
+
+  // Message or TextMessage
+  val jsonRpcParser: Flow[Message, Nothing, NotUsed] = Flow[Message].map {
+    case TextMessage.Strict(text) => JsonMessageParser.parseMessage(text) match {
+      case _ => ???
+    }
+    case TextMessage.Streamed(stream) => ??? // throw
```
pierluca, line 15: what's this?
you mentioned something about strict/streamed last night, but I was missing some context.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/PipelineError.scala
```diff
@@ -0,0 +1,4 @@
+package ch.epfl.pop.pubsub.graph
+
+final case class PipelineError(description: String)
```
pierluca, line 3: we might need an integer code, essentially saying what kind of error this is.
This error code might fit with the protocol specs.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/MessageEncoder.scala
```diff
@@ -0,0 +1,19 @@
...
+import akka.stream.scaladsl.Flow
+import ch.epfl.pop.json.JsonMessageParser
+import ch.epfl.pop.model.network.JsonRpcMessage
+
+import scala.util.Try
+
+object MessageEncoder {
+
+  val serializer: Flow[JsonRpcMessage, Nothing, NotUsed] = Flow[JsonRpcMessage].map { message =>
+    val serializedMessage: Try[String] = Try(JsonMessageParser.serializeMessage(message))
```
pierluca, line 13: Wait a second, there are 2 messages that need to be sent out at this point.
One is the broadcast (if we received a publish), the other is the response.
I don't think we defined the two (or more flows) following a publish, leading to sending:
- a broadcast to other clients
- a response to the publishing client
- a future as-of-yet-undefined message to other servers

be2-scala/src/main/scala/ch/epfl/pop/model/network/Validatable.scala
```diff
@@ -0,0 +1,7 @@
+package ch.epfl.pop.model.network
+
+import ch.epfl.pop.pubsub.graph.PipelineError
+
+trait Validatable {
```
pierluca, line 5: What a remarkably ugly word. I can't blame you, wiktionary says it exists and it seems appropriate :-)
https://en.wiktionary.org/wiki/validatable

pierluca (commented):
Minor remarks, but I think we're heading in the right direction with this ! 👍

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/MessageDecoder.scala
RaulinN: AFAIK, `TextMessage.Strict` sends (or receives) a **full** message on/from a websocket. 

The graph is able to start processing the streamed version before the whole message was received (probably for large amount of data). I don't know yet if it will be useful for us

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/PipelineError.scala
RaulinN: Good idea! I used the names we find in the protocol so that if Céline ends up serializing this error, the process would only take one line

be2-scala/src/main/scala/ch/epfl/pop/model/network/Validatable.scala
RaulinN: Hahah! ^^ If you find a better word, I'll happily change it! But I googled and couldn't find better :/

RaulinN: Hopefully the "Answerer" graph component will not hurt your eyes too much when I push the next commit ^^'

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/MessageEncoder.scala
RaulinN: If I understand correctly, this could be the job of the "answerer" in the diagram above. Depending on the situation, it might send 0 to N messages. I do not know yet how to contact others (broadcast), nor how to serialize the answers to be sent (since the encoder is tehcnically located before in the pipeline)

RaulinN:
In the 3 last commits, I changed the graph a little bit. I decided to add the "handlerMerger" graph component to make it more clean (even though there are 2 mergers one after the other). If you look at the green part, the input is now 1 path/channel and the output is 1 path/channel (instead of 2). It is way easier/cleaner to deal with 1-to-1 flows rather than 1-to-N in subgraphs ^^

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/MessageDecoder.scala
pierluca: I would say not.
I know a bunch of protocols that are very well adapted to this kind of "streaming" but our naive JSON-RPC approach isn't one of them. Let's not go down that path.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/MessageEncoder.scala
pierluca: I think the broadcast might be an additional "sink" from the LaoHandler (and others in that partition/merge group), following a separate route, as not all messages will be broadcast on the same channel (e.g. Lao/create creates its own channel, and the broadcast happens on that channel)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/PipelineError.scala
```diff
@@ -1,4 +1,20 @@
...
+  val InvalidAction: ErrorCodes = Value(-1, "invalid action")
+  // invalid resource (e.g. channel does not exist, channel was not subscribed to, etc.)
+  val InvalidResource: ErrorCodes = Value(-2, "invalid resource")
+  // resource already exists (e.g. lao already exists, channel already exists, etc.)
+  val AlreadyExists: ErrorCodes = Value(-3, "resource already exists")
+  // request data is invalid (e.g. message is invalid)
+  val InvalidData: ErrorCodes = Value(-4, "request data is invalid")
+  // access denied (e.g. subscribing to a “restricted” channel)
+  val AccessDenied: ErrorCodes = Value(-5, "access denied")
+}
```
pierluca, line 20: I believe we added a -6 saying "internal server error"

RaulinN: Changed! Thank you :)

be2-scala/src/main/scala/ch/epfl/pop/model/objects/Channel.scala
```diff
@@ -0,0 +1,20 @@
...
+  /**
+   * Extract the sub-channel (laoId) from a channel name
+   *
+   * @param channel full channel name
+   * @return an array of bytes corresponding to the decoded sub-channel name or None if an error occurred
+   */
+  def decodeSubChannel(channel: Channel.Channel): Option[Array[Byte]] = channel match {
+    case _ if channel.startsWith(rootChannelPrefix) => Some(Base64.getDecoder.decode(channel.getBytes)) // TODO may throw
+    case _ => None
+  }
```
pierluca, line 19: do we even need this? 
I don't think we have this logic anywhere else, on the other hand we do need to do the reverse, from a lao id getting the sub-channel

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/Validator.scala
```diff
@@ -2,17 +2,46 @@ package ch.epfl.pop.pubsub.graph
...
+      case message@(_: JsonRpcRequestUpdateLao) => validateUpdateLao(message)
+      case message@(_: JsonRpcRequestCreateMeeting) => validateCreateMeeting(message)
+      case message@(_: JsonRpcRequestStateMeeting) => validateStateMeeting(message)
+      case message@(_: JsonRpcRequestCreateLao) => validateCreateRollCall(message)
+      case message@(_: JsonRpcRequestStateLao) => validateOpenRollCall(message)
+      case message@(_: JsonRpcRequestUpdateLao) => validateReopenRollCall(message)
+      case message@(_: JsonRpcRequestUpdateLao) => validateCloseRollCall(message)
+      case message@(_: JsonRpcRequestWitnessMessage) => validateWitnessMessage(message)
+      case _ => Right(PipelineError(
+        ErrorCodes.SERVER_FAULT.id,
```
pierluca, line 39: I'm not sure this is strictly a server error. It might well be that whoever sent the message has a buggy front-end :) 
I feel this would be an INVALID_ACTION as far as the server is concerned.
Same remark applies to other similar cases below.

We had thought of the server error more as a way of reporting that we divided by 0 and crashed (for instance) than due to any data going across the network.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/PipelineError.scala
```diff
@@ -6,15 +6,17 @@ object ErrorCodes extends Enumeration {
...
-  val AlreadyExists: ErrorCodes = Value(-3, "resource already exists")
+  val ALREADY_EXISTS: ErrorCodes = Value(-3, "resource already exists")
   // request data is invalid (e.g. message is invalid)
-  val InvalidData: ErrorCodes = Value(-4, "request data is invalid")
+  val INVALID_DATA: ErrorCodes = Value(-4, "request data is invalid")
   // access denied (e.g. subscribing to a “restricted” channel)
-  val AccessDenied: ErrorCodes = Value(-5, "access denied")
+  val ACCESS_DENIED: ErrorCodes = Value(-5, "access denied")
+  // server fault
+  val SERVER_FAULT: ErrorCodes = Value(-6, "server fault")
```
pierluca, line 21: I'm going to be super annoying and quote the specs nearly verbatim :-P

```suggestion
  // internal server error (e.g. crashed while processing, database unavailable, etc.)
  val SERVER_ERROR: ErrorCodes = Value(-6, "internal server error")
```

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/LaoHandler.scala
```diff
@@ -2,19 +2,55 @@ package ch.epfl.pop.pubsub.graph.handlers
...
+  def handleCreateLao(rpcMessage: JsonRpcRequest): GraphMessage = {
+    val messageData: CreateLao = CreateLao.buildFromPartial(rpcMessage.getDecodedData.get, rpcMessage)
+    val channel: String = s"${Channel.rootChannelPrefix}${messageData.id}"
+
+    subActor.ask(ref => CreateMessage(channel, ref)) match {
+      case Success(_) =>
+        // Publish on the LAO main channel
+        dbActor.ask(ref => DbActorNew.Write(channel, rpcMessage.getParamsMessage.get, ref)) match {
+          case Success(_) => Left(rpcMessage)
+          case _ => Right(PipelineError(-10, "")) // FIXME add DbActor "answers" with error description if failed
```
pierluca, line 38: This would typically be a `SERVER_ERROR`, I believe.
If it's useful for processing, you might also consider having extra error codes (e.g., `DB_ERROR`, `RUNTIME_ERROR`, ...) that are then "merged" into a `SERVER_ERROR` response at the end.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/LaoHandler.scala
```diff
@@ -2,19 +2,55 @@ package ch.epfl.pop.pubsub.graph.handlers
...
+    }
+  }
+
+  def handleStateLao(rpcMessage: JsonRpcRequest): GraphMessage = {
+    val modificationId: Hash = rpcMessage.getDecodedData.asInstanceOf[StateLao].modification_id
+    dbActor.ask(ref => DbActorNew.Read(rpcMessage.getParamsChannel, modificationId, ref)) match {
+      case Some(_) => dbAskWritePropagate(rpcMessage)
+      case _ => Right(PipelineError(
+        ErrorCodes.INVALID_DATA.id,
+        s"Unable to request lao state: invalid modification_id '$modificationId' (no message associated to this id)"
```
pierluca, line 50: This is a hard case to process.
We might have to "queue" the message for later processing, as we might receive the messages out of order.
I don't think we can address this in this PR, but do write a big comment for now :)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/MeetingHandler.scala
```diff
@@ -2,17 +2,44 @@ package ch.epfl.pop.pubsub.graph.handlers
...
+      case _ => Right(PipelineError(
+        ErrorCodes.INVALID_DATA.id,
+        s"Unable to create meeting: invalid encoded laoId '${rpcMessage.getParamsChannel}'"
+      ))
+    }
+  }
+
+  def handleStateMeeting(rpcMessage: JsonRpcRequest): GraphMessage = {
+    val modificationId: Hash = rpcMessage.getDecodedData.asInstanceOf[StateMeeting].modification_id
+    dbActor.ask(ref => DbActorNew.Read(rpcMessage.getParamsChannel, modificationId, ref)) match {
```
pierluca, line 37: same as above + all of the mess in #283

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/LaoValidator.scala
```diff
@@ -0,0 +1,78 @@
...
+
+import ch.epfl.pop.model.network.JsonRpcRequest
+import ch.epfl.pop.model.network.method.message.Message
+import ch.epfl.pop.model.network.method.message.data.lao.{CreateLao, StateLao, UpdateLao}
+import ch.epfl.pop.model.objects.Hash
+import ch.epfl.pop.pubsub.graph.{GraphMessage, PipelineError}
+
+case object LaoValidator extends ContentValidator {
+  sealed def validateCreateLao(rpcMessage: JsonRpcRequest): GraphMessage = {
+    def validationError(reason: String): PipelineError = super.validationError(reason, "CreateLao")
```
pierluca, line 11: in these validators, how does one know which JSON-RPC id / message_id this refers to ? 
I assume we'll want to log this stuff, and it's important to be able to trace it back to a specific request.

pierluca (commented):
Wow. That's a whole lot of good work 👍 

I haven't yet been able to review the validation logic in detail (too tired at that point) but the overall thing seems pretty solid, despite some small things here and there.

I have two concerns at this point:
1. Making sure that logging happens throughout and that it can be traced back to a specific request. This can be the subject of a separate PR, but it's very important.
2. Making sure that messages received out-of-order can be processed whenever we get the full picture. Also for a separate PR, could be taken care of by a student, but it's pretty important.

be2-scala/src/main/scala/ch/epfl/pop/model/objects/Channel.scala
RaulinN: I'm not sure yet how useful the "Channel" case class is. Last semester, we emitted the possibility of having sub-sub-channels (e.g. root/xabc/def). Is this still up-to-date?

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/Validator.scala
RaulinN: Got it! :)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/LaoValidator.scala
RaulinN: I think this issue will be resolved if I manage to implement the UserActors we were talking about

be2-scala/src/main/scala/ch/epfl/pop/model/objects/Channel.scala
pierluca: Yes, students need this for the e-voting project :)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/LaoValidator.scala
pierluca: I'm not sure about this :)

gnarula (changes requested):
Awesome work 💯 ! I left a few comments, mostly about minor changes in the spec since you started working on this. LGTM otherwise 👍 

Edit: my bad, the comments in files were put as 'General Comments' and not 'Review Comments'.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/package.scala
```diff
@@ -0,0 +1,8 @@
+package ch.epfl.pop.pubsub
+
+import ch.epfl.pop.model.network.JsonRpcMessage
+
+package object graph {
+  type JsonString = String
+  type GraphMessage = Either[JsonRpcMessage, PipelineError]
```
gnarula, line 7: https://www.scala-lang.org/api/2.12.0/scala/util/Either.html

> Convention dictates that Left is used for failure and Right is used for success.

gnarula (approved):

PR #277, opened 2021-03-19
Fix Android CI

gnarula:
* Adds a mock injection class
* Uses JDK 11 since Sonar has dropped support for JDK 8


fe2-android/app/src/main/res/layout/fragment_launch.xml
```diff
@@ -1,78 +1,76 @@
...
+        android:layout_marginVertical="@dimen/margin_button"
+        android:gravity="center_horizontal"
+        android:orientation="horizontal">
 
-                <Button
-                        android:id="@+id/button_cancel_launch"
-                        android:layout_width="wrap_content"
-                        android:layout_height="wrap_content"
-                        android:gravity="center"
-                        android:onClick="onClick"
```
pierluca, line 61: why remove all the onClick's ?

fe2-android/app/src/mock/java/com/github/dedis/student20_pop/Injection.java
```diff
@@ -0,0 +1,87 @@
+package com.github.dedis.student20_pop;
```
pierluca, line 1: what is this file used for?

fe2-android/app/src/test/java/com/github/dedis/student20_pop/utility/security/HashTest.java
```diff
@@ -1,18 +1,11 @@
...
 
 import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertThrows;
 
 import org.junit.Test;
 
 public class HashTest {
 
-  @Test
-  public void hashNullDataTest() {
```
pierluca, line 11: I don't have context, what was the issue with this test? superficially it seems fine

pierluca (commented):
I feel I have more questions than answers here :)

fe2-android/app/src/main/res/layout/fragment_launch.xml
gnarula: The onClick handler's are set inside the fragments via databinding. As a result the declaration in the XML files is redundant

fe2-android/app/src/mock/java/com/github/dedis/student20_pop/Injection.java
gnarula: It *will* be used to provide mock instances for unit tests.

fe2-android/app/src/test/java/com/github/dedis/student20_pop/utility/security/HashTest.java
gnarula: I removed it because this was written with the old hashing function in mind but it does hold for the new one too.

Adding a commit that adds a tests for nulls and empty strings.

pierluca (approved):

pierluca:
After analysis, it seems the issue raised by Sonar is a matter of best practices from a security perspective.
http://cwe.mitre.org/data/definitions/397.html

In this case, I believe it's not a big deal to ignore it, though as a general rule we should probably set a good example for students :)

pierluca (approved):

PR #278, opened 2021-03-22
Work fe2 elctionsetup uximprovement johann

Warhedgehog:
Added the display of error on an empty roll-call title directly on the text field. Added the same function to be used in the election set-up interface but didn't call for the code I branched from didn't allow to see said interface.


Walfar (approved):
Looks all good to me ! Great refactoring of the input checker

PR #279, opened 2021-03-23
Add Json verification for messages

flmichel:
- Added Json verification for generic messages and data.
- Added Json schema for generic data.
- Added the possibility to send an error with null value for the id in case the id in not provided in the query.


be1-go/hub/organizer.go
```diff
@@ -54,12 +90,34 @@ func (o *organizerHub) handleIncomingMessage(incomingMessage *IncomingMessage) {
 	log.Printf("organizerHub::handleIncomingMessage: %s", incomingMessage.Message)
 
 	client := incomingMessage.Client
+	byteMessage := incomingMessage.Message
 
-	// unmarshal the message
+	// Check if the GenericMessage has a field "id"
 	genericMsg := &message.GenericMessage{}
-	err := json.Unmarshal(incomingMessage.Message, genericMsg)
+	id, ok := genericMsg.GetID(byteMessage)
```
cba33, line 97: Minor comment: would it make sense to rename `GetID` to `UnmarshallID`? The method name confused me, made me think that the ID wasn't unmarshalled anymore.

flmichel: I totally agree, it would make more sense. I just updated it.

Tesa27 (commented):
Seems good to me!

be1-go/hub/organizer.go
```diff
@@ -23,15 +25,49 @@ type organizerHub struct {
...
+	// Import the Json schemas defined in the protocol section
+	protocolPath, err := filepath.Abs("../protocol")
+	if err != nil {
+		return nil, xerrors.Errorf("failed to load the path for the json schemas: %v", err)
+	}
+	protocolPath = "file://" + protocolPath
+
+	schemas := make(map[string]*gojsonschema.Schema)
+
+	// Impot the schema for generic messages
```
gnarula, line 48:
```suggestion
	// Import the schema for generic messages
```

be1-go/hub/base_channel.go
```diff
@@ -84,7 +84,15 @@ func (c *baseChannel) VerifyPublishMessage(publish message.Publish) error {
 
 	// Check if the structure of the message is correct
 	msg := publish.Params.Message
-	err := msg.VerifyAndUnmarshalData()
+
+	// Verify the data
+	err := c.hub.verifyJson(msg.RawData, DataSchema)
+	if err != nil {
+		return xerrors.Errorf("failed to validate the data: %v", err)
```
gnarula, line 91: Perhaps this should be `message.Error{}`?

be1-go/hub/organizer.go
```diff
@@ -171,6 +239,30 @@ func (o *organizerHub) handleIncomingMessage(incomingMessage *IncomingMessage) {
 	client.SendResult(id, result)
 }
 
+func (o *organizerHub) verifyJson(byteMessage []byte, schemaName string) error {
+	// Validate the Json "byteMessage" with a schema
+	messageLoader := gojsonschema.NewBytesLoader(byteMessage)
+	resultErrors, err := o.schemas[schemaName].Validate(messageLoader)
+	if err != nil {
+		return err
```
gnarula, line 247: I think this should be wrapped in message.Error as well

gnarula (changes requested):
Looks good! I've a few questions about error handling. Since client.SendError assumes that the error will be of type message.Error a lot of xerrors.Errorfs are possibly ignored and never reach the client.

This might be the cause of #295 as well. Would be nice if you can confirm this - I believe you have some input to reproduce this behaviour.

Edit: please resolve the conflicts with master

be1-go/hub/organizer.go
```diff
@@ -171,6 +239,30 @@ func (o *organizerHub) handleIncomingMessage(incomingMessage *IncomingMessage) {
...
+	}
+	errorsList := resultErrors.Errors()
+	descriptionErrors := ""
+	// Concatenate all error descriptions
+	for index, e := range errorsList {
+		descriptionErrors += " (" + fmt.Sprintf("%d", index+1) + ") " + e.Description()
```
gnarula, line 253:
```suggestion
		descriptionErrors += fmt.Sprintf(" (%d) %s", index+1, e.Description())
```

gnarula (changes requested):

be1-go/hub/base_channel.go
flmichel: If I replace `%v` by `%w` the error can still be interpreted as an `message.Error{}`: [https://play.golang.org/p/S_QDU4Q0b4T](https://play.golang.org/p/S_QDU4Q0b4T). Should I do that?

flmichel: But the description of the error remains the same so it will not provide a long and complete description in the error message. So maybe it would be better to wrap it in `message.Error`.

flmichel:
> Looks good! I've a few questions about error handling. Since client.SendError assumes that the error will be of type message.Error a lot of xerrors.Errorfs are possibly ignored and never reach the client.
> 
> This might be the cause of #295 as well. Would be nice if you can confirm this - I believe you have some input to reproduce this behaviour.
> 
> Edit: please resolve the conflicts with master

It is correct, this is the also the cause of #295.

be1-go/message/answer.go
```diff
@@ -86,6 +87,17 @@ func (e *Error) Error() string {
...
+func Errorf(format string, err error) error {
+	msgError := &Error{}
+
+	if xerrors.As(err, &msgError) {
+		msgError.Description = fmt.Sprintf(format, msgError.Description)
+		return msgError
+	}
+	return xerrors.Errorf(format, err)
+}
+
```
flmichel, line 100: I created a method which is a little bit similar to `xerrors.Errorf` to be able to add more description to an error of type `Error{}` but I'm not sure it is a good idea. What do you think?

gnarula:
@flmichel Can you resolve the conflict please?

be1-go/message/answer.go
```diff
@@ -86,6 +87,17 @@ func (e *Error) Error() string {
 	return e.Description
 }
 
+// Errorf returns an error with the new formatting
+func Errorf(format string, err error) error {
+	msgError := &Error{}
+
+	if xerrors.As(err, &msgError) {
+		msgError.Description = fmt.Sprintf(format, msgError.Description)
```
gnarula, line 95: This assumes `format` will only have one format specifier and `msg.Description` will always satisfy it. That's easy to overlook.

I'd suggest

```go
func NewError(description string, parent error) error {
    msgError := &Error{}
    
    if xerrors.As(parent, &msgError) {
        msgError.Description = fmt.Sprintf("%s: %s", description, msgError.Description)
        return msgError
    }

    return xerrors.Errorf("%s: %v", description, parent)
}
```

Sample usage:

```go
return message.NewError("failed to validate the data", err)
```

gnarula (changes requested):
Looks good, just a small comment on `message.Errorf`

be1-go/message/answer.go
flmichel: Yes I agree! Thank you for the suggestion. I made the modifications.

be1-go/message/generic_message.go
```diff
@@ -47,3 +47,17 @@ func (m *GenericMessage) UnmarshalJSON(data []byte) error {
 	m.Query = query
 	return nil
 }
+
+type IdStruct struct {
```
gnarula, line 51: Sorry I overlooked this before, you might want to move this inside `UnmarshalID` and declare it as a private struct

be1-go/message/generic_message.go
```diff
@@ -47,3 +47,17 @@ func (m *GenericMessage) UnmarshalJSON(data []byte) error {
...
+	ID *int `json:"id"`
+}
+
+// Unmarshal the ID of a message
+func (m *GenericMessage) UnmarshallID(data []byte) (int, bool) {
```
gnarula, line 56:
```suggestion
func (m *GenericMessage) UnmarshalID(data []byte) (int, bool) {
```

gnarula (changes requested):
Last few comments, sorry I missed them earlier

be1-go/hub/organizer.go
```diff
@@ -205,6 +274,33 @@ func (o *organizerHub) Start(done chan struct{}) {
 	}
 }
 
+func (o *organizerHub) verifyJson(byteMessage []byte, schemaName string) error {
```
gnarula, line 277: Thinking about it again, you might want to refactor this part because it's going to be common across both Organizer and Witness Hubs

be1-go/message/generic_message.go
flmichel: No worries, I made the changes :)

be1-go/hub/organizer.go
flmichel: Could I let it like this in this PR and change it in #315 where I would replace `o *organizerHub` with `b *baseHub` once this PR is merged?

gnarula: Might even be worth having a separate package that deals with validation exclusively. But yeah, that's something that can be taken up in another PR 👍

gnarula (approved):
LGTM

PR #280, opened 2021-03-23
Implementation of wallet communication with IndexedDB

CaiMusso:
Content proposed in this Pull Request
----------------------------
**To start with**: the Digital Wallet in web front-end needs the to communicate with an IndexedDB browser database in order to securely store the secret key which is used to encrypt the wallet seed (used to generate PoP tokens) before storing it in it react state. This is what is done in this branch at the moment.

Important file in this PR
------------------------
+ [WalletCryptographyHandler.ts](https://github.com/dedis/student_21_pop/blob/cce9c9d1afdcb694294f0c9039affefb19a47ca7/fe1-web/model/objects/WalletCryptographyHandler.ts) - this is the cryptography handler for the wallet. This class stores and retrieves the secret key from the IndexedDB database and offers public methods to **encrypt/decrypt the wallet seed with the retrieved key.**

Crypto Library
--------------------------------
The utilized crypto library is [SubtleCrypto](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto) in the Web Crypto API. The encryption/decryption algorithm is RSA and the RSA key is generated with `extractable = false`.

Related issues that will be addressed when possible
------------------------------------------------------
See **issue** #297 .

The **handleWalletInitialization()** function in [WalletCryptographyHandler.ts](https://github.com/dedis/student_21_pop/blob/cce9c9d1afdcb694294f0c9039affefb19a47ca7/fe1-web/model/objects/WalletCryptographyHandler.ts) should **solve the problems** of wallet storage initialisation. The reasons for storage not correctly initialised are **two**:
+ it was never initialised.
+ it was initialised (already used for encryption) but then cleared for some reason (e.g. clear browser data).

In the **first case** the wallet is initialised by generating an RSA key and creating the wallet storage (implemented in this PR). In the **second case** a solution still has to be found.

New dependencies
-------------------------------------
+ **idb-keyval** - used for storage
+ **fake-indexeddb** - used for storage simulation testing

Tests
-----------------------------------------------
There is a jest test [WalletCryptographyHandler.test.ts](https://github.com/dedis/student_21_pop/blob/398657ddc6fa210d86191b6a5d98c417fe2d96e8/fe1-web/__tests__/model/objects/WalletCryptographyHandler.test.ts) for the handler. This test uses a MOCK SubtleCrypto library coded in-test and a simulated storage from **fake-indexeddb** package.

Diagram Overview of Digital Wallet **implemented** functionalities
----------------------------------------------------------------------
CryptographyHandler

Quick demo (only for visualisation)
-------------------------------
+ creating IndexedDB database and store the secret RSA key in it.
+ encrypting a test-token by retrieving the RSA key from database
+ decrypting the encrypted test-token by retrieving the RSA key from database

https://user-images.githubusercontent.com/47753346/114602848-5e3d4300-9c97-11eb-9055-83f5f21ed7da.mov


pierluca (changes requested):
There are a number of design issues, I've tried to point them out through the comments.
You might want to take extra time to read through the documentation of the libraries you use, as I feel they're addressing many of your needs already.

Furthermore, you might want to integrate this with the the reducer function dealing with the keypair, so you can be sure that your API is actually usable where it needs to be used. I'm quite sure that right now you will face a problem due to the async nature of your API.

pierluca:
On a separate note, I see a very large number of changes in the package-lock.json , are you sure those changes are legitimate and needed? I don't see why adding a dependency should change 30k lines. 🤔

CaiMusso:
@pierluca thank you for all the **very** useful feedback. I will work on all these remarks and ask for a new review later.

fe1-web/model/objects/WalletCryptographyHandler.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  /**
+   * decrypts the encrypted ed25519 token with the RSA key stored in the indexedDB database
+   * @param encryptedToken ed25519 encrypted token (ArrayBuffer)
+   */
+  public async decrypt(encryptedToken: ArrayBuffer): Promise<ArrayBuffer> {
+    const key = await this.getKeyFromDatabase(STRINGS.walletPrivateKey);
+    const plaintextToken = await window.crypto.subtle.decrypt(this.algorithm, key, encryptedToken);
+    return plaintextToken;
```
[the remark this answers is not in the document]
pierluca: Careful here @RaulinN , you might be a tiny bit mistaken :) 
The `await/async` is a way of having more readable promise chaining. We haven't used it so far, but it's a perfectly valid approach :) 

The whole function is basically the equivalent of doing:

```
  public async decrypt(encryptedToken: ArrayBuffer): Promise<ArrayBuffer> {
    return this.getKeyFromDatabase(STRINGS.walletPrivateKey)
      .then((key) => window.crypto.subtle.decrypt(this.algorithm, key, encryptedToken));
```

Your version has a bug, as you'd return the keyPromise and not the decryption results.

fe1-web/store/stores/IndexedDBStore.ts
```diff
@@ -0,0 +1,70 @@
...
+    const secKey = await get(this.privateKeyId);
+    return secKey;
+  }
+
+  /**
+   * this is used to update the encryption key in case it will
+   * every be necessary to change it
+   * @param key the new key for encryption/decryption
+   */
+  public async updateKey(key : WalletCryptoKey): Promise<void> {
```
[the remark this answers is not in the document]
pierluca: Indeed, I believe the eslint / tsconfig suggest using `key: WalletCryptoKey`

fe1-web/store/stores/IndexedDBStore.ts
```diff
@@ -0,0 +1,70 @@
+import { get, set, update } from 'idb-keyval';
```
pierluca, line 1: It seems this file is doing nothing but wrapping around idb-keyval, do we still need it at all? 
Why not directly doing the get/set/update in the Wallet file?

fe1-web/model/objects/WalletCryptographyHandler.ts
```diff
@@ -0,0 +1,76 @@
...
+    const key: WalletCryptoKey = await this.generateRSAKey();
+    await this.keyDatabase.putKey(key);
+  }
+
+  /**
+   * encrypts the given ed25519 token with the RSA key stored in the indexedDB database
+   * @param token ed25519 key toUint8Array()
+   */
+  public async encrypt(token: Uint8Array): Promise<ArrayBuffer> {
+    const key = await this.getKeyFromDatabase(STRINGS.walletPublicKey);
```
pierluca, line 37: to get a key, you could either do `await get(this.publicKeyId);` (with `get` defined as in idb-keyval),
or here you're calling another method, who then calls on the other class.

It seems there's a ton of indirection which only makes the actual behavior harder to analyze.

fe1-web/model/objects/WalletCryptographyHandler.ts
```diff
@@ -0,0 +1,76 @@
...
+  async (type: string) => {
+    const key: CryptoKey = await this.keyDatabase.getKey(type);
+    return key;
+  };
+
+  /**
+   * generates the RSA key according to specified algorithm
+   */
+  private generateRSAKey: () => Promise<WalletCryptoKey> =
+  async () => {
```
pierluca, line 67: you've used function syntax above, why use the arrow here? 

For consistency and readability: 
```suggestion
  private async generateRSAKey(): Promise<WalletCryptoKey> {
```

Same applies to `getKeyFromDatabase` if you want to keep that.

fe1-web/model/objects/WalletCryptographyHandler.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  /* the encryption/decryption key storage */
+  private keyDatabase: IndexedDBStore = new IndexedDBStore();
+
+  /**
+   * this method adds an RSA encryption/decryption key,
+   * interacts with IndexedDB browser database.
+   */
+  public async initWalletStorage(): Promise <void> {
+    const key: WalletCryptoKey = await this.generateRSAKey();
```
pierluca, line 28: this means that every time you initialize the wallet, you're going to overwrite the key? That seems very dangerous...

pierluca (changes requested):
Much better, though there still seems to be a lot of extra code that's not really needed to achieve the functionality. 

I would also strongly suggest adding a few unit tests, for instance one doing the following sequence of operations:
1. Creating the wallet object
2. encrypting data
3. destroying the wallet object
4. creating another wallet object (simulating an app restart)
5. decrypting the data
6. verifying that the initial and decrypted data match

pierluca:
Pointer for you :) 
https://stackoverflow.com/questions/52612122/how-to-use-jest-to-test-functions-using-crypto-or-window-mscrypto

Adalsteinnjons (commented):
LGTM. Well structured and well commented. Good Job

Adalsteinnjons (dismissed):

fe1-web/model/objects/WalletCryptographyHandler.ts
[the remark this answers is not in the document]
RaulinN: Yes indeed! Misread the documentation about await/async ^^' Sorry for that!

For the piece of code I wrote, I didn't expect it to follow javascript's syntax since I was very busy with Kotlin's futures lately ^^ (that's also why I put the disclaimer "approximate syntax")

fe1-web/model/objects/WalletCryptographyHandler.ts
```diff
@@ -1,12 +1,27 @@
...
  * This class has the job of handling the cryptography functions of the wallet.
  * It interacts with the IndexedDB database in order to store and retrieve the
- * secret key. It also encrypts and decrypts the tokens with the retrieved key.
+ * secret key. It also encrypts and decrypts the wallet seed (used to create the
+ * tokens) with the retrieved key.
+ * More info on this approach at https://blog.engelke.com/2014/09/19/saving-cryptographic-keys-in-the-browser/
  */
 export class WalletCryptographyHandler {
+  /* the crypto library - passed to constructor */
+  private cryptography!: object;
```
pierluca, line 19: you shouldn't need to assert that it's not-null with `!` here, right?
`object` is by definition not-null :)

fe1-web/model/objects/WalletCryptographyHandler.ts
```diff
@@ -15,62 +30,134 @@ export class WalletCryptographyHandler {
...
+  }
+
+  /**
+   * this is used to update the encryption key in case it will ever be
+   * necessary to change it, it updates the key since only one crypto-key
+   * is allowed in the wallet storage.
+   * @private
+   * @param key the new key for encryption/decryption
+   */
+  private async updateKeyInDatabase(key: WalletCryptoKey): Promise<void> {
```
pierluca, line 122: typically this would be useful if we need to restore the key from a backup, excellent! 👍

fe1-web/model/objects/WalletCryptographyHandler.ts
pierluca: you might also want to reference the actual interface: `Crypto`
https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_dom_d_.crypto.html#subtle

which itself contains `SubtleCrypto` 
https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_dom_d_.subtlecrypto.html

fe1-web/model/objects/WalletCryptographyHandler.ts
```diff
@@ -15,62 +30,134 @@ export class WalletCryptographyHandler {
...
 
+  /* usages for the RSA key */
   private readonly keyUsages: KeyUsage[] = ['encrypt', 'decrypt'];
 
-  /* the encryption/decryption key storage */
-  private keyDatabase: IndexedDBStore = new IndexedDBStore();
+  /* the crypto library is passed to constructor, this is necessary in order to test the
+     cryptography handler without the crypto.subtle library provided by the window object.
+     In jest context provide a MOCK crypto.subtle library, otherwise provide window.crypto */
+  constructor(cryptography: object) {
```
pierluca, line 39: same here, you might want to reference the `Crypto` interface

fe1-web/model/objects/WalletCryptographyHandler.ts
```diff
@@ -15,62 +30,134 @@ export class WalletCryptographyHandler {
...
    */
-  public async decrypt(encryptedToken: ArrayBuffer): Promise<ArrayBuffer> {
+  public async decrypt(ciphertext: ArrayBuffer): Promise<ArrayBuffer> {
     const key = await this.getKeyFromDatabase(STRINGS.walletPrivateKey);
-    const plaintextToken = await window.crypto.subtle.decrypt(this.algorithm, key, encryptedToken);
-    return plaintextToken;
+    if (key === undefined) {
+      throw Error('Error while retrieving decryption key from database: undefined');
+    }
+    // @ts-ignore
```
pierluca, line 83: no need for `ts-ignore` if you define `cryptography` with the right type (`Crypto`)

fe1-web/model/objects/WalletCryptographyHandler.ts
```diff
@@ -15,62 +30,134 @@ export class WalletCryptographyHandler {
...
 
   /**
    * generates the RSA key according to specified algorithm
+   * @private
    */
-  private generateRSAKey: () => Promise<WalletCryptoKey> =
-  async () => {
-    const keyPair = await window.crypto.subtle.generateKey(
+  private async generateRSAKey(): Promise<WalletCryptoKey> {
+    // @ts-ignore
```
pierluca, line 154: same, no need if `this.cryptography` has the right type

fe1-web/model/objects/WalletCryptographyHandler.ts
```diff
@@ -15,62 +30,134 @@ export class WalletCryptographyHandler {
...
-  public async encrypt(token: Uint8Array): Promise<ArrayBuffer> {
-    const key = await this.getKeyFromDatabase(STRINGS.walletPublicKey);
-    const cypheredToken = await window.crypto.subtle.encrypt(this.algorithm, key, token);
-    return cypheredToken;
+  public async encrypt(plaintext: Uint8Array): Promise<ArrayBuffer> {
+    const key: CryptoKey = await this.getKeyFromDatabase(STRINGS.walletPublicKey);
+    if (key === undefined) {
+      throw Error('Error while retrieving encryption key from database: undefined');
+    }
+    // @ts-ignore
```
pierluca, line 69: same

fe1-web/__tests__/model/objects/WalletCryptographyHandler.test.ts
```diff
@@ -0,0 +1,89 @@
...
+            const publicKey = await get(STRINGS.walletPublicKeyId);
+            /* undoing public key append by substituting the substring with empty string */
+            const plaintext: string = ciphertext.toString().replace(publicKey.toString(), '');
+            const plaintextBuffer: ArrayBuffer = Buffer.from(plaintext);
+            return new Promise((resolve) => {
+              resolve(plaintextBuffer);
+            });
+          },
+        },
+      });
```
pierluca, line 63: you might want to move this in the definition of the `crypto` object above, as this is independent on the test itself (unless I'm mistaken).

fe1-web/__tests__/model/objects/WalletCryptographyHandler.test.ts
```diff
@@ -0,0 +1,89 @@
...
+ * The scope is not testing the SubtleCrypto library but testing the function
+ * calls to it. It also tests the 'store' and 'retrieve' functions of indexedDB
+ * with an intermediate simulation of app restart.
+ */
+describe('=== Wallet Cryptography Handler tests ===', () => {
+  describe('encryption - decryption', () => {
+    it('should correctly encrypt and decrypt the generated seed', async () => {
+      /* this is the MOCK crypto manager, it uses the mock SubtleCrypto library */
+      // eslint-disable-next-line max-len
+      let mockCryptoManager : WalletCryptographyHandler | null = new WalletCryptographyHandler(crypto);
```
pierluca, line 24: if you apply the change I suggested to `WalletCryptographyHandler` (definining `cryptography` as type `Crypto`) you might need to do the following casting:
```suggestion
      let mockCryptoManager : WalletCryptographyHandler | null = new WalletCryptographyHandler(crypto as Crypto);
```

pierluca (changes requested):
Very good !
There are a few remarks to polish things up, but they should be very quick to fix !

pierluca (approved):
I'm off today and can't review, but I trust you :)

PR #286, opened 2021-03-30
refactored client to make sockets more generic to add different sockets

lpatel99:
I renamed the Client type Socket and created a type ClientSocket composed of a Socket (struct embedding) in order to allow the creation of types OrganizerSocket and WitnessSocket, to enable connections firstly between an organizer, witnesses and clients.

### DIFF
--- a/be1-go/cli/organizer/organizer.go
+++ b/be1-go/cli/organizer/organizer.go
@@ -13,7 +13,7 @@ import (
 	"golang.org/x/xerrors"
 )
 
-var upgrader = websocket.Upgrader{
+var Upgrader = websocket.Upgrader{
 	ReadBufferSize:  1024,
 	WriteBufferSize: 1024,
 	CheckOrigin:     func(r *http.Request) bool { return true },
@@ -48,7 +48,7 @@ func Serve(c *cli.Context) error {
 		serveWs(h, w, r)
 	})
 
-	log.Printf("Starting the WS server at %d", port)
+	log.Printf("Starting the organizer WS server at %d", port)
 	err = http.ListenAndServe(fmt.Sprintf(":%d", port), nil)
 	if err != nil {
 		return xerrors.Errorf("failed to start the server: %v", err)
@@ -60,20 +60,20 @@ func Serve(c *cli.Context) error {
 }
 
 func serveWs(h hub.Hub, w http.ResponseWriter, r *http.Request) {
-	conn, err := upgrader.Upgrade(w, r, nil)
+	conn, err := Upgrader.Upgrade(w, r, nil)
 	if err != nil {
 		log.Printf("failed to upgrade connection: %v", err)
 		return
 	}
 
-	client := hub.NewClient(h, conn)
+	client := hub.NewClientSocket(h, conn)
 
 	go client.ReadPump()
 	go client.WritePump()
 
 	// cleanup go routine that removes clients that forgot to unsubscribe
-	go func(c *hub.Client, h hub.Hub) {
+	go func(c *hub.ClientSocket, h hub.Hub) {
 		c.Wait.Wait()
-		h.RemoveClient(c)
+		h.RemoveClientSocket(c)
 	}(client, h)
 }
--- a/be1-go/cli/organizer/organizer.go
+++ b/be1-go/cli/organizer/organizer.go
@@ -20,9 +20,9 @@ var upgrader = websocket.Upgrader{
 }
 
 // Serve parses the CLI arguments and spawns a hub and a websocket server.
-func Serve(c *cli.Context) error {
-	port := c.Int("port")
-	pk := c.String("public-key")
+func Serve(context *cli.Context) error {
+	port := context.Int("port")
+	pk := context.String("public-key")
 
 	if pk == "" {
 		return xerrors.Errorf("organizer's public key is required")
--- a/be1-go/cli/organizer/organizer.go
+++ b/be1-go/cli/organizer/organizer.go
@@ -13,7 +13,7 @@ import (
 	"golang.org/x/xerrors"
 )
 
-var Upgrader = websocket.Upgrader{
+var upgrader = websocket.Upgrader{
 	ReadBufferSize:  1024,
 	WriteBufferSize: 1024,
 	CheckOrigin:     func(r *http.Request) bool { return true },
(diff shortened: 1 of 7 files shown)


be1-go/hub/organizer.go
```diff
@@ -53,7 +53,16 @@ func (o *organizerHub) Recv(msg IncomingMessage) {
...
+	socket := incomingMessage.Socket
+	client := ClientSocket{
+		Socket{
+			socketType: clientSocket,
+			hub:        socket.hub,
+			conn:       socket.conn,
+			send:       socket.send,
+			Wait:       sync.WaitGroup{},
+		},
+	}
```
cba33, line 65: Shouldn't handling of incoming messages depend on the socket type `incomingmessage.Socket.socketType` in the message? What I have in mind is a switch statement based on the socket type. Here you're straight away assuming the message came on a client socket, why is that?

be1-go/hub/socket.go
```diff
@@ -33,32 +36,32 @@ type Client struct {
...
-	return &Client{
+// NewSocket returns an instance of a Socket.
+func NewSocket(h Hub, conn *websocket.Conn) *Socket {
+	return &Socket{
 		hub:  h,
 		conn: conn,
 		send: make(chan []byte, 256),
 		Wait: sync.WaitGroup{},
 	}
 }
```
cba33, line 47: `NewSocket` doesn't initialize `socketType`, What is the type of the created socket? Did you intend that `NewSocket`creates a "default" socket that is none of {client,organizer,witness}?
Also, you create some socket in the file `hub/socket_types.go` (`NewClientSocket` etc), is the method `NewSocket`here necessary at all?

be1-go/hub/organizer.go
cba33: Another question: why not call client := NewClientSocket(...)` (defined in `socket_types.go` in this same PR) instead of creating it by hand?

cba33 (changes requested):

be1-go/hub/organizer.go
lpatel99: The handling of messages should indeed depend on the socket type and I will definitely add a switch! I did not yet as I just wanted the refactored version to work, and for now the only sockets used are clients but I will add a switch anyway, as it will be useful for future code.
I do not call NewClientSocket here as NewClientSocket creates a new send channel and I thought that the client variable should keep the same one. I'm not quite sure though whether this makes a difference, do you know?
Also instead of using type composition to create different types of sockets, I am wondering whether I should just keep the socket type and use the "socketType" attribute in order to tell them apart as there is no difference in the way they function. What do you think?

be1-go/cli/organizer/organizer.go
```diff
@@ -13,7 +13,7 @@ import (
 	"golang.org/x/xerrors"
 )
 
-var upgrader = websocket.Upgrader{
+var Upgrader = websocket.Upgrader{
```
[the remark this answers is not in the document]
lpatel99: That was unintentional, I will change that back!

be1-go/hub/socket.go
lpatel99: Indeed it is not necessary. When refactoring I simply changed the client type and all of its methods to match the socket type and did not think to remove this function. I will make sure to be more rigorous in the future.

cba33: Awesome, thanks!

be1-go/hub/organizer.go
cba33: Yes, I think adding a switch where the not-yet-implemented branches throw an error "Not implemented" or similar would be very useful in preventing bugs. Thanks!

cba33: You're right, here we don't want a new channel, which is what NewClientSocket does. So perhaps it's fine to leave it like this for now.

cba33: _Also instead of using type composition to create different types of sockets, I am wondering whether I should just keep the socket type and use the "socketType" attribute in order to tell them apart as there is no difference in the way they function. What do you think?_ Good point. I'm not sure though what's best from a coding practice point of view, so I'd like to ask for @gnarula's opinion. Gaurav, could I ask what you think about it? Thanks!

gnarula: I'd suggest something like:

```go
type Socket interface {
  ReadPump()
  WritePump()
  Type() SocketType
  // .. other methods
}

// baseSocket implements Socket interface
type baseSocket struct {
  // fields go in here
}

// ClientSocket represents a connection with a client.
type ClientSocket {
  socketType SocketType
  *baseSocket
}

func NewClientSocket() Socket {
  return &ClientSocket{
    Type: ClientSocketType,
    baseSocket: &newBaseSocket(...),
  }
}

type SocketType string

```

The advantage of having an interface there is it will allow you to mock it easily in your tests if you need to. unlike a struct where you'd be forced to use the concrete implementation.

be1-go/hub/socket_types.go
```diff
@@ -0,0 +1,62 @@
+package hub
+
+import (
+	"github.com/gorilla/websocket"
+	"sync"
+)
+
+// socket types
+const (
+	clientSocket = "client"
```
gnarula, line 10: You might want to consider using a custom type here to avoid any ambiguity

lpatel99: By custom type do you mean creating a structure with a string as an attribute ?

gnarula: I meant something like

`type SocketType string` and then having package level vars like `var ClientSocket = SocketType("client")`

cba33 (approved):

be1-go/hub/socket_types.go
```diff
@@ -0,0 +1,56 @@
...
+// socket types
+type SocketType string
+
+const (
+	clientSocket SocketType = "client"
+	organizerSocket SocketType= "organizer"
+	witnessSocket SocketType= "witness"
+)
+
+func newSocket(socketType SocketType, h Hub, conn *websocket.Conn) *baseSocket {
```
cba33, line 17: suggestion: `newBaseSocket` instead of `newSocket`

be1-go/hub/socket_types.go
```diff
@@ -0,0 +1,56 @@
...
+	return &baseSocket {
```
cba33, line 18: Where is the struct `baseSocket` defined? I would've expected to find it in this file, am I missing smth?

be1-go/hub/socket_types.go
lpatel99: Yes will do! Unfortunately I was not careful and was too quick to merge this PR but I will make sure to change this on my current branch.

be1-go/hub/socket_types.go
lpatel99: I defined it in socket.go, since all of the methods of Socket (interface) are implemented on baseSocket. I thought the socket_types should just contain the code relevant to different socket types. Do you think I should move baseSocket to socketType ?

PR #287, opened 2021-03-30
Roll Call: Update protocol specification

gnarula:
Please refer to https://docs.google.com/document/d/1fyNWSPzLhM6W9V0VTFf2waMLiJGcscy7wa4bQlLkySM/edit#heading=h.snprjk777bwe

### DIFF
--- a/protocol/query/method/message/data/dataCloseRollCall.json
+++ b/protocol/query/method/message/data/dataCloseRollCall.json
@@ -10,14 +10,14 @@
 		"update_id": {
 			"type": "string",
 			"contentEncoding": "base64",
-			"$comment": "Hash : SHA256('R'||lao_id||closes||end)"
+			"$comment": "Hash : SHA256('R'||lao_id||closes||closed_at)"
 		},
 		"closes": {
 			"type": "string",
 			"contentEncoding": "base64",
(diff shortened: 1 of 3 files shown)


protocol/query/method/message/data/dataCloseRollCall.json
```diff
@@ -29,12 +29,11 @@
 	"required": ["object", "action", "update_id", "closes", "end", "attendees"],
```
flmichel, line 29: I think "end" should be updated to "closed_at" here. Same thing for the updated fields in dataCreateRollCall.json and dataOpenRollCall.json.

flmichel (changes requested):
Looks good! I just have a minor comment.

Kar1213 (commented):
Since now all roll calls that we close have been opened I think we could remove this part of the comment in "closes" field: "or in its absence, the 'id' field of the roll call creation"

pierluca (approved):
LGTM !

flmichel (approved):

Kar1213 (approved):

pierluca:
Following this change, the Fe1 tests are failing.
You might want to update them as needed @Adalsteinnjons @CaiMusso

PR #288, opened 2021-03-31
Election Setup UI and a new Dropdown Selector component

Adalsteinnjons:
## UI screen for the UI election creation.
It should match all the fields in the image given in the documentation.
### Image from the documentation
image

## My Implementation
Currently, it is only displaying the UI and stores the values locally. 
When a user presses `CONFIRM` the values just get displayed in the console. 
(I am also working on sending the communication, that will come in a later PR)
There is some "button logic" implemented, such that a person can only click `CONFIRM` when all required fields are filled out.
### Demo Election UI
ElectionSetupDemo

## DropdownSelector
To create the choice of the Voting method dropdown, I created a new component called `DropdownSelector.tsx`, it uses a native picker so it should work natively and for the web.
It takes as input a string array of possible options and a function that stores the selected option.
DropDownSelector
This gif shows just how it displays multiple values which are given in the string array. 
Currently, we only have one voting method 'Plurality', but I still implemented this to be consistent with the documentation

Next steps from me (building on this):
- PR with a truly native datepicker
- PR which allows to send the correct message according to the protocol to the backend


fe1-web/parts/lao/organizer/eventCreation/events/CreateElection.tsx
```diff
@@ -0,0 +1,158 @@
+import React, { useState } from 'react';
```
CaiMusso, line 1: Looks **GREAT**! I will add comments on this file for an issue I encountered which is **not** a priority at the moment. For the purpose of this PR I approve the changes since my remark is more of an idea for improvement on **next** PRs. 

---------------------------------------------------------
The issue is the following: 
Supposing I add 20 ballot options and then I would like to remove ballot option number 3. I would have to remove all the options I created after option 3 before removing option 3. This happens because there is only one **remove** button. One idea (which I will explain commenting this file) would be to render a specific _per-ballot-option_ **remove** button when adding ballot options and removing the general remove button. 
Unfortunately (I tried it locally) the **WideButtonView** is pretty bulky on the screen to have 20 of them, it creates confusion. So another idea will be to do the same but creating a new (_small_) button component which stands on the right side of the **TextInput** of each ballot option.

fe1-web/parts/lao/organizer/eventCreation/events/CreateElection.tsx
```diff
@@ -0,0 +1,158 @@
...
+  };
+
+  const onRemoveBallotPress = () => {
+    // decrements counter
+    setElectionBallotCounter((prevCount) => prevCount - 1);
+    // removes value from ballot array
+    const newArr = [...electionBallots];
+    newArr.splice(electionBallotCounter - 1, 1);
+    setElectionBallots(newArr);
+  };
```
CaiMusso, line 84: Here this arrow function should get the **ballot option index** as a parameter and update the state according to the index. Could look something like:


```
const onRemoveBallotPress = (ballotId) => {

        setElectionBallotCounter((prevCount) => prevCount - 1);
        const newArr = electionBallors.filter(b => b.id !== ballotId);
        ......
        setElectionBallots(newArr);

```

fe1-web/parts/lao/organizer/eventCreation/events/CreateElection.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+  // Creates all the Ballot option text-fields based on the electionBallotCounter
+  for (let i = 0; i < electionBallotCounter; i += 1) {
+    const ballotPlaceholder = `${STRINGS.election_create_ballot_option} ${i + 1}`;
+    ballotOptionsUIComponents.push(<TextInput
+      style={styles.textInput}
+      placeholder={ballotPlaceholder} // Add index in the setState
+      onChangeText={(text: string) => { updateBallotArray(i, text); }}
+      key={ballotPlaceholder}
+    />);
```
CaiMusso, line 101: Here you could render the **specific per-ballot-option remove button** (done with WideButtonView for the moment but should be done with **small** button)

```
    ballotOptionsUIComponents.push(
      <>
        <TextInput
          style={styles.textInput}
          placeholder={ballotPlaceholder}
          onChangeText={(text: string) => { updateBallotArray(i, text); }}
          key={ballotPlaceholder}
        />
        <WideButtonView
          onPress={() => onRemoveBallotPress({ id of ballot option to delete })} // becomes arrow function
          .......
          title="-"
        />
      </>,
    );
```

fe1-web/parts/lao/organizer/eventCreation/events/CreateElection.tsx
```diff
@@ -0,0 +1,158 @@
...
+      <TextBlock text={STRINGS.election_create_ballot_options} />
+      <WideButtonView
+        onPress={onAddBallotPress}
+        title="+"
+      />
+      <WideButtonView
+        onPress={onRemoveBallotPress}
+        disabled={!removeButtonVisibility}
+        title="-"
+      />
```
CaiMusso, line 143: this could be removed if one **remove button** is done per ballot option

CaiMusso (approved):
Looks great! I just added some improvement remarks which could be addressed in next PRs, all good for this PR in my opinion.

fe1-web/parts/lao/organizer/eventCreation/events/CreateElection.tsx
Adalsteinnjons: Thanks for the review.
Great idea with the small button to remove a specific ballot option. That would make it much easier to use when creating elections with many options. 
This is a good point to have for future PR, as you mentioned. :)

fe1-web/parts/lao/organizer/eventCreation/events/CreateElection.tsx
```diff
@@ -0,0 +1,158 @@
...
+import DatePicker from 'components/DatePicker';
+import ParagraphBlock from 'components/ParagraphBlock';
+import WideButtonView from 'components/WideButtonView';
+import { Timestamp } from 'model/objects';
+import TextBlock from 'components/TextBlock';
+import DropdownSelector from 'components/DropdownSelector';
+
+function dateToTimestamp(date: Date): Timestamp {
+  return new Timestamp(Math.floor(date.getTime() / 1000));
+}
```
RaulinN, line 18: This function is replicated three times in the project (once here, once in Carlo's review, and once in one of the original DatePicker files). A solution could be adding this directly in the `Timestamp` mode/object file. You could have:
- `dateToTimestamp` (as a  static method)
- `timestampToDate` (since Carlo needs it and implements it in his PR) as a regular method

fe1-web/parts/lao/organizer/eventCreation/events/CreateElection.tsx
```diff
@@ -0,0 +1,158 @@
...
+  // Sets initial end date to 1 hour later than start date
+  initialEndDate.setHours(initialEndDate.getHours() + 1);
+
+  const [startDate, setStartDate] = useState(dateToTimestamp(initialStartDate));
+  const [endDate, setEndDate] = useState(dateToTimestamp(initialEndDate));
+
+  const [electionName, setElectionName] = useState('');
+  const [electionQuestion, setElectionQuestion] = useState('');
+  const [electionBallots, setElectionBallots] = useState(['']);
+  const [electionBallotCounter, setElectionBallotCounter] = useState(2);
```
RaulinN, line 39: You are using the magic number "2" a few times. Consider using a constant representing the minimum amount of ballot options

RaulinN (changes requested):
I found two bugs while testing this feature:

1. Fill in the name and question, add a ballot option and fill the two first ballot options:
    - click confirm. The console prints two ballot options selected (`Array [ "f", "b" ]`) which is what we expect
    - fill in the third ballot and then delete the text you just wrote (third box). Click confirm. This time, the console prints `Array(3) [ "f", "b", "" ]` (so 3 options instead of the expected 2)
2. Imagine you are on mobile and you create an election for $n = 26$ options (one for each canton) to query the attendees origins. After the tedious process of filling all the 26 cantons, you realise that you wrote "Fribourg" twice (once at position 5 and once at position 13). Removing one of the two makes the confirm button not clickable anymore and you have to correct all the subsequent ballot options so that there is no empty slot in the middle

These two issues can be resolved with one change: if there are `n` ballot options displayed with `k <= n` of them filled (aka not empty string), then the resulting array printed by the console should have size `k`

fe1-web/parts/lao/organizer/eventCreation/events/CreateElection.tsx
pierluca: I agree with @CaiMusso , it might make sense to have a tiny button next to each ballot option, to remove any of them. This would also address @RaulinN 's remark

fe1-web/parts/lao/organizer/eventCreation/events/CreateElection.tsx
pierluca: I tend to agree, but I see this was resolved as well, was there some discussion between you two? :)

fe1-web/parts/lao/organizer/eventCreation/events/CreateElection.tsx
pierluca: I tend to agree, but I see this was resolved as well, was there some discussion between you two? :)

fe1-web/parts/lao/organizer/eventCreation/events/CreateElection.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+  // Updates the array with the specified ballot entries in the textfields
+  const updateBallotArray = (index: number, text: string) => {
+    const newArr = [...electionBallots];
+    newArr[index] = text;
+    setElectionBallots(newArr);
+  };
+
+  // Creates all the Ballot option text-fields based on the electionBallotCounter
+  for (let i = 0; i < electionBallotCounter; i += 1) {
```
pierluca, line 94: why use a separate `electionBallotCounter` rather than the length of the electionBallots array ? 
`electionBallotCounter` seems like an unnecessary additional state

pierluca (changes requested):
I second @CaiMusso and @RaulinN 's comments.

I'm also wondering if it might make sense to split out the ballot-option logic into a separate part, which would get the electionBallots state as parameter and deal with its specific logic and display (`onAddBallotPress`, `onRemoveBallotPress`, `updateBallotArray`, etc.)

fe1-web/parts/lao/organizer/eventCreation/events/CreateElection.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+  const [electionName, setElectionName] = useState('');
+  const [electionQuestion, setElectionQuestion] = useState('');
+  const [electionBallots, setElectionBallots] = useState(['']);
+  const [electionBallotCounter, setElectionBallotCounter] = useState(2);
+  const votingMethods = [STRINGS.election_method_Plurality];
+  const [votingMethod, setVotingMethod] = useState(votingMethods[0]);
+  const ballotOptionsUIComponents = [];
+
+  const buildDatePickerWeb = () => {
```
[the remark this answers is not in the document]
Adalsteinnjons: To my knowledge, the [React DatePicker library](https://reactdatepicker.com/) only supports the Web. 
I made a native Component `DatePicker.native.tsx` on another branch but haven't been able to test it properly since I can't get the project to run on my iPhone yet. 
The issues I get are when the required file is outside the folder fe1-web (such as the ones in the protocol folder), then it can't find it and can't compile. 
I was planning on asking about this in our meeting tomorrow.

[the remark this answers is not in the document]
pierluca: This might take a fair amount of time to figure out, and I'd be happy to work with you on it, as it's something we're definitely interested in and the engineering team did not have time to look into proper react native support.

We can either meet earlier (3pm onwards) or stay longer in the meeting, after covering @CaiMusso 's needs, in case he wants to leave earlier. Let me know on Slack.

fe1-web/parts/lao/organizer/eventCreation/events/CreateElection.tsx
Adalsteinnjons: Very good idea! Thank you
Implemented your suggestion in commit `f04fa372fba889bb92f7ec416286f1c05f621b0c`

RaulinN:
Changes you made in f04fa37 regarding `Timestamp`s look good! 👍

Adalsteinnjons:
The new version allows for individually deletable options. It should address most of your concerns from before.
I made it similar to how the text-doodle works in [here](https://doodle.com/create/options#text). 
Here is a video showing the new version
NewVersionElectionSetup

fe1-web/parts/lao/organizer/eventCreation/events/CreateElection.tsx
Adalsteinnjons: This is now implemented

fe1-web/parts/lao/organizer/eventCreation/events/CreateElection.tsx
Adalsteinnjons: This is implemented

fe1-web/parts/lao/organizer/eventCreation/events/CreateElection.tsx
Adalsteinnjons: This is implemented

Adalsteinnjons:
> I found two bugs while testing this feature:
> 
> 1. Fill in the name and question, add a ballot option and fill the two first ballot options:
>    
>    * click confirm. The console prints two ballot options selected (`Array [ "f", "b" ]`) which is what we expect
>    * fill in the third ballot and then delete the text you just wrote (third box). Click confirm. This time, the console prints `Array(3) [ "f", "b", "" ]` (so 3 options instead of the expected 2)
> 2. Imagine you are on mobile and you create an election for $n = 26$ options (one for each canton) to query the attendees origins. After the tedious process of filling all the 26 cantons, you realise that you wrote "Fribourg" twice (once at position 5 and once at position 13). Removing one of the two makes the confirm button not clickable anymore and you have to correct all the subsequent ballot options so that there is no empty slot in the middle
> 
> These two issues can be resolved with one change: if there are `n` ballot options displayed with `k <= n` of them filled (aka not empty string), then the resulting array printed by the console should have size `k`

Thank you very much for the detailed comment. I re-made the structure and both bugs have been fixed and the structure is better now.

fe1-web/parts/lao/organizer/eventCreation/events/CreateElection.tsx
Adalsteinnjons: Now I don't use the counter anymore :)

fe1-web/parts/lao/organizer/eventCreation/events/CreateElection.tsx
```diff
@@ -0,0 +1,158 @@
...
+          selected={endTime}
+          onChange={(date: Date) => setEndDate(dateToTimestamp(date))}
+        />
+      </View>
+    );
+  };
+
+  // Confirm button only clickable when the Name, Question and 2 Ballot options have values
+  const buttonsVisibility: boolean = (electionQuestion !== '' && electionName !== ''
+    && electionBallots.length >= 2 && electionBallots[0] !== '' && electionBallots[1] !== '');
```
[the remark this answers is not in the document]
Adalsteinnjons: Thank you for this suggestion. I followed a very similar approach

fe1-web/components/RemovableTextInput.tsx
```diff
@@ -0,0 +1,70 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {
+  StyleSheet, TextInput, TextStyle, View, ViewStyle,
+} from 'react-native';
+import DeleteButton from 'components/DeleteButton';
+import STRINGS from 'res/strings';
+import { Spacing, Typography, Views } from '../styles';
```
pierluca, line 8:
```suggestion
import { Spacing, Typography, Views } from 'styles';
```

fe1-web/components/TextInputList.tsx
```diff
@@ -0,0 +1,78 @@
...
+ */
+
+const TextInputList = (props: IPropTypes) => {
+  const { onChange } = props;
+  const [idCount, setIdCount] = useState(1);
+  const [userOptions, setUserOptions] = useState([{ id: 0, value: '' }]);
+
+  const updateParent = (options: { id: number, value: string }[]) => {
+    // Gets the distinct options which are not empty ('')
+    // Set() keeps the order the same
```
pierluca, line 21: thanks, I learned something 😄

fe1-web/components/TextInputList.tsx
```diff
@@ -0,0 +1,78 @@
...
+ * Component which creates a list of Text input fields which can be deleted individually
+ * Returns array with non-empty Values of the text fields
+ * It uses the RemovableTextInput.tsx component to show the individual inputs
+ * Input: A function that stores the values that were inputted by the user
+ * Output: Array of the unique, nonempty values inputted by the user
+ */
+
+const TextInputList = (props: IPropTypes) => {
+  const { onChange } = props;
+  const [idCount, setIdCount] = useState(1);
```
pierluca, line 16: I think you could do without this, by simply referencing the length of userOptions, and using `map( (val,idx) => { ...} )` when you need to iterate.

On the other hand, if this is what you've tested and it works, don't waste extra time doing this for now :)

fe1-web/components/TextInputList.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const [userOptions, setUserOptions] = useState([{ id: 0, value: '' }]);
```
pierluca, line 17: `options` is counterintuitive in the context of a `TextInputList` , it's gotten me confused a few times.

I would suggest renaming to `userInputs, setUserInputs` and the functions: `addInput`, `removeInput`, etc.

fe1-web/parts/lao/organizer/eventCreation/events/CreateElection.tsx
```diff
@@ -30,16 +27,15 @@ const CreateElection = ({ route }: any) => {
...
   const [electionName, setElectionName] = useState('');
   const [electionQuestion, setElectionQuestion] = useState('');
   const [electionBallots, setElectionBallots] = useState(['']);
-  const [electionBallotCounter, setElectionBallotCounter] = useState(2);
   const votingMethods = [STRINGS.election_method_Plurality];
   const [votingMethod, setVotingMethod] = useState(votingMethods[0]);
-  const ballotOptionsUIComponents = [];
+  const minBallotOptions = 2;
 
   const buildDatePickerWeb = () => {
```
pierluca, line 40: so much cleaner ❤️

fe1-web/res/img/delete.svg
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0"?>
```
pierluca, line 1: did you draw this yourself or did you get it from somewhere?
If it's the latter, you might want to reference the source and include the license terms in the project.

fe1-web/res/img/delete.svg
```diff
@@ -0,0 +1,41 @@
...
+<g xmlns="http://www.w3.org/2000/svg">
+</g>
+<g xmlns="http://www.w3.org/2000/svg">
+</g>
+<g xmlns="http://www.w3.org/2000/svg">
+</g>
+<g xmlns="http://www.w3.org/2000/svg">
+</g>
+<g xmlns="http://www.w3.org/2000/svg">
+</g>
```
pierluca, line 40: I'm pretty sure these can be removed :)

fe1-web/res/img/delete.svg
```diff
@@ -0,0 +1,41 @@
...
+<script/></g></svg>
```
pierluca, line 41: and the script tag too
```suggestion
</g></svg>
```

pierluca (commented):
Minor remarks, we're almost there. It's great work !

fe1-web/components/TextInputList.tsx
Adalsteinnjons: I think we need a unique ID for each textField. 
When the `RemovableTextInput` gets created it gets the Id as a parameter from `TextInputList` which it uses in the removeInput and updateInput function
So if we delete an entry in the middle of the list of textfields, the indexes don't match anymore.  

**1. Current state**
Input_0, index: 0, remove(0)
Input_1, index: 1, remove(1)
Input_2, index: 2, remove(2)

**2. Delete Input_1**
Input_0, index: 0, remove(0)
Input_2, index: 1, remove(2)

**3. Try to delete Input_2**
remove(2) doesn't work since no Item there

At least this is what happens in my implementation. 
Perhaps by refactoring the logic of the functions it is possible to use only indexes, but I thought that having unique IDs would be better.

fe1-web/components/TextInputList.tsx
Adalsteinnjons: Yes, in python and C I think it messes the order up but in JS it doesn't :)

fe1-web/res/img/delete.svg
Adalsteinnjons: The author of the icon is Google.
https://www.flaticon.com/packs/action
Where should I include that source?

fe1-web/components/TextInputList.tsx
Adalsteinnjons: Very good suggestion, this will be updated.

fe1-web/res/img/delete.svg
Adalsteinnjons: You really check everything :D
This will get removed

fe1-web/components/RemovableTextInput.tsx
Adalsteinnjons: Implemented :)

fe1-web/components/TextInputList.tsx
pierluca: Fair enough. Thanks for the rationale 👍

fe1-web/res/img/delete.svg
Adalsteinnjons: Ah yes perfect. This seems to be the original source indeed.
Here it is: https://github.com/google/material-design-icons/blob/master/src/action/delete/materialicons/24px.svg

pierluca: If that's the original, you might want to use it as-is here :)

fe1-web/parts/lao/organizer/eventCreation/events/CreateElection.tsx
```diff
@@ -13,6 +13,7 @@ import { Timestamp } from 'model/objects';
 import TextBlock from 'components/TextBlock';
 import DropdownSelector from 'components/DropdownSelector';
 import TextInputList from 'components/TextInputList';
+import { requestCreateElection } from '../../../../../network';
```
pierluca, line 16:
```suggestion
import { requestCreateElection } from 'network';
```
would be more readable :)

fe1-web/model/network/method/message/data/election/CreateElection.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  public readonly end_time: Timestamp;
+
+  public readonly questions: {
+    id: string,
+    question: string,
+    voting_method: string,
+    ballot_options: string[],
+    write_in: boolean,
+  }[];
```
pierluca, line 33: you might want to define a sub-type for this.

fe1-web/model/network/method/message/data/election/CreateElection.ts
```diff
@@ -0,0 +1,94 @@
...
+    // FIXME: implementation not finished, get event from storage,
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const lao: Lao = OpenedLaoStore.get();
+    console.log('Create Election called: ', this);
+
+    // const expectedHash = Hash.fromStringArray(
+    //   EventTags.ELECTION, lao.id.toString(), lao.creation.toString(), msg.name,
+    // );
+    // eslint-disable-next-line max-len
+    // if (!expectedHash.equals(msg.id)) throw new ProtocolError('Invalid \'id\' parameter encountered during \'CreateMeeting\': unexpected id value');
```
pierluca, line 76: huh? why do you need this?

fe1-web/model/network/method/message/data/election/CreateElection.ts
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  public static fromJson(obj: any): CreateElection {
```
pierluca, line 79: you might want to add tests for this, just to make sure that it works properly :)

fe1-web/network/MessageApi.ts
```diff
@@ -208,3 +209,38 @@ export function requestCloseRollCall(rollCallId: Number, attendees: PublicKey[])
...
+  version: string,
+  start: Timestamp,
+  end: Timestamp,
+  questions: {
+    id: string,
+    question: string,
+    voting_method: string,
+    ballot_options: string[],
+    write_in: boolean,
+  }[],
```
pierluca, line 227: I'm even more convinced that you definitely want a type for this :)

fe1-web/parts/lao/organizer/eventCreation/events/CreateElection.tsx
pierluca, line 16:
```suggestion
import { requestCreateElection } from 'network';
```
would be more readable :)

fe1-web/parts/lao/organizer/eventCreation/events/CreateElection.tsx
```diff
@@ -71,6 +72,26 @@ const CreateElection = ({ route }: any) => {
...
+      electionName,
+      '1',
+      startDate,
+      endDate,
+      [{
+        id: 'idtest',
+        question: electionQuestion,
+        voting_method: votingMethod,
+        ballot_options: electionBallots,
+        write_in: false,
```
pierluca, line 86: I'd suggest splitting this initialization from the function call.

pierluca (commented):
A few quick remarks, I had a bit of time for it :)

fe1-web/model/network/method/message/data/election/CreateElection.ts
Adalsteinnjons: Forgot to delete this. :facepalm: :D

fe1-web/model/network/method/message/data/election/CreateElection.ts
```diff
@@ -0,0 +1,94 @@
+import {
+  Hash, Timestamp, Lao,
+} from 'model/objects';
+import { OpenedLaoStore } from 'store';
+import { ProtocolError } from 'model/network/ProtocolError';
+import { validateDataObject } from 'model/network/validation';
+import { ActionType, MessageData, ObjectType } from '../MessageData';
+import { checkTimestampStaleness } from '../Checker';
+
```
pierluca, line 9:
```suggestion
export interface ElectionQuestion {
     id: string,
     question: string,
     voting_method: string,
     ballot_options: string[],
     write_in: boolean,
   }
```

?

fe1-web/model/network/method/message/data/election/CreateElection.ts
Adalsteinnjons: Subtype is not defined as an interface in the model/objects/Election.ts file

fe1-web/__tests__/model/network/FromJsonData.test.ts
```diff
@@ -44,6 +52,7 @@ describe('=== fromJsonData checks ===', () => {
   const time = STANDARD_TIMESTAMP;
   const name = 'poof';
   const location = 'Lausanne';
+  const mockVersion = STRINGS.election_version_identifier;
```
Adalsteinnjons, line 55: I am not 100% sure what the Version field in the SetupElection should be, but having communicated with Victor, they just set it to '1.0.0', which I do here

fe1-web/parts/lao/organizer/eventCreation/events/CreateElection.tsx
```diff
@@ -0,0 +1,158 @@
...
+  return (
+    <ScrollView>
+      <TextBlock text={STRINGS.election_create_setup} bold />
+      <TextInput
+        style={styles.textInput}
+        placeholder={STRINGS.election_create_name}
+        onChangeText={(text: string) => { setElectionName(text); }}
+      />
+      { /* see archive branches for date picker used for native apps */ }
+      { Platform.OS === 'web' && buildDatePickerWeb() }
```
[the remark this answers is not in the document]
Adalsteinnjons: Yes

Adalsteinnjons:
I implemented tests for the Election Setup, and the communication with the mock server is working. The next step is to try it with the backend.

fe1-web/model/network/method/message/data/election/CreateElection.ts
Adalsteinnjons: This is implemented in the `objects/Election.ts` file

fe1-web/__tests__/model/network/FromJsonData.test.ts
pierluca: I think that's specified in the e-voting document :)

fe1-web/__tests__/model/network/FromJsonData.test.ts
```diff
@@ -456,7 +533,7 @@ describe('=== fromJsonData checks ===', () => {
 
     it('should fail when omitting a mandatory parameter', () => {
```
pierluca, line 534: it seems that this section, described as `should fail (throw) during object creation` , really only tests `CreateLao`.
I'd suggest renaming it to `should fail (throw) during CreateLao creation`
And split your tests apart in another describe, dealing only with the `SetupElection` object.

fe1-web/ingestion/handlers/Election.ts
```diff
@@ -0,0 +1,61 @@
...
+  const election = new Election({
+    lao: elecMsg.lao,
+    id: elecMsg.id,
+    name: elecMsg.name,
+    version: elecMsg.version,
+    created_at: elecMsg.created_at,
+    start: elecMsg.start_time,
+    end: elecMsg.end_time,
+    questions: elecMsg.questions,
+
```
pierluca, line 39: no need for this blank line :)

fe1-web/model/network/method/message/data/election/SetupElection.ts
```diff
@@ -0,0 +1,98 @@
...
+    if (msg.end_time < msg.created_at) throw new ProtocolError('Invalid timestamp encountered: \'end\' parameter smaller than \'creation\'');
+    if (msg.end_time < msg.start_time) throw new ProtocolError('Invalid timestamp encountered: \'end\' parameter smaller than \'start\'');
+    this.end_time = msg.end_time;
+    if (msg.created_at > msg.start_time) {
+      // This avoids errors when the start-time is set before creation time (Which is invalid)
+      // that can happen unwillingly when the organizer takes time creating an election
+      this.start_time = msg.created_at;
+    } else {
+      this.start_time = msg.start_time;
+    }
```
pierluca, line 60: This might be called when receiving a message from the network though, thus "corrupting" its original content.

Is there a better place to address this issue? (closer to the source of the problem, perhaps)

fe1-web/res/strings.ts
```diff
@@ -99,13 +99,15 @@ const STRINGS = {
   election_create_setup: 'Election Setup',
   election_create_name: 'Name*',
   election_create_start_time: 'Start time: ',
-  election_create_finish_time: 'End time: ',
+  election_create_finish_time: 'End time:  ',
```
pierluca, line 102: is the extra space on purpose?

pierluca (commented):
Most issues are very minor, overalls a pretty good job 👍 

Be careful with the SetupElection object though, as that seems to me like a serious issue.

fe1-web/model/network/method/message/data/election/SetupElection.ts
Adalsteinnjons: Good point! I didn't think about that being called when getting a message from the network.
I will do it at the source in the creation.

fe1-web/res/strings.ts
Adalsteinnjons: Yes, I tried to make the election setup look a tiny bit better. 
image
With the extra space, the datepicker lookes more centered. But this could probably also be solved with styling.

fe1-web/ingestion/handlers/Election.ts
Adalsteinnjons: Fixed

fe1-web/model/network/method/message/data/election/SetupElection.ts
Adalsteinnjons: Now its implemented as a ternary operator in the MessageApi.ts 
```
start_time: ((start > time) ? start : time),
```

fe1-web/__tests__/model/network/FromJsonData.test.ts
Adalsteinnjons: It also tests `stateLao`, `createMeeting` and `WitnessMessage` in the same describe. Should I put all of them in their own describe?

fe1-web/parts/lao/organizer/eventCreation/events/CreateElection.tsx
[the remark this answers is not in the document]
Adalsteinnjons: Marked as resolved bacause we decided to leave the native date-picker for future work. It will not be in this PR.

Adalsteinnjons:
This PR is getting quite big and I think now is a good time to stop adding funcitonality.
The Election Setup works with the back-end and the elections get displayed in the event view. 
(To test it with the back-end, use the branch: `work-be1-go-tesa-electionSetupToTestWithFrontEnd`)
There are also 2 bug fixes in the PR which make the events display in the event view. 

If anyone wants to review it with me on zoom to clarify questions I am always available. :)

pierluca:
Sonar is *really* not happy about some of the stuff. You might want to address at least the major and critical things (except for variable names)

fe1-web/model/network/method/message/data/election/SetupElection.ts
```diff
@@ -48,16 +48,10 @@ export class SetupElection implements MessageData {
     checkTimestampStaleness(msg.start_time);
     if (!msg.end_time) throw new ProtocolError('Undefined \'end_time\' parameter encountered during \'SetupElection\'');
     checkTimestampStaleness(msg.end_time);
-    if (msg.end_time < msg.created_at) throw new ProtocolError('Invalid timestamp encountered: \'end\' parameter smaller than \'creation\'');
+    if (msg.start_time < msg.created_at) throw new ProtocolError('Invalid timestamp encountered: \'start\' parameter smaller than \'created_at\'');
+    this.start_time = msg.start_time;
```
pierluca: you might want to use `{ ... }` around the if body

Adalsteinnjons:
> Sonar is _really_ not happy about some of the stuff. You might want to address at least the major and critical things (except for variable names)

Ok, most of the issues are just from lines like this:
`if (!msg.id) throw new ProtocolError('Undefined \'id\' parameter encountered during \'SetupElection\'');`
I did it in the same way it was already done in the other classes. But of course I can fix it.

pierluca (approved):
Sonar is unhappy, but I'm pretty happy with the latest changes 👍 
I'm approving to unblock (as I'll be on holidays), but I trust you'll fix those things before merging.

fe1-web/model/network/method/message/data/election/SetupElection.ts
Adalsteinnjons: Thanks! I just changed that and it resolved most of them.
Now the only code smells are:
- variable name (Need to be that way for clarity with the JSON)
- switch statement with only 1 case (in the future more cases will come, probably in next PR even)
- one severe one about a long ling which I can't see the issue with:
image

pierluca:
Okay, the quality gate actually fails because of "1 critical issue", which is the cognitive complexity of the SetupElection constructor.

If you validate the `msg.questions` object in a separate function, the code will be clearer and the complexity lower ;-)

https://sonarcloud.io/project/issues?id=dedis_student_21_pop_fe1&open=AXkK9ljsa56fLAtu5sF3&pullRequest=288&resolved=false&severities=CRITICAL&sinceLeakPeriod=true

Adalsteinnjons:
> Okay, the quality gate actually fails because of "1 critical issue", which is the cognitive complexity of the SetupElection constructor.
> 
> If you validate the `msg.questions` object in a separate function, the code will be clearer and the complexity lower ;-)
> 
> https://sonarcloud.io/project/issues?id=dedis_student_21_pop_fe1&open=AXkK9ljsa56fLAtu5sF3&pullRequest=288&resolved=false&severities=CRITICAL&sinceLeakPeriod=true

Thanks. Now it passed

fe1-web/components/DropdownSelector.tsx
```diff
@@ -0,0 +1,45 @@
...
+import PropTypes from 'prop-types';
+import { Picker } from '@react-native-picker/picker';
+import styles from 'styles/stylesheets/container';
+
+/**
+ * Simple dropdown where the user can select options
+ * Inputs:
+ *  - String array of all options (required)
+ *  - function that stores the selected option (required)
+ *      (e.g. {(method: string) => setSelectedElectionMethod(method)} )
```
RaulinN, line 11:
```suggestion
 *  - Callback function for when an item is selected (required)
 *      (e.g. {(method: string) => setSelectedElectionMethod(method)} )
```

fe1-web/components/DeleteButton.tsx
```diff
@@ -0,0 +1,35 @@
...
+  } as ImageStyle,
+});
+
+function DeleteButton({ action }: IPropTypes) {
+  return (
+    <Pressable onPress={action}>
+      <Image style={styles.icon} source={trashcan} />
+    </Pressable>
+  );
+}
```
RaulinN, line 26: Good! One little thing maybe: have you tried modifying the user cursor whenever you hover over the trashcan? Whenever a link/button/... is clickable, the cursor generally changes to the "hand cursor". Can we do this here too?

I didn't look into how to do it using react, but I'm asking this because I know that it is veeery easy to do it in CSS (closely related to the way react works)

fe1-web/components/TextInputList.tsx
```diff
@@ -0,0 +1,78 @@
...
+ */
+
+const TextInputList = (props: IPropTypes) => {
+  const { onChange } = props;
+  const [idCount, setIdCount] = useState(1);
+  const [userInputs, setUserInputs] = useState([{ id: 0, value: '' }]);
+
+  const updateParent = (options: { id: number, value: string }[]) => {
+    // Gets the distinct options which are not empty ('')
+    const distinctValues = [...new Set(options.map((option) => option.value))].filter((value) => value !== '');
```
RaulinN, line 21: 👍

fe1-web/network/MessageApi.ts
```diff
@@ -15,11 +16,12 @@ import {
 import {
   OpenedLaoStore, KeyPairStore,
 } from 'store';
+import { Question } from 'model/objects/Election';
 import { publish } from './JsonRpcApi';
 
 /** Send a server query asking for the creation of a LAO with a given name (String) */
 export function requestCreateLao(laoName: string): Promise<Channel> {
-  const time = Timestamp.EpochNow();
+  const time = new Timestamp(1619626038);
```
RaulinN, line 24: artifact of testing?

fe1-web/network/MessageApi.ts
```diff
@@ -201,3 +203,32 @@ export function requestCloseRollCall(rollCallId: Number, attendees: PublicKey[])
...
+  const message = new SetupElection({
+    lao: currentLao.id.toString(),
+    id: Hash.fromStringArray(
+      EventTags.ELECTION, currentLao.id.toString(), currentLao.creation.toString(), name,
+    ),
+    name: name,
+    version: version,
+    created_at: time,
+    start_time: ((start > time) ? start : time),
+    end_time: end,
```
RaulinN, line 228: You might have `start_time` > `end_time` with the protection you added at line 227. Might add such protection for `end_time` as well (cannot be smaller than `start_time`)

fe1-web/model/objects/Hash.ts
```diff
@@ -5,6 +5,8 @@ import { Base64Data } from './Base64';
 export enum EventTags {
   MEETING = 'M',
   ROLL_CALL = 'R',
+  ELECTION = 'Election',
+  QUESTION = 'Question',
 }
```
RaulinN, line 10: @pierluca did we agree on this? I think the point of having one byte as event tag is (in part) so that the hash doesn't leak any information rihgt?

RaulinN: Update: actually we indeed agreed on that on a [merged PR](https://github.com/dedis/student_21_pop/pull/265/files). I think we should open an issue for that. What do you think?

fe1-web/model/network/method/message/data/election/SetupElection.ts
```diff
@@ -0,0 +1,124 @@
...
+import { Question } from 'model/objects/Election';
+import { ActionType, MessageData, ObjectType } from '../MessageData';
+import { checkTimestampStaleness } from '../Checker';
+
+export class SetupElection implements MessageData {
+  public readonly object: ObjectType = ObjectType.ELECTION;
+
+  public readonly action: ActionType = ActionType.SETUP;
+
+  public readonly lao: string;
```
RaulinN, line 16: According to [the protocol](https://github.com/dedis/student_21_pop/pull/265/files), this field is the lao id. Thus, if I'm not wrong, the type should be a `Hash` like just below

fe1-web/model/objects/Election.ts
```diff
@@ -0,0 +1,99 @@
...
+export interface Question {
+  id: string,
+  question: string,
+  voting_method: string,
+  ballot_options: string[],
+  write_in: boolean,
+}
+
+export class Election implements LaoEvent {
+  public readonly lao: string;
```
RaulinN, line 24: Same as above I think

RaulinN (changes requested):
Really nice job overall :)

fe1-web/model/objects/Hash.ts
gnarula: @RaulinN Maybe I'm missing something - can you explain how |tag| > 1 leaks information?

fe1-web/components/DeleteButton.tsx
Adalsteinnjons: Really good point. Added it by doing

```const styles = StyleSheet.create({
  icon: {
    width: 26,
    height: 26,
    cursor: 'pointer',
  } as ImageStyle,
});

fe1-web/network/MessageApi.ts
Adalsteinnjons: Yes :worried:

fe1-web/network/MessageApi.ts
Adalsteinnjons: Good point!

Adalsteinnjons: I made it such that when a person does this it will put the end time 60 seconds after the start time. 
In the future it would be good to automatically block this behaviour in the UI.

fe1-web/model/network/method/message/data/election/SetupElection.ts
Adalsteinnjons: Yes that's right. It is changed now :)

fe1-web/components/DropdownSelector.tsx
Adalsteinnjons: Much better, thank you.

RaulinN (approved):
👍

PR #293, opened 2021-04-13
Roll call organizer's UI

Kar1213:
The following changes are mostly about the UI for managing roll calls. An organizer can now handle roll calls but the UI for attendees to join them is still not fully implemented.

+ EventState.java: enum used in roll call class to keep track of current state
+ RollCall.java: added state attribute
+ OpenRollCall.java: distinguish between OPEN and REOPEN action
+ Changes made to adapt to Gaurav’s PR #287: CreateRollCall, OpenRollCall, CloseRollCall, Injection, JsonCreateRollCallSerializer, JsonOpenRollCallSerializer, JsonCloseRollCallSerializer and LAORepository
+ layout\_event.xml: added fields for roll calls + added buttons depending on isOrganizer and roll call state
+ EventExpandableListViewAdapter.java: 
	- putEventsInMap method: fixed bug where eventsMap attribute was reset which prevented added events to be saved
	- getChildView method: setting up the display of roll call in the list through binding
+ RollCallEventCreationFragment.java:
	- based on the structure of the other fragments, I restructured the class by moving some code around
	- added/modified click listeners for roll call creation, opening and closing
	- added observer for roll call creation for the "scheduled" case to open LAO detail (see LaoDetailViewModel just below)
+ LaoDetailViewModel.java:
	- createNewRollCall method: creates a CreateRollCall instance from method paramaters and sends a publish message, for the "scheduled" case (parameter open set to false) go to LAO detail fragment, method returns roll call id to be used for opening (next method below)
	- openConnectRollCall method: creates a OpenRollCall instance based on roll call id, sends publish message and on response saves roll call id in mCurrentRollCallId to be used for closing, camera/permission is open
	- attendees are added in onQRCodeDetected method
	- closeRollCall method: similarly it closes the currently open roll call, attendees list is given to CloseRollCall instance and then cleared
	- opening the camera: inspired by what was previously done in HomeViewModel.java
+ QRCodeScanningFragment.java and fragment\_qrcode.xml:
	- exisiting code was for scanning LAO QR code, modified it such as it uses the same fragment and layout to also handle scanning for roll calls
	- added code to handle attendees counter and roll call closing when scanning
LaoDetailFragment.java: added setupEventListUpdates method to update list of events in fragment
+ LAORepository.java: 
	- handleCreateRollCall, handleOpenRollCall and handleCloseRollCall modified to adapt to new roll call fields and protocol
	- sendPublish method: to be able to test roll calls on the app one can uncomment the code that handles roll calls without waiting for a broadcast message from backend (as a temporary solution)


fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/adapters/EventExpandableListViewAdapter.java
```diff
@@ -242,15 +248,57 @@ public View getChildView(
...
     }
 
-    binding.setEvent(event);
+    //Event event = ((Event) getChild(groupPosition, childPosition));
+    Event event = (Event)getChild(groupPosition, childPosition);
+    if(event instanceof RollCall){
+      RollCall rollCall = (RollCall)event;
+      binding.setRollCall(rollCall);
+      binding.setIsRollCall(true);
+      SimpleDateFormat DATE_FORMAT = new SimpleDateFormat("HH:mm", Locale.FRENCH);
```
maximelaval, line 265: a date format constant is already defined above , I think you should use it to display the start/end  date + time

fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/adapters/EventExpandableListViewAdapter.java
```diff
@@ -242,15 +248,57 @@ public View getChildView(
...
+      binding.rollCallTime.setText("Time: "+DATE_FORMAT.format(new Date(1000*rollCall.getStart())));
+
+      binding.rollCallTitle.setText("Roll Call: "+rollCall.getName());
+
+      binding.setViewModel(viewModel);
+
+      boolean isOrganizer = viewModel.isOrganizer().getValue();
```
maximelaval, line 272: I think you should create a separate layout in xml for the Roll Calls , because in this layout you use a lot of attributes juste to know if it's a RollCall and it's a bit messy especially when there will more Events to use the same layout for all the events

fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/adapters/EventExpandableListViewAdapter.java
```diff
@@ -242,15 +248,57 @@ public View getChildView(
...
       LayoutInflater inflater = LayoutInflater.from(parent.getContext());
       binding = LayoutEventBinding.inflate(inflater, parent, false);
     } else {
       binding = DataBindingUtil.getBinding(convertView);
     }
 
-    binding.setEvent(event);
+    //Event event = ((Event) getChild(groupPosition, childPosition));
+    Event event = (Event)getChild(groupPosition, childPosition);
+    if(event instanceof RollCall){
```
maximelaval, line 261: I created an attribute EventType type for each event in the class Event so I think it would be nicer to do a switch case on this instead of having to do multiples if ( instanceof) else (instanceof) etc.. especially when there will be multiple events to handle.

maximelaval (changes requested):

fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/adapters/EventExpandableListViewAdapter.java
Kar1213: Right thanks I will use this one instead.

fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/adapters/EventExpandableListViewAdapter.java
Kar1213: Thanks for the idea, actually this layout should be similar to all events. The fields displayed are the same and there might only be tiny differences for the buttons. So I will remove the checks you mentioned and rename the elements with "event" instead of "roll_call".

fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/adapters/EventExpandableListViewAdapter.java
Kar1213: Honestly I don't think this is an issue and the lines of code stay the same anyway. I would prefer staying consistent with the style used overall in the code (for example in LAORepository for handleMessage) and this would also avoid the creation of an additional attribute.

Kar1213:
Please refer to PR #304

PR #294, opened 2021-04-13
Roll call logic

flmichel:
Implement the logic on the go backend for the roll call messages.


be1-go/hub/organizer_test.go
```diff
@@ -0,0 +1,326 @@
...
+	laoID, laoChannel, err := createLao(oHub, organizerKeyPair)
+	require.NoError(t, err)
+
+	keypair, err := generateKeyPair()
+	require.NoError(t, err)
+
+	// Create the roll call
+	dataCreate, err := newCorrectCreateRollCallData(laoID)
+	require.NoError(t, err)
+	err = laoChannel.processRollCallObject(keypair.publicBuf, dataCreate)
```
lpatel99, line 287: This is not working in the continuous integration. Perhaps you could use type composition to create a type similar to publicKey which implements the methods of the interface so that you can keep this as is.

lpatel99 (commented):
Otherwise the code seems good, unfortunately given a lack of familiarity with roll calls I am not best placed to properly understand the logic.

be1-go/hub/organizer_test.go
flmichel: Thank you for the review! I updated the tests now I think it is working.

be1-go/hub/organizer.go
```diff
@@ -228,6 +228,8 @@ func (o *organizerHub) createLao(publish message.Publish) error {
 	laoChannelID := "/root/" + encodedID
 
 	laoCh := laoChannel{
+		rollCall{},
+		make(map[string]struct{}),
```
gnarula, line 232: Please use the named syntax instead of positional one. Allows readers to know what field this map is for easily without looking up the struct :)

be1-go/message/answer.go
```diff
@@ -86,6 +87,14 @@ func (e *Error) Error() string {
 	return e.Description
 }
 
+// InvalidActionError
+func InvalidActionError(action interface{}) error {
```
gnarula, line 91:
```suggestion
func NewInvalidActionError(action DataAction) error {
```

be1-go/hub/organizer_test.go
```diff
@@ -0,0 +1,347 @@
...
+	point = point.Mul(secret, point)
+
+	pkbuf, err := point.MarshalBinary()
+	if err != nil {
+		return keypair{}, xerrors.Errorf("failed to create keypair: %v", err)
+	}
+	return keypair{point, pkbuf, secret}, nil
+}
+
+func getTime() message.Timestamp {
```
gnarula, line 45:
```suggestion
func timestamp() message.Timestamp {
```

be1-go/hub/organizer_test.go
```diff
@@ -0,0 +1,347 @@
...
+	}
+	// Creation of the data
+	data := &message.CreateLAOData{
+		GenericData: &message.GenericData{
+			Action: message.DataAction(message.CreateLaoAction),
+			Object: message.LaoObject,
+		},
+		ID:        laoID,
+		Name:      name,
+		Creation:  getTime(),
```
gnarula, line 66: You might want to reuse the one declared on line 53.

Offtopic, but this very likely indicates we're not verifying if laoID is consistent with `SHA256(organizer||creation||name)`

gnarula (changes requested):
Great work! Looks really good 👍  Added a few comments about the tests

be1-go/hub/organizer_test.go
```diff
@@ -0,0 +1,347 @@
...
+	return keypair{point, pkbuf, secret}, nil
+}
+
+func getTime() message.Timestamp {
+	return message.Timestamp(time.Now().Unix())
+}
+
+func createLao(o *organizerHub, oKeypair keypair) (string, *laoChannel, error) {
+	// Data of the Lao
+	name := strconv.Itoa(laoCounter)
```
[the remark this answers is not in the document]
flmichel: I added argument `name string`

be1-go/hub/organizer_test.go
flmichel: I agree. I think we're currently not verifying if any ID is consistent with its hash described in the protocol (expect for the roll call ids). Should I open an issue for that?

gnarula: Yes, please do create one for it

gnarula (dismissed):
LGTM

Edit: @flmichel we've a failing test. Can you look into that please? Good to merge after that

gnarula (approved):
LGTM

PR #296, opened 2021-04-16
Cast vote UI

Warhedgehog:
The UI seems to be working with static data for ballot choices and election name. For now does nothing on pressing the cast vote button.

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/fragments/CastVoteFragment.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/fragments/CastVoteFragment.java
@@ -38,7 +38,7 @@ public class CastVoteFragment extends Fragment implements AdapterView.OnItemClic
     private int numberOfChoices;
     String [] ballotOptions;
     Set<String> selectedOptions;
-    String uniqueSelectedOption;
+    int selectedOption;
     private TextView laoNameText;
     private TextView voteInText;
     private TextView electionNameText;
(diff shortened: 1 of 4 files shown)


fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/fragments/CastVoteFragment.java
```diff
@@ -0,0 +1,178 @@
...
+
+       // ArrayAdapter<String> ballotAdapter = new ArrayAdapter<String>(getActivity(), android.R.layout.simple_list_item_1, ballotOptions);
+        lvBallots.setAdapter(ballotAdapter);
+
+        lvBallots.setOnItemClickListener(this);
+        voteButton.setOnClickListener(buttonListener);
+        return mElectionDisplayFragBinding.getRoot();
+    }
+
+    private void setBallotOptions(){
```
Walfar, line 125: To use this function, you could use the method getBallotOptions() from the election associated. Now, to know which election this fragment is associated with, I guess you would need to retrieve the id corresponding from the ViewModel.

fe2-android/app/src/main/res/values/strings.xml
```diff
@@ -106,5 +106,8 @@
     <!-- Properties -->
     <string name="test_organization_name">Test Organization Name</string>
     <string name="edit_lao_name">Edit Lao Name</string>
+    <!-- TODO: Remove or change this placeholder text -->
+    <string name="hello_blank_fragment">Hello blank fragment</string>
```
Walfar, line 110: What is this for ?

Walfar (commented):
Looks all good to me ! I'm eager to see how we will merge that with the rest of our code and make it work per election. 
I would however suggest to resolve some basic code smells before merging with the master, e.g unused imports, duplicated code etc...

PR #301, opened 2021-04-22
witness connects to organizer

lpatel99:
The witness now connects to the organizer at ws://host.port/org/witness/, where host and port are entered on the command line (or have default values localhost and 9000). Logic for organizer handling messages from witness and witness handling messages is not implemented.


be1-go/cli/organizer/organizer.go
```diff
@@ -44,36 +44,51 @@ func OrganizerServe(context *cli.Context) error {
 	done := make(chan struct{})
 	go h.Start(done)
 
-	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
-		organizerServeWs(h, w, r)
+	go orgCreateSocket(hub.WitnessSocketType, h, port)
+	orgCreateSocket(hub.ClientSocketType, h, port)
+
+	done <- struct{}{}
```
Tesa27, line 50: I might be missing something, but why is this needed? And why does it matter if it's initialized before or after the socket creation?

be1-go/cli/witness/witness.go
```diff
@@ -39,19 +42,32 @@ func WitnessServe(context *cli.Context) error {
...
-	organizerSocket.ReadPump()
+	go organizerSocket.WritePump()
+	go organizerSocket.ReadPump()
+
+	done := make(chan struct{})
+	h.Start(done)
+
+	//TODO: connect to clients and other witnesses
+
+	done <- struct{}{}
```
Tesa27, line 60: Same here, there must a good reason I just don't see it.

Tesa27 (commented):
The rest looks fine to me!

be1-go/cli/organizer/organizer.go
lpatel99: This is because "done <- struct{}{}" signals the end of the program and tells the organizer to stop serving messages (check organizerHub.start).

be1-go/cli/witness/witness.go
lpatel99: Same as above for witness.

be1-go/cli/organizer/organizer.go
Tesa27: Ohh, yes of course wasn't aware of that, thanks!

flmichel (approved):
Looks good to me! I think you can do a `go fmt .` to add a new line at the end of the files. And it might be a good idea to update `README.md` with the new features.

flmichel (dismissed):

be1-go/cli/pop.go
```diff
@@ -4,42 +4,89 @@ import (
...
-			{
-				Name:  "organizer",
-				Usage: "manage the organizer",
-				Flags: []cli.Flag{
-					&cli.StringFlag{
-						Name:    "public-key",
-						Aliases: []string{"pk"},
-						Usage:   "base64 encoded organizer's public key",
+	var app *cli.App
+	switch os.Args[1] {
```
gnarula, line 14: Why not have `witness` as another command in the Commands[] slice?

be1-go/cli/pop.go
```diff
@@ -4,42 +4,89 @@ import (
...
+							Usage:   "base64 encoded witness's public key",
+						},
+					},
+					Subcommands: []*cli.Command{
+						{
+							Name:  "serve",
+							Usage: "start the organizer server",
+							Flags: []cli.Flag{
+								&cli.StringFlag{
+									Name:    "org-address",
```
gnarula, line 69:
```suggestion
									Name:    "organizer-address",
```

be1-go/cli/organizer/organizer.go
```diff
@@ -20,7 +20,7 @@ var upgrader = websocket.Upgrader{
 }
 
 // Serve parses the CLI arguments and spawns a hub and a websocket server.
-func Serve(context *cli.Context) error {
+func OrganizerServe(context *cli.Context) error {
```
gnarula, line 23: I believe we can do without the name change since witness and organizer are in separate packages

be1-go/cli/organizer/organizer.go
```diff
@@ -44,36 +44,51 @@ func Serve(context *cli.Context) error {
 	done := make(chan struct{})
 	go h.Start(done)
 
-	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
-		serveWs(h, w, r)
+	go orgCreateAndServeWs(hub.WitnessSocketType, h, port)
```
gnarula, line 47: I'd suggest having two different ports - one for communication with clients and another for communication with witnesses. This would allow easier debugging of traffic by just monitoring one of the ports than filtering based on the path

be1-go/cli/witness/witness.go
```diff
@@ -0,0 +1,72 @@
...
+
+	return nil
+}
+
+func witConnectToOrganizer(orgIpAddr string, port int) (*websocket.Conn, error) {
+	var addr = flag.String("addr", fmt.Sprintf("%s:%d", orgIpAddr, port), "http service address")
+	u := url.URL{Scheme: "ws", Host: *addr, Path: "/org/witness/"}
```
gnarula, line 66: https://golang.org/pkg/net/url/#Parse might be better here

gnarula (dismissed):
Looks good! I've added a few comments about the CLI and the endpoint at which you listen for requests.

be1-go/hub/witness.go
```diff
@@ -0,0 +1,77 @@
...
+type witnessHub struct {
+	messageChan chan IncomingMessage
+
+	sync.RWMutex
+	channelByID map[string]Channel
+
+	public kyber.Point
+}
+
+// NewOrganizerHub returns a Organizer Hub.
```
cba33, line 19: Comment should be `NewWitnessHub`

be1-go/hub/witness.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+func (w *witnessHub) RemoveClientSocket(client *ClientSocket) {
+	//TODO
+}
+
+func (w *witnessHub) Recv(msg IncomingMessage) {
+	log.Printf("witnessHub::Recv")
+	w.messageChan <- msg
+}
```
cba33, line 35: Looks good, but it's unclear to me who and where would call this function?

cba33 (changes requested):
Looks good! Just a few comments, please take a look. Thanks!

be1-go/cli/organizer/organizer.go
lpatel99: Yes good idea, I've changed that! Indeed that is the intention.

be1-go/README.md
```diff
@@ -12,13 +12,15 @@ NAME:
...
+   pop witness command [command options] [arguments...]
 
 COMMANDS:
-   serve    start the organizer server
+   serve    start the organizer or witness server
    help, h  Shows a list of commands or help for one command
 
 OPTIONS:
    --public-key value, --pk value  base64 encoded organizer's public key
+   --org-address, --org            organizer's IP address for witness to connect to organizer
```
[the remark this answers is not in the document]
lpatel99: Yes I need to add "value", I will do that.
What do you mean by bootstrap addresses of other witnesses?
Yes will modify the CLI to add the choice of port for witness and add the relevant documentation.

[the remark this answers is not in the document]
cba33: _What do you mean by bootstrap addresses of other witnesses?_ This is a suggest for a future iteration of the code, when the witness also connect to other witnesses. To do that, the witness receives an initial set of addresses of other witnesses, which is what I meant with these bootstrap addresses. The witness connect to these initial witnesses, and then through communication potentially learns of new witnesses. 
Because this is not relevant for this PR, no change is required. Just perhaps keep it in mind for the future.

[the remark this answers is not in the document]
lpatel99: Yes understood!

be1-go/hub/witness.go
lpatel99: I believe it is called by any socket (witness, organizer or client) to which the witness is connected in the ReadPump() function.

flmichel (approved):

PR #292, opened 2021-04-12
Added election setup fragment, organizer can now create an election

Walfar:
I added a Fragment for setting up an election (only 1 question per election available for now). When submitting the setup, a message is sent to the backend. When receiving the result, the fragment will bring the user back to the main screen of the corresponding LAO. 
Edit: This branch was first merged with Tesa's branch for the Go backend, to make the election setup work. To avoid any confusion, I deleted those changes by retrieving the be1-go folder from the master branch.

PR #304, opened 2021-04-24
Roll call organizer's UI

Kar1213:
I created this new PR after rebasing the old one with master (PR #293). This one includes new changes.

These are the comments from the old PR:

The following changes are mostly about the UI for managing roll calls. An organizer can now handle roll calls but the UI for attendees to join roll calls is still not implemented.

+ EventState.java: enum used in roll call class to keep track of current state
+ RollCall.java: added state attribute
+ OpenRollCall.java: distinguish between OPEN and REOPEN action
+ Changes made to adapt to Gaurav’s PR #287: CreateRollCall, OpenRollCall, CloseRollCall, Injection, JsonCreateRollCallSerializer, JsonOpenRollCallSerializer, JsonCloseRollCallSerializer and LAORepository
+ layout\_event.xml: added fields for roll calls + added buttons depending on isOrganizer and roll call state
+ EventExpandableListViewAdapter.java: 
	- putEventsInMap method: fixed bug where eventsMap attribute was reset which prevented added events to be saved
	- getChildView method: setting up the display of roll call in the list through binding
+ RollCallEventCreationFragment.java:
	- based on the structure of the other fragments, I restructured the class by moving some code around
	- added/modified click listeners for roll call creation, opening and closing
	- added observer for roll call creation for the "scheduled" case to open LAO detail (see LaoDetailViewModel just below)
+ LaoDetailViewModel.java:
	- createNewRollCall method: creates a CreateRollCall instance from method paramaters and sends a publish message, for the "scheduled" case (parameter open set to false) go to LAO detail fragment, method returns roll call id to be used for opening (next method below)
	- openConnectRollCall method: creates a OpenRollCall instance based on roll call id, sends publish message and on response saves roll call id in mCurrentRollCallId to be used for closing, camera/permission is open
	- attendees are added in onQRCodeDetected method
	- closeRollCall method: similarly it closes the currently open roll call, attendees list is given to CloseRollCall instance and then cleared
	- opening the camera: inspired by what was previously done in HomeViewModel.java
+ QRCodeScanningFragment.java and fragment\_qrcode.xml:
	- exisiting code was for scanning LAO QR code, modified it such as it uses the same fragment and layout to also handle scanning for roll calls
	- added code to handle attendees counter and roll call closing when scanning
LaoDetailFragment.java: added setupEventListUpdates method to update list of events in fragment
+ LAORepository.java: 
	- handleCreateRollCall, handleOpenRollCall and handleCloseRollCall modified to adapt to new roll call fields and protocol
	- sendPublish method: to be able to test roll calls on the app one can uncomment the code that handles roll calls without waiting for a broadcast message from backend (as a temporary solution)

These are the new changes:
	
+ EventExpandableListAdapter.java: the fix with the eventsMap I suggested initially was wrong, so I had to fix the fix

+ LAORepository.java bug fix: handleCloseRollCall returned true on success and should return false

+ RollCallEventCreationFragment: 
     - cleanup by removing comments
     - added call to addTextChangedListener for the roll call title, it fixes a little bug when editing title (inspired by PR #292 but the change was removed for some reason)

+ layout_event.xml and EventExpandableListAdapter.java: code cleanup to be adaptable to general events and moving strings to strings.xml
+ LaoDetailViewFragment.java: 
changed createNewRollCall to fix a bug to open roll call that appeared when tested with backend

+ changes made to add pop up warning messages when scanning: in QRCodeScanningFragment.java and onQRCodeDetected method in LaoDetailViewFragment.java

+ OpenRollCall.java: normally we should use the REOPEN action but OPEN is used for now because backend still doesn't recognize this action


fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/LaoDetailActivity.java
```diff
@@ -107,85 +114,102 @@ public void handleNewEvent(EventType eventType) {
...
+                    getSupportFragmentManager().findFragmentById(R.id.fragment_organizer_poll);
     if (pollCreationFragment == null) {
       pollCreationFragment = PollEventCreationFragment.newInstance();
       ActivityUtils.replaceFragmentInActivity(
-          getSupportFragmentManager(), pollCreationFragment, R.id.fragment_container_lao_detail);
+              getSupportFragmentManager(), pollCreationFragment, R.id.fragment_container_lao_detail);
+    }
+  }
+  private void setupScanFragmentRollCall() {
+    /*QRCodeScanningFragment scanningFragment =
```
celinecamacho, line 187: This should be deleted

fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/LaoDetailActivity.java
```diff
@@ -107,85 +114,102 @@ public void handleNewEvent(EventType eventType) {
...
+      scanningFragment =
+              QRCodeScanningFragment.newInstance(
+                      Injection.provideCameraSource(context, qrCodeDetector, width, height),
+                      qrCodeDetector);
+      ActivityUtils.replaceFragmentInActivity(
+              getSupportFragmentManager(), scanningFragment, R.id.fragment_container_lao_detail);
+    }
+  }
+  private void setupCameraPermissionFragmentRollCall() {
+    CameraPermissionFragment cameraPermissionFragment =
```
Walfar, line 206: Why is it necessary to have a fragment for the cam permission ? Can't we simply have a pop up asking for permissions in the QRcodeScanningFragment ?

fe2-android/app/src/main/java/com/github/dedis/student20_pop/model/network/method/message/data/rollcall/CreateRollCall.java
```diff
@@ -127,8 +127,8 @@ public int hashCode() {
         getId(),
         getName(),
         getCreation(),
-        getStartTime(),
-        getStartType(),
+        getProposedStart(),
```
Walfar, line 130: If I understand correctly, you removed the StartType enum, to avoid using NOW and SCHEDULED, and directly let the organizer set a start time which would categorize the roll call either in the present section or future section, depending on the time, right ?

fe2-android/app/src/main/java/com/github/dedis/student20_pop/utility/json/JsonCloseRollCallSerializer.java
```diff
@@ -0,0 +1,33 @@
+package com.github.dedis.student20_pop.utility.json;
```
Walfar, line 1: I don't think those serializers are needed. I might be mistaking, but from my understanding, as your data objects only have json serializable attributes, you can directly serialize them using gson, and there is no need to pass via a custom serializer.

fe2-android/app/src/main/java/com/github/dedis/student20_pop/qrcode/QRCodeScanningFragment.java
```diff
@@ -20,89 +22,114 @@
...
+          @NonNull LayoutInflater inflater,
+          @Nullable ViewGroup container,
+          @Nullable Bundle savedInstanceState) {
     mQrCodeFragBinding = FragmentQrcodeBinding.inflate(inflater, container, false);
-
     FragmentActivity activity = getActivity();
     if (activity instanceof HomeActivity) {
       mQRCodeScanningViewModel = HomeActivity.obtainViewModel(activity);
-    } else {
+      mQrCodeFragBinding.addAttendeeTotalText.setVisibility(View.GONE);
```
Walfar, line 56: When would this case happen ?

fe2-android/app/src/main/res/layout/fragment_add_attendee.xml
```diff
@@ -6,14 +6,14 @@
...
-            android:id="@+id/add_attendee_qr_code_fragment"
-            android:layout_width="wrap_content"
-            android:layout_height="wrap_content"
-            android:orientation="vertical"
-            app:layout_constraintBottom_toTopOf="@id/add_attendee_confirm_cancel"
-            app:layout_constraintLeft_toLeftOf="parent"
-            app:layout_constraintRight_toRightOf="parent"
-            app:layout_constraintTop_toTopOf="parent"></LinearLayout>
+        android:id="@+id/add_attendee_qr_code_fragment"
+        android:layout_width="3dp"
```
celinecamacho, line 10: You need to add all the dimensions from the layout files on the values/dimens.xml file

fe2-android/app/src/main/java/com/github/dedis/student20_pop/qrcode/QRCodeScanningFragment.java
```diff
@@ -20,89 +22,114 @@
...
+      mQrCodeFragBinding.addAttendeeNumberText.setVisibility(View.GONE);
+      mQrCodeFragBinding.addAttendeeConfirm.setVisibility(View.GONE);
```
celinecamacho, line 58: You can set the visibility to gone on the layout file and set it to visible only when you are scanning attendees

celinecamacho (changes requested):
Great job! The app is working well, but there are still some changes that I think you should make. There are some redundant methods for the roll call event creation and the `RollCallEventCreationFragment` is completely different, it seems you discarded all changes from master while merging. Maybe take a look and see if you can preserve the logic and methods from master

fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/LaoDetailActivity.java
Kar1213: This is what was done in the HomeActivity and to stay consistent I did something similar.

Kar1213: This is what was done in the HomeActivity and to stay consistent I did something similar.

fe2-android/app/src/main/java/com/github/dedis/student20_pop/model/network/method/message/data/rollcall/CreateRollCall.java
Kar1213: Exactly

fe2-android/app/src/main/java/com/github/dedis/student20_pop/utility/json/JsonCloseRollCallSerializer.java
Kar1213: You are right it was mainly about converting fields such as "openedAt" to "opened_at" but I found a way using gson annotations and I was also worried about lists. I removed those classes it seems to work fine as well.

fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/adapters/EventExpandableListViewAdapter.java
```diff
@@ -232,25 +236,63 @@ public void addEvent() {
...
 
-    binding.setEvent(event);
+    binding.openButton.setVisibility(View.GONE);
+    binding.reopenButton.setVisibility(View.GONE);
+    binding.scheduledButton.setVisibility(View.GONE);
+    binding.enterButton.setVisibility(View.GONE);
+    binding.closedButton.setVisibility(View.GONE);
+
+    Event event = (Event)getChild(groupPosition, childPosition);
+    if(event instanceof RollCall){
```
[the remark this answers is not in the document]
Kar1213: Defining a new layout would not simplify the code in this class because the visibility of the buttons depend on other factors than the type of event. The idea behind was to use the current layout for all the events because the same information and buttons are most likely to appear in several types of events. The layout would contain all possible button names and we set to visible only the one that corresponds to our case. Do you think this would be a good solution or a separate layout is still better?

fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/fragments/event/creation/RollCallEventCreationFragment.java
```diff
@@ -9,146 +9,108 @@
...
-    return mFragBinding.getRoot();
-  }
-
-  @Override
-  public void onActivityCreated(@Nullable Bundle savedInstanceState) {
-    super.onActivityCreated(savedInstanceState);
-
-    setupConfirmButton();
-    setupOpenButton();
-    setupCancelButton();
```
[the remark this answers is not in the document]
Kar1213: I see what you mean, I will add these. Regarding your comment above about trying to keep as much as possible the master version, I intentionally moved the code around to have a fragment structure consistent with the others because I found it very different. Is adding back these methods enough?

Warhedgehog (approved):
Nice work ! I have nothing to add on top of previous comments

fe2-android/app/src/main/java/com/github/dedis/student20_pop/qrcode/QRCodeScanningFragment.java
Walfar: thanks !

fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/adapters/EventExpandableListViewAdapter.java
```diff
@@ -264,10 +306,12 @@ public View getChildView(
    */
   private void putEventsInMap(List<Event> events) {
     Collections.sort(events);
-
+    this.eventsMap.get(PAST).clear();
```
maximelaval, line 309: I don't understand why you clear the previous events in PAST,PRESENT,FUTURE ? Didn't it worked without doing that ?

maximelaval (commented):
Great job overall ! I made some comments on ExpandableListViewAdapter

fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/fragments/event/creation/RollCallEventCreationFragment.java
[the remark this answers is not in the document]
celinecamacho: Moving the code around is ok, but some methods and events from the view model are completely different

fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/adapters/EventExpandableListViewAdapter.java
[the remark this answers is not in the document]
celinecamacho: I understand the issue. I think creating a method for the roll call changes can help with the readability of the method

fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/LaoDetailActivity.java
celinecamacho: There are two fragments, one for the camera permissions and the other for the QR code scanning, this way we have two corresponding View Models and we keep the logic separate

fe2-android/app/src/main/java/com/github/dedis/student20_pop/utility/json/JsonCloseRollCallSerializer.java
celinecamacho: If the classes are not used and the serialization/deserialization works fine without them, then you can delete them

celinecamacho (commented):
The .DS_Store files should be in the gitignore of the project, it would be good to add them

celinecamacho (changes requested):
Great job on the changes :) I still have some comments, mainly on the LAO detail view model. I also suggest some small changes on `AddAttendeeFragment`, `RollCallEventCreationFragment`, `LAORepository` and `OpenRollCall`

fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/LaoDetailViewModel.java
```diff
@@ -246,9 +261,104 @@ public String createNewRollCall(String title, String description, long start, lo
...
+                      .sendPublish(channel, msg)
+                      .subscribeOn(Schedulers.io())
+                      .observeOn(AndroidSchedulers.mainThread())
+                      .timeout(5, TimeUnit.SECONDS)
+                      .subscribe(
+                              answer -> {
+                                if (answer instanceof Result) {
+                                  Log.d(TAG, "closed the roll call");
+                                  mCurrentRollCallId = "";
+                                  attendees.clear();
```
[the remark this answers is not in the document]
Kar1213: The UI doesn't really depend on those values, they are used only to keep track of the state of the currently open roll call to be able to close it, where we need to know the roll call id and the list of attendees.

fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/fragments/event/creation/RollCallEventCreationFragment.java
```diff
@@ -3,152 +3,118 @@
...
-  }
-
-  private void setupConfirmButton() {
-    confirmButton.setOnClickListener(
-        v -> mLaoDetailViewModel.newLaoEventCreation(EventType.ROLL_CALL));
-  }
-
-  private void setupOpenButton() {
-    openButton.setOnClickListener(v -> mLaoDetailViewModel.openNewRollCall(true));
-  }
```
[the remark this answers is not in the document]
Kar1213: I am made the changes so that I use them now.

.gitignore
```diff
@@ -24,4 +24,6 @@
 *.db
 *.lock
 
-be1-go/pop
+
+.DS_Store
+**/.DS_Store
```
celinecamacho, line 29: The be1-go/cli/.DS_Store file is still present

celinecamacho (approved):

PR #305, opened 2021-04-26
Work be2 cleanup nicolas parser

RaulinN:
- bc2fe52 : Finished graph validators (rpc-layer + method layer + message layer)
- 9cda812 : Implemented parsers/serializers for all layers. Heavily inspired from Céline's work
- 28109ed : Included 9cda812's work into the DSL graph
- ee83bb7 : Applied Pierluca's review comments from #275


be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/MessageDecoder.scala
```diff
@@ -2,15 +2,31 @@ package ch.epfl.pop.pubsub.graph
...
-  // Message or TextMessage
   val jsonRpcParser: Flow[Either[JsonString, PipelineError], GraphMessage, NotUsed] = Flow[Either[JsonString, PipelineError]].map {
-    case Left(jsonString) =>
-      Left(JsonRpcRequest("2.0", MethodType.INVALID, ???, Some(0)))
-    // case _ => _ // TODO check if type ok. Else prob Right(pipelineError)
+    case Left(jsonString) => Try(jsonString.parseJson.asJsObject) match {
+      case Success(obj) =>
+        val fields: Set[String] = obj.fields.keySet
+
+        if (fields.contains("method")) { // FIXME check that the error (if any) is correctly propagated
```
pierluca, line 18: I don't understand the FIXME here

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/MessageDecoder.scala
```diff
@@ -2,15 +2,31 @@ package ch.epfl.pop.pubsub.graph
...
+          Left(obj.convertTo[JsonRpcRequest])
+        } else {
+          Left(obj.convertTo[JsonRpcResponse])
```
pierluca, line 21: I'm a bit confused as to the meaning of Left() and Right(), are we indicating that Left() is the left option in the `Either[JsonString, PipelineError]` ? Or should this be Right? 
Apologies for my Scala/Akka ignornace :)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/Validator.scala
```diff
@@ -2,28 +2,82 @@ package ch.epfl.pop.pubsub.graph
...
+import ch.epfl.pop.pubsub.graph.validators.LaoValidator._
+import ch.epfl.pop.pubsub.graph.validators.MeetingValidator._
+import ch.epfl.pop.pubsub.graph.validators.RollCallValidator._
+import ch.epfl.pop.pubsub.graph.validators.WitnessValidator._
 
 object Validator {
 
+  private val VALIDATOR_ERROR: PipelineError = PipelineError(
+    ErrorCodes.INVALID_ACTION.id,
+    "Internal server fault: Validator was given a message it could not recognize"
```
pierluca, line 23: `Internal server fault` might not be the best description here, how about `Unsupported action` ?

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/Validator.scala
```diff
@@ -2,28 +2,82 @@ package ch.epfl.pop.pubsub.graph
...
+    case Left(jsonRpcRequest: JsonRpcRequest) => jsonRpcRequest.getParams match {
+      case _: Broadcast => validateBroadcast(jsonRpcRequest)
+      case _: Catchup => validateCatchup(jsonRpcRequest)
+      case _: Publish => validatePublish(jsonRpcRequest)
+      case _: Subscribe => validateSubscribe(jsonRpcRequest)
+      case _: Unsubscribe => validateUnsubscribe(jsonRpcRequest)
+      case _ => Right(VALIDATOR_ERROR)
+    }
+    case Left(_) => Right(PipelineError(
+      ErrorCodes.INVALID_ACTION.id,
```
pierluca, line 48: probably an internal error rather than an invalid action here?

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/MethodValidator.scala
```diff
@@ -0,0 +1,23 @@
+package ch.epfl.pop.pubsub.graph.validators
+
+import ch.epfl.pop.model.network.JsonRpcRequest
+import ch.epfl.pop.pubsub.graph.{GraphMessage, PipelineError}
+
+object MethodValidator extends MethodContentValidator {
```
pierluca, line 6: I would expect `MethodValidator` to validate a method.
Am I correct in understanding here we're validating the parameters?
Could we rename it as ParameterValidator?

pierluca (changes requested):
Seems pretty good overalls!

I'm a bit shocked by the jsonFormat# calls, which really should have better names - but I don't understand where they come from either. I think that might really need a bit of fixing :)

I also don't understand where we validate the message structure: I don't mean the high-level message, but the mid-level object containing the base64-encoded data, signatures, message_id, etc.  Maybe the code is not included in the PR because it exists already?

Those details aside, it seems overalls pretty understandable.

Would it make sense to have @celinecamacho review this, as she's more Scala-fluent than the rest of us at DEDIS are?

be2-scala/src/main/scala/ch/epfl/pop/jsonNew/HighLevelProtocol.scala
```diff
@@ -0,0 +1,84 @@
...
+        case _ => throw new IllegalArgumentException(s"Can't parse json value $json to a Message object")
+      }
+
+    override def write(obj: Message): JsValue =
+      JsObject(
+        PARAM_DATA -> obj.data.toJson,
+        PARAM_SENDER -> obj.sender.toJson,
+        PARAM_SIGNATURE -> obj.signature.toJson,
+        PARAM_MESSAGE_ID -> obj.message_id.toJson,
+        PARAM_WITNESS_SIG -> obj.witness_signatures.toJson // FIXME does this work? Otherwise use "JsArray(obj.witness_signatures.map(_.toJson).toVector))"
```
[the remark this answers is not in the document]
RaulinN: Yes....... wheeeen the code will compile ^^'. Will answer other messages tomorrow, but in short, since the graph doesn't have a global reference to dbActorNew yet, the compilation fails. This is an easy fix but I'm not quite convinced we have the best approach for it yet

pierluca:
Minor merge conflicts here :)

be2-scala/src/main/scala/ch/epfl/pop/jsonNew/HighLevelProtocol.scala
```diff
@@ -0,0 +1,84 @@
...
+      )
+  }
+
+  implicit val broadcastFormat: JsonFormat[Broadcast] = jsonFormat2(Broadcast.apply)
+  implicit val catchupFormat: JsonFormat[Catchup] = jsonFormat1(Catchup.apply)
+  implicit val publishFormat: JsonFormat[Publish] = jsonFormat2(Publish.apply)
+  implicit val subscribeFormat: JsonFormat[Subscribe] = jsonFormat1(Subscribe.apply)
+  implicit val unsubscribeFormat: JsonFormat[Unsubscribe] = jsonFormat1(Unsubscribe.apply)
```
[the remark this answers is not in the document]
gnarula: Caught me by surprise as well. It's documented here https://github.com/spray/spray-json#providing-jsonformats-for-case-classes

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/MessageDecoder.scala
```diff
@@ -2,15 +2,31 @@ package ch.epfl.pop.pubsub.graph
...
+import ch.epfl.pop.model.network.{JsonRpcRequest, JsonRpcResponse}
+import ch.epfl.pop.jsonNew.HighLevelProtocol._
+import spray.json._
+
+import scala.util.{Success, Try}
 
 object MessageDecoder {
 
-  // Message or TextMessage
   val jsonRpcParser: Flow[Either[JsonString, PipelineError], GraphMessage, NotUsed] = Flow[Either[JsonString, PipelineError]].map {
```
gnarula, line 13: IIRC, `Either[PipelineError, JsonString]` would be the convention

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/LaoHandler.scala
```diff
@@ -35,7 +35,7 @@ case object LaoHandler extends MessageHandler {
         // Publish on the LAO main channel
         dbActor.ask(ref => DbActorNew.Write(channel, rpcMessage.getParamsMessage.get, ref)) match {
           case Success(_) => Left(rpcMessage)
-          case _ => Right(PipelineError(-10, "")) // FIXME add DbActor "answers" with error description if failed
+          case _ => Right(PipelineError(-10, "")) // FIXME add DbActor "answers" with error description if failed. Should be a SERVER_ERROR
```
gnarula, line 38: `ErrorCodes.SERVER_ERROR` here please. Very likely the clients won't give any feedback on seeing an unknown error code.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/MessageDecoder.scala
RaulinN: This is actually for the if-else below. AFAIK the `.convertTo` could fail ==> create an error which would be piped into Lefft(_)

```scala
        if (fields.contains("method")) { // FIXME check that the error (if any) is correctly propagated
          Left(obj.convertTo[JsonRpcRequest])
        } else {
          Left(obj.convertTo[JsonRpcResponse])
        }
```

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/MessageDecoder.scala
RaulinN: Everything (except the parser input) that goes through the graph is a `GraphMessage` that I defined as `Either[JsonRpcMessage, PipelineError]` (which has type parameters inverted as Gaurav mentionned... woops! ^^'). So everything that goes out of the parser is an instance of this weird "Either" object. If I want to output (success) a `JsonRpcMessage`, then I would use `Left(message)` to create this Either instance. If a problem occured, then I would use `Right(PipelineError(...))`

So in short, yes, Left creates an instance of `Either[V, W]` which holds an object of type `V`

PS: warning `Either[JsonString, PipelineError]` is the input, not the output!

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/Validator.scala
RaulinN: Truuuue because a response message ends up here, it's a bug in the server, not the client's fault!

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/MethodValidator.scala
RaulinN: LGTM

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/MessageDecoder.scala
RaulinN: Correct! I can change this one now since it's only used in this PR :)

RaulinN: EDIT : nvm :/ Will have to wait ffor the merge as well

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/LaoHandler.scala
RaulinN: Yep! Also added to my todo list :) Same as the last PR. I unfortunately need to wait for everything to be in a single PR to resolve that

gnarula (approved):

gnarula:
@RaulinN can you resolve the conflicts please

RaulinN:
Yes! Doing it now :)

PR #306, opened 2021-04-26
Wallet store and reducer and beginning of wallet UI

CaiMusso:
Content proposed in this Pull Request
----------------------------
On this branch progress for the **typescript web wallet** has beed implemented:
+ The wallet UI enables the user to ask for a **new wallet** or initialise their **existing wallet.**
+ If asked for a new wallet the wallet object creates a **12-word mnemonic** and displays it to the user.
+ If an existing wallet mnemonic is inserted the wallet object checks the **validity** of the mnemonic and **converts** it to the numerical seed.
+ At wallet initialization the **WalletCryptographyHandler** is initialized and the **RSA key is stored in indexedDB.**
+ The seed is **encrypted and stored in the redux state** (in WalletStore slice), through the WalletReducer.

Important files in this PR
------------------------

### UI components

+ **WalletHome.tsx**
+ **WalletSetSeed.tsx**
+ **WalletShowSeed.tsx**
+ **WalletSyncedSeed.tsx**

### Logic files

+ **HDWallet.ts -** the wallet **object**. 
+ **WalletStore.ts -** the wallet **store**.
+ **WalletReducer.ts -** the wallet **reducer**.

### Other small changes
+ made window.crypto (**Subtle Crypto**) default library in cryptography handler, no more need to specify it when creating handler.
+ integrated wallet reducer in root reducer
+ integrated wallet UI in main navigation

New dependencies
-------------------------------------
+ [**bip39**](https://www.npmjs.com/package/bip39) - mnemonic and seed creation/verification

Quick demo
-------------------------------
https://user-images.githubusercontent.com/47753346/116389849-b8084600-a81d-11eb-94cf-1d3547afff30.mov


pierluca:
I'll let @Adalsteinnjons review first, but how come the package-lock changed so much? 🤔

CaiMusso:
> I'll let @Adalsteinnjons review first, but how come the package-lock changed so much? 🤔

Good point, thanks @pierluca. I think I messed up something by accident. It is fixed now.

fe1-web/model/objects/HDWallet.ts
```diff
@@ -0,0 +1,127 @@
...
+    if (seedIsValid) {
+      this.cryptoManager = new WalletCryptographyHandler();
+      const seed: Uint8Array = await bip39.mnemonicToSeed(mnemonic);
+      await this.cryptoManager.initWalletStorage();
+      const encryptedSeed = await this
+        .encryptSeedToStoreInState(seed);
+      this.encryptedSeed = encryptedSeed;
+      WalletStore.store(encryptedSeed);
+    }
+    return seedIsValid;
```
pierluca, line 57: When you have a clear path which validates the inputs, it's often cleaner to immediately exit the function on a validation fail. Example:

```suggestion
    const seedIsValid: boolean = bip39.validateMnemonic(mnemonic);
    if (!seedIsValid) {
      return false;
    }
    
    this.cryptoManager = new WalletCryptographyHandler();
    const seed: Uint8Array = await bip39.mnemonicToSeed(mnemonic);
    await this.cryptoManager.initWalletStorage();
    const encryptedSeed = await this
      .encryptSeedToStoreInState(seed);
    this.encryptedSeed = encryptedSeed;
    WalletStore.store(encryptedSeed);
    return true;
```

fe1-web/model/objects/HDWallet.ts
```diff
@@ -0,0 +1,127 @@
...
+   * this method initializes the wallet:
+   * 1. creating the crypto-manager (RSA key in IndexedDB).
+   * 2. encrypting the seed.
+   * 3. storing it in the application state through reducer.
+   * @param mnemonic the 12-word representation of the wallet seed
+   * @return true if the given mnemonic is valid, false otherwise
+   */
+  public async initialize(mnemonic: string): Promise<boolean> {
+    if (mnemonic === undefined) {
+      throw new Error('Error while initializing the wallet seed: seed is undefined');
```
pierluca, line 45: why throw an error here when you could just return false ? 
do you expect to process it differently?
as far as I can think of, you could just log that error and return false.

fe1-web/model/objects/HDWallet.ts
```diff
@@ -0,0 +1,127 @@
...
+      WalletStore.store(encryptedSeed);
+    }
+    return seedIsValid;
+  }
+
+  /**
+   * retrieves and decrypts the wallet state (seed) from the application store.
+   * @return the plaintext wallet seed as a Uint8Array
+   */
+  // eslint-disable-next-line class-methods-use-this
```
pierluca, line 64: Don't disable eslint whenever you have a problem, there's generally a very good reason for those remarks.
Google the error and the solution will be immediately obvious:
https://eslint.org/docs/rules/class-methods-use-this

Also, I suspect that even if you remove the eslint-disable, there won't be any linting issue here.

fe1-web/store/stores/WalletStore.ts
```diff
@@ -0,0 +1,32 @@
...
+    dispatch(setWalletState(walletSeed));
+  }
+
+  /**
+   * returns the wallet state: the encrypted wallet seed or throws
+   * an error if the state was never initialized.
+   */
+  export async function get(): Promise<ArrayBuffer> {
+    console.log('PoP Application STATE');
+    console.log(getStore().getState());
```
pierluca, line 23: you might want to remove those logs if you're ready to merge :)

fe1-web/store/reducers/WalletReducer.ts
```diff
@@ -0,0 +1,47 @@
...
+
+/* the store slice in charge of the wallet state */
+const walletSlice = createSlice({
+  name: walletReducerPath,
+  initialState,
+  reducers: {
+    /* set global wallet state */
+    setWalletState: (state, action: PayloadAction<ArrayBuffer>) => {
+      if (!action.payload) {
+        console.log('KeyPair storage was set to: null');
```
pierluca, line 29:
```suggestion
        console.log('Wallet storage was set to: null');
```

fe1-web/store/reducers/WalletReducer.ts
```diff
@@ -0,0 +1,47 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+
+/**
+ * This file represents the reducer for the wallet state. Its
+ * job is to handle actions between the application and the
+ * storage (e.g. storing the wallet seed)
```
pierluca, line 6: I'm not sure that "Its job is to handle actions between the application and the storage" is very explicit 😅 
You might just want to skip ahead to "Its job is to store the wallet seed".

fe1-web/store/reducers/WalletReducer.ts
```diff
@@ -0,0 +1,47 @@
...
+  reducers: {
+    /* set global wallet state */
+    setWalletState: (state, action: PayloadAction<ArrayBuffer>) => {
+      if (!action.payload) {
+        console.log('KeyPair storage was set to: null');
+        state.walletState = undefined;
+      }
+
+      state.walletState = action.payload;
+      console.log(`Wallet storage was updated with new state: ${state.walletState}`);
```
pierluca, line 34: I'm wondering, `walletState` is not really human-readable, does it make sense to log it?
We could also just say `Wallet storage was update with new state` ... up to you :)

fe1-web/model/objects/HDWallet.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  /**
+   * creates a new wallet object from the state (encryptedSeed)
+   * @param encryptedSeed wallet's encrypted seed
+   * @return a new wallet object initialized from the given encrypted seed
+   */
+  public async fromState(encryptedSeed: ArrayBuffer): Promise<HDWallet> {
+    await this.cryptoManager.initWalletStorage();
+    const walletFromState = new HDWallet();
+    walletFromState.setEncryptedSeed(encryptedSeed);
```
pierluca, line 107: why not pass the `encryptedSeed` as a parameter (optionally) to the constructor?
something like `encryptedSeed: ArrayBuffer | undefined`

fe1-web/model/objects/WalletCryptographyHandler.ts
```diff
@@ -33,11 +32,19 @@ export class WalletCryptographyHandler {
...
+  /**
+   * the crypto library is passed to constructor, this is necessary in order to test the
+   * cryptography handler without the crypto.subtle library provided by the window object.
+   * In jest context provide a MOCK crypto.subtle library, otherwise window.crypto is
+   * selected by default.
+   * @param cryptography MOCK crypto.subtle library or nothing (window.crypto default)
+   */
+  constructor(cryptography?: Crypto) {
+    if (!cryptography) {
+      this.cryptography = window.crypto;
```
pierluca, line 44: This will not work in a react-native context.
If you accept a Crypto interface, whoever calls this can pass on a polyfill for react-native.
On the other hand, given that it's awkward to pass the crypto library around, you might want to have a separate file allowing you to obtain crypto (e.g. `getCrypto(): Crypto`) , which returns window.crypto in the browser, something else in react-native, and which can be mocked during tests.

fe1-web/model/objects/HDWallet.ts
```diff
@@ -0,0 +1,127 @@
...
+    const encryptedSeed = await this.cryptoManager.encrypt(seed);
+    return encryptedSeed;
+  }
+
+  /**
+   * creates a new wallet object from the state (encryptedSeed)
+   * @param encryptedSeed wallet's encrypted seed
+   * @return a new wallet object initialized from the given encrypted seed
+   */
+  public async fromState(encryptedSeed: ArrayBuffer): Promise<HDWallet> {
```
pierluca, line 104: who is going to call this? 
`fromState` in the application have typically been used to restore something from the reducer.
In this case, however, you're calling `WalletStore.store`, thereby overwriting whatever was passed by the reducer. I'm a bit confused here.

fe1-web/parts/Wallet.tsx
```diff
@@ -0,0 +1,185 @@
...
+    justifyContent: 'flex-end',
+  } as ViewStyle,
+});
+
+const wallet: HDWallet = new HDWallet();
+
+const Wallet = () => {
+  /* this is the UI state of the wallet, starts null (namely homepage)
+     the states can be: null | show_seed | set_seed | initialized */
+  const [walletState, setWalletState] = useState(STRINGS.wallet_state_null);
```
Adalsteinnjons, line 32: I think it would be better to use a string Enum for this state

fe1-web/parts/Wallet.tsx
```diff
@@ -0,0 +1,185 @@
...
+  /* used to set the mnemonic seed inserted by the user */
+  const [seed, setSeed] = useState('');
+  /* boolean set to true if the token recover process is finished */
+  const [tokensRecovered, setTokensRecovered] = useState(false);
+
+  /* all the below are temporarily used for demonstration purposes (showing seed in UI) */
+  const [showHideSeedTitle, setShowHideSeedTitle] = useState(STRINGS.show_seed_title);
+  const [visibleSeed, setVisibleSeed] = useState(false);
+  const [recoveredSeed, setRecoveredSeed] = useState('');
+  // eslint-disable-next-line max-len
```
pierluca, line 42: probably don't want to disable eslint :)

fe1-web/parts/Wallet.tsx
```diff
@@ -0,0 +1,185 @@
...
+  const [showHideEncryptedSeedTitle, setShowHideEncryptedSeedTitle] = useState(STRINGS.show_encrypted_seed_title);
```
pierluca, line 43: these variable names are extremely confusing to me.
if you're using this just to change a button title, you can just have two buttons, and you display either based on a boolean value.

fe1-web/parts/Wallet.tsx
```diff
@@ -0,0 +1,185 @@
...
+  /* this is the UI state of the wallet, starts null (namely homepage)
+     the states can be: null | show_seed | set_seed | initialized */
+  const [walletState, setWalletState] = useState(STRINGS.wallet_state_null);
+  /* used to set the mnemonic seed inserted by the user */
+  const [seed, setSeed] = useState('');
+  /* boolean set to true if the token recover process is finished */
+  const [tokensRecovered, setTokensRecovered] = useState(false);
+
+  /* all the below are temporarily used for demonstration purposes (showing seed in UI) */
+  const [showHideSeedTitle, setShowHideSeedTitle] = useState(STRINGS.show_seed_title);
```
pierluca, line 39: if you're using this just to change a button title, you can just have two buttons, and you display either based on a boolean value.

fe1-web/parts/Wallet.tsx
```diff
@@ -0,0 +1,185 @@
...
+  function getStartWalletDisplay() {
+    return (
+      <View style={styleContainer.centered}>
+        <TextBlock bold text={STRINGS.welcome_to_wallet_display} />
+        <TextBlock text={' '} />
+        <TextBlock text={STRINGS.info_to_set_wallet} />
+        <TextBlock text={' '} />
+        <TextBlock text={STRINGS.caution_information_on_seed} />
+        <TextBlock text={' '} />
+        <TextBlock text={' '} />
```
pierluca, line 56: Don't use empty text blocks for spacing. Use styling, that's what it's for :) 
(below as well)

Adalsteinnjons (commented):
Good progress. 
The main functionality seems to be working well which is most important.
I just added some minor comments. 
Also probably good to not have too many 'console.log()'s in there.

One curious question about the store: The encrypted seed is stored there and we can see it in the console when printing, but it doesn't show in redux dev tools. Do you know why?
image

fe1-web/parts/Wallet.tsx
```diff
@@ -0,0 +1,185 @@
...
+    );
+  }
+
+  function getDisplay() {
+    return (
+      <View style={styleContainer.centered}>
+        { walletState === STRINGS.wallet_state_null && getStartWalletDisplay() }
+        { walletState === STRINGS.wallet_state_show_seed && getShowSeedWalletDisplay() }
+        { walletState === STRINGS.wallet_state_set_seed && getInsertSeedWalletDisplay() }
+        { walletState === STRINGS.wallet_state_initialized && getWalletDisplay() }
```
pierluca, line 178: if I understand correctly what is going on, you're rolling your own state machines to display 4 completely different screens within one part. 

It seems to me like the correct approach would be to use React's navigation capabilities in this case.

fe1-web/parts/Wallet.tsx
```diff
@@ -0,0 +1,185 @@
+import React, { useState } from 'react';
```
pierluca, line 1: I've pointed out a few issues, but this file to me seems particularly messy.
It also seems to contain both demo-specific code and actually useful things.
You might want to clean it up:
1. If you need navigation, use React navigation, not your own state machine and conditional display logic.
2. If code is only for demo, maybe split it out in a file that can be easily removed later.
3. If the code is actually useful in the long run, keep it clean.
4. Make sure that variables are properly named, I would recommend reading [Code Complete, Chapter 11, PDF page 296 and following](http://aroma.vn/web/wp-content/uploads/2016/11/code-complete-2nd-edition-v413hav.pdf) in this particular case, and the whole book to any programmer anyways 😄

pierluca (changes requested):
Great progress overalls ! 
I feel the GUI might still need a lot of cleaning, but the rest isn't far off ! Great job.

fe1-web/parts/Wallet.tsx
```diff
@@ -0,0 +1,185 @@
...
+        <TextBlock text={' '} />
+        <WideButtonView
+          title={STRINGS.back_to_wallet_home}
+          onPress={() => setWalletState(STRINGS.wallet_state_null)}
+        />
+      </View>
+    );
+  }
+
+  function showHideSeed() {
```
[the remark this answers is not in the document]
CaiMusso: Yes totally right! This was only for the presentation demo, to show how the seed is retrieved from redux store. I deleted all the useless demonstration buttons now!

fe1-web/parts/Wallet.tsx
```diff
@@ -0,0 +1,185 @@
...
+      <View style={styleContainer.centered}>
+        <TextBlock bold text={STRINGS.show_seed_info} />
+        <TextBlock text={' '} />
+        <TextBlock text={seed} />
+        <TextBlock text={' '} />
+        <WideButtonView
+          title={STRINGS.back_to_wallet_home}
+          onPress={() => setWalletState(STRINGS.wallet_state_null)}
+        />
+      </View>
```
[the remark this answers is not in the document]
CaiMusso: This would be a cool feature! However I noticed in already implemented digital wallets that there should be no way for the user to copy the mnemonic, or more generally: **there should be no way for the user to avoid writing down the mnemonic on a piece of paper**. This really helps wallet access-loss issues given by a user which, in a hurry, decides to copy-paste the mnemonic without writing it down.

fe1-web/model/objects/HDWallet.ts
```diff
@@ -0,0 +1,127 @@
...
+ * The tokens are created and retrieved here.
+ */
+export class HDWallet {
+  private static readonly PREFIX: string = 'm';
+
+  private static readonly PURPOSE: string = '888';
+
+  private static readonly ACCOUNT: string = '0';
+
+  private static readonly SEPARATOR: string = '/';
```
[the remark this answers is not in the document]
CaiMusso: Definitely ! good point

fe1-web/model/objects/HDWallet.ts
CaiMusso: Yes totally cleaner

CaiMusso:
The quick fixes have been solved, all the unnecessary code written for demo purposes has been removed and the DEMO video in this PR has been updated accordingly. The more cumbersome fixes will be addressed soon.

fe1-web/__tests__/model/objects/WalletCryptographyHandler.test.ts
```diff
@@ -1,51 +1,11 @@
 import 'jest-extended';
 import { WalletCryptographyHandler } from 'model/objects/WalletCryptographyHandler';
 import { sign } from 'tweetnacl';
-import STRINGS from 'res/strings';
-import { get } from 'idb-keyval';
+import { WalletCryptoLibrary } from 'res/adhoc-implementations/__mocks__/WalletCryptoLibrary';
 
 /* used to simulate indexedDB database to test store/retrieve functions */
 require('fake-indexeddb/auto');
```
pierluca, line 7: you should be able to do 
```suggestion
import 'fake-indexeddb/auto';
```

fe1-web/res/adhoc-implementations/WalletCryptoLibrary.native.ts
```diff
@@ -0,0 +1,8 @@
+/**
```
pierluca, line 1: I'm a bit surprised of the location.
`res` stands for `resources`, i.e. images, data, assets, etc.
Not a typical location for source code, and `adhoc-implementations` seems an odd name for a folder: technically all implementations are ad hoc :)
You might want to put this simply in a folder like `fe1-web/utils/myfile.ts`

fe1-web/res/adhoc-implementations/WalletCryptoLibrary.native.ts
```diff
@@ -0,0 +1,8 @@
...
+ * TODO: should return the crypto library for native context
+ */
+export class WalletCryptoLibrary {
+  public static getCrypto() : undefined {
+    return undefined;
+  }
+}
```
pierluca, line 8: why the `WalletCryptoLibrary` name?
why a class just for a single static method? 
why return undefined when we know the interface we want to respect?  

```suggestion
export  getCrypto(): Crypto {
    throw new Exception('not yet implemented');
}
```

fe1-web/res/adhoc-implementations/WalletCryptoLibrary.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * @return the SubtleCrypto library on browser for browser context
+ */
+export class WalletCryptoLibrary {
+  public static getCrypto() : Crypto {
+    return window.crypto;
+  }
+}
```
pierluca, line 8: same comments as above.

fe1-web/res/adhoc-implementations/__mocks__/WalletCryptoLibrary.ts
```diff
@@ -0,0 +1,53 @@
...
+  },
+});
+
+/**
+ * returns the MOCK SubtleCrypto library on browser for browser context
+ */
+export class WalletCryptoLibrary {
+  public static getCrypto() : object {
+    return crypto;
+  }
```
pierluca, line 52: same as above

fe1-web/model/objects/WalletCryptographyHandler.ts
```diff
@@ -33,17 +34,17 @@ export class WalletCryptographyHandler {
...
-    if (!cryptography) {
-      this.cryptography = window.crypto;
+  constructor(mockCrypto?: Crypto | object) {
+    if (mockCrypto !== undefined) {
+      // @ts-ignore
+      this.cryptography = mockCrypto;
     } else {
-      this.cryptography = cryptography;
+      this.cryptography = WalletCryptoLibrary.getCrypto();
     }
```
pierluca: Now that you've split out the `getCrypto` function, you don't need to take it as a parameter or store it as `this.cryptography`. 

You can just do `getCrypto()` wherever you need it.

fe1-web/model/objects/HDWallet.ts
```diff
@@ -98,30 +116,48 @@ export class HDWallet {
...
-   * @param encryptedSeed wallet's encrypted seed
+   * Transforms the given encrypted seed (ArrayBuffer) in string,
+   * only primitive types can be stored in redux state
+   * @param encryptedSeed the encrypted seed as an ArrayBuffer
    * @private
    */
-  private setEncryptedSeed(encryptedSeed: ArrayBuffer) {
-    this.encryptedSeed = encryptedSeed;
+  private static serializeEncryptedSeed(encryptedSeed: ArrayBuffer): string {
+    return new Int8Array(encryptedSeed).toString();
```
pierluca, line 146: It seems you're converting rather liberally between Uint8 and Int8 , does this actually work?

fe1-web/__tests__/model/objects/WalletCryptographyHandler.test.ts
```diff
@@ -1,51 +1,11 @@
 import 'jest-extended';
 import { WalletCryptographyHandler } from 'model/objects/WalletCryptographyHandler';
 import { sign } from 'tweetnacl';
-import STRINGS from 'res/strings';
-import { get } from 'idb-keyval';
+import { WalletCryptoLibrary } from 'res/adhoc-implementations/__mocks__/WalletCryptoLibrary';
```
pierluca, line 4: you should just be able to call the jest mocking function, as we had discussed.
no need to pass this as a parameter or reference the mock explicitly, I believe.

fe1-web/model/objects/HDWallet.ts
```diff
@@ -98,30 +116,48 @@ export class HDWallet {
...
   }
 
   /**
    * returns the current wallet object to state (encryptedSeed)
+   * serialized as a String in order to be stored in redux state
    */
-  public toState(): ArrayBuffer {
-    return this.encryptedSeed;
+  public toState(): String {
+    return HDWallet.serializeEncryptedSeed(this.encryptedSeed);
```
pierluca, line 136: why not move the logic of that static function in here?
It's private, one line, and only used here.
You're saving yourself 1 indirection and improving the readability a lot.

fe1-web/model/objects/HDWallet.ts
```diff
@@ -18,20 +18,35 @@ export interface WalletState {
...
   private static readonly PREFIX: string = 'm';
 
   private static readonly PURPOSE: string = '888';
 
   private static readonly ACCOUNT: string = '0';
 
-  private static readonly SEPARATOR: string = '/';
+  private static readonly PATH_SEPARATOR: string = '/';
+
+  private static readonly BUFFER_SEPARATOR: string = ',';
```
pierluca, line 31: I'm not sure I understand why you need this ?

fe1-web/model/objects/HDWallet.ts
```diff
@@ -98,30 +116,48 @@ export class HDWallet {
...
+    return new Int8Array(encryptedSeed).toString();
+  }
+
+  /**
+   * Transforms the serialized encrypted seed stored in state back to the
+   * original ArrayBuffer encrypted seed used by the cryptography handler
+   * @param encryptedSeedEncoded
+   * @private
+   */
+  private static deSerializeEncryptedSeed(encryptedSeedEncoded: string): ArrayBuffer {
```
pierluca, line 155: deserialize is a word, it's a bit counterintuitive to capitalize the S in `deSerialize`

fe1-web/navigation/bars/WalletNavigation.tsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
```
pierluca, line 1: Great job on the navigation and GUI, so much cleaner !!

fe1-web/parts/wallet/WalletSetSeed.tsx
```diff
@@ -0,0 +1,74 @@
...
+          placeholder={STRINGS.type_seed_example}
+          onChangeText={(input: string) => setSeed(input)}
+        />
+        <View style={styles.smallPadding} />
+        <WideButtonView
+          title={STRINGS.setup_wallet}
+          onPress={() => wallet.initialize(seed)
+            .then((seedIsValid) => ((seedIsValid)
+              ? navigation.navigate(STRINGS.navigation_synced_wallet)
+              : navigation.navigate(STRINGS.navigation_home_tab_wallet)))}
```
pierluca, line 54: here the logic is non-trivial, you might want to extract a function to improve readability.

fe1-web/parts/wallet/WalletSyncedSeed.tsx
```diff
@@ -0,0 +1,70 @@
...
+      <View>
+        <TextBlock bold text={STRINGS.wallet_synced_info} />
+        <View style={styles.largePadding} />
+        <WideButtonView
+          title={STRINGS.recover_tokens_title}
+          onPress={() => {
+            wallet.getDecryptedSeed()
+              .then(() => setTokensRecovered(true));
+            // TODO : recover all PoP tokens for this walletHome
+          }}
```
pierluca, line 52: same comment :)

pierluca (changes requested):
The GUI and navigation are much cleaner, great job ! 

There are still a few things that struck me, but we're getting close to having something clean, keep going !

You'll notice that a few of the comments are things that we've already discussed, so they shouldn't really surprise you.

pierluca:
You might also own to have a look at the sonarcloud remarks :) most of them seem applicable and quickly fixed.

pierluca (approved):

PR #308, opened 2021-04-27
Fix error messages

### LINKED ISSUE #295: Be1: The server does not always send an error when needed
For example if a client sends a publish query on a lao channel with data `{"object":"lao","action":"invalidAction"}`, we expect to have an error message like : 
```
{
  "jsonrpc": "2.0",
  "error": {
      "code": -1,
      "description": "invalid action: invalidAction" // or a longer description
  }
}
```
But actually the client get no answer from the server. And this is the same for many other incorrect queries.

I believe this is because the method [`SendError(id int, err error)`](https://github.com/dedis/student_21_pop/blob/ef74df3bbf0f8efb5389ecb80f5080b45a5016cd/be1-go/hub/client.go#L131) is sometimes called with the param `err error` different from the error that we expect (with a code and a description) defined [here](https://github.com/dedis/student_21_pop/blob/ef74df3bbf0f8efb5389ecb80f5080b45a5016cd/be1-go/message/answer.go#L25).

flmichel:
Fixes #295 
- All error generated by a client message are now sent back to the client in an error message.
- Some error codes were corrected
- The log of the error was moved inside the `sendError` method
- Some duplicate logs of errors were removed


flmichel:
This PR depends on #279.

lpatel99 (approved):
Looks good!

be1-go/hub/organizer.go
```diff
@@ -142,15 +151,20 @@ func (o *organizerHub) handleMessageFromClient(incomingMessage *IncomingMessage)
...
 		// Unmarshal the data
 		err = query.Publish.Params.Message.VerifyAndUnmarshalData()
 		if err != nil {
-			log.Printf("failed to verify and unmarshal data: %v", err)
+			// Return a error of type "-4 request data is invalid" for all the verifications and unmarshalling problems of the data
+			err = &message.Error{
+				Code:        -4,
+				Description: fmt.Sprintf("failed to verify and unmarshal data: %v", err),
+			}
+			log.Printf("%v", err)
```
gnarula, line 167: This can be removed since there's a log in `client.SendError`

be1-go/hub/organizer.go
```diff
@@ -159,7 +173,9 @@ func (o *organizerHub) handleMessageFromClient(incomingMessage *IncomingMessage)
 			query.Publish.Params.Message.Data.GetObject() == message.DataObject(message.LaoObject) {
 			err := o.createLao(*query.Publish)
 			if err != nil {
-				log.Printf("failed to create lao: %v", err)
+				err = message.NewError("failed to create lao", err)
+				log.Printf("%v", err)
```
gnarula, line 177: Same here

gnarula (changes requested):
Looks good! I think a few `log.Printfs` can be removed. Please feel free to merge this after removing those.

be1-go/hub/organizer.go
flmichel: Oh yes that is correct!

lpatel99 (approved):

gnarula (approved):

PR #315, opened 2021-05-05
Implementation of a generic hub

flmichel:
Implementation of a generic hub called `baseHub` that will be used in the 2 different types of hub: `organizerHub` and `witnessHub`.


lpatel99 (commented):
All good for me, test and it works correctly.

be1-go/hub/base_hub.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+}
+
+// Recv accepts a message and enques it for processing in the hub.
+func (h *baseHub) Recv(msg IncomingMessage) {
+	log.Printf("Hub::Recv")
+	h.messageChan <- msg
+}
+
+func start(h Hub, done chan struct{}, messageChan chan IncomingMessage) {
```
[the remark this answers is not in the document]
flmichel: I tried to create a function like this `func start(h *baseHub, ...` but I encountered a problem: I couldn't find a way to call the correct method `handleIncomingMessage`. This method is different for a witness and an organizer. This is because the `baseHub` is inside the `organizerHub struct` and `witnessHub struct`.  So the function has a interface Hub `h` as an argument so I couldn't do `h.messageChan`. This is why I passed `messageChan` as an argument.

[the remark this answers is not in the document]
flmichel: But I'm not sure if this is a good idea how I did that.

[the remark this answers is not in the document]
gnarula: I think it's easier to follow this if baseHub implement `Start(done chan struct{})` with the organizer's implementation in it (it looks like the more generic way of handling a message) and `witnessHub` can override it with its custom implementation.

That way we'd
* Have baseHub implement the Hub interface
* Not require handleIncomingMessage() on the Hub interface

be1-go/hub/base_hub.go
```diff
@@ -0,0 +1,117 @@
...
+
+const (
+	GenericMsgSchema string = "genericMsgSchema"
+	DataSchema       string = "dataSchema"
+)
+
+// NewBaseHub returns a Base Hub.
+func NewBaseHub(public kyber.Point) (*baseHub, error) {
+	// Import the Json schemas defined in the protocol section
+	protocolPath, err := filepath.Abs("../protocol")
```
gnarula, line 35: I reckon the validation can be isolated in a separate package.

```go
type struct SchemaValidator {
    schemas map[string]*gojson.Schema
}

// RegisterSchema adds a JSON schema specified by the path
func (s *SchemaValidator) RegisterSchema(path string) error { .. }

func (s *SchemaValidator) VerifyJSON(byteMessage []byte, schema string) error { ..}

func NewSchemaValidator() (*SchemaValidator, error) {
    return NewSchemaValidatorWithSchemas("path/to/genericMessage.json", "path/to/data.json")
}

func NewSchemaValidatorWithSchemas(paths ...string) (*SchemaValidator, error) {
    // instantiate schema and register the paths using RegisterPath
}
```

The baseHub can then use an instance of a SchemaValidator. This way we decouple schema validation from hub and can test the validations independently. Does that sound like a good idea?

be1-go/hub/base_hub.go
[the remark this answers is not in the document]
flmichel: Yes I agree! I made the changes.

be1-go/hub/base_hub.go
flmichel: Yes good idea! I implemented it. I just did a small modification. Instead of using `path string` for the agument in RegisterSchema I used a `schema struct`:
```
type schema struct {
	name string
	path string
}
```
I did that to be able to pass the `name` and the `path` of a schema to the function `RegisterSchema` and `NewSchemaValidatorWithSchemas`

flmichel: Do you think it is a good idea?

be1-go/validation/schema_validator.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+func (s *SchemaValidator) VerifyJson(byteMessage []byte, schemaName string) error {
+	// Validate the Json "byteMessage" with a schema
+	messageLoader := gojsonschema.NewBytesLoader(byteMessage)
+	resultErrors, err := s.schemas[schemaName].Validate(messageLoader)
+	if err != nil {
+		return &message.Error{
+			Code:        -1,
+			Description: err.Error(),
```
gnarula, line 46:
```suggestion
			Description: fmt.Sprintf("failed to validate schema: %s", err.Error()),
```

be1-go/validation/schema_validator.go
```diff
@@ -0,0 +1,93 @@
...
+
+func NewSchemaValidator() (*SchemaValidator, error) {
+	// Import the Json schemas defined in the protocol section
+	protocolPath, err := filepath.Abs("../protocol")
+	if err != nil {
+		return nil, xerrors.Errorf("failed to load the path for the json schemas: %v", err)
+	}
+	protocolPath = "file://" + protocolPath
+
+	genericMsgSchema := schema{GenericMsgSchema, protocolPath + "/genericMessage.json"}
```
gnarula, line 75: Is the path compatible across all environments? Specially Windows?

be1-go/hub/base_hub.go
```diff
@@ -0,0 +1,280 @@
...
+
+	data, ok := publish.Params.Message.Data.(*message.CreateLAOData)
+	if !ok {
+		return &message.Error{
+			Code:        -4,
+			Description: "failed to cast data to CreateLAOData",
+		}
+	}
+
+	encodedID := base64.StdEncoding.EncodeToString(data.ID)
```
gnarula, line 258: Are these changing to base64url as well? Is that in a separate PR?

be1-go/hub/base_hub.go
```diff
@@ -0,0 +1,280 @@
...
+
+	laoChannelID := "/root/" + encodedID
+
+	laoCh := laoChannel{
+		rollCall:    rollCall{},
+		attendees:   make(map[string]struct{}),
+		baseChannel: createBaseChannel(h, laoChannelID),
+	}
+
+	messageID := base64.StdEncoding.EncodeToString(publish.Params.Message.MessageID)
```
gnarula, line 274: Same as above

be1-go/validation/schema_validator.go
gnarula: Also, please use the field names while declaring a struct

be1-go/validation/schema_validator.go
```diff
@@ -0,0 +1,93 @@
...
+	msgSchema, err := gojsonschema.NewSchema(msgLoader)
+	if err != nil {
+		return xerrors.Errorf("failed to load the json schema with path `%s`: %v", schema.path, err)
+	}
+
+	s.schemas[schema.name] = msgSchema
+	return nil
+}
+
+func (s *SchemaValidator) VerifyJson(byteMessage []byte, schemaName string) error {
```
gnarula, line 39: A comment here please

// VerifyJson ...

be1-go/validation/schema_validator.go
```diff
@@ -0,0 +1,93 @@
...
+		return &message.Error{
+			Code:        -1,
+			Description: descriptionErrors,
+		}
+	}
+
+	return nil
+}
+
+func NewSchemaValidator() (*SchemaValidator, error) {
```
gnarula, line 67: Comment here as well
// NewSchemaValidator ...

be1-go/validation/schema_validator.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	protocolPath = "file://" + protocolPath
+
+	genericMsgSchema := schema{GenericMsgSchema, protocolPath + "/genericMessage.json"}
+	dataSchema := schema{DataSchema, protocolPath + "/query/method/message/data/data.json"}
+
+	return NewSchemaValidatorWithSchemas(genericMsgSchema, dataSchema)
+}
+
+func NewSchemaValidatorWithSchemas(schemas ...schema) (*SchemaValidator, error) {
```
gnarula, line 81: Same as above

gnarula (changes requested):
Looks good! Have a comment about the schema paths because I heard there were some issues with validation on windows.

I'm going to modify the CI pipeline to run the Go tests on Windows as well so a quick unit test for `NewSchemaValidator` should do the job. Can you write that please?

be1-go/hub/base_hub.go
flmichel: I think I can change it in #329  (if this PR is merged first)

be1-go/validation/schema_validator.go
flmichel: The windows path had `\\` instead of `/`, I just replaced all `\\` by `/`.

be1-go/validation/schema_validator.go
```diff
@@ -0,0 +1,107 @@
...
+// NewSchemaValidatorWithSchemas returns a Schema Validator for the schemas 'schemas'
+func NewSchemaValidatorWithSchemas(schemas ...schema) (*SchemaValidator, error) {
+	// Instantiate schema
+	schemaValidator := &SchemaValidator{
+		schemas: make(map[string]*gojsonschema.Schema),
+	}
+
+	// Register the paths
+	for _, schema := range schemas {
+		schemaValidator.RegisterSchema(schema)
```
gnarula, line 103: please check for an error here

be1-go/validation/schema_validator_test.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestOrganizer_SchemaLoading(t *testing.T) {
+	_, err := NewSchemaValidator()
+	require.NoError(t, err)
```
gnarula, line 11: I'd suggest adding more assertions

```go
	require.Equal(t, 2, len(validator.schemas))
	require.NotNil(t, validator.schemas[GenericMsgSchema])
	require.NotNil(t, validator.schemas[DataSchema])
```

be1-go/validation/schema_validator.go
```diff
@@ -0,0 +1,107 @@
...
+		}
+	}
+
+	return nil
+}
+
+// NewSchemaValidator returns a Schema Validator
+func NewSchemaValidator() (*SchemaValidator, error) {
+	// Import the Json schemas defined in the protocol section
+	protocolPath, err := filepath.Abs("../protocol")
```
gnarula, line 73: This would fail the test if you check for errors returned from `RegisterSchema`.

gnarula:
```
=== RUN   TestOrganizer_SchemaLoading
    schema_validator_test.go:12:
                Error Trace:    schema_validator_test.go:12
                Error:          Received unexpected error:
                                failed to register schema:
                                    student20_pop/validation.NewSchemaValidatorWithSchemas
                                        /Users/narula/dev/student20_pop/be1-go/validation/schema_validator.go:106
                                  - failed to load the json schema with path `file:///Users/narula/dev/student20_pop/be1-go/protocol/genericMessage.json`:
                                    student20_pop/validation.(*SchemaValidator).RegisterSchema
                                        /Users/narula/dev/student20_pop/be1-go/validation/schema_validator.go:33
                                  - open /Users/narula/dev/student20_pop/be1-go/protocol/genericMessage.json: no such file or directory
                Test:           TestOrganizer_SchemaLoading
--- FAIL: TestOrganizer_SchemaLoading (0.00s)
```

gnarula: This has another assumption that the executable is always invoked from `be1-go` subdirectory. That's not the case for executing tests and might also not hold when the server is deployed.

I'd suggest accepting the protocol directory as a command line argument instead and passing it as an argument to `NewSchemaValidator` and `NewSchemaValidatorWithSchemas`.

gnarula (changes requested):

be1-go/validation/schema_validator.go
flmichel: I added a command line argument for the protocol path and if no value is passed the server loads the schemas using the URL of the protocol (on the GitHub website)

gnarula:
@flmichel Can you resolve the conflicts please?

Also @pierluca suggested a great alternative to using CLI arguments/HTTP urls which is worth exploring - Go 1.16 supports embedding files/folders at compile time. https://blog.carlmjohnson.net/post/2021/how-to-use-go-embed. gojsonschema does have `NewReferenceLoaderFilesystem` at https://github.com/xeipuuv/gojsonschema/blob/b076d39a02e5015af0a2a96636e4cc479ecd9f45/jsonLoader.go#L125. Can you give this a shot?

gnarula (approved):
Tested the batch script on Windows and it works well :)

PR #300, opened 2021-04-18
Elections Display + Manage Elections

maximelaval:
Implemented the display of the elections by creating a specific layout for those. 
Implemented the manage election fragment where an organizer can manage an election by clicking on the edit button next to an election and then terminate it which will successfully make that election disappear in the display and update the list of elections .
Changes made in : 
- LaoDetailViewModel
- LaoRepository
- ElectionSetupFragment
- LaoDetailActivity
- EventExpandableListViewAdapter
- com.github.dedis.student20_pop.model.event where I created an attribute EventType type
- Election in com.github.dedis.student20_pop.model where I set the type .
- RollCall in com.github.dedis.student20_pop.model where I set the type .
- Lao in com.github.dedis.student20_pop.model where I implemented a method to delete an election from the Map of ongoing elections and the same for the RollCalls. 

I also merged with Celine's last PR.

PR #318, opened 2021-05-05
Election Display + Manage Election

maximelaval:
This PR is based from PR #300 . I created this new one due to git problems in my previous PR branch. 
Implemented the display of the elections by creating a specific layout for those.
Implemented the manage election fragment where an organizer can manage an election by clicking on the edit button next to an election and then terminate it which will successfully make that election disappear in the display and update the list of elections .
Changes made in :

LaoDetailViewModel
LaoRepository
ElectionSetupFragment
LaoDetailActivity
EventExpandableListViewAdapter
com.github.dedis.student20_pop.model.event where I created an attribute EventType type
Election in com.github.dedis.student20_pop.model where I set the type .
RollCall in com.github.dedis.student20_pop.model where I set the type .
Lao in com.github.dedis.student20_pop.model where I implemented a method to delete an election from the Map of ongoing elections and the same for the RollCalls.


celinecamacho (commented):
The sonar cloud code analysis failed, you have 5 critical smells and some others that you should work on

fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/adapters/EventExpandableListViewAdapter.java
```diff
@@ -13,280 +8,321 @@
...
+        } else {
+            binding = DataBindingUtil.getBinding(convertView);
+        }
+
+        EventCategory category = (EventCategory) getGroup(groupPosition);
+
+        // we use a switch case to handle all the different type of actions we want when we click on a certain event
+
+        switch (event.getType()) {
+            case ELECTION:
```
celinecamacho, line 257: You can refactor everything inside this case into a method, this way the method `getChildView` doesn't become extremely long with other events arrive

fe2-android/app/src/main/java/com/github/dedis/student20_pop/model/data/LAORepository.java
```diff
@@ -514,6 +517,14 @@ public static void destroyInstance() {
       CreateLao data = (CreateLao) message.getData();
       createLaoRequests.put(id, "/root/" + data.getId());
     }
+    // Uncomment to test display without message from Backend
+    else {
+      if(message.getData() instanceof ElectionSetup) {
+        handleElectionSetup(channel,(ElectionSetup) message.getData());
+      }
+    }
```
celinecamacho, line 525: This "testing hack" for the display should be deleted, this way the election will be handled twice when we receive the broadcast from the backend

fe2-android/app/src/main/res/layout/fragment_manage_election.xml
```diff
@@ -0,0 +1,230 @@
...
+            android:layout_height="match_parent"
+            android:orientation="horizontal"
+            app:layout_constraintGuide_percent="@dimen/guideline_horizontal_tab" />
+
+
+        <ScrollView
+            android:id="@+id/manage_election_scrollview"
+            android:layout_width="match_parent"
+            android:layout_height="match_parent"
+            android:layout_marginTop="50dp"
```
celinecamacho, line 30: You have some dimensions left in this file, layout_election_display.xml and layout_event.xml that should be in the dimens.xml file

fe2-android/app/src/main/res/layout/tab_back.xml
```diff
@@ -0,0 +1,32 @@
...
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        android:orientation="vertical"
+        app:layout_constraintGuide_percent="@dimen/guideline_vertical_tab_connect" />
+
+    <Button
+        android:id="@+id/tab_back"
+        android:layout_width="0dp"
+        android:layout_height="0dp"
+        android:text="Back"
```
celinecamacho, line 26: This string should be in the strings.xml file

celinecamacho (changes requested):
Good job! I have only 4 requested changes, which should not take long to apply :)

fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/adapters/EventExpandableListViewAdapter.java
```diff
@@ -38,311 +40,376 @@
...
+    public View getChildView(
+            int groupPosition,
+            int childPosition,
+            boolean isLastChild,
+            View convertView,
+            ViewGroup parent) {
+
+        Event event = ((Event) getChild(groupPosition, childPosition));
+        EventCategory category = (EventCategory) getGroup(groupPosition);
+        LayoutEventBinding layoutEventBinding;
```
Kar1213, line 254: I think it is better to create a different binding for each event and not use LayoutEventBinding at all. What you can do is to put all the code from creating the binding until returning the view in the functions setUpElectionElement/setUpRollCallElement. Then in the if branches you call the setUp method and return what this method returns.

maximelaval: Thank You I was able to fix the roll call creation problem with the Bindings .

Walfar (approved):
Looks all good to me ! Tested with the back-end and seems to work well. Code looks clean as well.

celinecamacho (approved):

PR #320, opened 2021-05-06
witness opens socket for clients

lpatel99:


be1-go/hub/serveWs.go
```diff
@@ -0,0 +1,56 @@
...
+
+		go client.ReadPump()
+		go client.WritePump()
+
+		// cleanup go routine that removes clients that forgot to unsubscribe
+		go func(c *ClientSocket, h Hub) {
+			c.Wait.Wait()
+			h.RemoveClientSocket(c)
+		}(client, h)
+	case WitnessSocketType:
```
flmichel, line 50: I'm was wondering: do we need to really need to do a different between the `WitnessSocketType` and the `OrganizerSocketType`? The communication mechanism between a witness and a organizer server should exactly the same than between 2 witness servers, right?
But maybe this can be change in an other PR.

flmichel (dismissed):
Looks good! I have minor comments. I think there is some unintentional modification in `message/query.go` and `message/query_test.go`was deleted.

flmichel:
And maybe it would be good write in the README the new websocket full path: `ws://host/organizer/client/` and `ws://host/organizer/witness/`

be1-go/hub/serveWs.go
lpatel99: Good point! For now the only reason for the socket types are in order to determine the addresses at which to open the web sockets (/organizer/client/ vs /witness/client/ for example)

be1-go/README.md
```diff
@@ -32,6 +32,9 @@ OPTIONS:
 You may start the organizer server at ports `9000` for clients and `9001` for witnesses by executing `./pop organizer --pk "<base64 encoded pk>" serve`.
 Please use the `-cp` and `-wp` flags to specify an alternative port.
 
+You may start the witness server at ports `9002` for clients and `9001` for organizer by executing `./pop organizer --pk "<base64 encoded pk>" serve`.
```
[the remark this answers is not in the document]
lpatel99: Indeed!

be1-go/cli/organizer/organizer.go
```diff
@@ -2,23 +2,13 @@ package organizer
...
 	"github.com/urfave/cli/v2"
 	"golang.org/x/xerrors"
 )
 
-var upgrader = websocket.Upgrader{
-	ReadBufferSize:  1024,
-	WriteBufferSize: 1024,
-	CheckOrigin:     func(r *http.Request) bool { return true },
-}
-
```
cba33, line 21: What's the reason for deleting these lines?

cba33 (approved):
Nice, looks good to me!

be1-go/cli/organizer/organizer.go
lpatel99: I moved them to the file serveWs.go in the hub package along with the functions that use the upgrader. I did this to avoid duplicating code because these functions are identical for witness and organizer.

PR #323, opened 2021-05-06
Implemented DSL graph answerer + clientActor

RaulinN:
- 333e17f : DSL graph answerer + client actor
  - The answerer gives an answer a client (front-end) query
  - The client actor represents a client (front-end) that is currently
  connected to the server

### DIFF
--- a/be2-scala/src/main/scala/ch/epfl/pop/Server.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/Server.scala
@@ -26,7 +26,7 @@ object Server {
     val PORT = 8000
     val PATH = ""
 
-    val system = akka.actor.ActorSystem("magic")
+    val system = akka.actor.ActorSystem("pop-be2-inner-actor-system")
     implicit val typedSystem: ActorSystem[Nothing] = system.toTyped
 
     val root = Behaviors.setup[Nothing] { _ =>
--- a/be2-scala/src/main/scala/ch/epfl/pop/Server.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/Server.scala
@@ -2,6 +2,7 @@ package ch.epfl.pop
 
 import java.util.concurrent.TimeUnit
 
+import akka.actor.ActorRef
 import akka.actor.typed.ActorSystem
 import akka.actor.typed.scaladsl.adapter._
 import akka.actor.typed.scaladsl.Behaviors
@@ -9,7 +10,7 @@ import akka.http.scaladsl.Http
 import akka.http.scaladsl.server.Directives._
 import akka.http.scaladsl.server.{RequestContext, RouteResult}
 import akka.util.Timeout
-import ch.epfl.pop.pubsub.PublishSubscribeNew
+import ch.epfl.pop.pubsub.{PubSubMediator, PublishSubscribeNew}
 import org.iq80.leveldb.Options
 
 import scala.concurrent.{ExecutionContextExecutor, Future}
@@ -34,8 +35,9 @@ object Server {
       val options: Options = new Options()
       options.createIfMissing(true)
 
+      val pubSubMediator: ActorRef = system.actorOf(PubSubMediator.props)
       def publishSubscribeRoute: RequestContext => Future[RouteResult] = path(PATH) {
-        handleWebSocketMessages(PublishSubscribeNew.buildGraph(system))
+        handleWebSocketMessages(PublishSubscribeNew.buildGraph(pubSubMediator)(system))
       }
 
       implicit val executionContext: ExecutionContextExecutor = typedSystem.executionContext
--- a/be2-scala/src/main/scala/ch/epfl/pop/Server.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/Server.scala
@@ -3,16 +3,16 @@ package ch.epfl.pop
 import java.util.concurrent.TimeUnit
 
 import akka.actor.typed.ActorSystem
+import akka.actor.typed.scaladsl.adapter._
 import akka.actor.typed.scaladsl.Behaviors
 import akka.http.scaladsl.Http
 import akka.http.scaladsl.server.Directives._
-import akka.stream.scaladsl.{BroadcastHub, Keep, MergeHub}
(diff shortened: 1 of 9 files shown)


be2-scala/src/main/scala/ch/epfl/pop/Server.scala
```diff
@@ -25,37 +25,34 @@ object Server {
     val PORT = 8000
     val PATH = ""
 
-    val root = Behaviors.setup[Nothing] { context =>
-      implicit val system: ActorSystem[Nothing] = context.system
+    val system = akka.actor.ActorSystem("magic")
```
pierluca, line 28: We should probably use another string :)

RaulinN: Done

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ParamsHandler.scala
```diff
@@ -1,5 +1,71 @@
...
+      // ActorFlow.ask(clientActorRef)(makeMessage = (el, replyTo: ActorRef) => SubscribeTo(channel))
+      clientActorRef ! ClientActor.SubscribeTo(jsonRpcMessage.channel)
+      Right(PipelineError(-100, "FIXME: should use akka ask pattern to create an ActorFlow"))
+    case graphMessage@_ => graphMessage
+  }
+
+  def unsubscribeHandler(clientActorRef: ActorRef): Flow[GraphMessage, GraphMessage, NotUsed] = Flow[GraphMessage].map {
+    case Left(jsonRpcMessage: Unsubscribe) =>
+      clientActorRef ! ClientActor.UnsubscribeFrom(jsonRpcMessage.channel)
+      Right(PipelineError(-100, "FIXME: should use akka ask pattern to create an ActorFlow"))
```
gnarula, line 63:
```suggestion
      Right(PipelineError(ErrorCodes.SERVER_FAULT.id, "FIXME: should use akka ask pattern to create an ActorFlow"))
```

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ParamsHandler.scala
```diff
@@ -1,5 +1,71 @@
...
+    }
+  })
+
+  case class Asking(g: GraphMessage, replyTo: ActorRef)
+
+  def subscribeHandler(clientActorRef: ActorRef): Flow[GraphMessage, GraphMessage, NotUsed] = Flow[GraphMessage].map {
+    case Left(jsonRpcMessage: Subscribe) =>
+      // ActorFlow.ask(clientActorRef)(makeMessage = (el, replyTo: ActorRef) => SubscribeTo(channel))
+      clientActorRef ! ClientActor.SubscribeTo(jsonRpcMessage.channel)
+      Right(PipelineError(-100, "FIXME: should use akka ask pattern to create an ActorFlow"))
```
gnarula, line 56:
```suggestion
      Right(PipelineError(ErrorCodes.SERVER_FAULT.id, "FIXME: should use akka ask pattern to create an ActorFlow"))
```

gnarula: I'm not sure, but is this where we'd handle a Nack?

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/Answerer.scala
```diff
@@ -1,12 +1,51 @@
...
+      .collect { case graphMessage: GraphMessage => ClientAnswer(graphMessage) }
+      // Send the ClientAnswer to clientActorRef. Whenever the stream between the client
+      // actor and the actual client (front-end) is broken, the message DisconnectWsHandle
+      // is sent to clientActorRef
+      .to(Sink.actorRef(clientActorRef, DisconnectWsHandle))
+
+    // Integration point between Akka Streams and above actor
+    val source: Source[TextMessage, NotUsed] = Source
+      // By using .actorRef, the source emits whatever the actor "wsHandle" sends
+      .actorRef(bufferSize = 10, overflowStrategy = OverflowStrategy.dropBuffer)
```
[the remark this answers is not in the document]
RaulinN: I upgraded the size of the buffer to 50 for now. Moreover, here are all the possible overflow strategies. I decided to go for the `backpressure` strategy as this is the only one that doesn't drop messages (or fails)

```scala
object OverflowStrategy {

  /**
   * If the buffer is full when a new element arrives, drops the oldest element from the buffer to make space for
   * the new element.
   */
  def dropHead: OverflowStrategy = DropHead(Logging.DebugLevel)

  /**
   * If the buffer is full when a new element arrives, drops the youngest element from the buffer to make space for
   * the new element.
   */
  def dropTail: OverflowStrategy = DropTail(Logging.DebugLevel)

  /**
   * If the buffer is full when a new element arrives, drops all the buffered elements to make space for the new element.
   */
  def dropBuffer: OverflowStrategy = DropBuffer(Logging.DebugLevel)

  /**
   * If the buffer is full when a new element arrives, drops the new element.
   */
  def dropNew: OverflowStrategy = DropNew(Logging.DebugLevel)

  /**
   * If the buffer is full when a new element is available this strategy backpressures the upstream publisher until
   * space becomes available in the buffer.
   */
  def backpressure: OverflowStrategy = Backpressure(Logging.DebugLevel)

  /**
   * If the buffer is full when a new element is available this strategy completes the stream with failure.
   */
  def fail: OverflowStrategy = Fail(Logging.ErrorLevel)
}
```

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ParamsHandler.scala
RaulinN: This is something that should be modified asap. This is the last problem I have with the pubsub. In theory, there is a way to make a flow out of an actor (with Actorflow on line 54) but I cannot yet manange to make it compile

gnarula (approved):

Issue #281, opened 2021-03-25
Protocol: Channel id encoding


flmichel:
The channel id for a lao is stored as follow: `“channel”: “/root/<lao_id>”`. With `<lao_id>` encoded in base64 (At least the go backend is expecting this encoding). And the channel id for an election would be `“channel”: “/root/<lao_id>/<election_id>”` with `<lao_id>`and `<election_id>`in base64. We might have `/` in the in the base64 encoding which might be a bit confusing for the election channel id. Should we used [base64url](https://tools.ietf.org/html/rfc4648#section-5) instead?

pierluca:
That seems like a potentially interesting option. You might want to consider evaluating its support for it across the different stacks :)

pierluca:
We need to come to a decision on this, @dedis/student_21-1_pop !

Can at least one person from each tech stack take responsibility for this, and agree to using base64url or suggest an alternative ? 

Thanks !

lhmerino:
+1 for base64url

Adalsteinnjons:
I can take care of this for the Typescript frontend. 
Should we also use base64url for all the other IDs, such as the Ids of the events (Meetings, Roll-Calls, Elections,...) or just change the channel ID? 
On the web front-end only the channel ID causes problems, so it might be ok to keep the others as is.

Walfar:
In Java, every IDs are encoded using the same hashing method. If I change the encoding of this hash from a base64 encoding to a base64url, it will change the IDs for all events aswell. So I think it's best if we use base64url everywhere, even if it only causes problem for the channel ID.

Adalsteinnjons:
> In Java, every IDs are encoded using the same hashing method. If I change the encoding of this hash from a base64 encoding to a base64url, it will change the IDs for all events aswell. So I think it's best if we use base64url everywhere, even if it only causes problem for the channel ID.

You're right. The same applies to the TS front-end. 
Best to use the same encoding everywhere.

Adalsteinnjons:
@gnarula It is in the same place in the front-end. 
Everything is encoded in this file
https://github.com/dedis/student_21_pop/blob/f2d0721aa17155ad20cc21975430555ce9899242/fe1-web/model/objects/Base64.ts#L1-L22

pierluca:
As this was agreed upon and all stacks have implemented this, I'm closing down this issue.

PR #326, opened 2021-05-10
Changed all usage of Base64 to Base64Url on fe1-web

Adalsteinnjons:
All Ids and data encodings are now in base64Url instead of normal base64.
This is linked to issue #281.


RaulinN (dismissed):

pierluca (dismissed):

Adalsteinnjons:
Tested with @flmichel base64 url branch from the backend and it works.

fe1-web/model/objects/Base64Url.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  public toBuffer(): Buffer {
+    return base64url.toBuffer(this.valueOf());
+  }
+
+  public static fromBase64(b64: string): string {
+    return this.addPadding(base64url.fromBase64(b64));
+  }
+
+  public static addPadding(str: string): string {
```
pierluca, line 31: this looks like it probably shouldn't be public.

fe1-web/model/objects/Base64Url.ts
```diff
@@ -0,0 +1,38 @@
...
+  public equals(o: Base64UrlData): boolean {
+    return this.toString() === o.toString();
+  }
+
+  public toBuffer(): Buffer {
+    return base64url.toBuffer(this.valueOf());
+  }
+
+  public static fromBase64(b64: string): string {
+    return this.addPadding(base64url.fromBase64(b64));
```
pierluca, line 28: shouldn't this return a `Base64UrlData` ?

pierluca (changes requested):
A couple of minor points need addressing, but it seems pretty good overalls 👍

fe1-web/model/objects/Base64Url.ts
Adalsteinnjons: Correct

fe1-web/model/objects/Base64Url.ts
Adalsteinnjons: Correct

pierluca (approved):
LGTM

PR #327, opened 2021-05-10
CastVote Fragment

Warhedgehog:
Implemented CastVote fragment and its layout. 
Implemented a custom adapter - QuestionViewPagerAdapter - for swiping through each election question
Implemented the swiped layout with listView for election ballots
Implemented multiple selection on each question
Adapted Election, LaoDetailViewModel, LAO Repository to accept multiple question per election

No testing so far


fe2-android/app/src/main/java/com/github/dedis/student20_pop/model/Election.java
```diff
@@ -12,15 +14,10 @@
...
-    private String question;
-    private List<String> ballotOptions;
-
-    //votes as attribute ?
-
 
+    private List<ElectionQuestion> electionQuestions;
     public Election() {
-        this.ballotOptions = new ArrayList<>();
+      this.electionQuestions = new ArrayList<>();
```
celinecamacho, line 20: I understand that you want to allow multiple questions in an election, right? The problem here is that you have an election with only a list of questions, deleting all the other important information (ballot options, write in and the actual question). This means that in `LaoRepository` when you handle the ElectionSetup message you are ignoring this information. A solution for this is to create a separate model for questions with this information or keep the old implementation with only one question

celinecamacho (changes requested):
Good job with the fragment! I left some comments about the Election class and how you manage the ElectionSetup

fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/LaoDetailActivity.java
```diff
@@ -314,32 +316,46 @@ private void setupAttendeesListFragment(String id) {
...
+                                  CastVoteFragment castVoteFragment =
+                                          (CastVoteFragment) getSupportFragmentManager().findFragmentById(R.id.fragment_cast_vote);
+                                  if(castVoteFragment == null){
+                                      castVoteFragment = castVoteFragment.newInstance();
+                                      ActivityUtils.replaceFragmentInActivity(
+                                              getSupportFragmentManager(), castVoteFragment, R.id.fragment_container_lao_detail);
+                                  }
+                              }
+                          }
+                  );
```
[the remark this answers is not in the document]
Warhedgehog: Indeed, I don't know what happened here, I changed it back as to it is on master

fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/adapters/QuestionViewPagerAdapter.java
```diff
@@ -0,0 +1,130 @@
...
+        //setting the question
+        ElectionQuestion question = election.getElectionQuestions().get(position);
+        holder.questionView.setText(question.getQuestion());
+
+
+
+        //todo change when implemented in the setup
+        //this will determine the number of option the user can select and vote for
+       // int numberOfChoices = numbersOfChoices.get(position);
+        int numberOfChoices = 1; //by default
```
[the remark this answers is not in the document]
Warhedgehog: For now we have only one voting method which admits voting for one candidate only for each question. More voting vethods could be added in the future and in that case the cast vote fragment is already ready to accept any number of selectable candidates.
Nonetheless, the comment didn't make that much sense so I changed it

fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/fragments/ManageElectionFragment.java
```diff
@@ -66,7 +66,7 @@ public View onCreateView(
         endTime.setText(dateFormat.format(dEnd));
         laoName.setText(laoDetailViewModel.getCurrentLaoName().getValue());
         electionName.setText(laoDetailViewModel.getCurrentElection().getName());
-        question.setText("Election Question : " + laoDetailViewModel.getCurrentElection().getQuestion());
+        question.setText("Election Question : " + laoDetailViewModel.getCurrentElection().getElectionQuestions().get(0).getQuestion()); //Todo change that
```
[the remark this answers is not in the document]
Warhedgehog: This was a temporary comment that should not have been left for the PR. When the setUp allows multiple question the whole layout of this fragment will have to be modified anyway

fe2-android/app/src/main/java/com/github/dedis/student20_pop/model/Election.java
Warhedgehog: The ElectionQuestion object has all the fields that I deleted in the Election class. Then all the relevant information are kept together by question in an object instead of having lists of lists in Election.

celinecamacho: I must have missed the fact that the list of questions was a list of `ElectionQuestion` and not String, my bad. Indeed the implementation is ok and there's nothing to change here

fe2-android/app/src/test/java/com/github/dedis/student20_pop/model/ElectionTest.java
```diff
@@ -14,29 +13,20 @@
...
     private long endTime = 1;
     private long creationTime = 0;
-    private List<String> ballotOptions = Arrays.asList("candidate1", "candidate2");
+    private List<List<String>> ballotsOptions = Arrays.asList(Arrays.asList("candidate1", "candidate2"));
     private Election election = new Election();
 
     @Test
     public void settingNullParametersThrowsException() {
         assertThrows(IllegalArgumentException.class, () -> election.setName(null));
-        assertThrows(IllegalArgumentException.class, () -> election.setBallotOptions(null));
```
celinecamacho, line 28: Instead of deleting all these tests you can move them to a file dedicated to test the class `ElectionQuestion` :)

Warhedgehog: There already are tests for ElectionQuestion (that I didn't write) :)

celinecamacho (approved):
Looks good to me 👍

PR #329, opened 2021-05-10
Changed all base64 encoding/decoding to base64url on be1-go

flmichel:
This is linked to issue #281.


gnarula (commented):
Is the encoding of the `Data` field for a message being changed as well? https://github.com/dedis/student_21_pop/blob/master/protocol/query/method/message/messageGeneral.json

If so, then you'd have to change the Marshaler/Unmarshaler in `messages/message.go`

gnarula (approved):
Looks good to me!

PR #331, opened 2021-05-12
Changed all Base64 encode/decode to Base64Url on Java Front-end

Walfar:
Replaced every call to Base64.encode to Base64.getUrlEncoder().encodeToString. Same for the decoder. No skip for padding "=" used, as it doesn't concern our initial problem with the slashes. The ids, the keys and the data should now be url safe.
I also replaced calls of android.util.Base64 by java.util.Base64, as it seemed more consistent to use the same encoder/decoder everywhere, although the result should be equivalent.


fe2-android/app/src/main/java/com/github/dedis/student20_pop/utility/security/Keys.java
```diff
@@ -23,8 +23,8 @@ public static String getEncodedKey(KeysetHandle handle) throws IOException {
     JsonObject keyData = keyObject.get("keyData").getAsJsonObject();
 
     String encoded = keyData.get("value").getAsString();
-    byte[] buf = Base64.decode(encoded, Base64.NO_WRAP);
+    byte[] buf = Base64.getUrlDecoder().decode(encoded);
 
-    return Base64.encodeToString(buf, 2, 32, Base64.NO_WRAP);
```
celinecamacho, line 28: Here we have 2 as an offset and 32 as the length, so we encode 32 bytes of `buf` from the position 2. You can use the flag `Base64.URL_SAFE` and keep the [android.util.Base64](https://developer.android.com/reference/android/util/Base64) method `encodeToString` or you can slice the byte array before encoding it with `getUrlEncoder().encodeToString()`

celinecamacho (commented):
Good job! I just found one decoding in LaoDetailViewModel.java, line `286`, that wasn't modified

celinecamacho (commented):
The line `568` on `LaoDetailViewModel` still uses Base64

fe2-android/app/src/prod/java/com/github/dedis/student20_pop/Injection.java
```diff
@@ -47,7 +47,12 @@
 import okhttp3.logging.HttpLoggingInterceptor;
 
 public class Injection {
+<<<<<<< HEAD
+
+  private static String SERVER_URL = "ws://10.0.2.2:8080/organizer/client/";
+=======
   private static String SERVER_URL = "ws://10.0.2.2:9000/organizer/client";
+>>>>>>> master
```
celinecamacho, line 55: It seems you had some issues while merging master on your branch, this is why the fe2 build failed. As for the Base64 changes, after making a grep everything seems to be modified

Walfar: Yes, indeed ! I still have to test with Francois's changes before merging anyway, to make sure everything works as expected. The problem is that, with the json schema verification, there are some incompatibilities with Windows (the OS I use). He's working on making it work with Windows so that I can test with the backend. If it's still not ready on Monday, I will ask someone using MacOs to test my branch.

Walfar:
Tested with the back-end in base64url, everything seems to work as expected :)

celinecamacho (approved):
Looks good!

celinecamacho (approved):
All good 👍

PR #333, opened 2021-05-17
Witness socket for witnesses

lpatel99:
The witness now opens a socket for other witnesses. It is supposed to also connect to another witness if its address and port are passed on the command line but this does not yet work and I do not know why. It always returns an error: either "websocket: bad handshake" or "dial tcp [::1]:9001: connect: connection refused".


be1-go/cli/pop.go
```diff
@@ -10,6 +10,42 @@ import (
...
+		Name:    "witness-port",
+		Aliases: []string{"wp"},
+		Usage:   "port to listen websocket connections from witnesses on",
+		Value:   9002,
+	}
+	otherWitnessFlag := &cli.StringSliceFlag{
+		Name:    "other-witness",
+		Aliases: []string{"ow"},
+		Usage:   "address and port to connect to other witness",
+		Value:   nil,
```
Tesa27, line 46: Why is the value of this post nil? Is it the same thing if we suppress this line?

Tesa27 (commented):
Nice work, I only have one small comment bellow.

be1-go/cli/pop.go
lpatel99: oh good point!

PR #335, opened 2021-05-17
Handling of ElectionEnd/ElectionResult data

Walfar:
- ElectionEnd class that represents the data received from the BackEnd when the election terminates
- ElectionResult class that represents the data received from the BackEnd when the result should be displayed
- Handling of those events in the LAO repository. When receiving ElectionEnd, the organizer's registered votes are stored in the election. In the future, this will be used by the witnesses to compare with their own registered votes (also stored in the election). When receiving ElectionResult, the results are sorted and stored in the election (to be used by the UI for the Election result fragment)
- Added some tests for  ElectionEnd, ElectionResult, ElectionQuestionResult (representing a question field in the electionResult data) + additional tests for ElectionSetup


fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/LaoDetailViewModel.java
```diff
@@ -140,6 +143,109 @@ protected void onCleared() {
...
+        }
+
+        String channel = election.getChannel();
+        String laoId = lao.getChannel().substring(6); // removing /root/ prefix
+        ElectionEnd electionEnd = new ElectionEnd(election.getId(), laoId, election.computerRegisteredVotes());
+
+        try {
+            KeysetHandle publicKeysetHandle = mKeysetManager.getKeysetHandle().getPublicKeysetHandle();
+            String publicKey = Keys.getEncodedKey(publicKeysetHandle);
+            byte[] sender = Base64.getDecoder().decode(publicKey);
```
celinecamacho, line 162: Shouldn't it be Base64URL now?

Walfar: Yes, it was merged with the old version of the master that still had some base64 in the ViewModel, but I will resolve those conflicts and merge with the latest version in an instant :)

fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/LaoDetailViewModel.java
```diff
@@ -219,19 +267,22 @@ public void sendVote(List<ElectionVote> votes) {
                                     answer -> {
                                         if (answer instanceof Result) {
                                             Log.d(TAG, "sent a vote successfully");
+                                            // Toast ? + send back to election screen or details screen ?
```
celinecamacho, line 270: Is this comment an implementation to be done? You can change this to a TODO

fe2-android/app/src/main/java/com/github/dedis/student20_pop/model/Election.java
```diff
@@ -78,12 +128,39 @@ public long getStartTimestamp() {
...
+     */
+    public String computerRegisteredVotes() {
+        List<String> listOfVoteIds = new ArrayList<>();
+        //Since messageMap is a TreeMap, votes will already be sorted in the alphabetical order of messageIds
+        for (String senderPk: messageMap.values()) {
+            for (ElectionVote vote: voteMap.get(senderPk)) {
+                listOfVoteIds.add(vote.getId());
+            }
+        }
+        if (listOfVoteIds.isEmpty()) return Hash.hash("novotes");
```
celinecamacho, line 147: Is this convention of hashing "novotes" decided for when there are no votes?

fe2-android/app/src/main/java/com/github/dedis/student20_pop/model/Lao.java
```diff
@@ -45,11 +45,11 @@ public void updateRollCall(String prevId, RollCall rollCall) {
   }
 
   public void updateElection(String prevId, Election election) {
-    if (elections.containsKey(prevId)) {
-      elections.remove(prevId);
-    }
-    String newId = election.getId();
-    elections.put(newId, election);
```
celinecamacho, line 52: Why has this check been deleted?

fe2-android/app/src/main/java/com/github/dedis/student20_pop/model/network/method/message/ElectionQuestion.java
```diff
@@ -1,5 +1,4 @@
-package com.github.dedis.student20_pop.model.network.method.message.data.election;
-
+package com.github.dedis.student20_pop.model.network.method.message;
```
celinecamacho, line 1: I think keeping this file in the package `data/election`is better

fe2-android/app/src/main/java/com/github/dedis/student20_pop/model/network/method/message/ElectionResultQuestion.java
```diff
@@ -0,0 +1,24 @@
+package com.github.dedis.student20_pop.model.network.method.message;
```
celinecamacho, line 1: Also, adding this to the `data/election` package is better

fe2-android/app/src/main/java/com/github/dedis/student20_pop/model/network/method/message/ElectionVote.java
```diff
@@ -1,4 +1,4 @@
-package com.github.dedis.student20_pop.model.network.method.message.data.election;
+package com.github.dedis.student20_pop.model.network.method.message;
```
celinecamacho, line 1: Same thing about the package

fe2-android/app/src/main/java/com/github/dedis/student20_pop/model/network/method/message/QuestionResult.java
```diff
@@ -0,0 +1,20 @@
+package com.github.dedis.student20_pop.model.network.method.message;
```
celinecamacho, line 1: Same thing about the package

celinecamacho (commented):
There are some classes that should be in different packages but otherwise it looks good

fe2-android/app/src/main/java/com/github/dedis/student20_pop/model/network/method/message/ElectionQuestion.java
Walfar: I will change it right now

fe2-android/app/src/main/java/com/github/dedis/student20_pop/model/Election.java
Walfar: I put this just to prevent an election without any votes to make the hashing process crash, but after further discussion with the other teams, I'll just send an empty string

fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/LaoDetailViewModel.java
Walfar: Indeed, I forgot about that line. I'll just change it to a Toast

fe2-android/app/src/main/java/com/github/dedis/student20_pop/model/Lao.java
Walfar: If I'm not mistaking, when you use .put on a map, the old value will be replaced by the new one, or added if no such mapping exists. As the id of the election doesn't change, the prevId parameter will always correspond to election.getId() (since we simply want to update an election), so it seemed like this check was useless.

fe2-android/app/src/main/res/layout/layout_election_display.xml
```diff
@@ -59,8 +59,6 @@
             android:id="@+id/election_action_button"
             android:layout_width="wrap_content"
             android:layout_height="@dimen/event_buttons_height"
-            android:text="@{eventCategory == EventCategory.PRESENT ? `Cast Vote ` : `Tally up Votes `}"
-            android:enabled="@{eventCategory == EventCategory.PRESENT ? true : view_model.isOrganizer() ? true : false}"
```
celinecamacho: This election action button doesn't have a text anymore and is always enabled, is this the desired action?

Walfar: Yes, all of that is now set in the setupElectionElement method in EventExpandableListViewAdapter. I know that it is better to have layout-related code directly in the xml, but here, the text changes when an event happens (hence requiring use of an observer). If the text was already defined in the xml, when clicking on another layout categroy (e.g PRESENT, whereas the election is in PAST), this would reload the xml, and "override" the text. With the old version, that would then change the text from "Waiting for results" to "Tally up votes". As for enabling, it seemed easier to enable/disable depending on the category and the event in the method as well.

celinecamacho (dismissed):

Warhedgehog (approved):
Looks good to me

PR #337, opened 2021-05-18
Order the message in a catchup answer

### LINKED ISSUE #336: Be1: Message are not sorted in time (ascending) in the catchup answer
I think the message are sorted with their ID before being send in a catchup answer. It should be sorted in time instead.

flmichel:
Fixes #336.
I created a new struct: 
```
type messageInfo struct {
	message    message.Message
	storedTime message.Timestamp
}
```
And now the `inbox` of a channel is a map from the `messageID` to the `messageInfo` in order to know when the message was stored.
The `Catchup` methods now sort the messages in  `storedTime` (ascending).


lpatel99 (commented):
Looks good to me.

gnarula (changes requested):
I'd suggest having something like

```go
messages := make([]*messageInfo, len(c.inbox))

// iterate over map and collect all the values (messageInfo instances)

// sort.Slice on messages based on the timestamp

// iterate and extract the messages[i].message field and
// append it to the result slice
```

This way we avoid losing messages for common timestamps

be1-go/hub/base_channel.go
```diff
@@ -70,14 +77,41 @@ func (c *baseChannel) Catchup(catchup message.Catchup) []message.Message {
 	c.inboxMu.RLock()
 	defer c.inboxMu.RUnlock()
 
+	timestampMap := make(map[message.Timestamp]message.Message)
+	for _, msgInfo := range c.inbox {
+		timestampMap[msgInfo.storedTime] = msgInfo.message
```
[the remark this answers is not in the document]
flmichel: Oh yes that's right! I made the changes.

be1-go/hub/organizer.go
```diff
@@ -594,7 +590,7 @@ func (c *laoChannel) processMessageObject(public message.PublicKey, data message
 			c.inboxMu.Unlock()
 			return nil
 		}
-		msg.WitnessSignatures = append(msg.WitnessSignatures, message.PublicKeySignaturePair{
+		msgInfo.message.WitnessSignatures = append(msgInfo.message.WitnessSignatures, message.PublicKeySignaturePair{
```
gnarula, line 593: I think this wouldn't work ebcause `msgInfo` is not a pointer to the struct.

Please have a look at: https://play.golang.org/p/2uHpMO2lCzh

Please try declaring `inbox` as `map[string]*messageInfo` instead

gnarula (changes requested):

be1-go/hub/base_channel.go
```diff
@@ -70,14 +77,42 @@ func (c *baseChannel) Catchup(catchup message.Catchup) []message.Message {
...
+	// append it to the result slice
+	for _, msgInfo := range messages {
+		result = append(result, msgInfo.message)
 	}
 
 	return result
 }
 
+func (c *baseChannel) storeMessage(msg message.Message) {
+	msgIDEncoded := base64.StdEncoding.EncodeToString(msg.MessageID)
```
gnarula, line 103: URLEncoding here :)

be1-go/hub/organizer.go
nkcr: At this stage it might be interesting to create a new inbox type:

```
type inbox {
  sync.Mutex
  msgs map[string]*messageInfo
}

func (i inbox) addWitnessSig(key string, sig ...) {
  i.msgs[key] = append(i.msgs[key], sig)
}
```

then

```
sig := message.PublicKeySignaturePair{
  ...
}
c.inbox.addWitnessSig(msgEncoded, sig)

File: be1-go/hub/inbox.go [hunk 4]
```diff
@@ -0,0 +1,64 @@
...
+func (i *inbox) addWitnessSig(messageID []byte, public message.PublicKey, signature message.Signature) {
+	msg, ok := i.getMessage(messageID)
+	if !ok {
+		// TODO: We received a witness signature before the message itself.
+		// We ignore it for now but it might be worth keeping it until we
+		// actually receive the message
+		log.Printf("failed to find message_id %s for witness message", messageID)
+	}
+	i.mutex.Lock()
+	msg.WitnessSignatures = append(msg.WitnessSignatures, message.PublicKeySignaturePair{
```
gnarula, line 34: I don't think this will update the existing message since `getMessage` doesn't return a pointer to the struct

```go
func TestAddWitnessSig_AddWitnessSignature(t *testing.T) {
	inbox := createInbox()

	msg, err := message.NewMessage(message.PublicKey{1, 2, 3}, message.Signature{1, 2, 3}, nil, nil)
	require.NoError(t, err)

	inbox.storeMessage(*msg)

	require.Equal(t, 1, len(inbox.msgs))

	buf, err := base64.URLEncoding.DecodeString("oJYBapM5ZuVrnggAwzQMa3oBLrFSjEQY-hv_JQRgs1U=")
	require.NoError(t, err)

	inbox.addWitnessSig(buf[:], message.PublicKey{4, 5, 6}, message.Signature{7, 8, 9})

	storedMsg, ok := inbox.getMessage(buf[:])
	require.Equal(t, true, ok)

	require.Equal(t, 1, len(storedMsg.WitnessSignatures))
}
```

be1-go/hub/inbox.go
```diff
@@ -0,0 +1,64 @@
...
+func createInbox() *inbox {
+	return &inbox{
+		mutex: sync.RWMutex{},
+		msgs:  make(map[string]*messageInfo),
+	}
+}
+
+// addWitnessSig adds a signature of witness to a message of ID `messageID`.
+// if no message has ID `messageID`, does nothing
+func (i *inbox) addWitnessSig(messageID []byte, public message.PublicKey, signature message.Signature) {
```
gnarula, line 25:
```suggestion
func (i *inbox) addWitnessSignature(messageID []byte, public message.PublicKey, signature message.Signature) {
```

gnarula (changes requested):
Please add more tests for `inbox::addWitnessSignature` - to begin with please consider the scenarios where:

1. The inbox is empty and you're adding a witness signature for a message id that doesn't exist
2. The message id exists and you're appending to the witness signature slice

be1-go/hub/base_channel_test.go
```diff
@@ -0,0 +1,39 @@
...
+	for i := 0; i < messageNumber; i++ {
+		// Create a new message containing only an id
+		message := message.Message{MessageID: []byte{byte(i)}}
+		messages = append(messages, message)
+
+		// Store the message in the inbox
+		channel.inbox.storeMessage(message)
+
+		// Wait before storing a new message
+		time.Sleep(time.Microsecond)
```
gnarula, line 28: I believe we need this to get a unique timestamp for each message in the test?

be1-go/hub/inbox_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+	// Add a message to the inbox
+	inbox.storeMessage(*msg)
+
+	require.Equal(t, 1, len(inbox.msgs))
+
+	buf, err := base64.URLEncoding.DecodeString("oJYBapM5ZuVrnggAwzQMa3oBLrFSjEQY-hv_JQRgs1U=")
+	require.NoError(t, err)
+
+	sinaturesNumber := 100
```
gnarula, line 68:
```suggestion
	signaturesNumber := 100
```

be1-go/hub/inbox_test.go
```diff
@@ -0,0 +1,80 @@
...
+	for i := 0; i < sinaturesNumber; i++ {
```
gnarula, line 69:
```suggestion
	for i := 0; i < signaturesNumber; i++ {
```

be1-go/hub/inbox_test.go
```diff
@@ -0,0 +1,80 @@
...
+		// Add the witness signature to the message in the inbox
+		ok := inbox.addWitnessSignature(buf[:], message.PublicKey{byte(i)}, message.Signature{byte(i)})
+		require.True(t, ok, fmt.Sprintf("message n:%d", i))
+	}
+
+	// Check if the message was updated
+	storedMsg, ok := inbox.getMessage(buf[:])
+	require.True(t, ok)
+
+	require.Equal(t, sinaturesNumber, len(storedMsg.WitnessSignatures))
```
gnarula, line 79:
```suggestion
	require.Equal(t, signaturesNumber, len(storedMsg.WitnessSignatures))
```

be1-go/hub/organizer.go
```diff
@@ -590,21 +579,7 @@ func (c *laoChannel) processMessageObject(public message.PublicKey, data message
...
-			log.Printf("failed to find message_id %s for witness message", msgEncoded)
-			c.inboxMu.Unlock()
-			return nil
-		}
-		msg.WitnessSignatures = append(msg.WitnessSignatures, message.PublicKeySignaturePair{
-			Witness:   public,
-			Signature: witnessData.Signature,
-		})
-		c.inboxMu.Unlock()
+		c.inbox.addWitnessSignature(witnessData.MessageID, public, witnessData.Signature)
```
gnarula, line 582: We're not checking the return value of `addWitnessSignature` here. Is that safe?

be1-go/hub/base_channel_test.go
flmichel: Yes! Maybe it would be better to just increment the timestamp

be1-go/hub/organizer.go
flmichel: No I don't think this is safe! I'm gonna add a check!

gnarula (dismissed):
LGTM

gnarula (approved):

PR #343, opened 2021-05-22
Witness socket for witnesses

lpatel99:


be1-go/cli/witness/witness.go
```diff
@@ -42,36 +44,59 @@ func Serve(context *cli.Context) error {
...
+	case hub.WitnessSocketType:
+		if address == "" {
+			return nil
+		}
+		urlString = fmt.Sprintf("ws://%s/%s/witness/", address, socketType)
+	}
+	u, err := url.Parse(urlString)
 	if err != nil {
-		return nil, xerrors.Errorf("failure to connect to organizer: %v", err)
+		return xerrors.Errorf("parsing, %v", err)
```
gnarula, line 83:
```suggestion
		return xerrors.Errorf("failed to parse connection url %s %v", url, err)
```

be1-go/cli/witness/witness.go
```diff
@@ -42,36 +44,59 @@ func Serve(context *cli.Context) error {
...
 	}
 	ws, _, err := websocket.DefaultDialer.Dial(u.String(), nil)
 	if err != nil {
-		return ws, xerrors.Errorf("failure to connect to organizer: %v", err)
+		return xerrors.Errorf("dialing, %v", err)
```
gnarula, line 87:
```suggestion
		return xerrors.Errorf("failed to dial %v", err)
```

be1-go/cli/witness/witness.go
```diff
@@ -42,36 +44,59 @@ func Serve(context *cli.Context) error {
...
 	done <- struct{}{}
 
 	return nil
 }
 
-func connectToOrganizer(organizerAddr string, port int) (*websocket.Conn, error) {
-	address := fmt.Sprintf("ws://%s:%d/organizer/witness/", organizerAddr, port)
-	u, err := url.Parse(address)
+func connectToSocket(socketType hub.SocketType, address string, h hub.Hub, port int) error {
+	var urlString = ""
```
gnarula, line 71:
```suggestion
	var url string
```

be1-go/cli/witness/witness.go
```diff
@@ -42,36 +44,59 @@ func Serve(context *cli.Context) error {
...
-	go organizerSocket.WritePump()
-	go organizerSocket.ReadPump()
+	go hub.CreateAndServeWs(hub.WitnessHubType, hub.ClientSocketType, h, clientPort)
+	go hub.CreateAndServeWs(hub.WitnessHubType, hub.WitnessSocketType, h, witnessPort)
 
 	done := make(chan struct{})
-	go h.Start(done)
-
-	hub.CreateAndServeWs(hub.WitnessHubType, hub.ClientSocketType, h, clientPort)
+	h.Start(done)
```
gnarula, line 63: Maybe I'm missing something but I reckon `h.Start()` runs infinitely so this would never unblock, making lines 65-67 unreachable

gnarula (changes requested):
Looks good, I've some questions about exiting cleanly from the `Serve` method.

Another thing which applies to the organizer's Serve method too and should be taken up in a separate PR possibly in the future is cleanly shutting down - right now we rely on the main go routine ending to stop other go routines. That works but it doesn't perform a graceful shutdown for the clients (both frontends and other servers)

be1-go/cli/witness/witness.go
lpatel99: I have implemented a solution for this, could you have a look and tell me what you think?

be1-go/hub/witness.go
```diff
@@ -63,14 +65,18 @@ func (w *witnessHub) handleIncomingMessage(incomingMessage *IncomingMessage) {
...
+
 	log.Printf("started witness...")
 
 	for {
 		select {
 		case incomingMessage := <-w.messageChan:
 			w.handleIncomingMessage(&incomingMessage)
-		case <-done:
+		case <-ctx.Done():
+			fmt.Println("closing the hub...")
```
gnarula, line 79:
```suggestion
			log.Println("closing the hub...")
```

be1-go/hub/serveWs.go
```diff
@@ -14,21 +16,33 @@ var upgrader = websocket.Upgrader{
...
 
-func CreateAndServeWs(hubType HubType, socketType SocketType, h Hub, port int) error {
-	http.HandleFunc("/"+string(hubType)+"/"+string(socketType)+"/", func(w http.ResponseWriter, r *http.Request) {
-		serveWs(socketType, h, w, r)
+func CreateAndServeWS(ctx context.Context, hubType HubType, socketType SocketType, h Hub, port int, wg *sync.WaitGroup) *http.Server {
+	wg.Add(1)
+	srv := &http.Server{Addr: fmt.Sprintf(":%d", port)}
+
+	path := fmt.Sprintf("/%s/%s/", hubType, socketType)
+	log.Printf("handling: %s", path)
```
gnarula, line 24: Please remove this

be1-go/hub/serveWs.go
```diff
@@ -14,21 +16,33 @@ var upgrader = websocket.Upgrader{
...
+	http.HandleFunc(path, func(w http.ResponseWriter, r *http.Request) {
+		log.Println("In here")
```
gnarula, line 26: And this

gnarula (dismissed):
Looks good, just a few minor changes regarding the log statements

be1-go/hub/serveWs.go
```diff
@@ -37,20 +52,43 @@ func serveWs(socketType SocketType, h Hub, w http.ResponseWriter, r *http.Reques
...
 	case WitnessSocketType:
-		witness := NewWitnessSocket(h, conn)
+		witness := NewWitnessSocket(h, conn, wg)
+
+		go witness.ReadPump(ctx)
+		go witness.WritePump(ctx)
+	}
+}
+
+func ShutDownServers(ctx context.Context, cancel context.CancelFunc, witnessSrv *http.Server, clientSrv *http.Server, wg *sync.WaitGroup) {
```
gnarula, line 73: I think it's better if this is in the cli package instead - this isn't exactly tied to the hub. A nit about naming but I'd prefer `ShutdownServers` instead :)

be1-go/cli/witness/witness.go
```diff
@@ -41,38 +41,80 @@ func Serve(context *cli.Context) error {
...
-	go organizerSocket.WritePump()
-	go organizerSocket.ReadPump()
+	// increment wait group and create and serve servers for witnesses and clients
+	clientSrv := hub.CreateAndServeWS(ctx, hub.WitnessHubType, hub.ClientSocketType, h, clientPort, wg)
+	witnessSrv := hub.CreateAndServeWS(ctx, hub.WitnessHubType, hub.WitnessSocketType, h, witnessPort, wg)
 
-	done := make(chan struct{})
-	go h.Start(done)
+	// shut down client server and witness server when ctrl+c received
+	go hub.ShutDownServers(ctx, cancel, clientSrv, witnessSrv, wg)
```
gnarula, line 73: This should probably be after `go h.Start(ctx, wg)` and can be a blocking call i.e. without the `go` keyword.

be1-go/hub/serveWs.go
gnarula: We can also avoid passing in `cancel` and `wg` if this is invoked as a blocking function

gnarula (changes requested):

be1-go/hub/serveWs.go
lpatel99: Indeed good idea! I have renamed it and made it a blocking call to avoid passing it cancel and wg.

be1-go/cli/pop.go
```diff
@@ -10,6 +11,41 @@ import (
...
+		Name:    "client-port",
+		Aliases: []string{"cp"},
+		Usage:   "port to listen websocket connections from clients on",
+		Value:   9000,
+	}
+	witnessPortFlag := &cli.IntFlag{
+		Name:    "witness-port",
+		Aliases: []string{"wp"},
+		Usage:   "port to listen websocket connections from witnesses on",
+		Value:   9002,
```
gnarula, line 41: This should be a different value than `9002` because it conflicts with the default for `organizerPortFlag`

be1-go/cli/witness/witness.go
```diff
@@ -41,38 +42,80 @@ func Serve(context *cli.Context) error {
...
-	go h.Start(done)
+	// increment wait group and launch organizer hub
+	go h.Start(ctx, wg)
 
-	hub.CreateAndServeWs(hub.WitnessHubType, hub.ClientSocketType, h, clientPort)
+	// shut down client server and witness server when ctrl+c received
+	network.ShutdownServers(ctx, clientSrv, witnessSrv)
 
-	done <- struct{}{}
+	// wait for all goroutines to finish
```
gnarula, line 79:
```suggestion

	// cancel the context
	cancel()
	
	// wait for all goroutines to finish
```

be1-go/cli/organizer/organizer.go
```diff
@@ -1,51 +1,67 @@
...
+	// increment wait group and launch organizer hub
+	go h.Start(ctx, wg)
 
-	go hub.CreateAndServeWs(hub.OrganizerHubType, hub.WitnessSocketType, h, witnessPort)
-	hub.CreateAndServeWs(hub.OrganizerHubType, hub.ClientSocketType, h, clientPort)
+	// shut down client server and witness server when ctrl+c received
+	network.ShutdownServers(ctx, witnessSrv, clientSrv)
 
-	done <- struct{}{}
+	// wait for all goroutines to finish
```
gnarula, line 63:
```suggestion
	// cancel the context
	cancel()
	
	// wait for all goroutines to finish
```

gnarula (changes requested):
We're missing a call to `cancel()` which causes a deadlock on `wg.Done()`

gnarula (approved):

PR #344, opened 2021-05-24
Encrypt the seed

filippo888:
## Content 
This PR manages the encryption of the seed inside the wallet class. The principal changes are:
- The init of the AndroidKeysetManager when we call import/ export method of the wallet
- Linked to the previous point: adding a applicationContext parameter to the two method signature allowing the initialization of the AndroidKeysetManager, add comment the unit test for the wallet because there is no way to have the context without an activity.


fe2-android/app/src/main/java/com/github/dedis/student20_pop/model/Wallet.java
```diff
@@ -271,5 +285,17 @@ private String convertStringToPath(String string){
...
+      throws IOException, GeneralSecurityException {
+    AesGcmKeyManager.register(true);
+    AeadConfig.register();
+    keysetManager =
+        new AndroidKeysetManager.Builder()
+            .withSharedPref(applicationContext, "POP_KEYSET_2",  "POP_KEYSET_SP_2")
+            .withKeyTemplate(AesGcmKeyManager.rawAes256GcmTemplate())
+            .withMasterKeyUri("android-keystore://POP_MASTER_KEY_2")
+            .build();
+    aead = keysetManager.getKeysetHandle().getPrimitive(Aead.class);
```
celinecamacho, line 298: From what I see you are creating a new Keyset Manager each time you want to import/export the seed in order to encrypt/decrypt it, why so?

fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/LaoDetailViewModel.java
```diff
@@ -617,6 +617,8 @@ public void enterRollCall(String id) {
         mPkRollCallEvent.postValue(new Event<>(pk));
       } catch (NoSuchAlgorithmException | InvalidKeyException | ShortBufferException e) {
         Log.d(TAG, "failed to retrieve public key from wallet", e);
+      } catch (GeneralSecurityException e) {
+        Log.d(TAG, "failed to retrieve public key from wallet", e);;
```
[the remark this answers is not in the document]
filippo888: nice catch!

fe2-android/app/src/main/java/com/github/dedis/student20_pop/model/Wallet.java
filippo888: yes you are right! is better init only one time!

fe2-android/app/src/test/java/com/github/dedis/student20_pop/model/WalletTest.java
```diff
@@ -19,6 +19,7 @@
   public void importSeedAndExportSeedAreCoherent()
       throws NoSuchAlgorithmException, InvalidKeyException, ShortBufferException {
 
+    /*
```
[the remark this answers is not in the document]
filippo888: I have a problem with this unit test, I don't have the applicationContext so I don't know how I can test the wallet...

fe2-android/app/src/main/java/com/github/dedis/student20_pop/model/Wallet.java
```diff
@@ -223,24 +248,29 @@ private String convertStringToPath(String string){
...
+      for (String i : words)
+        joiner.add(i);
+      seed = aead.encrypt(new SeedCalculator().calculateSeed(joiner.toString(), ""),
+          new byte[0]);
+      Log.d(TAG, "ExportSeed: new seed initialized: " + Utils.bytesToHex(seed));
 
-    return words;
+      return words;
+    } else {
+      throw new IllegalArgumentException("key set manager not init!");
```
celinecamacho, line 272: I don't think IllegalArgumentException is the right exception, if you need an exception at all

filippo888: Did you think that instead of throw an error I can make a log.d() ?

celinecamacho: Sure, and return null like in `importSeed`

celinecamacho (approved):
Looks good to me!

PR #345, opened 2021-05-24
Wallet content on fe2

Kar1213:
The changes consist of an implementation of the wallet content containing essentially:
- the list of LAOs the user is subscribed to (similar to Home page)
- when clicking an a LAO we get the list of roll calls the user either attended or was the organizer of, only closed roll calls are listed
- when clicking an a roll call we get the list of users who participated to the roll call, the public key is displayed in text as well as in QR code format

Main additions:
- ContentWalletFragment.java: displays the list of LAOs, belongs to HomeActivity
- LaoWalletFragment.java: displays the list of roll calls, belongs to LaoDetailActivity
- WalletListAdapter.java: used by LaoWalletFragment
- LaoDetailViewModel.java: the main addition is the LiveData containing the list of roll calls which is filtered so that we can display the appropriate one in the LaoWalletFragment
- AttendeesListFragment.java: displays the list of attendees, belongs to LaoDetailActivity
- AttendeesListAdapter.java: used by the AttendeesListFragment

In the subscribeToLao function of the LaoDetailViewModel I replaced the postValue() calls of mIsOrganizer to setValue() because isOrganizer() was sometimes returning null (for me in the attendedOrOrganized function). Gaurav suggested to me this "fix" because (from my understanding and correct me if I am wrong) using "post" on a LiveData schedules the task of setting the value to be executed on the main thread and it might not be executed immediately because it is queued. On the contrary using "set" directly sets the value but has the drawback of requiring the call to come from the main thread. So he told me for now we could assume it will always be called from the main and he will come back to me with updates.


fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/LaoDetailViewModel.java
```diff
@@ -460,6 +512,14 @@ public Lao getCurrentLaoValue() {
     return mLaoEvents;
   }
 
+  public LiveData<List<com.github.dedis.student20_pop.model.RollCall>> getLaoRollCalls() {
+    return mLaoRollCalls;
```
celinecamacho, line 516: Instead of having a new variable you can filter the list of events to extract only the roll calls

fe2-android/app/src/main/java/com/github/dedis/student20_pop/home/HomeActivity.java
```diff
@@ -296,10 +308,15 @@ private void setupSeedWalletFragment() {
     }
   }
 
-  private void openLaoDetails(String laoId) {
+  private void openLaoDetails(String laoId, boolean openLaoDetail) { //change to enum??
```
celinecamacho, line 311: You can have two different methods, `openLaoDetails` and `openContentWallet`, that call this method with the respective true or false. If you do that you'll need to generalise the name of this method, you can name it `openLaoDetailActivity` for example

fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/LaoDetailViewModel.java
```diff
@@ -106,6 +109,49 @@
...
+              mKeysetManager.getKeysetHandle().getPublicKeysetHandle();
+      isOrganizer =
+              lao.getOrganizer().equals(Keys.getEncodedKey(publicKeysetHandle));
+    } catch (GeneralSecurityException e) {
+      Log.d(TAG, "failed to get public keyset handle", e);
+      return false;
+    } catch (IOException e) {
+      Log.d(TAG, "failed to get public key", e);
+      return false;
+    }
```
celinecamacho, line 142: If the idea here is to know if the user attended or organised the roll calls of the current LAO then you can use the method `isOrganizer`

celinecamacho (commented):
Good job! I mostly left some comments about minor details to modify and one question about the method `attendedOrOrganized`

fe2-android/app/src/main/res/values/strings.xml
```diff
@@ -145,4 +146,8 @@
     <string name="text_seed_wallet">This is the only backup for your PoP tokens - store it securely</string>
     <string name="confirm_button">Confirm</string>
     <string name="content_title">Tokens</string>
+    <string name="wallet_empty">Your wallet is empty</string>
+    <string name="tokens_introduction">If you would like to receive tokens, please connect to a LAO and participate in a roll call.\n
+        If you would like to be an organizer, you can look at the attendees\' public tokens of the roll calls you organized.</string>
```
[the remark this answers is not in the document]
Kar1213: Yes basically if you want to organize events, maybe this is better:
"If you are an organizer, you can look at the attendees\' public tokens of the roll calls you organize"

fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/fragments/AttendeesListFragment.java
```diff
@@ -0,0 +1,80 @@
...
+import com.github.dedis.student20_pop.detail.LaoDetailActivity;
+import com.github.dedis.student20_pop.detail.LaoDetailViewModel;
+import com.github.dedis.student20_pop.detail.adapters.AttendeesListAdapter;
+import com.github.dedis.student20_pop.model.RollCall;
+
+import java.util.ArrayList;
+import java.util.Optional;
+
+public class AttendeesListFragment extends Fragment {
+    public static final String TAG = LaoWalletFragment.class.getSimpleName();
```
celinecamacho, line 24: This seems to be a typo

fe2-android/app/src/main/res/values/strings.xml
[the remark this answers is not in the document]
celinecamacho: Yes that is better

celinecamacho (approved):
Looks good 👍

PR #346, opened 2021-05-26
fe2-android: Subscribe to LAOs on Websocket connect event

gnarula:
Android frontend does reconnect to the server automatically - however it does not re-subscribe to existing channels.

This PR adds changes to `LAORepository` which listens to Websocket events and sends a subscribe request to existing channels on websocket connect event.


Walfar (approved):
Tested on my phone, and seems to work well now ! Roll call can now be opened/closed, even after putting the app in background.

Walfar (approved):

PR #359, opened 2021-06-02
Add test for Wallet UI & cross-validation with Fe1-web

filippo888:
## Content:
In this PR, I made a small refactor on the wallet class. In particular I refactor the method  `private String convertStringToPath(String path)`, now the method is more readable. Secondly I write the cross-validation test with the fe1-web and after resolve some issue with Carlo now the two Wallet (fe1-web and fe2-android) work exactly the same. In particular if we init a wallet with the same 12 word and we generate a token with the same LaoId and RollCallId the two wallet generate the same sequence of byte for the token. Finally I added a general test for the Wallet UI whereI tested if all element of the UI are really display and their content are corrects.


fe2-android/app/src/androidTest/java/com/github/dedis/student20_pop/home/WalletFragmentTest.java
```diff
@@ -0,0 +1,262 @@
...
+
+@LargeTest
+@RunWith(AndroidJUnit4ClassRunner.class)
+public class WalletFragmentTest {
+
+  @Rule
+  public ActivityScenarioRule<HomeActivity> mActivityTestRule = new ActivityScenarioRule<>(HomeActivity.class);
+
+  @Test
+  public void walletFragmentTest() {
```
celinecamacho, line 40: You can create a method for each check you make, this way it is easier to know what works and what doesn't

celinecamacho (approved):

PR #363, opened 2021-06-03
Fix Hash encoding in Fe1-Web

pierluca:


fe1-web/model/objects/Base64Url.ts
```diff
@@ -8,8 +8,8 @@ export class Base64UrlData extends String {
     base64url.decode(value);
   }
 
-  public static encode(text: string | Buffer) : Base64UrlData {
-    return new Base64UrlData(this.addPadding(base64url.encode(text)));
+  public static encode(text: string | Buffer, encoding: string = 'utf8') : Base64UrlData {
+    return new Base64UrlData(this.addPadding(base64url.encode(text, encoding)));
```
pierluca, line 12: the `base64url.encode` by default assumes that the string is utf8 encoded.
It's perfect if you're hashing a string, not so much a binary array.

fe1-web/model/objects/Hash.ts
```diff
@@ -37,7 +37,7 @@ export class Hash extends Base64UrlData {
     const bString = hash.update(data).array();
     const str = String.fromCharCode(...bString);
 
-    return Base64UrlData.encode(str);
+    return Base64UrlData.encode(str, 'binary');
```
pierluca, line 40: and this is the actual fix ...

gnarula (dismissed):
LGTM

PR #368, opened 2021-06-06
Election end

Tesa27:


gnarula:
@Tesa27 can you resolve the conflict and run `go fmt ./...` please? Also, is it okay if I review this after #337? You might have to make some changes because of a new `inbox` type...

Tesa27:
Replaced by #399

PR #371, opened 2021-06-07
fixed the RC id

Adalsteinnjons:
It was using the lao creation in the id but was supposed to use the lao creation. 

image


CaiMusso (approved):
This totally makes sense. It shouldn't refer to the LAO here, thanks Adi

PR #372, opened 2021-06-07
Witness UI

maximelaval:
Organizer is now able to add new witness to a Lao by scanning their QR code with the add button in  show/hide properties.
Implemented a new fragments for witness messages where witness can sign messages manually.
Refactored the class QrCodeScanningFragment and created a new enum ScanningAction so that we know what type of action we want to do when scanning a QR code. 
Implemented the methods IsWitness(), IsSignedByCurrentWitness() , updateLaoWitnesses() and signMessage() in viewModel.
Created a class WitnessMessage in model to represent a message needing to be signed. 
Renamed the class WitnessMessage in com.github.dedis.student20_pop.model.network.method.message.data.message to WitnessMessageSignature to avoid confusions with the model class.
The message signing works with the backend , we receive a broadcast for the WitnessMessageSignature which update graphically the state of the message to " Signed ".


fe2-android/app/src/main/java/com/github/dedis/student20_pop/model/WitnessMessage.java
```diff
@@ -0,0 +1,76 @@
...
+    public WitnessMessage(String messageId) {
+        witnesses = Collections.emptySet();
+        this.messageId = messageId;
+    }
+
+    /**
+     * Method that checks whether the current message has been signed by the Witness public key provided
+     *
+     * @param pk public key of the witness
+     */
```
celinecamacho, line 39: Nice doc but where's the method? 😅

fe2-android/app/src/main/java/com/github/dedis/student20_pop/model/data/LAORepository.java
```diff
@@ -55,504 +53,593 @@
...
+    /**
+     * Helper method to verify signature of a message
+     *
+     * @param messageId    Base 64 URL encode Id of the message to sign
+     * @param senderPkBuf  Base 64 decoded public key of the signer
+     * @param signatureBuf Base 64 URL decoded signature of the signer
+     * @return false if there was a problem signing the message
+     */
+
+    private boolean verifySignature(String messageId, byte[] senderPkBuf, byte[] signatureBuf) {
```
celinecamacho, line 548: Same as the `generateSignature` method, these are utility methods to sign and verify that correspond in a separate class in the package utility/security

fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/LaoDetailViewModel.java
```diff
@@ -496,6 +575,45 @@ public void setCurrentLaoName(String laoName) {
...
     }
 
+    public LiveData<Boolean> isWitness() {
+        try {
+            KeysetHandle publicKeysetHandle =
+                    mKeysetManager.getKeysetHandle().getPublicKeysetHandle();
+            boolean isWitness =
+                    getCurrentLaoValue().getWitnesses().contains(Keys.getEncodedKey(publicKeysetHandle));
+            Log.d(TAG, "isWitness: " + isWitness);
+            mIsWitness.setValue(isWitness);
```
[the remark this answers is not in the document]
maximelaval: The thing is that the isOrganizer will never change once the lao is already created , but for the witness suppose we scan a new witness after the lao has already been subscribed to then isWitness() will return false every time . Maybe a more compliant solution would be to update mIsWitness when we get the result of updateLaoWitnesses() ?

fe2-android/app/src/main/java/com/github/dedis/student20_pop/model/WitnessMessage.java
maximelaval: Yeah in the end this method was deleted just forgot to delete the doc .

fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/LaoDetailViewModel.java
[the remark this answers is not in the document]
celinecamacho: Yes that sounds like a good solution

fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/LaoDetailViewModel.java
```diff
@@ -685,6 +827,64 @@ public void updateLaoName() {
...
+                                                    .subscribe(
+                                                            answer2 -> {
+                                                                if (answer2 instanceof Result) {
+                                                                    Log.d(TAG, "updated lao state with new witnesses");
+                                                                } else {
+                                                                    Log.d(TAG, "failed to update lao state with new witnesses");
+                                                                }
+                                                            },
+                                                            throwable -> Log.d(TAG, "timed out waiting for result on update lao witnesses", throwable)
+                                                    );
```
celinecamacho, line 875: When updating the name and the witness list of a LAO a `StateLao` message is sent, creating a method for this avoids the code duplication in `updateLaoWitnesses` and `updateLaoName`

fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/LaoDetailViewModel.java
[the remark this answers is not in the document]
maximelaval: Actually it wouldn't work because even in updateLaoWitnesses the new witness list  is only being updated in handleStateLao of LaoRepository after the broadcast has been received . And because we can't have access to laoDetailViewModel.setWitness() in the LaoRepository I don't really see where we could set the witnesses elsewhere than here .

celinecamacho (dismissed):

celinecamacho (commented):
Did you verify that the QR code scanning works well for the connect and attendee scan after these changes?

celinecamacho (approved):

PR #366, opened 2021-06-04
Centralise id generation and recheck protocol specifications

celinecamacho:
- Creation of an id generator class used for the message/data classes.
- Recheck carefully the protocol specifications and fixed some id inconsistencies.

PR #375, opened 2021-06-08
Fix LAO connection problems in fe2 reported by fe1

Kar1213:
- remove /root/ prefix in LAO QR code
- if proposed_start<creation_time then proposed_start is automatically set to creation_time
- the message_id of MessageGeneral was applying base64url encoding on the hash output twice
- remove the 'L' in the hash for the createLao (this is fixed by PR #366 but is not yet on master)

Adi and I tested that on fe1 we can connect to a LAO created by fe2 and create rollcalls. This PR contains only changes regarding the fe2 side.


celinecamacho (approved):
Looks good to me 👍

PR #376, opened 2021-06-08
election setup saved on wrong channel

Tesa27:


flmichel (commented):
I think the message should also be saved in the election channel: "A duplicate of this election creation message should be sent to this channel to self-contain the entire election history." https://docs.google.com/document/d/1wJK0L_oO4dtLIlXGh7D8p9_Gc4_k-_zz7BslVqMUo2E. Is this done somewhere else in the code?

Tesa27:
> I think the message should also be saved in the election channel: "A duplicate of this election creation message should be sent to this channel to self-contain the entire election history." https://docs.google.com/document/d/1wJK0L_oO4dtLIlXGh7D8p9_Gc4_k-_zz7BslVqMUo2E. Is this done somewhere else in the code?

Nope but was definitely going to ask today since I wasn't sure thanks for pointing this out will save it on the election channel too then!

be1-go/hub/election.go
```diff
@@ -91,6 +91,10 @@ func (c *laoChannel) createElection(msg message.Message) error {
 
 	// Add the SetupElection message to the new election channel
```
flmichel, line 92: Maybe this comment should be updated

Tesa27: done

flmichel (approved):

PR #377, opened 2021-06-08
check only attendees can cast votes

Tesa27:


be1-go/hub/election.go
```diff
@@ -23,6 +23,9 @@ type electionChannel struct {
 	// Questions asked to the participants
 	//the key will be the string representation of the id of type byte[]
 	questions map[string]question
+
+	// attendees that took part in the roll call string of their PK
+	attendees []string
```
flmichel, line 28: I think you can use a map here as it is done in the LaoChannel: `attendees map[string]struct{}`. It will make the check easier.

be1-go/hub/election.go
```diff
@@ -87,13 +90,14 @@ func (c *laoChannel) createElection(msg message.Message) error {
...
 	}
 
 	// Add the SetupElection message to the new election channel
 	messageID := base64.URLEncoding.EncodeToString(msg.MessageID)
-	electionCh.inboxMu.Lock()
-	electionCh.inbox[messageID] = msg
-	electionCh.inboxMu.Unlock()
+	c.inboxMu.Lock()
+	c.inbox[messageID] = msg
+	c.inboxMu.Unlock()
```
flmichel, line 100: Maybe those change can be removed since it is already in #376.

be1-go/hub/election.go
```diff
@@ -157,6 +161,20 @@ func (c *electionChannel) castVoteHelper(publish message.Publish) error {
 		}
 	}
 
+	senderPK := msg.Sender.String()
```
flmichel, line 164: In #370 I updated the encoding of the attendees' key that are stored on the server. The public keys are store in Base64URL in that PR. So we will need to change this accordingly later.

be1-go/hub/election.go
```diff
@@ -157,6 +161,20 @@ func (c *electionChannel) castVoteHelper(publish message.Publish) error {
...
+	ok = false
+	for _, attendee := range c.attendees {
+		if attendee == senderPK {
+			ok = true
+		}
+	}
```
flmichel, line 170: If `attendees` is a map you can simply do `ok := c.attendees[senderPK]`

be1-go/hub/election.go
Tesa27: done

be1-go/hub/election.go
Tesa27: done

be1-go/hub/election.go
Tesa27: done

be1-go/hub/election.go
flmichel: Yes exactly

be1-go/hub/election.go
```diff
@@ -87,6 +90,7 @@ func (c *laoChannel) createElection(msg message.Message) error {
 		data.EndTime,
 		false,
 		getAllQuestionsForElectionChannel(data.Questions),
+		getAllAttendees(c.attendees),
```
flmichel, line 93: Maybe you can just pass `c.attendees` to the struct and there is no need for the function `getAllAttendees`

be1-go/hub/election.go
```diff
@@ -87,6 +90,7 @@ func (c *laoChannel) createElection(msg message.Message) error {
 		data.EndTime,
 		false,
 		getAllQuestionsForElectionChannel(data.Questions),
+		c.attendees,
```
gnarula, line 93: Are the attendees for an election supposed to change over time with every roll call or should they remain the same as they were at the time the election was created?

Tesa27: I think the attendees should remain the same on election creation time, it would be like creating an election for the currently present participants and if someone is part of a roll call after the election has been created it should not be considered as attendee for now

gnarula: I see - maps are passed by reference in Go, so your election's attendees change every time `c.attendees` changes (which it does after every roll call).

I'd suggest creating a custom data type with a Copy method. 

```go
type Attendees struct {
    sync.Mutex,
    store map[string]struct{}
}

func NewAttendees() Attendees {

}

func (a Attendees) IsPresent(key string) ok {

}

func (a Attendees) Add(key string) {

}

func (a Attendees) Copy() Attendees {
     a.Lock()
     defer a.Unlock()

     copy := NewAttendees()

     for key := range a {
         copy.store[key] = struct{}
     }

     return copy
}

Comment (Tesa27) [reply be1-go/hub/election.go:93]: But I even think that case is acceptable too, arriving at the role call late shouldn't penalize you from voting right? Both scenarios seem actually to make sense, so I am not sure which is better

Comment (gnarula) [reply be1-go/hub/election.go:93]: Seems like a question for the evoting channel on slack ;)

File: be1-go/hub/election.go [hunk 7]
```diff
@@ -156,6 +194,15 @@ func (c *electionChannel) castVoteHelper(publish message.Publish) error {
 			Description: fmt.Sprintf("Vote cast too late, vote casted at %v and election ended at %v", voteData.CreatedAt, c.end),
 		}
 	}
+	senderPK := base64.URLEncoding.EncodeToString(msg.Sender)
+	senderPoint := student20_pop.Suite.Point()
```
gnarula, line 198:
```suggestion
	senderPK := base64.URLEncoding.EncodeToString(msg.Sender)

	senderPoint := student20_pop.Suite.Point()
	err := senderPoint.UnmarshalBinary(msg.Sender)
	if err != nil {
		return &message.Error{
			Code:        -4,
			Description: "Invalid sender public key",
		}
	}

```

gnarula (changes requested):
I reckon you need to unmarshal the sender's PK to a point

be1-go/hub/election.go
Tesa27: Yes this makes sense totally

gnarula (approved):

PR #379, opened 2021-06-08
Manage Election UI

maximelaval:
Finished the fragment for Manage Election . In the future will have to send an message like UpdateElection in order to effectively modify some attributes of the election.


fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/fragments/ManageElectionFragment.java
```diff
@@ -75,43 +113,229 @@ public View onCreateView(
...
+        // On click, edit start time button
+        editStartTimeButton.setOnClickListener(
                 v -> {
-                   // TODO implement the cancel election action with backend when the backend is ready
-                    laoDetailViewModel.terminateCurrentElection();
-                    laoDetailViewModel.openLaoDetail();
+                    // we set the request code
+                    requestCode = START_TIME_CODE;
+                    // show the timePicker
+                    timePickerDialog.show();
```
celinecamacho, line 299: These four lines are used multiple times, might be a good idea to refactor and create a method that takes the requestCode as an argument and then implements those lines

fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/fragments/ManageElectionFragment.java
```diff
@@ -75,43 +113,229 @@ public View onCreateView(
...
+                        case CANCEL_CODE: {
+                            // TODO : In the future send a UpdateElection message with a modified end time as the current time
+                            laoDetailViewModel.openLaoDetail();
+                            break;
+                        }
+
+
+                        case START_TIME_CODE: {
+                            Calendar startTimeCalendar = Calendar.getInstance();
+                            startTimeCalendar.setTimeInMillis(laoDetailViewModel.getCurrentElection().getStartTimestamp());
```
celinecamacho, line 134: Why milliseconds and not seconds?

fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/fragments/ManageElectionFragment.java
```diff
@@ -75,43 +113,229 @@ public View onCreateView(
...
+        editEndDateButton.setOnClickListener(
+                v -> {
+                    // we set the request code
+                    requestCode = END_DATE_CODE;
+                    // show the timePicker
+                    datePickerDialog.show();
+                });
 
-    }
-    public void setupHomeButton() {
```
celinecamacho, line 104: Why is the access to home from this fragment deleted?

celinecamacho (commented):
There are some sonar cloud code smells and a bug that should not take much time to solve. Other than that I would suggest creating methods for the different implementations in `onActivityCreated`, I left some comments about this

fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/fragments/ManageElectionFragment.java
maximelaval: because there is only a method setTimeInMillis. To fix this I did *1000 to the timestamp .

fe2-android/app/src/main/java/com/github/dedis/student20_pop/detail/fragments/ManageElectionFragment.java
maximelaval: This setupHomeButton was a mistake as this button is already set in the LaoDetailActivity

celinecamacho (approved):

PR #381, opened 2021-06-08
test witness and servers

lpatel99:


be1-go/hub/witness_test.go
```diff
@@ -0,0 +1,72 @@
...
+	return h
+}
+
+// to complete
+func TestNewWitnessHub(t *testing.T) {
+	pk := "invalid pk"
+	_, err := base64.URLEncoding.DecodeString(pk)
+	if err == nil {
+		t.Errorf("decoded invalid string")
+	}
```
flmichel, line 38: For the other tests we are using the `require` package. But maybe we can let it like this.

be1-go/hub/witness_test.go
```diff
@@ -0,0 +1,72 @@
...
+	"context"
+	"encoding/base64"
+	"log"
+	"strings"
+	"student20_pop"
+	"sync"
+	"testing"
+)
+
+func CreateWitnessHub() Hub {
```
flmichel, line 14: Maybe you could return an error in case there the function fails to create a wintess hub: ` func CreateWitnessHub() (Hub, error)` but that is a detail

flmichel (approved):
Looks good to me. I just added few comments but feel free to merge the PR since you might not have the time to do too much changes.

be1-go/hub/witness_test.go
lpatel99: a good idea, but when used the test checks that the witness is not nil and returns an error if it is

PR #382, opened 2021-06-08
Wallet UI modification and extension

Kar1213:
- added fragment to display tokens
- added wallet logout to delete seed
- moved the requirement of setting up a wallet to the moment where users want to attend rollcalls
- added isSetup attribute to Wallet class to know whether the seed has been initialized

For rollcalls we can now close the rollcall when clicking on Home or Identity during scanning and a confirmation popup shows up when closing rollcalls.

I will later need to adapt the wallet test (now commented) for the new implementation.


fe2-android/app/src/main/java/com/github/dedis/student20_pop/qrcode/QRCodeScanningFragment.java
```diff
@@ -146,13 +138,37 @@ private void startCamera() throws SecurityException {
...
+    ((LaoDetailViewModel)mQRCodeScanningViewModel)
+            .getAskCloseRollCallEvent()
+            .observe(
+                    this,
+                    integerEvent -> {
+                      Integer nextFragment = integerEvent.getContentIfNotHandled();
+                      if (nextFragment != null) {
+                        setupClickCloseListener(nextFragment);
+                      }
+                    });
```
celinecamacho, line 156: This doesn't really correspond to the setup of the close roll call button, maybe you can create another method for it and call it on `onCreateView`

celinecamacho (commented):
Great job! There is just one small detail that I would suggest to change

celinecamacho (approved):

PR #384, opened 2021-06-11
Disabled html escaping, converted election setup's creation time to seconds

Walfar:
- Disabling html escaping prevents the "=" character from being encoded into "\u003d", which causes a parsing failure 
- the variable for the date of creation in ElectionSetup was in milliseconds, which was inconsistent with the rest of the timestamps that are in seconds


celinecamacho (approved):
Looks good 👍

PR #385, opened 2021-06-12
Cleaning and fix double click for dialog

Kar1213:
- changed name of app to "Student 21 PoP"
- changed the persistentId in RollCall to final
- use persistentId to display token in RollTokenFragment
- bug: when double clicking on a button that shows a dialog it might show up twice and could crash the app if we click on confirm the first time and click on one of the buttons on the second one. The problem was always appearing for the close roll-call dialog and very rarely to the others. It probably depends on how much time the code for setting the builder takes. Is this a clean solution ? If it is I could do something similar for the others.


celinecamacho (commented):
Everything seems good, you just have one unused variable you should delete

celinecamacho (approved):

PR #387, opened 2021-06-15
Fix Message Id

maximelaval:
Issue : the message Id was never Base 64 encoded so when we transformed it into a string it gave some weird UFT8 character. 
AF30E74C-1F29-45A6-9E18-87E07883D1FD


gnarula (changes requested):
Looking at the other `byte[]` in the class I think none of them are Base64 representations. The conversion happens in the getters instead. Could you please change `getMessageId()` to use `Base64.getUrlEncoder().encodeToString()` instead and revert this?

gnarula (approved):

PR #388, opened 2021-06-15
Remove schemas in "maybe" folder

gnarula:
This interferes with schema validation in Go because the empty JSON files throw an EOF.


pierluca (approved):

gnarula:
Merging this since the CI failures on OSX are totally unrelated

PR #389, opened 2021-06-15
Fix JsonMessageGeneralSerializer message_id issue

celinecamacho:
The issue from PR #387.

- MessageGeneral: `messageId` is not encoded twice on `getMessageId` 
- JsonMessageGeneralSerializer: `message_id` is not decoded, because the message id is kept Base64URL encoded in MessageGeneral

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/student20_pop/model/network/method/message/MessageGeneral.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/student20_pop/model/network/method/message/MessageGeneral.java
@@ -1,6 +1,5 @@
 package com.github.dedis.student20_pop.model.network.method.message;
 
-import java.util.Base64;
 import android.util.Log;
 
 import com.github.dedis.student20_pop.model.network.method.message.data.Data;
@@ -11,8 +10,10 @@ import com.google.crypto.tink.PublicKeyVerify;
 import com.google.crypto.tink.subtle.Ed25519Verify;
 import com.google.gson.Gson;
 
+import java.nio.charset.StandardCharsets;
 import java.security.GeneralSecurityException;
 import java.util.ArrayList;
+import java.util.Base64;
 import java.util.List;
 
 /**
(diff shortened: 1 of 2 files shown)


Kar1213 (dismissed):
Nice catch ! It seems to solve the problem, thank you.

fe2-android/app/src/test/java/com/github/dedis/student20_pop/model/network/method/message/MessageGeneralTest.java
```diff
@@ -1,81 +1,40 @@
...
+    private final byte[] signature = new byte[] {-55, 121, 102, -88, -17, 87, -125, -87, 93, -127, -56, -125, -14, -61, 56, 117, 35, -113, -82, 57, 107, 85, 100, -30, 43, 69, 22, 42, -25, 66, -70, -64, 20, -31, -32, -112, -78, 115, 9, 13, -37, 59, -29, 45, 12, 54, 71, -73, 119, 89, 119, 106, 24, -115, 67, 103, -91, 29, -122, 83, -38, 101, 101, 11};
+    private final CreateLao data = new CreateLao(IdGenerator.generateLaoId(organizer, creation, name), name, creation, organizer, new ArrayList<>());
+    private final MessageGeneral messageGeneral = new MessageGeneral(sender, dataBuf, data, signature, messageId, new ArrayList<>());
+
+    @Test
+    public void messageGeneralWithNonBase64URLIdTest() {
+        assertThrows(IllegalArgumentException.class, () -> new MessageGeneral(sender, dataBuf, data, signature, new byte[] {1, 2, 3}, new ArrayList<>()));
```
Kar1213, line 32: Is the exception expected to be thrown by the Log call in the constructor? If yes it would mean that we are testing the behavior of something which is not part of the actual algorithm and for example if the Log will be removed later the test would fail. Maybe we could do the same thing but put it outside of the Log (and the output would be unused). What do you think?

celinecamacho: I can change it into two steps, but I don't think this log will be deleted because it seems helpful for debugging

Kar1213 (approved):
Everything looks good for me

PR #391, opened 2021-06-15
Fix election selection bug

### LINKED ISSUE #390: Election display always selects latest election whichever button we click on
If I setup two elections : election 1 and election 2, in that order, then when pressing on cast votes - or manage - on election 1 it opens election 2

Warhedgehog:
- Before the current election in the view model was not set in the button listener but outside. It means that when a button was pressed the current election was always the last one created because it is the last one getChildView in EventExpandableListViewAdapter calls.
- fixes  #390


Walfar (approved):
Looks good to me ! Nice fix for a last minute bug haha

PR #392, opened 2021-06-16
checks if organizer sends the election creation message

Tesa27:
We should only let the organizer create an election any other sender should be denied when sending the election setup message.


flmichel:
Maybe it could be good to also add the same check for the creation of a lao

gnarula (approved):

PR #394, opened 2021-06-16
Multiple questions in election setup

Warhedgehog:
Implemented the possibility to set elections with multiple questions, tested with casting votes, and election with with organizer and an attendee with webfrontend. All working features of master seemed to be working.

- Adapted setup fragment
- Implemented Adapter for viewPager 
- Adapted the existing tests

setup


Walfar (approved):
All fine to me,  looked good when testing

Walfar (approved):

PR #396, opened 2021-06-17
Update dataElectionResult.json

Adalsteinnjons:
This protocol needs an update according to the e-voting specifications on google drive.
This is needed for the verification of the election result on the front-end.


CaiMusso (dismissed):
Good fix

protocol/query/method/message/data/dataElectionResult.json
```diff
@@ -19,10 +19,10 @@
 						"$comment": "ID of the question"
 					},
 					"result": {
-						"description": "[Array[String]] results",
+						"description": "[Array[object]] results",
 						"type": "array",
 						"items": {
-							"type": "string",
+							"type": "object",
```
pierluca, line 25: you might want to clarify the format of that object (i.e. which properties, etc.)

Adalsteinnjons: Done

gnarula:
@pierluca I think a test in be1-go/network is flakey but that can be taken up as a separate PR

gnarula (approved):

PR #397, opened 2021-06-17
Election Result UI

Warhedgehog:
Implemented UI for the election results

- Implemented fragment and its layout
- Implemented adapter for ViewPager and its layout
- Implemented adapter for ListView in ViewPager and its layout

Is only mergeable once the PR#335 about election result protocol is merged


Walfar (approved):
All good, works fine when tested on emulator and phone

PR #400, opened 2021-06-20
Fragment Tests

maximelaval:
Added the tests to test the fragment with Espresso library .
The tests to launch a LAO and Setup an Election work well only when run with the backend .
In the future we should mock the server so that the tests work on their own.


pierluca:
@celinecamacho what do you think, should this be closed or fixed for merge ?

pierluca:
Based on the current state of this PR, it is being closed. 
The branch will be kept for future reference (if need be).

PR #401, opened 2021-07-22
be1-go: Add documentation

gnarula:


be1-go/hub/mod.go
```diff
@@ -0,0 +1,6 @@
+// Package hub defines a Hub that is used for processing incoming messages
+// and replying back to the client.
+//
+// A concrete instance of a Hub may be an Organizer or a Witness. A baseHub
+// type contains the implementation common across both types.
+package hub
```
pierluca, line 6: The package also defines the sockets , would it make sense to cover all of that here?

be1-go/hub/inbox.go
```diff
@@ -9,11 +9,13 @@ import (
 	"time"
 )
 
+// inbox represnts an in-memory data store to record incoming messages.
```
pierluca, line 12:
```suggestion
// inbox represents an in-memory data store to record incoming messages.
```

be1-go/README.md
```diff
@@ -3,7 +3,8 @@ Proof-of-personhood, Spring 2021. Go back-end version.
 
 ## Overview
 
-You may build the `pop` CLI to interact with the server by executing `go build -o pop ./cli/`.
+You may build the `pop` CLI to interact with the server by executing
+`make build` on Linux/OSX or `make.bat build` in Windows.
```
pierluca, line 7: at this point, it will fail for a new student who's discovering this repo for the first time.
requirements and all :)

be1-go/docs/README.md
```diff
@@ -0,0 +1,164 @@
...
+Please ensure you have Go >= 1.16 installed on your machine. Instructions for
+doing so are available [here](https://golang.org/doc/install).
+
+Linux/OSX users also need to install GNU Make. OSX users may install it
+using homebrew. Linux users may do so using their package manager:
+
+```bash
+brew install make # OSX
+sudo apt-get install build-essential # Ubuntu/Debian
+```
```
pierluca, line 24: I'd consider moving this to the main readme

be1-go/docs/README.md
```diff
@@ -0,0 +1,164 @@
...
+└── validation          # module to validate incoming/outgoing messages
+```
+
+Depending on which component you're working on, the entry point would
+either be cli/organizer or cli/witness, with bulk of the implementation
+logic in the hub module.
+
+#### Architecture
+
+The PoP Go backend expects actors to establish long lived websocket connections
```
pierluca, line 71: `actors` might be a confusing (or overloaded) term. How about clients ? attendees ?

be1-go/docs/README.md
```diff
@@ -0,0 +1,164 @@
...
+* `ClientSocket`: Used to denote a connection to a user participating in a PoP Party
+* `WitnessSocket`: Used to represent a connection to a witness server.
+* `OrganizerSocket`: Used to represent a connection to the organizer server.
+
+The `ReadPump` and `WritePump` are low-level methods which allow
+reading/writing data over the wire. Most users would instead use the `Send(msg []byte)`,
+`SendError(id int, err error)` and `SendResult(id int, res message.Result)` APIs
+
+Each incoming message read by `ReadPump` is passed to the `Hub interface`'s
+(refer hub/hub.go) `Recv` method for processing
```
pierluca, line 92:
```suggestion
(refer hub/hub.go) `Recv` method for processing.
```

be1-go/docs/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+Depending on the type of end user, a `Socket` has three concrete implementations:
+
+* `ClientSocket`: Used to denote a connection to a user participating in a PoP Party
+* `WitnessSocket`: Used to represent a connection to a witness server.
+* `OrganizerSocket`: Used to represent a connection to the organizer server.
+
+The `ReadPump` and `WritePump` are low-level methods which allow
+reading/writing data over the wire. Most users would instead use the `Send(msg []byte)`,
+`SendError(id int, err error)` and `SendResult(id int, res message.Result)` APIs
```
pierluca, line 89:
```suggestion
`SendError(id int, err error)` and `SendResult(id int, res message.Result)` APIs.
```

be1-go/docs/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+Each incoming message read by `ReadPump` is passed to the `Hub interface`'s
+(refer hub/hub.go) `Recv` method for processing
+
+The `Hub` interface has two concrete implementations - one for the organizer
```
pierluca, line 94: At this point you haven't yet described what a `Hub` is.

be1-go/docs/README.md
```diff
@@ -0,0 +1,164 @@
...
+and another for the witness. Since both share common implementations they embed
+the `baseHub` implementation with a few methods "overriden" where custom
+implementation is needed.
+
+The `baseHub` on receiving a message, processed it by invoking the
```
pierluca, line 99:
```suggestion
The `baseHub` on receiving a message, processes it by invoking the
```

be1-go/docs/README.md
```diff
@@ -0,0 +1,164 @@
...
+├── test                # sample test data
+└── validation          # module to validate incoming/outgoing messages
+```
+
+Depending on which component you're working on, the entry point would
+either be cli/organizer or cli/witness, with bulk of the implementation
+logic in the hub module.
+
+#### Architecture
+
```
pierluca, line 70: The notion of channel hasn't been introduced at all, though it's used.
Speaking of the general architecture, it's worth to mention (even if it's mentioned elsewhere) that this is a server that needs to support a publish-subscribe pattern over websockets, and that channels are channels in a pub/sub sense.

pierluca (commented):
I think the documentation overalls is getting better as reference documentation, I understand it well because I'm familiar with the project, even if I'm not familiar with the Go backend specifics.

However, I don't think it's fit-for-purpose yet as introduction documentation for people who are not familiar with the project. It might be worth duplicating some information so that the documentation is somewhat self-sufficient in understanding what the backend is trying to achieve. Otherwise, we might end up confusing the students more than anything.

Don't get me wrong, I believe a lot is already here. What it lacks is introduction / context in most places, and depth in some places.

be1-go/docs/README.md
```diff
@@ -68,13 +53,35 @@ logic in the hub module.
...
+using a publish/subscribe pattern.
 
 <div align="center">
   <img alt="Communication Stack" src="images/comm_stack.jpeg" width="600" />
 </div>
 
+On a higher level of abstraction, a client may publish messages or subscribe
+for messages on a *channel*. You may think of a channel as a topic which
+stores events that occur over time. For instance, every LAO is denoted by a unique
+channel (also called the root channel) and contains messages about all events
```
pierluca, line 68: isn't the root channel the one named `/root` , as opposed to LAO channels `/root/{lao-id}` ?

pierluca (approved):
Looks like a great improvement to me !

What do you think would be a meaningful next step, if we're to put in writing the introduction you would give the students at the beginning of the semester?

be1-go/docs/README.md
gnarula: Nice catch! Updated

PR #402, opened 2021-08-06
Refactor Hub and Socket

gnarula:
Moves `socket` to its own package and removes checks on logs in tests.

PR #405, opened 2021-08-10
Ignore directory in fs.WalkDir

### LINKED ISSUE #404: be1-go: validation can include unwanted elements
https://github.com/dedis/student_21_pop/blob/b07f7246a685881dea3770687020992a7dcf2eff/be1-go/validation/schema_validator.go#L95

Just checking if the extension is .json is not enough, as it would also include any folder with such extension.

gnarula:
Fixes #404


nkcr (approved):

gnarula:
@nkcr tests are flakey on Github Actions and #402 should fix them 🤞

PR #406, opened 2021-08-13
Adds the ability to store/restore the organiser state on a DB. 

nkcr:
By providing a HUB_DB env variable the organizer will try to load its state and save it on an SQLite db.

Right now the code is not decoupled and there is no use of a storage abstraction. I leave it as is for now so as to not bring too much changes at once.

It also changes the internal use of the channelID to always include the path, ie. with /root/... This is to avoid confusion.

Depends on #402.


be1-go/hub/inbox.go
```diff
@@ -75,3 +114,73 @@ func (i *inbox) getMessage(messageID []byte) (*message.Message, bool) {
...
+	stmt, err := db.Prepare(query)
+	if err != nil {
+		return xerrors.Errorf("failed to prepare query: %v", err)
+	}
+
+	defer stmt.Close()
+
+	msg := messageInfo.message
+
+	_, err = stmt.Exec(msg.MessageID, msg.Sender, msg.Signature, msg.RawData, messageInfo.storedTime, i.channelID)
```
gnarula, line 146: I'm not sure if the underlying []byte will be converted to base64url encoded string representations here automatically. Did you try to observe the rows here in a viewer?

be1-go/hub/inbox.go
```diff
@@ -75,3 +114,73 @@ func (i *inbox) getMessage(messageID []byte) (*message.Message, bool) {
...
+	if err != nil {
+		return xerrors.Errorf("failed to exec query: %v", err)
+	}
+
+	for _, pubKeySigPair := range messageInfo.message.WitnessSignatures {
+		err = addWitnessInDB(db, string(messageInfo.message.MessageID),
```
gnarula, line 152: Similarly here and all other places where we save/read these values

be1-go/hub/inbox.go
```diff
@@ -75,3 +114,73 @@ func (i *inbox) getMessage(messageID []byte) (*message.Message, bool) {
 	}
 	return msgInfo.message, true
 }
+
+func (i *inbox) storeMessageInDB(messageInfo *messageInfo) error {
+	db, err := sql.Open("sqlite3", os.Getenv("HUB_DB"))
```
gnarula, line 119: This will eventually be abstracted away but it would be nice if we can have a long living *sql.DB instance and pass that everywhere for the life of the hub.

gnarula (changes requested):

be1-go/hub/inbox.go
nkcr: MessageID, Sender, and Signature are saved as binary. I tried saving/loading and it shows that it can successfully restore back the data to the correct format.

be1-go/hub/inbox.go
gnarula: Conclusion from slack: to be taken up separately when dealing with Storage abstraction

gnarula (approved):
Looks good!

PR #411, opened 2021-08-23
be2 scala description

RaulinN:


be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
```diff
@@ -1,44 +1,89 @@
...
 
   def getInstance: AskableActorRef = INSTANCE
 
   def getTimeout: Timeout = TIMEOUT
 
   def getDuration: FiniteDuration = DURATION
+
+  def apply(): DbActor = DbActor()
+
+  sealed case class DbActor() extends Actor with ActorLogging {
```
pierluca, line 58: @nkcr created a relational database schema in be1 to store all the data.
Would it be a useful inspiration? I'm available to discuss

PR #412, opened 2021-08-24
Wrap errors using %w specifier

gnarula:
Cleaned up the way errors propagate up the stack. All leaf
methods on hub/channels should return a *message.Error and
intermediate methods should wrap them using xerrors.Errorf


nkcr (approved):
Would be nice to update the documentation with when to use message.Error or wrap them. Otherwise LGTM 🚀 .

nkcr (approved):

PR #413, opened 2021-08-28
Test LAORepository

celinecamacho:
Use Mockito library to test the class LAORepository. This is an example of how to mock the needed classes for unit tests.


fe2-android/app/src/test/java/com/github/dedis/student20_pop/model/data/LAORepositoryTest.java
```diff
@@ -0,0 +1,156 @@
...
+    Mockito.verify(remoteDataSource, Mockito.times(1)).sendMessage(Mockito.any());
+
+    TestObserver<Answer> testObserverCatchup = TestObserver.create();
+
+    answerCatchup.subscribe(testObserverCatchup);
+
+    testScheduler.advanceTimeBy(950, TimeUnit.MILLISECONDS);
+    testObserverCatchup.assertNotComplete();
+
+    testScheduler.advanceTimeBy(100, TimeUnit.MILLISECONDS);
```
pierluca, line 80: assuming we have a `RESPONSE_DELAY` constant set at 1000 ms

```suggestion
    testScheduler.advanceTimeTo(RESPONSE_DELAY-1, TimeUnit.MILLISECONDS);
    testObserverCatchup.assertNotComplete();

    testScheduler.advanceTimeTo(RESPONSE_DELAY+1, TimeUnit.MILLISECONDS);
```

fe2-android/app/src/main/java/com/github/dedis/student20_pop/model/data/LAORepository.java
```diff
@@ -61,119 +62,119 @@
...
-                if (INSTANCE == null) {
-                    INSTANCE = new LAORepository(laoRemoteDataSource, localDataSource, keysetManager, gson);
-                }
-            }
+          INSTANCE = new LAORepository(laoRemoteDataSource, localDataSource, keysetManager, gson,
+              schedulerProvider);
         }
-        return INSTANCE;
+      }
     }
```
pierluca, line 151: SonarSource seems to suggest something like this:

```suggestion
  public static synchronized LAORepository getInstance(
      LAODataSource.Remote laoRemoteDataSource,
      LAODataSource.Local localDataSource,
      AndroidKeysetManager keysetManager,
      Gson gson, SchedulerProvider schedulerProvider) {
    if (INSTANCE == null) {
          INSTANCE = new LAORepository(laoRemoteDataSource, localDataSource, keysetManager, gson,
              schedulerProvider);
    }
    return INSTANCE;
  }
```

pierluca:
I would try to focus on these issues:
https://sonarcloud.io/project/issues?id=dedis_student_21_pop_fe2&pullRequest=413&resolved=false&severities=BLOCKER%2CCRITICAL%2CMAJOR&types=CODE_SMELL

Are some false-positives? We can mark them as such in Sonar if that's the case.

fe2-android/app/src/test/java/com/github/dedis/student20_pop/model/data/LAORepositoryTest.java
```diff
@@ -37,121 +43,201 @@
...
+    repository.sendSubscribe(LAO_CHANNEL);
+    testScheduler.advanceTimeBy(RESPONSE_DELAY, TimeUnit.MILLISECONDS);
 
-    testScheduler.advanceTimeBy(100, TimeUnit.MILLISECONDS);
-    testObserverUnsubscribe.assertComplete().assertNoErrors()
-        .assertValue(r -> r.getId() == 42);
+    repository.getAllLaos().subscribe(subscriber);
+
+    // Check the LAO is present in the laoById map of LAORepository
+    List<Lao> repositoryLaoList = (List<Lao>) subscriber.getEvents().get(0).get(0);
```
pierluca, line 188: what's `get(0).get(0)` doing here? 😅 
it's not exactly obvious.
it might not hurt to actually assign intermediate steps to named variables.

fe2-android/app/src/test/java/com/github/dedis/student20_pop/model/data/LAORepositoryTest.java
```diff
@@ -37,121 +43,201 @@
...
-    testScheduler.advanceTimeBy(1500, TimeUnit.MILLISECONDS);
+    // Check the LAO is present in the allLaoSubject list of LAORepository
+    Lao repositoryLao = (Lao) subscriber.getEvents().get(0).get(0);
+    assertEquals(repositoryLao.getChannel(), LAO_CHANNEL);
+  }
+
+  @Test
+  public void testBroadcast() throws GeneralSecurityException {
+    // Destroy the instance from the setup
+    LAORepository.destroyInstance();
```
pierluca, line 212: if the setup is not generic enough, maybe it's worth either repeating the setup in the relevant tests, OR splitting this test in a separate class.

fe2-android/app/src/test/java/com/github/dedis/student20_pop/model/data/LAORepositoryTest.java
```diff
@@ -37,121 +43,201 @@
...
+
+    // Create message to broadcast
+    byte[] dataBuf = Injection.provideGson().toJson(CREATE_LAO, Data.class).getBytes();
+    Mockito.when(signer.sign(dataBuf)).thenReturn(dataBuf);
```
pierluca, line 216: why do you need to mock the signer here?

pierluca: and if you're mocking it, should you verify that the function was called with some specific arguments?

fe2-android/app/src/test/java/com/github/dedis/student20_pop/model/data/LAORepositoryTest.java
```diff
@@ -37,121 +43,201 @@
...
+    MessageGeneral message = new MessageGeneral(Base64.getUrlDecoder().decode(ORGANIZER), CREATE_LAO, signer, Injection.provideGson());
+
+    // Simulate a network response and then a broadcast from the server after the response delay
+    Observable<GenericMessage> upstream = Observable.fromArray((GenericMessage) new Result(REQUEST_ID), (GenericMessage) new Broadcast(LAO_CHANNEL, message))
```
pierluca, line 220:
```suggestion
    Observable<GenericMessage> upstream = Observable.fromArray(
        (GenericMessage) new Result(REQUEST_ID),
        (GenericMessage) new Broadcast(LAO_CHANNEL, message))
```
might be more readable

fe2-android/app/src/test/java/com/github/dedis/student20_pop/model/data/LAORepositoryTest.java
```diff
@@ -37,121 +43,201 @@
...
+    repository.getLaoObservable(LAO_CHANNEL).subscribe(subscriber);
 
-    testScheduler.advanceTimeBy(1500, TimeUnit.MILLISECONDS);
+    // Check the LAO is present in the allLaoSubject list of LAORepository
+    Lao repositoryLao = (Lao) subscriber.getEvents().get(0).get(0);
+    assertEquals(repositoryLao.getChannel(), LAO_CHANNEL);
+  }
+
+  @Test
+  public void testBroadcast() throws GeneralSecurityException {
```
pierluca, line 210: this test seems to be doing a lot of things, and it's actually quite hard to follow.
I can't quite put the finger on what's the issue, a few hypotheses:
- is the LaoRepository not easily testable? 
- are there too many dependencies and we end up having to mock things left and right? 
- would this test warrant a separate class with its own setup?
- would the testing code warrant some refactoring? In particular it seems the TestObserver could benefit from having a wrapper class, with a simple API along the lines of:
```
subscriber = new TestLaoSubscriber(repository, LAO_CHANNEL);
assertEquals(1, subscriber.GetEventsCount());
assertEquals(xxx, subscriber.GetEvent<T>(0).getXxx());
```

fe2-android/app/src/test/java/com/github/dedis/student20_pop/model/data/LAORepositoryTest.java
```diff
@@ -37,121 +43,201 @@
...
+    testScheduler.advanceTimeBy(RESPONSE_DELAY, TimeUnit.MILLISECONDS);
+
+    // Set subscriber of the LAO list to test from
+    TestObserver<Lao> subscriber = TestObserver.create();
+    repository.getLaoObservable(LAO_CHANNEL).subscribe(subscriber);
 
-    testScheduler.advanceTimeBy(1500, TimeUnit.MILLISECONDS);
+    // Check the LAO is present in the allLaoSubject list of LAORepository
+    Lao repositoryLao = (Lao) subscriber.getEvents().get(0).get(0);
+    assertEquals(repositoryLao.getChannel(), LAO_CHANNEL);
```
pierluca, line 206: how come the LAO is an event, in the absence of any createLao ? 🤔

pierluca (commented):
A few remarks on the testing side, happy to discuss further (zoom or in person)

fe2-android/app/src/test/java/com/github/dedis/student20_pop/model/data/LAORepositoryTest.java
celinecamacho: `getEvents` returns a list of `(values, errors, completion-notifications)`, the idea is to get the first value received from `getAllLaos` after the subscription. I will change this to make it more readable

fe2-android/app/src/test/java/com/github/dedis/student20_pop/model/data/LAORepositoryTest.java
celinecamacho: Sending the subscribe request and mocking the result simulates a LAO subscription which adds the new LAO to the LAORepository

fe2-android/app/src/test/java/com/github/dedis/student20_pop/model/data/LAORepositoryTest.java
celinecamacho: The mock is needed because the data is signed when creating a `MessageGeneral`, I will add some comments and a check

celinecamacho:
The blocker and critical issues of SonarCloud are being solved in another branch. @pierluca can you review again?

fe2-android/app/src/test/java/com/github/dedis/student20_pop/model/data/LAORepositoryTest.java
```diff
@@ -0,0 +1,308 @@
...
+
+  @After
+  public void destroy() {
+    // Ensure every test has a new LAORepository instance with a different TestSchedulerProvider
+    LAORepository.destroyInstance();
+  }
+
+  @Test
+  public void testSendCatchup() {
+    // Simulate a network response from the server after the response delay
```
nkcr, line 100: (I'm not experienced with this code base, so I might be very wrong) It feels like this is more testing the testing framework than the system itself. Here we are testing the sending of a message. The *only* thing we should test is if our system, given we sent a catchup message and we then got an answer (which we can assume since we are mocking it), did what we expected it to do (i.e use the Answer, process it, etc...). (And then we could also test if the system reacts correctly in the case the answer doesn't come back)

nkcr (dismissed):
Ok for me, we can merge. My comment is more for later discussion, if it makes sense. But it shouldn't prevent this PR from being merged.

nkcr (approved):

PR #414, opened 2021-09-02
Update dependencies & fix catch-up

pierluca:
- [ ] test that all functions still work after the dependencies upgrade
- [ ] test if fixed catchup works


nkcr (approved):

fe1-web/network/NetworkManager.ts
```diff
@@ -31,14 +30,13 @@ class NetworkManager {
...
-   * @param host the server's host
-   * @param port the server's port
+   * @param base the server's protocol + host + port (e.g. ws://host:clientport)
    * @param path the path at which the websocket can be established
    *
    * @returns a new connection to the server, or an existing one if it's already established
    */
-  public connect(host: string, port: number = 9000, path: string = 'organizer/client/'): NetworkConnection {
-    const address: string = NetworkManager.buildAddress(host, port, path);
+  public connect(base: string, path: string = 'organizer/client/'): NetworkConnection {
```
nkcr, line 38: I'm not convinced that trying to be smart and automatically provide a default path, if not provided, is a good idea. I would assume that the client knows the correct url and provide it. It makes the implementation simpler and the logic easier (if I provide a URL I'm not expecting the APP to modify it and add a path). Also what if I want to host the client at the root ?

pierluca: Fixed :)

nkcr (approved):

PR #417, opened 2021-09-03
Adds new protocol documentation

nkcr:
Based on the words from @pierluca:

> Making a change on the message format is not a big deal as such, as there would be only a few places in the code to reflect that change. However the painful part is to update the documentation accordingly...

That's not what he exactly said but the idea is there: right now the protocol specification is documented on the "5. Protocol - Detailed specifications" gdoc document, and reflected (hopefully) by what is defined in `/protocol`. Plus there are a bunch of extra READMEs on that same folder (that copies content from the google doc...).

I'm proposing to have a single source of documentation for the protocol specification in `/docs/protocol.md`, where that document directly refers to the JSON files in `/protocol`. This allows to keep the documentation closer to the implementation and always synced. We can then get rid of the "5. Protocol - Detailed specifications" gdoc document on the drive.

Aim is also to develop a common taxonomy of messages across implementation and in the doc. For example with:

- `RootRPC`: The root RPC message
- `RPCPayload`: The kind of message embedded in a` RootRPC` message (Query, Response)
- `QueryPayload`: The kind of message embedded in a `Query` message (Publish, Broadcast, ...)
- `MidLevelMessage` (probably a bad name): Message containing a base64URL encoded message, along with its signature, sender, witness_signatures, ...
- `HighLevelMessage` (probably a bad name): Application message encoded in a `MidLevelMessage` (`lao#create`, `lao#update_properties`, ...)

See the protocol.md 👉 https://github.com/dedis/student_21_pop/blob/proposal-global-protocol-documentation/docs/protocol.md

**Note**: I also took the opportunity to format the JSON files.


pierluca (changes requested):
TL;DR: I think it's a great initiative, but I'd advise against it in its *current* form. We can rework it to make it work, but I'd suggest delaying it a bit until we have a number of things ready (detailed below).

First of all, please do keep in mind that all remarks are specific to a software repository that is being developed mostly by bachelor/master students, and I wouldn't necessarily have the same point of view in the context of a more "homogenous" software team using, say, company-mandated processes, tools, configurations, IDEs, etc. 

Instead of reviewing on a file-by-file basis, I'd like to submit a few global remarks, as I believe these are broader and may concern most files. Here goes nothing:

-> I'd like to point out that there are a number of elements that make the JSON-Schema less readable than they should be, and negatively affect the proposed documentation. We probably should:
1. rework the JSON-Schema before proceeding to the documentation, as some of its structure is purely an artifact of past students' decision to (re)factor things in a certain way. This, in turn, led you to define a terminology which is in contradiction with the established one (e.g. regarding JSON-RPC). I have several ideas in this regard, but it'll be easier to talk about that in person.
2. improve the JSON-Schema documentation (`$comment` field), as it's by far not as good as the specification document in Google Drive. As such, if we're moving away from Google Drive, we'll have to rework all the protocol files to make sure the comments are crystal clear.

-> The current proposal assumes that anyone can read the JSON-Schema protocol and infer a valid example, that assumption probably doesn't hold for someone new to the project. Replacing the examples by the schema induces a loss of readability. I'd suggest adding the schema *in addition to* the examples.

Creating example files would be necessary to document, but they may also be used for testing purposes. As such, we could have a "continuous integration" validation of the examples (and thus the documentation), by checking with a super simple tool that the examples match the protocol schema.

The same examples could be used by unit and integration tests, and probably help the students get started.

-> I'm worried that this proposal actually complicates updating the documentation, as it now involves the setup and learning of new tools (embedme, doctoc,  ...) for each newcomer to the project (and each new computer), followed by a PR process.

I think we can easily solve this by adding a Github Action that automatically pushes a commit back on a development branch (or creates a PR on it) if it detects that embedme or doctoc would update a file.

Tiny downside: even so, we'll need to let the students know how this thing works.
That is, the documentation itself will need a bit of documentation 😟 

-> As I mentioned earlier, the vocabulary scared me off 😅
More seriously, there's already a sort of homogenous vocabulary in the project code bases, that we should document much better. It's mostly inspired by the standards we use and the field structures:
- JSON-RPC query (used for everything except broadcast)
- JSON-RPC notification (used for broadcast)
- JSON-RPC answer (used for sever response)
- Message (the publish/broadcast JSON-RPC parameter
- MessageData (the decoded content of the `data` field in `Message`)

I'd love the JSON-Schema to reflect this, in particular at the JSON-RPC level.
If you agree, let's work on that refactoring together @nkcr !

-> I would strive to keep at least pointers to the documentation in each folder in `/protocol`, as moving everything to `/docs` will not help making the documentation discoverable, or making sure that it gets updated when students change something in the `/protocol` folder.

(nice to have) This could also benefit from a Github Action which notices if a file was added in `/protocol` and "fails" if the `/docs` folder was not updated. It won't cover 100% of the cases, but it should be enough to force a tiny bit of documentation.

Another benefit of keeping pointers in each folder in `/protocol`, even if it's a tiny bit redundant, is that it helps the students navigate more easily in the `/protocol` folder and understanding what they're looking at. 

---

I think that's about it for now 😅 it's a long-ish response, but your work is massive and it deserved an in-depth response to address why I'm not comfortable with it in its current form, and also defining ways forward to make sure it gets integrated !

In conclusion, I really feel it's a great initiative, it's forcing us to address some underlying issues and we'll be all the stronger for it once it gets integrated !

nkcr:
Discussed with Pierluca the mentioned issues and we could agree on the way to move forward, notably with the nomenclature.
Here are some whiteboards results:
IMG_0590
IMG_0591

pierluca (approved):
As discussed, merging this as-is for now. Many more PRs to come :)

PR #419, opened 2021-09-03
Protocol: refactoring, examples, documentation, and tests

nkcr:
Based on our discussion with @pierluca I refactored a bit the protocol files. To ensure I didn't break things (too much... 🤭) I added examples that are tested against the specification.

👉 Examples: https://github.com/dedis/student_21_pop/tree/proposal-global-protocol-update/protocol/examples
👉 Documentation: https://github.com/dedis/student_21_pop/tree/proposal-global-protocol-update/docs

The documentation has been updated to include the examples, still with the goal of having a good-enough documentation so we can get rid of the corresponding google doc file.


protocol/answer/negativeAnswer/error.json
```diff
@@ -3,26 +3,24 @@
...
-			"minimum": -5, "maximum": -1,
-			"$comment": "Note: should be in { -5, -4, ..., -1 }",
-			"note": [
-				"0  operation was successful (should never be used)",
-				"-1 invalid action",
-				"-2 invalid resource (e.g. channel does not exist, channel was not subscribed to, etc.)",
-				"-3 resource already exists (e.g. lao already exists, channel already exists, etc.)",
-				"-4 request data is invalid (e.g. message is invalid), use the data field object to provide extra information",
-				"-5 access denied (e.g. subscribing to a “restricted” channel)"
-			]
```
pierluca, line 21: where are you moving this? 
Also, there's a `-6` code, equivalent to a `500 internal server error`

protocol/answer/answer.json
```diff
@@ -5,28 +5,7 @@
...
-				}
-			],
-			"$comment": "Note: this field is absent if the answer is negative"
-		},
-		"error": {
-			"description": "In case of negative answer, error generated by the client query",
-			"$ref": "negativeAnswer/error.json"
-		},
-		"id": {
+		"requestID": {
```
pierluca, line 8: `id` is mandated by the JSON-RPC standard.
https://www.jsonrpc.org/specification

pierluca (changes requested):
Okay, this is funny.
You're already addressing some of the points I raised in the other PR.

Let's talk about it first thing on Monday, as I have a clear idea on how to go a step further at the JSON-RPC level.

Also, you've just broken compliance with the JSON-RPC standard ;-)

.github/workflows/autoformat.yaml
```diff
@@ -0,0 +1,58 @@
...
+        with:
+          ref: ${{ github.head_ref }}
+
+      - name: Setup node 12.x
+        uses: actions/setup-node@v1
+        with:
+          node-version: 12.x
+
+      - name: install prettier
+        run: npm install -g --save-exact prettier
```
pierluca, line 21: you might want to install a given version number, to avoid things suddenly breaking

protocol/README.md
```diff
@@ -4,7 +4,22 @@ Proof-of-personhood, Spring 2021: Protocol specifications
 
 See `/docs/protocol.md` for documentation about the protocol.
 
+See `./examples` to see some examples.
+
+See `./test` to run tests on the examples.
+
 **Note**: There might a way to avoid repeating all the wrapping information by
 using refs, but for now, it may be safer to just repeat it so that every json
 file is guaranteed to be complete.
```
pierluca, line 13: what is this? I'm a bit confused

protocol/answer/answer.json
```diff
@@ -1,65 +1,59 @@
...
+        }
+    },
+
+    "oneOf": [
+        {
+            "required": ["result"]
+        },
+        {
+            "required": ["error"]
+        }
```
pierluca, line 56: we should also require all the other fields.

protocol/answer/answer.json
```diff
@@ -1,65 +1,59 @@
...
+                    "type": "integer",
+                    "const": 0
+                },
+                {
+                    "type": "array",
+                    "items": {
+                        "$ref": "../query/method/message/message.json"
+                    },
+                    "minItems": 1,
+                    "$comment": "if there are no items, then `0` must be returned"
```
pierluca, line 22: Warning: this is a change of spec, and probably a wrong one too.
It doesn't make sense to change the return type based on whether the array is empty or not.

```suggestion
                    "$minItems": 0,
                    "$comment": "Return value for a `catchup` request"
```

protocol/answer/answer.json
```diff
@@ -1,65 +1,59 @@
...
+    "description": "Match a positive or negative server answer",
+    "type": "object",
+    "additionalProperties": false,
+    "properties": {
+        "result": {
+            "description": "In case of positive answer, result of the client query",
+            "oneOf": [
+                {
+                    "type": "integer",
+                    "const": 0
```
pierluca, line 14:
```suggestion
                    "const": 0,
                    "$comment": "Default return value indicating a success",
```

protocol/answer/answer.json
pierluca: I'd suggest doing it inline (with some repetition) for clarity, rather than using the `allOf` + `oneOf` combination

protocol/answer/answer.json
```diff
@@ -1,65 +1,59 @@
...
+                },
+                {
+                    "type": "null"
+                }
+            ],
+            "$comment": "The id matches the request id. If there was an error in detecting the id, it must be null"
+        },
+
+        "jsonrpc": {
+            "$comment": "Defined by the parent, but needed here for the validation"
```
pierluca, line 46: This seems wrong. It's either included in the parent (and doesn't need to be repeated) or it's included here, but then we might as well put the full `jsonrpc` field information here.

protocol/answer/error.json
```diff
@@ -0,0 +1,31 @@
...
+                "-3 resource already exists (e.g. lao already exists, channel already exists, etc.)",
+                "-4 request data is invalid (e.g. message is invalid), use the data field object to provide extra information",
+                "-5 access denied (e.g. subscribing to a “restricted” channel)"
+            ]
+        },
+        "description": {
+            "description": "[String] error description",
+            "type": "string"
+        }
+    },
```
pierluca, line 27: we should probably add the `data` field of the error object, for rich error information.

protocol/answer/error.json
```diff
@@ -0,0 +1,31 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "https://raw.githubusercontent.com/dedis/student_21_pop/master/protocol/answer/error.json",
+    "title": "If an error occurred, the \"result\" field is absent and the error object must adhere to the following (extensible) format",
```
pierluca, line 4:
```suggestion
    "title": "If an error occurred, the returned error object must adhere to the following format",

```

protocol/examples/answer/error.json
```diff
@@ -0,0 +1,8 @@
+{
+    "jsonrpc": "2.0",
+    "id": 999,
+    "error": {
+        "code": -1,
+        "description": "This is an error"
```
pierluca, line 6: I'd add one example with a `data` field too.

protocol/examples/answer/general_message.json
```diff
@@ -0,0 +1,20 @@
...
+            "signature": "XXX",
+            "message_id": "XXX",
+            "witness_signatures": []
+        },
+        {
+            "data": "XXX",
+            "sender": "XXX",
+            "signature": "XXX",
+            "message_id": "XXX",
+            "witness_signatures": []
```
pierluca, line 17: this is invalid and will be rejected in our implementation.
the data needs to be base64url encoded and represent valid keys, hashes, etc.
if we want this to be usable for tests, we need it to be correct.

that's also why I suggested using examples generated from the application :-)

protocol/answer/answer.json
```diff
@@ -1,65 +1,59 @@
...
+        "id": {
+            "oneOf": [
+                {
+                    "type": "integer"
+                },
+                {
+                    "type": "null"
+                }
+            ],
+            "$comment": "The id matches the request id. If there was an error in detecting the id, it must be null"
```
pierluca, line 42:
```suggestion
                {
                    "type": "integer"
                    "$comment": "The id matching the request id."
                },
                {
                    "type": "null"
                    "$comment": "If there was an error in detecting the request id, it must be null."
                }
            ],
```

protocol/examples/answer/wrong_error_status.json
```diff
@@ -0,0 +1,8 @@
+{
+    "jsonrpc": "2.0",
+    "id": "xxx",
```
pierluca, line 3: we said `id` must be an integer. this will fail for the wrong reason.

protocol/examples/messageData/election_end.json
```diff
@@ -0,0 +1,8 @@
+{
+    "object": "election",
+    "action": "end",
+    "lao": "XXX",
+    "election": "XXX",
+    "created_at": 123,
+    "registered_votes": "XXX"
+}
```
pierluca, line 8: same problem, incorrect data

protocol/examples/messageData/vote_cast_write_in.json
```diff
@@ -0,0 +1,14 @@
...
+    "object": "election",
+    "action": "cast_vote",
+    "lao": "XXX",
+    "election": "XXX",
+    "created_at": 123,
+    "votes": [
+        {
+            "id": "XXX",
+            "question": "XXX",
+            "write_in": "XXX"
```
pierluca, line 11: is this really supported? 🤔 I don't think so, to be honest.
It may be best to remove this for now (incl. from the documentation and the application, as needed)

protocol/jsonRPC.json
```diff
@@ -0,0 +1,27 @@
...
+    },
+    "oneOf": [
+        {
+            "$ref": "answer/answer.json"
+        },
+        {
+            "$ref": "query/query.json"
+        }
+    ],
+    "required": ["jsonrpc", "id"]
```
pierluca, line 26: `id` is not required for a JSON-RPC notification. It's wrong to add it at this level.

protocol/jsonRPC.json
```diff
@@ -0,0 +1,27 @@
...
+    "properties": {
+        "jsonrpc": {
+            "type": "string",
+            "const": "2.0",
+            "description": "JSON RPC version"
+        },
+
+        "id": {
+            "$comment": "defined by the children"
+        }
```
pierluca, line 16:
```suggestion
        }
```

protocol/query/method/broadcast.json
```diff
@@ -1,43 +1,42 @@
...
+            "required": ["channel", "message"]
+        },
+
+        "jsonrpc": {
+            "$comment": "Defined by the parent, but needed here for the validation"
+        },
+
+        "id": {
+            "$comment": "Defined by the parent, but needed here for the validation"
+        }
```
pierluca, line 38: Not part of a broadcast.

protocol/query/method/message/data/data.json
```diff
@@ -1,47 +1,47 @@
...
-			"$ref": "dataWitnessMessage.json"
-		}
-	]
-}
\ No newline at end of file
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "https://raw.githubusercontent.com/dedis/student_21_pop/master/protocol/query/method/message/data/data.json",
+    "description": "Match one type of data",
+    "type": "object",
+    "oneOf": [
```
pierluca, line 6: would it make sense to factor this out in separate subdirectories, e.g. at this level 
```
        {
            "$ref": "election/index.json"
        },
        {
            "$ref": "lao/index.json"
        },
```
and so on, while in `election/index.json` there would be :
```
        {
            "$ref": "CastVote.json"
        },
        {
            "$ref": "EndElection.json"
        },
        {
            "$ref": "ResultElection.json"
        },
        {
            "$ref": "SetupElection.json"
        },
```
?

protocol/examples/answer/wrong_general_message_empty.json
```diff
@@ -0,0 +1,5 @@
+{
+    "jsonrpc": "2.0",
+    "id": 999,
+    "result": []
```
pierluca, line 4: that's actually a valid result for the `catchup` RPC.

protocol/query/query.json
```diff
@@ -1,44 +1,44 @@
...
-		}
-	]
-}
\ No newline at end of file
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "https://raw.githubusercontent.com/dedis/student_21_pop/master/protocol/query/query.json",
+    "title": "Match a custom JsonRpc 2.0 query message",
+    "description": "Match a client query",
+    "type": "object",
+    "additionalProperties": false,
```
pierluca, line 7: I'd keep this file to a minimum and just reference the children (i.e. the `oneOf` section).

On a side note, `params` is an optional field in JSON-RPC, and while we don't have the case for now, I don't see the reason to make it mandatory at this level.

protocol/query/method/subscribe.json
```diff
@@ -1,33 +1,35 @@
...
+
+            "required": ["channel"]
+        },
+
+        "jsonrpc": {
+            "$comment": "Defined by the parent, but needed here for the validation"
+        },
+
+        "id": {
+            "$comment": "Defined by the parent, but needed here for the validation"
```
pierluca, line 32: I see you've added these comments everywhere, but I'm pretty sure it's not necessary for validation.

pierluca (changes requested):
Stopping here for now. My brain has turned to mush.
A lot of comments are applicable throughout the code base.

We might also want to address this:
https://github.com/dedis/student_21_pop/issues/309
(in a separate PR, I guess :-))

protocol/query/query.json
pierluca: Feel free to reinterpret this in the light of our conversation :-) possibly minimal changes needed

docs/doctoc.sh
```diff
@@ -23,7 +23,7 @@ set -u
 INVALID_CHARS="'[]/?!:\`.,()*\";{}+=<>~$|#@&–—"
 
 # For Mac user, we need to use 'gsed'
-SED=sed
+SED=gsed
```
pierluca, line 26: we can't expect all users to have a Mac though :-/ 
Could we select between sed or used based on a detection of the OS?

pierluca (commented):
Minor comments, LGTM otherwise

pierluca (approved):
LGTM

PR #426, opened 2021-09-09
Message implementation clean-up

nkcr:
Cleans up the message implementation.

- Starts using zerolog.Logger
- Removes a lot complexities due to weird marshal/unmarshal of messages
- Uses sub-packages to better mimic the protocol specification
- Removes a lot of unecessary custom types

The most interesting element to review is the `message/` folder.


be1-go/message2/jsonRPC.go
```diff
@@ -0,0 +1,79 @@
...
+// RPCType ...
+type RPCType int
+
+// JSONRPC ...
+type JSONRPC struct {
+	JSONRPC string
+
+	query.Query
+	answer.Answer
+}
```
pierluca, line 31: 1. Tiny detail, but I'd call this a `JsonRpcMessage`
2. This feels semantically wrong, from an OOP perspective, a `JsonRpcMessage` does not `have` a query and/or an answer, it `is` a query `or` an answer. 

As such, I would suggest making this an interface and using a simple type switch where we need to use it. 
i.e. `type JsonRpcMessage interface {}`
That means that wherever needed, logically only in one place, we can do something along the lines of this
```
switch m := jsonRpcMsg.(type) {
    query.Query: processQuery(m)
    answer.Answer: processAnswer(m)
    default: return errors.Errorf("JsonRpcMessage is invalid")
}
```

and the Unmarshalling might look like this:
```
func UnmarshalJsonRpc(buf []byte) JsonRpcMessage, error {
    var q query.Query
    if err := json.Unmarshal(data, &q); err == nil {
        return q, nil
    }

    var a answer.Answer
    if err := json.Unmarshal(data, &a); err == nil {
        return a, nil
    }

    return nil, err
}
```

I feel it also makes the marshaling and unmarshaling logic a lot simpler and the objects contain their own fields in an obvious way.

the same logic could maybe also be applicable to `Query` vs. `Subscribe`, `Publish`, etc. but I'm not sure if that's very relevant there. I'll comment on that code directly :)

be1-go/message2/messagedata/vote_cast_write_in.go
```diff
@@ -0,0 +1,18 @@
+package messagedata
+
+// VoteCastWriteIn ...
```
pierluca, line 3: I don't think this is fully implemented . I'm wondering if we should remove it for the time being, in order to minimize the amount of dead/unsupported/low-priority code.

be1-go/message2/query/query.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"encoding/json"
+	"student20_pop/message2/query/method"
+
+	"golang.org/x/xerrors"
+)
+
+// Query ...
+type Query struct {
+	Method string
```
pierluca, line 12: I have a suggestion based also on my comment for the `JsonRpcMessage` struct/interface.

Can we add the `Id` here and make it optional (or its semantic equivalent)? The JSON-Schema will validate anyways that it's present where it's needed and absent where it's not. Only the marshaling logic will have to be aware that the `Id` should be absent in some cases.

If so, the *only* difference between the methods will be the params. Thus, each method object could be defined as implementing a `QueryParams` interface which is implemented like the `JsonRpcMessage` suggestion.
This would also be trivial to use, because based on the `Method` field, we can simply cast the `QueryParams` to the right type.

The beauty of this solution is that it also massively simplifies the current method objects, as they can all be aliases of just 2 types (params = a channel, and params = channel+message), which minimizes the amount of repetition.

pierluca (commented):
I haven't reviewed the full 1340 LOCs, but overalls it seems pretty good.
I have just a couple of thoughts on the implementation of the JSON-RPC layer.

be1-go/message2/messagedata/vote_cast_write_in.go
pierluca: Fully implemented in the rest of the system, I mean.

be1-go/message2/jsonRPC.go
nkcr: None of those solutions are ideal. 

I definitely don't like the idea of bringing interfaces as it complexities and hides the mapping between a go struct and the json structure. The message package should provides a simple and static representation of json messages to go structs. This is why I'd like to avoid bringing interfaces and (whenever possible...) weird marshal/unmarshal functions with hidden logic. And when needed, they should be as much explicit as possible.

Your proposal to "let's try to unmarshal this message into this struct `S1`, and if it doesn't work let's try with `S2`" is fragile as two structs `SX` and `SY` could both result in a successful unmarshalling. Ok, in this case this is trivial, but...

pierluca: I agree with you on most points if we were developing a *generic solution*.

In particular, I agree with you regarding `S1` and `S2` in *generic* terms (i.e. I wouldn't write a library like this), but for the JSON-RPC protocol, it is indeed a trivial and accurate solution.

More generally, I would also argue that, for the situation at hand, interfaces make relationships more obvious, by respecting semantics (`is` vs. `has`).
There's no `JsonRpcMessage` entity in the protocol, it's an abstraction because we do need a single unmarshaling point for both query and answer. In practice they're altogether different elements.
As such, I believe with interfaces there'll actually be less weird marshaling/unmarshaling logic.

I'm tempted to say: do try and write the corresponding marshaling function to see what I mean, I'm 99% sure that the current solution will make it more complex than a trivial interface. If I'm wrong, and if the Marshal/Unmarshal logic is indeed as simple as you seem to believe, I'm happy to stick to the current solution.

be1-go/hub/base_channel.go
```diff
@@ -113,25 +118,41 @@ func (c *baseChannel) Catchup(catchup message.Catchup) []message.Message {
...
+				}{
+					Channel: c.channelID,
+					Message: msg,
+				},
+			},
+		},
 	}
 
-	buf, err := json.Marshal(query)
+	buf, err := json.Marshal(&rpcMessage)
```
pierluca, line 155: doesn't this result in an object looking like this? 
```
{
  "jsonrpc": "2.0",
  "Query": {
    ...
  }
}
```

be1-go/hub/rollCall.go
```diff
@@ -5,29 +5,28 @@ import (
...
+	c.rollCall.id = string(msg.ID)
 	c.rollCall.state = Created
 	return nil
 }
 
 // processOpenRollCall processes an open roll call object.
-func (c *laoChannel) processOpenRollCall(data message.Data, action message.RollCallAction) error {
-	if action == message.RollCallAction(message.OpenRollCallAction) {
+func (c *laoChannel) processOpenRollCall(msg messageX.Message, action string) error {
+	if action == "open" {
```
pierluca, line 29: What do you think about having a big string enum somewhere with the different options ? 
It seems a bit brittle to have the strings hardcoded in possibly multiple places.

nkcr: Yes, there is definitely something to do about that. I'm waiting to see how things come up before entering the "code polishing phase" 😄.

be1-go/hub/base_channel.go
nkcr: No, because it is embedded and not declared as a property.

be1-go/message2/query/query.go
nkcr: I'd rather not. IMO it's only adding useless complexity, as the actual implementation already simple and explicit.

pierluca (changes requested):
I'd really really avoid having magic strings and use constants.
Rest LGTM 👍

pierluca (approved):
Decided to do the necessary changes in the next PR

PR #427, opened 2021-09-10
Incoming message handler

celinecamacho:
There are 6 new classes for the incoming message handling:

- `GenericHandler`: incoming messages are redirected from `LAORepository` to this class, `GenericHandler` handles the subscribe, catchup, broadcast and create LAO requests
- `MessageHandler`: the broadcast and catchup messages are redirected to this class, `MessageHandler` sends the messages to the corresponding data handlers
- `LaoHandler`: handles all LAO messages (CreateLao, UpdateLao, StateLao)
- `RollCallHandler`: handles all roll call messages (CreateRollCall, OpenRollCall, EndRollCall)
- `ElectionHandler`: handles all election messages (ElectionSetup, ElectionResult, ElectionEnd, CastVote)
- `WitnessMessageHandler`: handles all WitnessMessageSignature messages


fe2-android/app/src/main/java/com/github/dedis/student20_pop/model/Election.java
```diff
@@ -38,7 +39,10 @@
     //Results of an election (associated to a question id)
     private Map<String, List<QuestionResult>> results;
 
-    public Election() {
+    public Election(String laoId, long creation, String name) {
+        this.id =  IdGenerator.generateElectionSetupId(laoId, creation, name);
```
pierluca, line 43: extra space :)

fe2-android/app/src/main/java/com/github/dedis/student20_pop/model/data/LAORepository.java
```diff
@@ -61,744 +46,313 @@
...
-      is after (hence preventing reordering attacks) */
-      Optional<String> previousMessageIdOption = election.getMessageMap().entrySet().stream()
-              .filter(entry -> senderPk.equals(entry.getValue())).map(Map.Entry::getKey).findFirst();
-      //If there is no previous message, or that this message is the last of all received messages, then we consider the votes
-      if (!previousMessageIdOption.isPresent() ||
-              ((CastVote) messageById.get(previousMessageIdOption.get()).getData()).getCreation() <= data.getCreation()) {
-        election.putVotesBySender(senderPk, data.getVotes());
-        election.putSenderByMessageId(senderPk, messageId);
-        lao.updateElection(election.getId(), election);
+  public void sendStateLao(Lao lao, MessageGeneral msg, String messageId, String channel) {
```
pierluca, line 189: why do we need to keep this in LAORepository ? 
I'm probably missing something very obvious here :)

fe2-android/app/src/main/java/com/github/dedis/student20_pop/utility/handler/GenericHandler.java
```diff
@@ -0,0 +1,155 @@
...
+
+  /**
+   * Delete the pending requests with the id of the error
+   *
+   * @param genericMessage the message received
+   * @param subscribeRequests the pending subscribe requests
+   * @param catchupRequests the pending catchup requests
+   * @param createLaoRequests the pending create lao requests
+   */
+  public static void handleError(
```
pierluca, line 36: I infer from this that once you request the LAORepository to do something over the network, the LaoRepository RPC has no way of telling you if it failed. 

It might be interesting to implement a Promise-like pattern, where I can call the RPC and the LaoRepository makes sure that I get (asynchronous) feedback from the RPC, so that I can display (for instance) a spinner while the operation gets processed, or retry if there's an error.

I seem to remember that a LAORepository user can `subscribe` to an RPC to get the response, isn't a similar mechanism available for errors? If so, a `Promise`-like solution might be more appropriate.

Obviously, this is a completely separate issue from this PR :)

fe2-android/app/src/main/java/com/github/dedis/student20_pop/utility/handler/ElectionHandler.java
```diff
@@ -0,0 +1,180 @@
...
+/** Election messages handler class */
+public class ElectionHandler {
+
+  public static final String TAG = ElectionHandler.class.getSimpleName();
+
+  public static final String ELECTION_NAME = "Name : ";
+  public static final String MESSAGE_ID = "Message ID : ";
+  public static final String ELECTION_ID = "Election ID : ";
+  public static final String ELECTION_QUESTION = "Question : ";
+  public static final String ELECTION_SETUP = "New Election Setup ";
```
pierluca, line 32: It might make sense to move these constants (and the `WitnessMessage` creation) in a dedicated function.
Same applies to other handlers.

fe2-android/app/src/main/java/com/github/dedis/student20_pop/utility/handler/MessageHandler.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    // Trigger an onNext
+    if (!(data instanceof WitnessMessageSignature) && laoRepository.isLaoChannel(channel)) {
+      LAOState laoState = laoRepository.getLaoById().get(channel);
+      laoState.publish();
+      if (data instanceof StateLao || data instanceof CreateLao) {
+        laoRepository.setAllLaoSubject();
+      }
+    }
```
pierluca, line 61: this might benefit from being moved into its own function and getting its own JavaDoc.
I get the general idea, but it's really not trivial to understand its function the first time around

fe2-android/app/src/mock/java/com/github/dedis/student20_pop/Injection.java
```diff
@@ -38,13 +45,50 @@
...
+      KeysetHandle publicKeysetHandle = keysetManager.getKeysetHandle().getPublicKeysetHandle();
+
+      try {
+        String publicKey = Keys.getEncodedKey(publicKeysetHandle);
+      } catch (IOException e) {
+        Log.e("INJECTION", "failed to retrieve public key", e);
+      }
+
+      KEYSET_MANAGER = keysetManager;
+    }
```
pierluca, line 90: does any of this exist in the mocked context? 
should we implement a mock ? is it even feasible ?

fe2-android/app/src/test/java/com/github/dedis/student20_pop/model/data/LAORepositoryTest.java
```diff
@@ -0,0 +1,308 @@
...
+  private TestObserver<Lao> laoTestObserver;
+  private TestObserver<List<Lao>> laoListTestObserver;
+  private SchedulerProvider testSchedulerProvider;
+
+  @Before
+  public void setup() throws GeneralSecurityException {
+    testSchedulerProvider = new TestSchedulerProvider();
+    testScheduler = (TestScheduler) testSchedulerProvider.io();
+
+    // Mock the signing of of the data CreateLao
```
pierluca, line 70: of of

fe2-android/app/src/test/java/com/github/dedis/student20_pop/model/data/LAORepositoryTest.java
```diff
@@ -0,0 +1,308 @@
...
+    byte[] dataBuf = Injection.provideGson().toJson(CREATE_LAO, Data.class).getBytes();
+    Mockito.when(signer.sign(dataBuf)).thenReturn(dataBuf);
+
+    // Create the message containing a CreateLao data
+    createLaoMessage = new MessageGeneral(Base64.getUrlDecoder().decode(ORGANIZER), CREATE_LAO, signer, Injection.provideGson());
+
+    // Check the correct data is signed during the creation of MessageGeneral
+    Mockito.verify(signer, Mockito.times(1)).sign(dataBuf);
```
pierluca, line 78: these lines almost look like their own test of `MessageGeneral`

fe2-android/app/src/main/java/com/github/dedis/student20_pop/utility/handler/MessageHandler.java
```diff
@@ -0,0 +1,64 @@
...
+    throw new IllegalStateException("Utility class");
+  }
+
+  /**
+   * Send messages to the corresponding handler.
+   *
+   * @param laoRepository the repository to access the messages and LAOs
+   * @param channel the channel on which the message was received
+   * @param message the message that was received
+   * @return true if the message cannot be processed and false otherwise
```
pierluca, line 28: This is a very atypical return value.
In general `true` signifies success, and `false` failure.
If possible, I'd switch it the other way around.

fe2-android/app/src/test/java/com/github/dedis/student20_pop/utility/handler/MessageHandlerTest.java
```diff
@@ -0,0 +1,304 @@
+package com.github.dedis.student20_pop.utility.handler;
```
pierluca, line 1: Not necessary for this PR, but keep this in mind as-needed : 

Do not hesitate to split this up by tape of message handlers, even if you keep the exact same code.
i.e. `LaoHandlerTest`,  `RollCallHandler`, etc. even if you still use the `MessageHandler` entry point.

Also, for now, these handlers depend on the `LaoRepository` for the input, but maybe we'll later be able to split them up as follows:
1. Verify that LaoRepository calls generic message handler on an incoming message
2. Verify that generic message handler calls the right message handler for any given message
3. Verify that the specific message handler processes a given message correctly, mocking the LaoRepository
This would effectively move the tests from an integration-level closer to a unit-level.

Again, no need to do this now, it's not the highest priority :)

pierluca (commented):
Can't review this fully now, but here are a few remarks.

fe2-android/app/src/main/java/com/github/dedis/student20_pop/model/data/LAORepository.java
celinecamacho: I thought about moving this method to the WitnessMessageHandler class because its only usage is in handleWitnessMessage. For that I would need to create getters for the keyset manager and Gson in the LAORepository or pass them as arguments on all handle methods. I am not sure what's best

fe2-android/app/src/main/java/com/github/dedis/student20_pop/utility/handler/MessageHandler.java
celinecamacho: The return value represents if the message is queued or not, I can swap it to represent success instead

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/WitnessMessageHandler.java
```diff
@@ -0,0 +1,116 @@
...
+
+    // Verify signature
+    if (!Signature.verifySignature(messageId, senderPkBuf, signatureBuf)) {
+      return false;
+    }
+
+    if (laoRepository.getMessageById().containsKey(messageId)) {
+      // Update the message
+      MessageGeneral msg = laoRepository.getMessageById().get(messageId);
+      msg.getWitnessSignatures().add(new PublicKeySignaturePair(senderPkBuf, signatureBuf));
```
pierluca, line 56: Nothing for you to do in this PR, but this bunch of code really makes me think the Java front-end could stand to benefit from some of the things we introduced in fe1-web, such as having `PublicKey`, `Hash` and `Signature` classes, abstracting away the representation/encoding (bytes vs string).

The code would then look like this, don't you think it reads better?
```suggestion
      String channel,
      PublicKey senderPk,
      WitnessMessageSignature message) {
    Log.d(TAG, "Received Witness Message Signature Broadcast");
    Hash messageId = message.getMessageId();
    Signature signature = message.getSignature();

    // Verify signature
    if (!signature.verify(messageId, senderPk)) {
      return false;
    }

    if (laoRepository.getMessageById().containsKey(messageId)) {
      // Update the message
      MessageGeneral msg = laoRepository.getMessageById().get(messageId);
      msg.getWitnessSignatures().add(new PublicKeySignaturePair(senderPk, signature));
```

pierluca (changes requested):
Only minor changes needed.
I've reviewed everything except the test classes at this point.

pierluca (approved):

PR #428, opened 2021-09-13
Reformat code and change applicationId to popstellar

celinecamacho:


celinecamacho:
The SonarCloud bugs are fixed in PR #427

pierluca (commented):
Thanks so much for taking this on !

Tiny remark concerning changing applicationId: 
-> Shouldn't that also entail fixing the package names? 
-> How about the paths?

Maybe it's best to do this in a coordinated way with be1-go, which is trying to fix a similar problem, and drop the `student20_` and just call it `pop` so that we don't need to change it against next year. What did you decide on your end @nkcr ?

pierluca:
As soon as the `applicationId` / package names things are fixed, I'm happy to merge this.

Careful though: you might have lots of merge conflicts with #427

nkcr:
Result of discussing the applicationID with Pierluca: `pop` is too generic since it is a broad concept, and shouldn't be used for our particular implementation. Therefore, we agreed with the name **popstellar**, which refers to the concept of a N-version programing system implementing PoP, with its multiple inter-connected elements.

I'm going to update the go back-end with this name. The repos is left unchanged for the moment.

pierluca (changes requested):
@celinecamacho  see message above by Noémien.
Please ensure that the folder names get updated too  (i.e. `src/main/java/com/github/dedis/popstellar/...` )

pierluca (approved):
As discussed, we can merge this now.

PR #429, opened 2021-09-13
Protocol tiny fix (?)

nkcr:
I feel this should be updated as follow. Just quickly mentioning that here to be fixed in the spec if true.


pierluca (commented):
You're absolutely right.

pierluca (approved):
LGTM

PR #399, opened 2021-06-18
Election End and Election Result server

Tesa27:

PR #431, opened 2021-09-14
Includes PR#399 (election#end, election#result)

nkcr:
- Uses zerolog.Logger
- Plus some formatting there and there

Depends on #426 
Closes #399


pierluca (approved):
LGTM

PR #435, opened 2021-09-15
Resolved be2 Hash issues

RaulinN:


be2-scala/src/test/scala/ch/epfl/pop/model/objects/HashSuite.scala
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  test("Hash 'fromStrings' works with array of strings") {
+    var hash: Hash = Hash.fromStrings("salut", "toi")
+    var expected: Hash = Hash(Base64Data("d_mpef1MwgCQoph0wjZDFDmhngFMXAAROLp6XNDii5c="))
+
+    hash should equal (expected)
+
+    hash = Hash.fromStrings("1", "2", "3", "4")
+    expected =   Hash(Base64Data("yru1zNKR23-3CidGmsdNH53tuvyLM0iftp2aPib-KVU="))
```
nkcr, line 46: Missing an assertion.

be2-scala/src/test/scala/ch/epfl/pop/model/objects/HashSuite.scala
```diff
@@ -0,0 +1,57 @@
...
+  test("Hash 'fromStrings' works with no string") {
+    val hash: Hash = Hash.fromStrings()
+    val expected: Hash = Hash(Base64Data("47DEQpj8HBSa-_TImW-5JCeuQeRkm5NMpJWZG3hSuFU="))
+
+    hash should equal (expected)
+  }
+
+  test("Hash 'fromStrings' works with one string") {
+    var hash: Hash = Hash.fromStrings("PoP")
+    var expected: Hash = Hash(Base64Data("fAoSEwSI4qx6brSVryjql0GdkwbMNidcIeJHa8TxGR4="))
```
nkcr, line 29: I don't understand why the expected result is not the same as with the first test, since the string is the same. Shouldn't `fromStrings` with one string yield the same as `fromString` ?

be2-scala/src/test/scala/ch/epfl/pop/model/objects/HashSuite.scala
```diff
@@ -0,0 +1,57 @@
...
+    hash = Hash.fromStrings("1", "2", "3", "4")
+    expected =   Hash(Base64Data("yru1zNKR23-3CidGmsdNH53tuvyLM0iftp2aPib-KVU="))
+  }
+
+  test("Hash 'fromStrings' is deterministic") {
+    val hash1: Hash = Hash.fromStrings("PoP", "project")
+    val hash2: Hash = Hash.fromStrings("PoP", "project")
+    val expected: Hash = Hash(Base64Data("6FCAKniEfwWAX1lkzUaCq3ZT2TUKnmz6jGmOl_c0PRg="))
+
+    hash1 should equal (expected)
```
pierluca, line 54: this part of the test is redundant with the others, maybe we can skip it.

be2-scala/src/test/scala/ch/epfl/pop/model/objects/HashSuite.scala
pierluca: agreed

be2-scala/src/test/scala/ch/epfl/pop/model/objects/HashSuite.scala
pierluca: I'd also split out the second assertion in a separate test

be2-scala/src/test/scala/ch/epfl/pop/model/objects/HashSuite.scala
RaulinN: Mhmmm... I think that's what we are doing in fe1 right ^^' ? `fromString` hashes a string using SHA-256 whereas `fromStrings` hashes using our custom hashing technique (length of the string || string)

pierluca (approved):

be2-scala/src/test/scala/ch/epfl/pop/model/objects/HashSuite.scala
pierluca: okay, thanks for the renaming. maybe `fe1` could benefit from the disambiguation too 😅

PR #436, opened 2021-09-15
Corrected potential future encoding problems

RaulinN:
Removed unsafe `mkString`


nkcr (approved):
Nice!
Makes me realize, @pierluca, that the spec should mention the use of UTF-8 as the charset.

PR #439, opened 2021-09-15
PipelineErrors generate the correct reply rpc id

RaulinN:
Up until now, if something went wrong in the graph, a PipelineError
was issued and the server would respond to the client with a 'null'
id (json-rpc id)

### DIFF
--- a/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ParamsHandler.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ParamsHandler.scala
@@ -54,14 +54,14 @@ object ParamsHandler {
     case Left(jsonRpcMessage: Subscribe) =>
       // ActorFlow.ask(clientActorRef)(makeMessage = (el, replyTo: ActorRef) => SubscribeTo(channel))
       clientActorRef ! ClientActor.SubscribeTo(jsonRpcMessage.channel)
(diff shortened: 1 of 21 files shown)


be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ParamsHandler.scala
```diff
@@ -54,14 +54,14 @@ object ParamsHandler {
     case Left(jsonRpcMessage: Subscribe) =>
       // ActorFlow.ask(clientActorRef)(makeMessage = (el, replyTo: ActorRef) => SubscribeTo(channel))
       clientActorRef ! ClientActor.SubscribeTo(jsonRpcMessage.channel)
-      Right(PipelineError(-100, "FIXME: should use akka ask pattern to create an ActorFlow"))
+      Right(PipelineError(-100, "", Some(-111))) // FIXME: should use akka ask pattern to create an ActorFlow
```
pierluca, line 57: what's going on here? I don't remember how this should be fixed 🤔

pierluca (approved):
LGTM overalls.
Just a question for a subsequent fix,

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ParamsHandler.scala
RaulinN: Was waiting for akka ask pattern to work in the system. I resolved the issue before even seeing your message ^^. See #441

PR #457, opened 2021-09-23
Avoid network redirections with default fe1/fe2 paths

pierluca:
This was introduced during the refactoring and created some problems for both front-ends. 

Essentially, in this scenario, the `mux.HandleFunc()` call was redirecting `/a/b` to `/a/b/ `, but not all WebSocket clients support redirection. It was hidden away by the fact that we were often running the server behind a Nginx proxy.

It's easier to do the modification here than on both front-ends, especially as it's a temporary fix and bound to be configurable later on.


nkcr (dismissed):

nkcr:
It won't work anymore if the URL has a trailing "/". Will it cause problems ?

nkcr (approved):

PR #458, opened 2021-09-23
Fe1 - re-integrating election code

pierluca:
Most of the code is contributed by the students from last semester.

No need to review in detail. Let's talk about it @nkcr


nkcr (approved):
Let's go to master 💣

PR #461, opened 2021-09-24
Enhanced documentation

RaulinN:


be2-scala/docs/README.md
```diff
@@ -224,22 +224,117 @@ val f: Future[GraphMessage] = (dbActor ? DbActor.Write(channel, message)).map {
...
+// Await.result waits for <duration> for the future <f> to complete. It returns the value contained by the future (here `GraphMessage`) if the latter is successful, or throws if the Future terminates without being successful (i.e. either Failure or Timeout) 
 Await.result(f, duration)
 ```
 
 
 
 :information_source: the database may easily be reset/purged by either deleting a folder corresponding to a channel or by deleting the `database` folder entirely
 
-## Debugging Tips
+## 4.	Git(Hub)
```
pierluca, line 235: I would move this section to the global docs of the project, as it's true for everyone.
I would also consider simply pointing to one of the many Git tutorials.

I'm concerned providing a little bit of information just to get started might delay learning about Git properly, which is arguably what students should be doing at this point :) 
What do you think?

be2-scala/docs/README.md
```diff
@@ -224,22 +224,117 @@ val f: Future[GraphMessage] = (dbActor ? DbActor.Write(channel, message)).map {
...
+
+```bash
+git checkout -b <branchName>
+```
+
+If you rather want to make a copy of branch `x` (e.g. a branch that is currently being reviewed, i.e. not merged), use this instead
+
+```bash
+git checkout <x>
+git chekcout -b <branchName>
```
pierluca, line 267: typo

```suggestion
git checkout -b <branchName>
```

be2-scala/docs/README.md
```diff
@@ -224,22 +224,117 @@ val f: Future[GraphMessage] = (dbActor ? DbActor.Write(channel, message)).map {
...
+
+---
+
+### Push changes on the remote server
+
+When pushing changes, you first want to **commit** the changes you've applied using
+
+```bash
+git add -A									# add all modified and new files to the commit
+git commit -m"<message>"    # add a title to the commit (switch <message> for the actual message)
```
pierluca, line 278: It's not a bad idea to put a longer (i.e. multi-line) message than what is convenient in the command line.
-m can also be omitted in those circumstances.

be2-scala/docs/README.md
```diff
@@ -224,22 +224,117 @@ val f: Future[GraphMessage] = (dbActor ? DbActor.Write(channel, message)).map {
...
+## Note: You may also add files separately using
+git add <fileName>
+
+## Note: You may also add all modified (!= new) files to the commit and set a title using
+git commit -am"<message>"
+
+## Note: You may also set a title and description for a specific commit using your favorite IDE (obviously doom emacs ^^). Then save and quit
+git add -A
+git commit
+```
```
pierluca, line 289: I would avoid doing this as it's very easy to commit the wrong things. 
I would suggest looking at `git add -p` , `git add -u`, `git diff --cached` and `git status` to make sure that you know exactly what is being committed.

be2-scala/docs/README.md
```diff
@@ -224,22 +224,117 @@ val f: Future[GraphMessage] = (dbActor ? DbActor.Write(channel, message)).map {
...
 The best way to "intercept" a `GraphMessage` being processed in the graph is to launch the server in debug mode, and then sending an isolated message to the server triggering the bug.
 
 :information_source: [Hoppscotch](https://hoppscotch.io/realtime/) (Realtime => WebSocket => `ws://localhost:8000/`) is a useful tool to achieve this result
 
 
-## Coding Styles
+## 6.	Coding Styles
 
-??? link + lint?
+A simple way to have a coherent style across the codebase is to use the IDE features of "code cleanup". For example, in IntelliJ, click on the `src/main/scala` folder and then on `Code -> Reformat Code`. You can then check "include subdirectories", "optimize imports", and "cleanup code" checkbox options. Be careful to not apply these changes to `src/test` folder as it transforms the scalatest syntax into a mess difficult to understand.
```
pierluca, line 339: is there a standard configuration file for the Scala IDEs?
Otherwise different IDEs might format things differently depending on the computer's settings.

If there's such a configuration (such as [EditorConfig](https://editorconfig.org/)), consider committing it.

pierluca (changes requested):
A few tiny remarks, feel free to disagree depending on what was discussed with the Scala team :)

be2-scala/docs/README.md
RaulinN: Yep! Putting this in the global docuementation file could be a good idea :) 

I only wrote very basic git commands in my opinion, but I think linking a git guide with what we have currently could be a good compromise ^^

be2-scala/docs/README.md
RaulinN: You mean simply doing something like this?
```bash
git commit -m"Start of a long comment
    \ end of the comment"
```

Not sure I grasp what you are saying 🤔

RaulinN:
Will do these changes very soon @pierluca  :)

be2-scala/docs/README.md
pierluca: If you just type `git commit`, it'll prompt you for a full commit message. The first line will show up in the git log, the following lines will be part of the extended description.

BadrTad (approved):

BadrTad (approved):

PR #464, opened 2021-09-25
JSON and JSON schema for the social media part

rjaccard:
I've added the JSON and JSON schema message data for the social media part.


lhmerino:
Decided on part 1 only for now. Will keep the work on the other parts in a separate branch.

PR #465, opened 2021-09-29
Use zerolog for logging 

celinecamacho:
All uses of `fmt.Println` and `log.Printf` have been modified to use `zerolog.Logger`. 

The only file that still needs to be modified is `channel/inbox/mod.go`, I hesitate between adding a log to the `Inbox` type or having a log in every function.

### DIFF
--- a/be1-go/channel/election/mod.go
+++ b/be1-go/channel/election/mod.go
@@ -3,7 +3,8 @@ package election
 import (
 	"encoding/base64"
 	"encoding/json"
-	"log"
+	"github.com/rs/zerolog"
+	"os"
 	"popstellar/channel"
 	"popstellar/channel/inbox"
 	"popstellar/crypto"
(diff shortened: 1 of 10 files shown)


rjaccard:
I went through the code and here are a few things that came to my attention :


(1) In cli/witness/witness.go we could add the role in the log when creating the hub, the clientSrv and the witnessSrv. Instead of giving log as argument we could give log.With().Str("role", "organizer").Logger()


(2) When we log an error the level of the log is not specified I believe. We should write log.Error().Err(err).Msg("") instead of log.Err(err).Msg("")


(3) Now we have :

const defaultLevel = zerolog.InfoLevel

var logout = zerolog.ConsoleWriter{
	Out:        os.Stdout,
	TimeFormat: time.RFC3339,
}
at the beginning of multiple files. We could maybe declare those var/const somewhere else and import it if needed. It would facilitate changing one of those variables.

nkcr (commented):
Couple of remarks:

- To avoid creating a log in multiple places we should have a global logger. For that we can add a `mod.go` at the root with an available logger. See https://github.com/dedis/dela/blob/master/mod.go
- Passing a log in argument to a function (except in initialization functions) feels wrong, so I would make that each package has an instance of a logger available (passed during initialization, of retrieved from the global logger). The advantage of passing a logger to an initialization function compared to using the global logger is that the logger can have additional context. For example the organizer logger has "role=organizer", and when it passes it to the lao channel, the lao channel can add additional context like "channel=lao" and, in all, have "role=organizer, channel=lao".

nkcr:
>  When we log an error the level of the log is not specified I believe. We should write log.Error().Err(err).Msg("") instead of log.Err(err).Msg("")

`.Err(err)` is a shortcut for `.Error().Err(err)`.

PR #471, opened 2021-10-01
Update valid data from protocol/examples

celinecamacho:
The values in commit [`3f03e5b`](https://github.com/dedis/student_21_pop/commit/3f03e5b734e20bcba4af22403d141daa934a1ea2), [`a0b6d84`](https://github.com/dedis/student_21_pop/commit/a0b6d84870ef6d0405007c1dfd903baf8b02e622) and [`4b5cc97`](https://github.com/dedis/student_21_pop/commit/4b5cc97009c1ed5e9d317b2f235ce25e10c949c0) were generated by using the fe1-web.

The remaining valid data was generated manually using the functions of fe2-android, you can see the files in commit [`a681d90`](https://github.com/dedis/student_21_pop/commit/a681d90edfce54d4c7d1554c28bd7eecde3ece51).

The files `lao_state.json`, `lao_update.json` and `meeting_state.json` are not updated, they still contain `XXX` as the field values. The reason for this is some confusion about the id, does the id field contain the LAO channel or the modified id?

The file `message_witness` has not been updated, I still have to generate the signature for the message.

The wrong and bad examples will come in another PR.


protocol/examples/messageData/election_setup.json
```diff
@@ -1,19 +1,19 @@
...
-    "name": "XXX",
-    "version": "XXX",
-    "created_at": 123,
-    "start_time": 123,
-    "end_time": 123,
+    "id": "JVehtpaFJwPc2dp8VuiskXwxmjTQGT6OiIVbei-EbBI=",
+    "lao": "fzJSZjKf-2cbXH7kds9H8NORuuFIRLkevJlN7qQemjo=",
+    "name": "Election",
+    "version": "1.0.0",
+    "created_at": 1633098941,
```
jbsv, line 8: Do we have a convention in the project on when to use "created_at" instead of "creation" (like in lao_create.json) ? Maybe the couple "begin"/"end" is clearer for the naming.

PR #472, opened 2021-10-02
[UX] Improved DatePicker

Xelowak:
To get more familiar with React and TypeScript, I decided to make the DatePicker cleaner and easier to use. 
Here is a description of what I've done :

- UI to create a meeting is more consistent with the other event creations
- Past dates are unclickable and displayed in gray
- End date is automatically set to start date + one hour at each change
- Handling of start < creation and end < start is now handled before sending the message to the network, to avoid possible issues
- Little bit of refactoring

I know I wrote code that is quite similar, but I don't really know how to be able to use a function in multiple files, like we do in Java. Moreover, it seems that it has not been done yet for a lot of functions. Feel free to tell me if you know how I can handle this better.

### DIFF
--- a/fe1-web/components/DatePicker.tsx
+++ b/fe1-web/components/DatePicker.tsx
@@ -12,6 +12,8 @@ const DatePicker = (props: IPropTypes) => {
       onChange={(date: any) => onChange(date)}
       dateFormat="MM/dd/yyyy HH:mm"
       showTimeInput
+      minDate={new Date()}
+      showDisabledMonthNavigation
     />
   );
 };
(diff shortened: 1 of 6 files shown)


fe1-web/components/DatePicker.tsx
```diff
@@ -30,3 +35,63 @@ DatePicker.defaultProps = {
...
 
 export default DatePicker;
+
+/**
+ * Transforms a date into a Timestamp.
+ *
+ * @param date the date to transform
+ */
+export function dateToTimestamp(date: Date): Timestamp {
+  return new Timestamp(Math.floor(date.getTime() / 1000));
```
jiabaow, line 45: divide by 1000 here is to convert from millisecond to second?

Issue #469, opened 2021-10-01
Add more logs to the Android application


GabrielFleischer:
Currently, the application lake some logs to find and understand anomalies.

The main ideas are the followings :

 - Understanding everything that comes in and out of the app.
 - Have a grasp of the path each execution takes

PR #480, opened 2021-10-05
Work on the logging system of the android application

GabrielFleischer:
The pull request aims to improve the logs of the android application.

It adds logs at some key point of the app
And it adds a lot of `toString()` methods to the data models.

Closing #469


jbsv:
Fe2 doesn't build: probably because startSubscription() should be changed to subscribeToUpstream() in LAORepository()

PR #487, opened 2021-10-09
Copying the token made easier on the web

Xelowak:
The wallet screen now displays a text input filled with the token, that we can copy to clipboard by clicking on the copy button.
If you try to select the text, it is still interpreted as a swipe though, so I strongly suggest using the button.
I don't know if it's implemented in fe2, but I think it might be a good idea to do so. 
Here is a screenshot :

Capture d’écran 2021-10-09 à 16 29 30
:

### DIFF
--- a/fe1-web/components/CopiableTextInput.tsx
+++ b/fe1-web/components/CopiableTextInput.tsx
@@ -32,7 +32,7 @@ const CopiableTextInput = (props: IPropTypes) => {
 
   return (
     <View style={styles.view}>
-      <TextInput value={text} style={styles.textInput} />
+      <TextInput value={text} style={styles.textInput} selectTextOnFocus />
       <CopyButton data={text} />
     </View>
   );


Xelowak:
> If I understand correctly, it seems that you're using the `TextInput` as a workaround for making the text copiable, as opposed to addressing the more fundamental issue of swiping.
> 
> However, none of the data you're displaying is actually for input, it's just displaying. Are you sure it's a good idea to address it as such, from a user interaction perspective? Wouldn't it make sense to address the swiping problem instead? Or am I missing something?

No, this PR is not about solving the swiping issue, otherwise I would have mentioned it in the description. The problem is still there, even if you try to select text in the TextInput.
Actually, Céline and I (UX team) talked about the fact that it wasn't clear enough visually that the token was something that needed to be copied in clipboard. When I first used the application, I didn't realize that this was the actual token. Moreover, there wasn't any button that allowed the user to copy the text, even if there was already a React component capable of it. That's why we decided that a small task could be to do it in fe1 and fe2. On Java specifically, we had some issue trying to copy it, so I think it might be nice to do it there too (Céline told me she would talk with the Java team). 

Yes, that's true, it's not meant to be an input. However, when you display text that needs to be copied in an applicaion, it is more common to show it in an input box that cannot be modified, than as a simple text. That's why we decided to change this.

pierluca:
I'm convinced :) merging

PR #526, opened 2021-10-22
Wrong/bad LAO data examples

celinecamacho:
The wrong examples represent the messages that don't pass the JSON Schema validation, for examples adding new parameters to a lao#create message.
The bad examples represent the messages that pass the JSON Schema validation but are not valid for the application, for example a lao#create id that is not computed as `Hash : SHA256(organizer||creation||name)`.


BadrTad (dismissed):

protocol/test/package-lock.json
```diff
@@ -2430,7 +2429,6 @@
                 "@types/node": "*",
                 "anymatch": "^3.0.3",
                 "fb-watchman": "^2.0.0",
-                "fsevents": "^2.3.2",
```
celinecamacho, line 2433: Is modifying this file an issue? It was modified automatically after running `npm install`. @pierluca

pierluca: that's okay. @Xelowak should now be able to tell you all you ever want to know about package-lock.json :-P

pierluca (approved):
LGTM

PR #537, opened 2021-10-27
Add value to json-rpc and id field of error 

celinecamacho:
The json-rpc field is now 2.0 for every error answer sent.
The id field is not null when possible. The errors from invalid schema, invalid RPC type, invalid method and unmarshal issues will have a null value. 

Will test in another PR when PR #526 will be merged.


rjaccard (dismissed):
looks fine to me 👍

Anghul (dismissed):
Everything seems fine

nkcr (approved):

Issue #540, opened 2021-10-27
Information missing about the Question ID in an election


tuomas-p:
### Description
Unlike the other base64 hashes, which are described in the documentation, the ElectionQuestion's (be2-scala/src/main/scala/ch/epfl/pop/model/network/method/message/data/election/ElectionQuestion.scala) id hash doesn't seem to be described in the documentation. It is however necessary for validations in the JsonSchemas relating to elections. I haven't been able to find the information in other front-ends or back-ends yet.

##### Back-ends:
- [ ] Be2-Scala

### Possible root cause
Missing comments from previous semesters (probably).

PR #541, opened 2021-10-28
Add election data on protocol documentation

celinecamacho:
As pointed out on the issue #540 the election data is missing on the protocol documentation. 

With this PR I added the documentation for every election data (setup, cast vote, end and result). I also added the question id description to the election setup JSON Schema for clarity.


docs/messageData.md
```diff
@@ -1025,3 +1030,496 @@ the organizer forgets to scan an attendee’s public key.
...
+    "object": "election",
+    "action": "cast_vote",
+    "lao": "fzJSZjKf-2cbXH7kds9H8NORuuFIRLkevJlN7qQemjo=",
+    "election": "QWTmcWMMMiUdWdZX7ib7GyqH6A5ifDYwPaMpKxIZm1k=",
+    "created_at": 1633098996,
+    "votes": [
+        {
+            "id": "DtIsj7nQ0Y4iLJ4ETKv2D0uah7IYGyEVW7aCLFjaL0w=",
+            "question": "WBVsWJI-C5YkD0wdE4DxnLa0lJzjnHEd67XPFVB9v3g=",
+            "write_in": "Computer Science"
```
pierluca, line 1247: isn't this functionality unsupported in practice?

docs/messageData.md
```diff
@@ -101,7 +110,7 @@ broadcast”).
         "id": {
             "type": "string",
             "contentEncoding": "base64",
-            "$comment": "Hash : SHA256(organizer||creation||name)"
+            "$comment": "Hash : HashLen(organizer, creation, name)"
```
pierluca, line 113: is HashLen defined somewhere?

docs/messageData.md
```diff
@@ -1025,3 +1030,496 @@ the organizer forgets to scan an attendee’s public key.
...
+```
+
+## Casting a vote (election#cast_vote)
+
+🧭 **RPC Message** > **RPC payload** (*Query*) > **Query payload** (*Publish*) >
+**Mid Level** > **High level** (*election#cast_vote*)
+
+A member of the LAO can cast a vote by publishing an election/cast_vote message to the
+election’s channel. Each member may cast multiple votes, only the last one will be counted.
+If write-in is allowed for the election then the vote has to have a write-in.
```
pierluca, line 1208: Sentence may be worth clarifying:
Is it that the write-in is mandatory for the election, and so the field is mandatory,
or is it that the write-in is allowed for the election, and so the field is mandatory (but may be null) ?

pierluca (commented):
Had a cursory look. Seems good in principle. Just minor comments

docs/messageData.md
celinecamacho: Yes indeed. Should I explain that write-in is unsupported in practice or delete the example?

docs/messageData.md
celinecamacho: Yes, in the [`docs/protocol.md`](https://github.com/dedis/student_21_pop/blob/master/docs/protocol.md#concatenation-for-hashing) file

docs/messageData.md
celinecamacho: The cast vote either has a vote field or a write-in field, in the schema it is defined as a `oneOf`

pierluca (approved):
Sorry, I'd missed that you had replied. Chase me if a PR gets stuck because of me.

I'll merge this because it's an improvement already.

Let's deal with the write in in a separate PR, as temporarily disabling it or removing it from the schema might break the application without additional changes.

PR #543, opened 2021-10-28
Shutdown server on SIGTERM

BadrTad:
Now, server terminates on` SIGTERM (Ctrl+C)` instead of `RETURN`.

PR #550, opened 2021-10-28
Fixes the bug of open roll call

BadrTad:
Double bugs:
A null ID gets included in the RPC message, causing the front-end-1 to reject it, the method was incorrect (`message `instead of `broadcast`, it has now been fixed according to the protocol specifications. 
This patch corrects this issue, I'll be providing more tests soon to cover the format specifications 😃

### DIFF
--- a/be2-scala/src/main/scala/ch/epfl/pop/config/application.conf
+++ b/be2-scala/src/main/scala/ch/epfl/pop/config/application.conf
@@ -1,10 +1,7 @@
 ## Akka configs 
-akka.http.server.idle-timeout = infinite   #  
-akka.http.server.idle-timeout = infinite   # for testing purposes only
 
-#akka.http.client.idle-timeout =  30 s 
-#akka.http.server.idle-timeout =  30 s  
-
+akka.http.client.idle-timeout =  600 s 
+akka.http.server.idle-timeout =  600 s  
 akka.http.server.request-timeout = 600 s
 
 #POP Server config default
--- a/be2-scala/src/main/scala/ch/epfl/pop/config/application.conf
+++ b/be2-scala/src/main/scala/ch/epfl/pop/config/application.conf
@@ -1,7 +1,9 @@
 ## Akka configs 
-akka.http.server.idle-timeout =  5 s  
-#akka.http.server.idle-timeout = infinite   # for testing purposes only
-akka.http.client.idle-timeout =  5 s 
+akka.http.server.idle-timeout = infinite   #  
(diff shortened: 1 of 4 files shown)


be2-scala/src/main/scala/ch/epfl/pop/json/HighLevelProtocol.scala
```diff
@@ -163,17 +163,20 @@ object HighLevelProtocol extends DefaultJsonProtocol {
...
-        case Some(idx) => idx.toJson
-        case _ => JsNull
+
+      var jsObjectContent: ListMap[String, JsValue] = ListMap.empty[String, JsValue]
+
+      jsObjectContent += (PARAM_JSON_RPC -> obj.jsonrpc.toJson)
+      jsObjectContent += (PARAM_METHOD -> obj.method.toJson)
+      jsObjectContent += (PARAM_PARAMS -> obj.params.toJson)
+
+      /*Add the id key iif it's non null*/
```
pierluca, line 173: `if` or `iff` I assume? :-)

be2-scala/src/main/scala/ch/epfl/pop/config/application.conf
```diff
@@ -1,8 +1,7 @@
 ## Akka configs 
-akka.http.server.idle-timeout =  5 s  
-#akka.http.server.idle-timeout = infinite   # for testing purposes only
-akka.http.client.idle-timeout =  5 s 
 
+akka.http.client.idle-timeout =  600 s 
+akka.http.server.idle-timeout =  600 s
```
pierluca, line 4: is there a reason for switching from 5s to 600s to fix the open roll-call bug?

pierluca (changes requested):
A couple of minor remarks.

Other than that, it seems the commit history is completely mixed with #543 , you might want to clean-up the PR so it does only one thing.

be2-scala/src/main/scala/ch/epfl/pop/config/application.conf
BadrTad: No the timeouts does not make any difference

be2-scala/src/main/scala/ch/epfl/pop/json/HighLevelProtocol.scala
BadrTad: yes it should be iff

BadrTad (commented):
The github history has been cleaned.

pierluca (approved):
Very clean ! 👍 
LGTM

Issue #552, opened 2021-10-28
[BUG] Camera preview crash the app (Connect to LAO or scan)


arnauds5:
### Description (Actual behavior)
When accessing a view with the camera preview for example connecting to a LAO or adding a witness, the app crash with this error 
```
2021-10-28 21:53:54.161 5423-5423/com.github.dedis.popstellar E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.github.dedis.popstellar, PID: 5423
    java.lang.NullPointerException: Attempt to invoke virtual method 'int com.google.android.gms.common.images.Size.getWidth()' on a null object reference
        at com.github.dedis.popstellar.ui.qrcode.CameraPreview.onLayout(CameraPreview.java:101)
        at android.view.View.layout(View.java:21912)
        at android.view.ViewGroup.layout(ViewGroup.java:6260)
        at androidx.constraintlayout.widget.ConstraintLayout.onLayout(ConstraintLayout.java:1873)
        ...
```

### Expected behavior
Not crashing

### How to reproduce
1. Click on Connect

### Version & Environment
This bug was reproduced on:
- Commit: 8e499053f87fcdb506f020b65180c4338367f7ce

##### Front-ends:
- [x] Fe2-Android

##### Environment (as applicable):
 - Device 1: Emulator, Pixel 2 API 29
 - Device 2: Oneplus 2, Android 10

### Workaround
Replace the code on CameraPreview at [line 100](https://github.com/dedis/student_21_pop/blob/86783c5fc01538ae85c471daed920c27cdcdf306/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/qrcode/CameraPreview.java#L100) by that (it was that before PR #512)
```
if (size != null) {
  height = size.getWidth();
  width = size.getHeight();
}
```

### Impact
Not possible to scan anything.

### Possible root cause
CameraPreview, line 100 : [mCameraSource.getPreviewSize()](https://github.com/dedis/student_21_pop/blob/86783c5fc01538ae85c471daed920c27cdcdf306/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/qrcode/CameraPreview.java#L100), even if it has the annotation `RecentlyNonNull`, and even if Android Studio says that `Condition 'size != null' is always 'true'`, it can return null. And if it's null, the app crash the line just after.

Issue #546, opened 2021-10-28
[BUG] The android application cannot open the Witness Messages UI


GabrielFleischer:
### Description (Actual behavior)
When clicking on the Witness Messages in an LAO ui, the app crashes.

### Expected behavior
The UI should be opened

### How to reproduce
1. Go to any LAO UI
2. Click on the 'Witness Messages' button

### Version & Environment
This bug was reproduced on:
- Commit: 9433e1731f0ebde5c934beb3e0b303db89a0db27

##### Front-ends:
- [ ] Fe1-Web (please include browser's names & version)
- [x] Fe2-Android (emulation, Pixel 2 API 29)
- [ ] Not applicable

##### Back-ends:
- [x] Be1-Go
- [ ] Be2-Scala
- [ ] Not applicable

##### Environment (as applicable):
 - Device: Pixel 2 emulator on Windows

### Workaround
No workaround found

### Impact
The witness messages cannot be seen

### Possible root cause
It seems that there are duplicated ids in the UIs. And the tab_back id points to both a Button and a ConstraintLayout.

This create a class cast exception when a Button is expected where a ConstraintLayout is given.

PR #553, opened 2021-10-28
Fix two bug that crash the app on some view

arnauds5:
- Add a null check to getPreviewSize() in CameraPreview.java => #552 
- Rename tab_back to tab_home_properties in the tab_properties.xml => #546


Lugiasc (approved):

jbsv (approved):
LGTM. Thanks

PR #558, opened 2021-11-02
Disable swipes in a non-deprecated way

Xelowak:
It seems that my way of disabling swipes in #525 was deprecated and I didn't notice (my bad). Here is a non-deprecated version of it.


pierluca (approved):
LGTM

PR #561, opened 2021-11-03
Add test package and some tests for TextInputChirp

Xelowak:
I added the [testing  library for React Native](https://testing-library.com/docs/react-native-testing-library/intro/), along with some tests for TextInputChirp. I added the react-renderer package too, as it is a peer dependency of the testing library.

I realized that the publish button should not be enabled until you write something in the text input, so I changed this along the way.


jiabaow (approved):

fe1-web/components/__tests__/TextInputChirp.test.tsx
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
+import { fireEvent, render } from '@testing-library/react-native';
+import STRINGS from 'res/strings';
+import TextInputChirp from '../TextInputChirp';
+
+const onChangeText = jest.fn();
+const onPress = jest.fn();
```
pierluca, line 7: you might want to have these initialized in a `beforeEach` call, to avoid mocks being shared across tests (and thus making the tests not-independent)

pierluca (changes requested):
Tiny suggestion, the rest LGTM :)

fe1-web/components/__tests__/TextInputChirp.test.tsx
Xelowak: True, I didn't think about that since my functions don't do anything.

pierluca (approved):

PR #563, opened 2021-11-04
AnswerValidatorTestSuite and JRPC broadcast parse bug fix   

BadrTad:
+ This PR is still a draft, because more dependency testing should be done ;)
+ Your review is always appreciated


be2-scala/src/test/scala/ch/epfl/pop/pubsub/graph/validator/SchemaValidatorSuite.scala
```diff
@@ -0,0 +1,62 @@
...
+
+    validateSchema(subscribeJson) should be (Left(subscribeJson))
+  }
+
+  test("Incorrect subscribe query: additional params"){
+    val subscribePath = "../protocol/examples/query/subscribe/wrong_subscribe__additional_params.json"
+    val subscribeJson = getJsonStringFromFile(subscribePath)
+
+    validateSchema(subscribeJson) shouldBe a [Right[_,PipelineError]]
+  }
```
RaulinN, line 32: Nice start! Glanced over the file real quick and noticed that you should probably check that the pipeline error is indeed what we expect (i.e. correct error number or indeed contains an id)

PR #568, opened 2021-11-07
Add tests for lao channel

rjaccard:


be1-go/channel/lao/mod_test.go
```diff
@@ -25,7 +26,114 @@ import (
...
+
+	var message method.Subscribe
+	err = json.Unmarshal(buf, &message)
+	require.NoError(t, err)
+
+	socket := &fakeSocket{id: "socket"}
+
+	channel.Subscribe(socket, message)
+
+	require.True(t, laoChannel.sockets.Delete("socket"))
```
celinecamacho, line 57: Wait, this means that we let clients subscribe to LAOs that do not exist?

rjaccard: No, the hub does the check that the channel, we want to subscribe to, exists and if it exists then the LAO has been created. In the test, we bypass the hub by calling Subscribe directly on the channel. Maybe we can add another check in Subscribe to be sure that the channel, we want to subscribe to, corresponds to the channel that we actually subscribe to.

Anghul: I want to add that as the function is channel.Subscribe(), channel has to exist to be able to call the function

Anghul (dismissed):
Good work

Anghul (approved):

be1-go/channel/lao/mod_test.go
```diff
@@ -21,7 +26,115 @@ import (
...
+
+	file := filepath.Join(relativePath, "broadcast.json")
+	buf, err := os.ReadFile(file)
+	require.NoError(t, err)
+
+	var message method.Broadcast
+	err = json.Unmarshal(buf, &message)
+	require.NoError(t, err)
+
+	require.Error(t, channel.Broadcast(message))
```
nkcr, line 134: please explain why it should fail

be1-go/channel/lao/mod_test.go
```diff
@@ -21,7 +26,115 @@ import (
...
+	require.NoError(t, err)
+
+	var message method.Unsubscribe
+	err = json.Unmarshal(buf, &message)
+	require.NoError(t, err)
+
+	socket := &fakeSocket{id: "socket"}
+	laoChannel.sockets.Upsert(socket)
+
+	require.NoError(t, channel.Unsubscribe("socket", message))
```
nkcr, line 87: Some additional checks you can do: 
- check that the socket has been deleted (can use socket.Delete and check for false)
- check that a second call to unsubscribe returns an error

be1-go/channel/lao/mod_test.go
```diff
@@ -67,22 +180,139 @@ func TestBaseChannel_RollCallOrder(t *testing.T) {
...
+
+	m1 := message.Message{
+		Data:              bufb64,
+		Sender:            publicKey64,
+		Signature:         "h",
+		MessageID:         messagedata.Hash(bufb64, publicKey64),
+		WitnessSignatures: []message.WitnessSignature{},
+	}
+
+	relativePathPub := filepath.Join("..", "..", "..", "protocol",
```
nkcr, line 211: we should move the protocol examples relative path to a constant

be1-go/channel/lao/mod_test.go
```diff
@@ -67,22 +180,139 @@ func TestBaseChannel_RollCallOrder(t *testing.T) {
...
+	require.NoError(t, err)
+
+	var messagePublish method.Publish
+
+	err = json.Unmarshal(bufPub, &messagePublish)
+	require.NoError(t, err)
+
+	messagePublish.Params.Message = m1
+
+	require.NoError(t, channel.Publish(messagePublish))
```
nkcr, line 225: The test a rather weak, as we only check for the absence of an error, and not expected actions performed. 

We can't do better right now and we don't have to change it for this PR, but this is an interesting exercice: could you think about how to change the channel implementation, so that it can be better tested ? In this case we want to check that, provided with a publish/lao#update message, the channel calls the expected function associated to the message data.

nkcr (changes requested):
I have some minor comments, but very nice overall! :)

nkcr (approved):

PR #569, opened 2021-11-09
Upgrade android libraries versions

GabrielFleischer:
While writing some tests, I added a library that needed SDK 31.

So I upgraded it and took time to upgrade other libraries in the meantime.

### DIFF
--- a/fe2-android/app/src/main/AndroidManifest.xml
+++ b/fe2-android/app/src/main/AndroidManifest.xml
@@ -18,7 +18,7 @@
     android:theme="@style/AppTheme">
     <activity
       android:name="com.github.dedis.popstellar.ui.home.HomeActivity"
-      android:exported="false">
+      android:exported="true">
       <intent-filter>
         <action android:name="android.intent.action.MAIN" />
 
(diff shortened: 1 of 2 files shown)


fe2-android/app/src/main/AndroidManifest.xml
```diff
@@ -16,7 +16,9 @@
     android:roundIcon="@mipmap/ic_launcher_round"
     android:supportsRtl="true"
     android:theme="@style/AppTheme">
-    <activity android:name="com.github.dedis.popstellar.ui.home.HomeActivity">
+    <activity
+      android:name="com.github.dedis.popstellar.ui.home.HomeActivity"
+      android:exported="false">
```
arnauds5, line 21: With exported="false", the app won't launch and the IDE says "A launchable activity must be exported as of Android 12"
```suggestion
      android:exported="true">
```

arnauds5 (changes requested):

fe2-android/app/src/main/AndroidManifest.xml
GabrielFleischer: This is ... weird ... I don't have this issue and from the documentation, and exported activity can be launched from other apps.

I will fix this

arnauds5: I was using the emulator with Pixel 2, API 29.
Do we need to update something in android studio ?

arnauds5 (approved):

jbsv (approved):
LGTM

PR #574, opened 2021-11-10
Consensus stage 2 frontend

arnauds5:
- Add message data : Prepare/Propose/Promise/Accept (with separate classes to hold value)
- Update Learn message
- Update ElectionStartFragment
- Update ConsensusHandler (proposer no longer send the learn message in stage 2)
- Add tests for message data (mostly identical and testing property/equals/toString/jsonValidation)
- Add tests for ConsensusHandler/ConsensusNode
- Fix a missing constraint on dataAccept json schema and a test
- Fix the issue #630 by replacing base url in json files (locally)


fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/consensus/LearnValue.java
```diff
@@ -0,0 +1,42 @@
+package com.github.dedis.popstellar.model.network.method.message.data.consensus;
+
+import com.google.gson.annotations.SerializedName;
+
+import java.util.Objects;
+
+public final class LearnValue {
+
+  @SerializedName("decision")
```
GabrielFleischer, line 9: Is this necessary ?

GabrielFleischer (approved):
Looks good to me

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/consensus/LearnValue.java
arnauds5: No, the @SerializedName is redundant because the variable name is the same.
But keeping it could maybe avoid problems if someone in a few years try to refactor some of theses classes.

fe2-android/app/build.gradle
```diff
@@ -186,6 +186,11 @@ copy {
     include 'examples/**'
     include 'query/**'
     include 'jsonRPC.json'
+
+    // Replace the remote part of the base URL ($id) to the local resource path
+    filter {
+        String line -> line.replaceAll("https://raw\\.githubusercontent\\.com/dedis/[a-zA-Z0-9_-]+/master/", "resource:/")
+    }
```
GabrielFleischer, line 193: This seems pretty straightforward.

It should still fix the issue.

Maybe add a link to the issue to explain the reason behind this replacement.

GabrielFleischer (dismissed):
LGTM

jbsv:
Nice job. I like the clean code. I am just missing some documentation. Can you please add documentation for the principal methods (and their arguments) of Consensus, as well as document the purpose of each test ?

GabrielFleischer:
Why is the coverage so low all of a sudden ?

PR #575, opened 2021-11-10
Refactor Modals

Xelowak:
As stated in PR #472, I refactored the Modals by creating 2 generic React components, namely DismissModal and ConfirmModal. 
I redesigned both of them too, as you can see on the screenshots :

Capture d’écran 2021-11-09 à 18 39 13

Capture d’écran 2021-11-09 à 18 39 50


jiabaow (approved):

jiabaow (approved):
LGTM

fe1-web/components/ConfirmModal.tsx
```diff
@@ -0,0 +1,75 @@
...
+};
+
+const propTypes = {
+  visibility: PropTypes.bool.isRequired,
+  setVisibility: PropTypes.func.isRequired,
+  title: PropTypes.string.isRequired,
+  description: PropTypes.string.isRequired,
+  buttonCancelText: PropTypes.string,
+  buttonConfirmText: PropTypes.string,
+  onConfirmPress: PropTypes.func.isRequired,
```
pierluca, line 55: I'm going to be very annoying here (same thing applies to the other modal), but it seems to me that the API is mixing different level of abstractions.

On one hand we have `onConfirmPress`, which is indicating the result of the modal. It seems like an obvious abstraction.
On the other hand, we have `setVisibility`, which seems like a concern for the caller of the modal.

I'd really consider having just `onConfirm` and `onCancel`, and let the callers handle the visibility (ideally in *both* cases, not just on cancel).

I understand this would lead to a tiny bit of code repetition, but it should also lead to a clearer separation of concerns and easier re-usability for the modals.

pierluca (commented):
Looks very very good to me, I just have a minor API concern for the modals.

PR #581, opened 2021-11-12
Add check for correct signature

rjaccard:


be1-go/hub/organizer/mod_test.go
```diff
@@ -460,12 +462,11 @@ type keypair struct {
 }
 
 var nolog = zerolog.New(io.Discard)
-var suite = crypto.Suite
+//var suite = crypto.Suite
```
Anghul, line 465: Does it still work if you remove the variable ? It seems to be used in generateKeyPair

be1-go/hub/organizer/mod.go
```diff
@@ -291,6 +295,30 @@ func (h *Hub) handleMessageFromClient(incomingMessage *socket.IncomingMessage) e
 	return nil
 }
 
+func verifySignature(publicKey64 string, signature64 string, data64 string) error {
+	pk, err := base64.URLEncoding.DecodeString(publicKey64)
+	if err != nil {
+		return xerrors.Errorf("failed to decode the public key")
```
Anghul, line 301: It could be a good idea to return err too, and same comment for the returns below

Anghul (changes requested):
Good job overall, just a few errors that could be returned to have more complete logs

be1-go/hub/organizer/mod_test.go
rjaccard: I moved it to mod.go

be1-go/hub/organizer/mod.go
rjaccard: nice catch!

Anghul (approved):

celinecamacho (approved):
LGTM

be1-go/hub/organizer/mod.go
```diff
@@ -291,6 +295,30 @@ func (h *Hub) handleMessageFromClient(incomingMessage *socket.IncomingMessage) e
...
+	}
```
nkcr, line 302: missing return

be1-go/hub/organizer/mod.go
```diff
@@ -291,6 +295,30 @@ func (h *Hub) handleMessageFromClient(incomingMessage *socket.IncomingMessage) e
...
+	if err != nil {
+		return xerrors.Errorf("failed to decode the signature: %v", err)
+	}
+
+	data, err := base64.URLEncoding.DecodeString(data64)
+	if err != nil {
+		return xerrors.Errorf("failed to decode the data: %v", err)
+	}
+
+	return schnorr.VerifyWithChecks(crypto.Suite, pk, data, signature)
```
nkcr, line 319: I'm not a big fan of returning an invocation of a function. It makes the code harder to follow and can make debugging harder, with less meaningful error message.

nkcr (approved):

PR #582, opened 2021-11-12
Add consensus frontend tests

arnauds5:
## What is tested
- display of election title, status, start button, nodes grid
- the status and start button are updated when start time is passed
- receiving messages update the nodes buttons and text
- messages sent by buttons
- consensus message handler

### DIFF
--- a/fe2-android/app/src/androidTest/java/com/github/dedis/popstellar/pages/qrcode/ElectionStartPageObject.java
+++ b/fe2-android/app/src/androidTest/java/com/github/dedis/popstellar/pages/qrcode/ElectionStartPageObject.java
@@ -1,0 +1,34 @@
+package com.github.dedis.popstellar.pages.qrcode;
+
+import static androidx.test.espresso.Espresso.onData;
+import static androidx.test.espresso.Espresso.onView;
+import static androidx.test.espresso.matcher.ViewMatchers.withId;
+import static org.hamcrest.CoreMatchers.instanceOf;
+import static org.hamcrest.Matchers.allOf;
+import static org.hamcrest.Matchers.is;
+
+import androidx.test.espresso.DataInteraction;
+import androidx.test.espresso.ViewInteraction;
+
+import com.github.dedis.popstellar.R;
+import com.github.dedis.popstellar.model.objects.ConsensusNode;
+
+public class ElectionStartPageObject {
+
+  public static ViewInteraction electionTitle() {
+    return onView(withId(R.id.election_title));
+  }
+
+  public static ViewInteraction electionStatus() {
+    return onView(withId(R.id.election_status));
+  }
+
+  public static ViewInteraction electionStartButton() {
+    return onView(withId(R.id.election_start));
+  }
+
+  public static DataInteraction nodesGrid() {
+    return onData(allOf(is(instanceOf(ConsensusNode.class))))
+        .inAdapterView(withId(R.id.nodes_grid));
+  }
+}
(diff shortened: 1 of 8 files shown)


fe2-android/app/src/androidTest/java/com/github/dedis/popstellar/pages/qrcode/ElectionStartPageObject.java
```diff
@@ -0,0 +1,34 @@
+package com.github.dedis.popstellar.pages.qrcode;
```
GabrielFleischer, line 1: I think this should be placed in `com.github.dedis.popstellar.pages.detail.event.consensus`

fe2-android/app/src/androidTest/java/com/github/dedis/popstellar/ui/detail/event/consensus/ElectionStartFragmentTest.java
```diff
@@ -0,0 +1,278 @@
...
+    lao.setWitnesses(Sets.newSet(node2Key, node3Key));
+
+    laoRepository.getLaoById().put(laoChannel, new LAOState(lao));
+    laoRepository.updateNodes(laoChannel);
+  }
+
+  @After
+  public void clean() {
+    LAORepository.destroyInstance();
+  }
```
GabrielFleischer, line 149: The @After is called after every tests.

As there is only one, this is not a big issue. But we might want to pass to JUnit 5 to get access to @BeforeAll and @AfterAll.

Or, you can create and destroy the LAORepository instance for each tests with @Before and @After, which makes them completely independent.

fe2-android/app/src/androidTest/java/com/github/dedis/popstellar/ui/detail/event/consensus/ElectionStartFragmentTest.java
```diff
@@ -0,0 +1,278 @@
...
+  private String getPublicKey() throws IOException, GeneralSecurityException {
+    KeysetHandle publicKeysetHandle =
+        Injection.provideAndroidKeysetManager(ApplicationProvider.getApplicationContext())
+            .getKeysetHandle()
+            .getPublicKeysetHandle();
+    return Keys.getEncodedKey(publicKeysetHandle);
+  }
+
+  @Test
+  public void test() throws InterruptedException {
```
GabrielFleischer, line 160: This test name is not really meaningful.
What are you testing exactly ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/LaoDetailViewModel.java
```diff
@@ -951,6 +951,10 @@ public Lao getCurrentLaoValue() {
     return mCurrentLao.getValue();
   }
 
+  public void setCurrentLao(Lao lao) {
```
GabrielFleischer, line 954: I guess this is a solution. It could be a good idea to add 
```suggestion
  @VisibleForTesting
  public void setCurrentLao(Lao lao) {
```

So that the method is only visible for testing.

GabrielFleischer (changes requested):
Great overall, some slight changes are needed for the test to be truly maintainable.

fe2-android/app/src/androidTest/java/com/github/dedis/popstellar/ui/detail/event/consensus/ElectionStartFragmentTest.java
arnauds5: I changed it to `displayWithUpdatesIsCorrectAndButtonsProduceCorrectMessages`,
I'm testing that everything is correctly display, with some updates. I also test  that the buttons produces the correct messages.

fe2-android/app/src/androidTest/java/com/github/dedis/popstellar/ui/detail/event/consensus/ElectionStartFragmentTest.java
arnauds5: I made a single test because most of the test is sequential and need the previous updates to work.
The setup function need to be called before the fragment is fully initialized, so it can't be used with @before.
And the clean function is used only to be sure that it won't interfere with test in others files.

GabrielFleischer (approved):
LGTM

GabrielFleischer (approved):

Lugiasc (approved):

jbsv (approved):

PR #583, opened 2021-11-12
Deactivate the DEBUG logs and provide resources to jar

BadrTad:
Provide logback.xml to configure logger

### DIFF
--- a/be2-scala/.gitignore
+++ b/be2-scala/.gitignore
@@ -19,3 +19,6 @@ database
 
 # Test
 database_test
+
+#Protocol folder
+src/main/resources/protocol
--- a/be2-scala/build.sbt
+++ b/be2-scala/build.sbt
@@ -30,6 +30,7 @@ copyProtocolTask := {
 }
 //Add task to compile time
 (Compile/ compile) := ((Compile/ compile) dependsOn copyProtocolTask).value
+resourceDirectory in (Compile, packageBin) := file(".") / "./src/main/resources"
 
 //Setup main calass task context/confiuration
 mainClass in (Compile, run) := Some("ch.epfl.pop.Server")
@@ -39,7 +40,6 @@ lazy val scoverageSettings = Seq(
   coverageEnabled in Compile := true,
   coverageEnabled in Test := true,
   coverageEnabled in packageBin := false,
-
 )
 
 
@@ -65,6 +65,12 @@ sonarProperties := Map(
   "sonar.scala.scapegoat.reportPaths" -> "./target/scala-2.13/scapegoat-report/scapegoat.xml"
 )
 
+assemblyMergeStrategy in assembly := {
+    case PathList("module-info.class") => MergeStrategy.discard
+    case PathList("reference.conf") => MergeStrategy.concat
+    case PathList("META-INF","MANIFEST.MF") => MergeStrategy.discard
+    case _ => MergeStrategy.defaultMergeStrategy("")
+}
 
 // For websockets
 val AkkaVersion = "2.6.8"
--- a/be2-scala/build.sbt
+++ b/be2-scala/build.sbt
@@ -1,4 +1,6 @@
+import scala.util.{Try, Success, Failure}
 import sbtsonar.SonarPlugin.autoImport.sonarProperties
+import sbt.IO._
 
 name := "pop"
 
@@ -6,6 +8,30 @@ version := "0.1"
 
 scalaVersion := "2.13.5"
 
+//Create task to copy the protocol folder to resources
+lazy val copyProtocolTask = taskKey[Unit]("Copy protocol to resources")
+copyProtocolTask := {
+    val log = streams.value.log
+    log.info("Executing Protocol folder copy...")
+    val scalaDest = "be2-scala"
+    baseDirectory.value.name
+    if(! baseDirectory.value.name.equals(scalaDest)){
+        log.error(s"Please make sure you working dir is $scalaDest !")
+    }else{
+        val source = new File("../protocol")
+        val dest   = new File("./src/main/resources/protocol")
+        Try(IO.copyDirectory(source,dest, true)) match {
+            case Success(_) => log.info("Copied !!")
+            case Failure(exception) =>
+                log.error("Could not copy protocol to ressource folder")
+                exception.printStackTrace()
+        }
+    }
+}
+//Add task to compile time
+(Compile/ compile) := ((Compile/ compile) dependsOn copyProtocolTask).value
+
+//Setup main calass task context/confiuration
 mainClass in (Compile, run) := Some("ch.epfl.pop.Server")
 mainClass in (Compile, packageBin) := Some("ch.epfl.pop.Server")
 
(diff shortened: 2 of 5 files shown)


be2-scala/src/main/resources/logback.xml
```diff
@@ -0,0 +1,14 @@
...
+<configuration debug="false" scan="true" scanPeriod="15 seconds">
+  <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+    <encoder>
+      <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
+    </encoder>
+  </appender>
+
+<!--Set root(general) logs level to INFO/DEBUG on STDOUT-->
+  <root level="INFO">
+    <appender-ref ref="STDOUT"/>
```
pierluca, line 12: This would be a separate PR, but if you want to take advantage of [Apache Chainsaw](https://logging.apache.org/chainsaw/2.x/), you could configure a socket appender.
It's really super helpful when debugging locally.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/Validator.scala
```diff
@@ -60,7 +59,8 @@ object Validator {
     val factory: JsonSchemaFactory = JsonSchemaFactory.getInstance(SpecVersion.VersionFlag.V7)
     // Creation of a JsonNode using the readTree function from the file query.json (at queryPath)
     // FIXME: error handling for queryPath
-    val jsonSchemaNode: JsonNode = objectMapper.readTree(new File(queryPath))
+    val jsonSchemaNode: JsonNode = objectMapper.readTree(queryFile)
+    queryFile.close()
```
pierluca, line 63: won't this lead to a failure the next time `validateSchema` is called? 

Also, for readability, it might make sense to create queryFile right before it's used, use it, and close it afterwards.

be2-scala/build.sbt
```diff
@@ -1,11 +1,40 @@
...
+
+//Create task to copy the protocol folder to resources
+lazy val copyProtocolTask = taskKey[Unit]("Copy protocol to resources")
+copyProtocolTask := {
+    val log = streams.value.log
+    log.info("Executing Protocol folder copy...")
+    val scalaDest = "be2-scala"
+    baseDirectory.value.name
+    if(! baseDirectory.value.name.equals(scalaDest)){
+        log.error(s"Please make sure you working dir is $scalaDest !")
```
pierluca, line 21: very minor point, but is there no way to do away with this ? can't we take the `build.sbt` location as a reference ?
below there are many other files that seem to be referenced with relative paths.

pierluca (commented):
Had a quick look, looks nice !
I have a few minor questions, but I'll let @RaulinN do the proper review :-)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/Validator.scala
BadrTad: No it does not fail....however it's unnecessary as the `objectMapper.readTree` does the closing.
Additionally, the current state of the method is computationally heavy, meaning to validate every message the validator has to read and process json skeleton from `query.json` (constant) to validate the `jsonString`.  
Calls to readTree are expensive. A fix to this was made in commit `9a6be31fad59817fa369969d26efc60c3217e759` to do this in a nice way. This is a major issue as I have repeatedly noticed a delay of the responses. now the delay is only noticeable in the first message all the others are instant.
A improvement could be computing the schema at compile time for good performance.

be2-scala/build.sbt
BadrTad: Assuming this:
 ..
 |___protocol
 |___be-scala
       |____src/main/resource/protocol
       |____build.sbt
 The project "be2-scala" should be open in IDE such that `build.sbt` is at the root of the workspace and the working directory is `be2-scala` . If not `sbt `does not recognize the build at all.
 
Plus there is no need to worry about this since it's done at compile time. once the jar is properly assembled.

be2-scala/src/main/resources/logback.xml
BadrTad: 👍

BadrTad (commented):
Careful consideration of `readTree` calls

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/Validator.scala
pierluca: Thank you for the nice analysis, it seems your new solution is much better indeed (though I haven't looked into all the tiny details) ! 

As opposed to compile-time evaluation, it could also be done on the startup of the server (before the evaluation of the first message), but don't bother with it. There's no need to push the optimization that far ;-)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/Validator.scala
```diff
@@ -31,44 +31,33 @@ import scala.util.{Failure, Success, Try}
...
-      case Success(Seq(JsNumber(id))) => Some(id.toInt)
-      case _ => None
-    }
-  }
-
-  // contains the relative path to our "main" query JsonSchema file
-  final val queryPath = "../protocol/query/query.json"
-
-  def validateSchema(jsonString: JsonString): Either[JsonString, PipelineError] = {
+  private final val queryPath = "protocol/query/query.json" //With respect to resource folder
```
tuomas-p, line 39: looks good

be2-scala/src/main/scala/ch/epfl/pop/config/application.conf
```diff
@@ -1,8 +1,21 @@
...
+akka {
+  #Setup list off loggers to config
+  loggers = ["akka.event.slf4j.Slf4jLogger"]
+  #Setup
+  logging-filter = "akka.event.slf4j.Slf4jLoggingFilter"
+  # Options: OFF, ERROR, WARNING, INFO, DEBUG
+  loglevel = "INFO"
+  #Set to OFF to turn off logs for sys startup & shuttdown logs
+  stdout-loglevel = "INFO"
+}
```
tuomas-p, line 17: :+1:

be2-scala/.gitignore
```diff
@@ -19,3 +19,6 @@ database
 
 # Test
 database_test
+
+#Protocol folder
+src/main/resources/protocol
```
tuomas-p, line 24: :+1:

tuomas-p (approved):
Looks very good to me!

be2-scala/src/main/resources/logback.xml
```diff
@@ -0,0 +1,14 @@
...
+  </root>
```
RaulinN, line 13: So this was the secret! I remember seeing this for the SimpleLogger we looked at. Good job making it work with that!

be2-scala/build.sbt
```diff
@@ -1,11 +1,40 @@
+import scala.util.{Try, Success, Failure}
```
RaulinN, line 1: Seems to work (copyProtocolTask) as expected

RaulinN (approved):

PR #585, opened 2021-11-13
Add chirp object

Xelowak:
I added a Chirp object, along with its tests.

I wanted to a test for line 37 of Chirp.ts (to achieve a coverage of 100%), but the TSlint told me that Object could not be null/undefined, so it would mean 2 lines for eslint-disable comments, which I prefer to avoid.

Please tell me there is a better way of handling unused constants than disabling it in the linter... I couldn't find anything that worked without declaring an unused variable.

### DIFF
--- a/fe1-web/model/objects/Chirp.ts
+++ b/fe1-web/model/objects/Chirp.ts
@@ -1,5 +1,6 @@
 import { Timestamp } from './Timestamp';
 import { Hash } from './Hash';
+import { PublicKey } from './PublicKey';
 
 /**
  * Object to represent a Chirp.
@@ -18,7 +19,7 @@ export class Chirp {
   public readonly id: Hash;
 
   // The sender's public key
-  public readonly sender: string;
+  public readonly sender: PublicKey;
 
   // The text of the chirp
   public readonly text: string;
@@ -51,14 +52,15 @@ export class Chirp {
       throw new Error("Undefined 'id' when creating 'Chirp'");
     }
     if (obj.likes === undefined) {
-      throw new Error("Undefined 'likes' when creating 'Chirp'");
+      this.likes = 0;
+    } else {
+      this.likes = obj.likes;
     }
 
     this.id = obj.id;
     this.sender = obj.sender;
     this.text = obj.text;
     this.time = obj.time;
-    this.likes = obj.likes;
     this.parentId = obj.parentId;
   }
 
@@ -70,7 +72,7 @@ export class Chirp {
   public static fromState(chirpState: ChirpState): Chirp {
     return new Chirp({
       id: new Hash(chirpState.id),
-      sender: chirpState.sender,
+      sender: new PublicKey(chirpState.sender),
       text: chirpState.text,
       time: new Timestamp(chirpState.time),
       likes: chirpState.likes,
--- a/fe1-web/model/objects/Chirp.ts
+++ b/fe1-web/model/objects/Chirp.ts
@@ -1,0 +1,82 @@
+import { Timestamp } from './Timestamp';
+import { Hash } from './Hash';
+
+/**
+ * Object to represent a Chirp.
+ */
+
+export interface ChirpState {
+  id: string;
+  sender: string;
+  text: string;
+  time: number;
+  likes: number;
+  parentId: string;
+}
+
+export class Chirp {
+  public readonly id: Hash;
+
+  public readonly sender: string;
+
+  public readonly text: string;
+
+  public readonly time: Timestamp;
+
+  public readonly likes: number;
+
+  public readonly parentId?: Hash;
+
+  constructor(obj: Partial<Chirp>) {
+    if (obj === undefined || obj === null) {
+      throw new Error('Error encountered while creating a Chirp object: '
+        + 'undefined/null parameters');
+    }
+
+    if (obj.id === undefined) {
+      throw new Error("Undefined 'id' when creating 'Chirp'");
+    }
+    if (obj.sender === undefined) {
+      throw new Error("Undefined 'sender' when creating 'Chirp'");
+    }
+    if (obj.text === undefined) {
+      throw new Error("Undefined 'text' when creating 'Chirp'");
+    }
+    if (obj.time === undefined) {
+      throw new Error("Undefined 'id' when creating 'Chirp'");
+    }
+    if (obj.likes === undefined) {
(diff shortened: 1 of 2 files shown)


jiabaow (approved):
LGTM

fe1-web/model/objects/__tests__/Chirp.test.ts
```diff
@@ -0,0 +1,96 @@
...
+        parentId: HASH_PARENT,
+      });
+    };
+    expect(createWrongChirp).toThrow(Error);
+  });
+
+  it('throws an error when text is undefined', () => {
+    const createWrongChirp = () => {
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      const wrongChirp = new Chirp({
```
pierluca, line 58: I believe you can avoid this pattern like so:

```suggestion
    const createWrongChirp = () => new Chirp({
```

The arrow functions would then return a chirp, instead of declaring an unused variable.

fe1-web/model/objects/Chirp.ts
```diff
@@ -0,0 +1,87 @@
...
+  public readonly likes: number;
+
+  // The id of the parent chirp (if it is a reply)
+  public readonly parentId?: Hash;
+
+  constructor(obj: Partial<Chirp>) {
+    if (obj === undefined || obj === null) {
+      throw new Error('Error encountered while creating a Chirp object: '
+        + 'undefined/null parameters');
+    }
```
pierluca, line 39: Per your question, you can test this behavior by forcing a cast. 
E.g. `const obj = null as Partial<Chirp>`.

This is basically the TypeScript way of telling the language "shut up and trust me, I know what I'm doing"

fe1-web/model/objects/Chirp.ts
```diff
@@ -0,0 +1,87 @@
...
+  time: number;
+  likes: number;
+  parentId: string;
+}
+
+export class Chirp {
+  public readonly id: Hash;
+
+  // The sender's public key
+  public readonly sender: string;
```
pierluca, line 21: if it's a public key, why not use the `PublicKey` object for clarity (and consistency) ?

fe1-web/model/objects/Chirp.ts
```diff
@@ -0,0 +1,87 @@
...
+    }
+    if (obj.text === undefined) {
+      throw new Error("Undefined 'text' when creating 'Chirp'");
+    }
+    if (obj.time === undefined) {
+      throw new Error("Undefined 'id' when creating 'Chirp'");
+    }
+    if (obj.likes === undefined) {
+      throw new Error("Undefined 'likes' when creating 'Chirp'");
+    }
```
pierluca, line 55: Do we have this information when creating the object the first time around? 
If not, 0 likes might be a reasonable default too.

fe1-web/model/objects/Chirp.ts
```diff
@@ -0,0 +1,87 @@
...
+  public readonly sender: string;
+
+  // The text of the chirp
+  public readonly text: string;
+
+  // The time where the chirp was posted
+  public readonly time: Timestamp;
+
+  // The number of likes
+  public readonly likes: number;
```
pierluca, line 30: I seem to remember that the social media project has this overall definition of reactions, which could be likes (👍) or any other Unicode character (e.g., 🤮).

Has this been simplified (on both front-ends) to just handling likes ? Or should we support more than just a number here ?

pierluca (commented):
A few remarks, but overalls very clean code. Good job !

fe1-web/model/objects/Chirp.ts
jiabaow: We decided to simplify it to support only likes for the midterm presentation and add the dislike later.

fe1-web/model/objects/Chirp.ts
Xelowak: Thanks ! I didn't think about that.

fe1-web/model/objects/Chirp.ts
Xelowak: That's true.

fe1-web/model/objects/__tests__/Chirp.test.ts
Xelowak: That's exactly what I was searching for, thanks a lot !

fe1-web/model/objects/Chirp.ts
Xelowak: Yes, for now we kept only likes, as it seems easier to implement.

fe1-web/model/objects/Chirp.ts
Xelowak: That's true !

fe1-web/model/objects/Chirp.ts
pierluca: Yup, sorry, that's the kind of details I tend to forget when switching between languages 😅

fe1-web/model/objects/__tests__/Chirp.test.ts
```diff
@@ -5,92 +5,94 @@ import {
...
 
 const TIMESTAMP = new Timestamp(12345);
 const HASH_ID = new Hash('id');
 const HASH_PARENT = new Hash('parent');
+const PK = new PublicKey('publicKey');
+const TEXT = 'text';
 
 describe('Chirp object', () => {
-  it('can do a state round trip', () => {
+  it('does a state round trip correctly', () => {
```
pierluca, line 17: No need to change anything, this is good enough.

However, as a general rule, it's a "harder" problem to do `object->plain data->object` than `plain data->object->plain data` , because the `object->plain data` conversion can be lossy. As such, round-tripping with objects is often a tougher test on the code (and more realistic with respect to the application's needs)

pierluca (approved):
Some remarks, but nothing that should block the merge :-)

fe1-web/model/objects/Chirp.ts
Xelowak: You're right, but actually I don't think we're going to support any reaction other than likes. We might need to talk about it next social media meeting though.

fe1-web/model/objects/__tests__/Chirp.test.ts
Xelowak: Ok, seems fair. I'll remember that. Actually I just wanted to change the name of the test to make it clearer

PR #588, opened 2021-11-13
Change the channel at which the chirps are sent

Xelowak:
I corrected the channel at which the chirps are sent, and added something to get the global social channel in Channel.ts.
I added some tests for it.

### DIFF
--- a/fe1-web/model/objects/Channel.ts
+++ b/fe1-web/model/objects/Channel.ts
@@ -18,7 +18,7 @@ export function channelFromIds(...args: Hash[]) : Channel {
  * @param laoIdHash - The hash containing the laoID of the currently opened LAO
  */
 export function userSocialChannel(laoIdHash: Hash): Channel {
-  const userPublicKey = KeyPairStore.get().publicKey.valueOf();
+  const userPublicKey = KeyPairStore.getPublicKey().valueOf();
   return `${ROOT_CHANNEL}/${laoIdHash.valueOf()}/social/${userPublicKey}`;
 }
 
@@ -33,8 +33,10 @@ export function generalChirpsChannel(laoIdHash: Hash): Channel {
 
 /** Returns the last part of the channel which is usually an event id
  * Example:
- * Input: /root/laoid/electionid
- * Output: electionId
+ * Input: /root/laoID/electionID
+ * Output: electionID
+ *
+ * @param channel - The channel we want the last port
  */
 export function getLastChannel(channel: Channel): Hash {
   const channels = channel.split('/');
--- a/fe1-web/model/objects/Channel.ts
+++ b/fe1-web/model/objects/Channel.ts
@@ -1,4 +1,5 @@
 import { Hash } from './Hash';
+import { KeyPairStore } from '../../store';
 
 export type Channel = string;
 export const ROOT_CHANNEL: Channel = '/root';
@@ -9,6 +10,25 @@ export function channelFromIds(...args: Hash[]) : Channel {
   return `${ROOT_CHANNEL}/${
     args.map((c) => c.valueOf())
       .join('/')}`;
+}
+
+/**
+ * Returns the social channel of the current user.
+ *
+ * @param laoIdHash - The hash containing the laoID of the currently opened LAO
+ */
+export function userSocialChannel(laoIdHash: Hash): Channel {
+  const userPublicKey = KeyPairStore.get().publicKey.valueOf();
+  return `${ROOT_CHANNEL}/${laoIdHash.valueOf()}/social/${userPublicKey}`;
+}
+
+/**
+ * Returns the general channel of chirps.
+ *
+ * @param laoIdHash - The hash containing the laoID of the currently opened LAO
+ */
+export function generalChirpsChannel(laoIdHash: Hash): Channel {
+  return `${ROOT_CHANNEL}/${laoIdHash.valueOf()}/social/chirps`;
 }
 
 /** Returns the last part of the channel which is usually an event id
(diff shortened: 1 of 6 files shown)


jiabaow (approved):
LGTM

fe1-web/model/objects/Channel.ts
```diff
@@ -11,10 +12,31 @@ export function channelFromIds(...args: Hash[]) : Channel {
       .join('/')}`;
 }
 
+/**
+ * Returns the social channel of the current user.
+ *
+ * @param laoIdHash - The hash containing the laoID of the currently opened LAO
+ */
+export function userSocialChannel(laoIdHash: Hash): Channel {
+  const userPublicKey = KeyPairStore.getPublicKey().valueOf();
```
pierluca, line 21: My understanding is that this basically de-anonymizes the user, because `getPublicKey()` is going to be constant over time. Is this desired ? I don't remember the social media specification in this regard.

fe1-web/model/objects/Channel.ts
```diff
@@ -11,10 +12,31 @@ export function channelFromIds(...args: Hash[]) : Channel {
...
+}
+
 /** Returns the last part of the channel which is usually an event id
  * Example:
- * Input: /root/laoid/electionid
- * Output: electionId
+ * Input: /root/laoID/electionID
+ * Output: electionID
+ *
+ * @param channel - The channel we want the last port
```
pierluca, line 39:
```suggestion
 * @param channel - The channel whose last component we want to obtain
```

By the way, do we still need this? It seems like a very easy function to misuse

fe1-web/model/objects/__tests__/Channel.test.ts
```diff
@@ -1,11 +1,43 @@
...
+    expect(channelFromIds()).toEqual('/root');
+    expect(channelFromIds(FAKE_ID)).toStrictEqual(`/root/${FAKE_ID}`);
+  });
+
+  it('userSocialChannel should return the correct channel', () => {
+    const keyPair = KeyPair.fromState({
+      publicKey: PUBLIC_KEY,
+      privateKey: PRIVATE_KEY,
+    });
+    KeyPairStore.store(keyPair);
```
pierluca, line 29: I don't remember: is there a global cleanup of the stores across tests?
Or have you just permanently modified the keyPair for all the other tests? 
Ideally, tests should not affect each other.
If there's no global cleanup, it might make sense to either:
1. look into it :-)  or 
2. ensure that the store contains the same information before and after the tests.

fe1-web/network/MessageApi.ts
```diff
@@ -289,16 +289,17 @@ export function terminateElection(
...
 
   const message = new AddChirp({
     text: text,
-    parent_id: parent_id,
+    parent_id: parentId,
     timestamp: timestamp,
   });
 
-  return publish(channelFromIds(currentLao.id), message);
+  return publish(userSocialChannel(currentLao.id), message);
```
pierluca, line 304: does this channel already exist before the first chirp ? 
Is the server creating it automatically upon receiving a publish ? 🤔

fe1-web/network/__tests__/MessageApi.test.ts
```diff
@@ -4,12 +4,18 @@ import '__tests__/utils/matchers';
 import 'store/Storage';
```
pierluca, line 4: not your fault, but this file is really really ugly.
Don't feel forced to comply with its current logic, you can start breaking it apart if you think it'll make your life easier.

pierluca (commented):
The code itself is pretty clean 👍 
I have more questions than answers here though, so I'll let you go through those points and let me know if I should merge or not yet 😄

fe1-web/model/objects/Channel.ts
Xelowak: Yes, this is what we chose to do. Each user sends chirp to his own channel, namely root/<laoID>/social/<userPublicKey>. This way, the channel of a user is constant over time.

fe1-web/model/objects/Channel.ts
Xelowak: It seems that it is used in Election.ts, and might be useful elsewhere if we want to retrieve an id. I think I'll just rename it because its name is misleading, I first thought it would return the last used channel.

fe1-web/model/objects/__tests__/Channel.test.ts
Xelowak: From what I've seen, I think that each test that uses the KeyPairStore stores his own key pair in the beforeEach function, so it is overwritten before each test. Is it a problem ?

fe1-web/network/MessageApi.ts
Xelowak: The channel is created when the user attends a roll call. For testing purposes, it is currently opened for the organizer as soon as he launches an LAO.

fe1-web/network/__tests__/MessageApi.test.ts
Xelowak: Yes, it might be a really good idea to refactor it if we have the time to do it, but it will probably be a lot of work. For now, I think it's usable but not practical at all.

Xelowak: I think It might be a really good idea to refactor it if we have the time to do it.
You're right, I moved my code in another file so that this file doesn't get any longer.

fe1-web/model/objects/__tests__/Channel.test.ts
pierluca: Sounds okay, if that's what's happening.

pierluca (approved):
LGTM

PR #589, opened 2021-11-14
Work be2 tuomas p database rework

tuomas-p:
The code may need a bit of cleaning up, however, the functionalities available before are available again now


be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/ElectionValidator.scala
```diff
@@ -25,10 +25,10 @@ object ElectionValidator extends MessageDataContentValidator with EventValidator
           Right(validationError(s"'start_time' (${data.start_time}) timestamp is smaller than 'created_at' (${data.created_at})"))
         } else if (!validateTimestampOrder(data.start_time, data.end_time)) {
           Right(validationError(s"'end_time' (${data.end_time}) timestamp is smaller than 'start_time' (${data.start_time})"))
-        } else if (expectedHash != data.id) {
+        } /*else if (expectedHash != data.id) {
```
BadrTad, line 28: What is going on here ?

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
```diff
@@ -137,157 +180,269 @@ object DbActor extends AskPatternConstants {
...
-      val channelDb: DB = channelsMap.get(channel) match {
-        case Some(channelDb) => channelDb
-        case _ => createDatabase(channel)
+      
+      //encapsulate this inside function
+      Try(db.get(channel.toString.getBytes)) match {
+        case Success(bytes) if bytes != null =>
+          log.info(s"Channel '$channel' already exists.")
+        case _ =>
+          //FIXME: handle type of message (is it actually a problem or is the MessageExample coded in a weird way?)
```
BadrTad, line 218: If you mean the `util.examples.MessageExample` in the test folder then yes it looks fine to me .

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
```diff
@@ -137,157 +180,269 @@ object DbActor extends AskPatternConstants {
...
+          //val objectType = message.decodedData.getOrElse(null)._object //or some other way, then match using match case and establish channel type
```
BadrTad, line 219: Carefull here if null is return then `null._object `....

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
```diff
@@ -137,157 +180,269 @@ object DbActor extends AskPatternConstants {
...
+          val objectType = ObjectType.LAO
+          val actualObjectType = if (objectType == ObjectType.ELECTION || objectType == ObjectType.CHIRP){
```
BadrTad, line 221: I don't know what it's really going on here but in either case the `actualObjectType ` will hold `ObjectType.LAO`

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
```diff
@@ -93,7 +118,7 @@ object DbActor extends AskPatternConstants {
   /** Response for a [[Write]] db request Receiving [[DbActorWriteAck]] works as
     * an acknowledgement that the write request was successful
     */
-  final case object DbActorWriteAck extends DbActorMessage
+  final case class DbActorWriteAck() extends DbActorMessage
```
BadrTad, line 121: Any particular reason why you switched from object to class

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
```diff
@@ -137,157 +180,269 @@ object DbActor extends AskPatternConstants {
...
     options.createIfMissing(true)
-    val channelNamesDb: DB =
-      try { factory.open(new File(CHANNELS_FOLDER), options) }
+    
+    val db: DB =
+      try { factory.open(new File(DATABASE_FOLDER), options) }
       catch {
         case e: IOException => {
-          logger.error("Could not open channels folder {}", CHANNELS_FOLDER)
+          logger.error("Could not open database folder {}", DATABASE_FOLDER)
```
BadrTad, line 193: Great you kept the log 👍

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
```diff
@@ -137,157 +180,269 @@ object DbActor extends AskPatternConstants {
...
+    DbActor(mediatorRef, db, DATABASE_FOLDER)
   }
 
-  sealed case class DbActor(mediatorRef: ActorRef, initialChannelsMap: Map[Channel, DB], channelNamesDb: DB, DATABASE_FOLDER: String, CHANNELS_FOLDER: String) extends Actor with ActorLogging {
-    private val channelsMap: mutable.Map[Channel, DB] = initialChannelsMap.to(collection.mutable.Map)
+  // In the arguments, we might remove the initialChannelsMap and replace the channelNamesDb by the general Db??
+  // FIXME: find way to keep count of channel nb. Idea: store key "channelCount" with value int.getBytes to db and increment it with each channel created
+  sealed case class DbActor(mediatorRef: ActorRef, db: DB, DATABASE_FOLDER: String) extends Actor with ActorLogging {
 
     override def preStart(): Unit = {
```
BadrTad, line 205: I don't think this method should be kept at all. If of course there is no need to run some code before deploying the DBactor. More info here ` https://www.javatpoint.com/akka-actor-life-cycle`

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
```diff
@@ -137,157 +180,269 @@ object DbActor extends AskPatternConstants {
...
-      log.info(
-        s"Actor $self (db) was initialised with a total of ${initialChannelsMap.size} recovered channels"
-      )
-      if (initialChannelsMap.size > DATABASE_MAX_CHANNELS) {
-        log.warning(
-          s"Actor $self (db) has surpassed a large number of active lao channels (${initialChannelsMap.size} > $DATABASE_MAX_CHANNELS)"
-        )
-      }
-
+      //FIXME: check if this should be kept
```
BadrTad, line 206: Nope see comment above !

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
```diff
@@ -137,157 +180,269 @@ object DbActor extends AskPatternConstants {
...
-
+      
       val messageId: Hash = message.message_id
-      Try(channelDb.put(messageId.getBytes, message.toJsonString.getBytes)) match {
-        case Success(_) =>
-          log.info(s"Actor $self (db) wrote message_id '$messageId' on channel '$channel'")
-          DbActorWriteAck
+      Try(db.get(channel.toString.getBytes)) match {
+        case Success(bytes) if bytes != null => {
+          val key: String = channel + "/" + messageId.toString
```
BadrTad, line 237: It would be nice to have some kind of help methods to compute the key from its parameters. 
So it will convey more abstraction and less confusion (and thus errors) for future developers.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
```diff
@@ -137,157 +180,269 @@ object DbActor extends AskPatternConstants {
...
+                case Success(_) =>
+                  log.info(s"Actor $self (db) wrote object 'objectId' and message_id '$messageId' on channel '$channel'") //change with object and objectId/name/smth like that
+                  DbActorWriteAck()
+                case Failure(exception) =>
+                  log.info(s"Actor $self (db) encountered a problem while writing message_id '$messageId' and object 'objectId' on channel '$channel' because of the batch write")
+                  DbActorNAck(ErrorCodes.SERVER_ERROR.id, exception.getMessage)
+              }      
+            }
+            case Failure(exception) =>
+              log.info(s"Actor $self (db) encountered a problem while writing message_id '$messageId' and object 'objectId' on channel '$channel' because of a batch creation")
```
BadrTad, line 254: Good log 👍

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
```diff
@@ -137,157 +180,269 @@ object DbActor extends AskPatternConstants {
...
+          val key: String = channel + "/" + messageId.toString
+          Try(db.createWriteBatch()) match {
+            case Success(batch) => {
+              val json = new String(bytes, StandardCharsets.UTF_8)
+              batch.put(channel.toString.getBytes, ChannelData.buildFromJson(json).addMessage(messageId).toJsonString.getBytes)
+              batch.put(key.getBytes, message.toJsonString.getBytes)
+              //allows writing all data atomically
+              Try(db.write(batch)) match {
+                case Success(_) =>
+                  log.info(s"Actor $self (db) wrote object 'objectId' and message_id '$messageId' on channel '$channel'") //change with object and objectId/name/smth like that
```
BadrTad, line 246: This means the new implemented Object types should have an ID. 
`LaoData `and `ChannelData `don't for now !

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
```diff
@@ -137,157 +180,269 @@ object DbActor extends AskPatternConstants {
...
+            case Failure(exception) =>
+              log.info(s"Actor $self (db) encountered a problem while writing message_id '$messageId' and object 'objectId' on channel '$channel' because of a batch creation")
+              DbActorNAck(ErrorCodes.SERVER_ERROR.id, exception.getMessage)
+          }
+        }
         case Failure(exception) =>
           log.info(s"Actor $self (db) encountered a problem while writing message_id '$messageId' on channel '$channel'")
           DbActorNAck(ErrorCodes.SERVER_ERROR.id, exception.getMessage)
+        case _ =>
+          log.info(s"Actor $self (db) encountered a problem while writing message_id '$messageId' on channel '$channel', as the channel does not exist")
```
BadrTad, line 262: I'd recommend turning this log into `log.error`

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
```diff
@@ -137,157 +180,269 @@ object DbActor extends AskPatternConstants {
...
+                  DbActorNAck(ErrorCodes.SERVER_ERROR.id, exception.getMessage)
+              }      
+            }
+            case Failure(exception) =>
+              log.info(s"Actor $self (db) encountered a problem while writing message_id '$messageId' and object 'objectId' on channel '$channel' because of a batch creation")
+              DbActorNAck(ErrorCodes.SERVER_ERROR.id, exception.getMessage)
+          }
+        }
         case Failure(exception) =>
           log.info(s"Actor $self (db) encountered a problem while writing message_id '$messageId' on channel '$channel'")
```
BadrTad, line 259: Use log.error instead

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
```diff
@@ -137,157 +180,269 @@ object DbActor extends AskPatternConstants {
...
+              DbActorNAck(ErrorCodes.SERVER_ERROR.id, s"Error when creating channel '$channel'")
+          }
+      }
+      Try(db.get(channel.toString.getBytes)) match {
+        case Success(bytes) if bytes != null => {
+          Try(db.createWriteBatch()) match {
+            case Success(batch) => {
+              val json = new String(bytes, StandardCharsets.UTF_8)
+              //this is the easiest way I found until now
+              val dataKey: String = new String(dataOpt.get, StandardCharsets.UTF_8) + "/data"
```
BadrTad, line 308: Same remark as above for the data key generation

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/Validator.scala
```diff
@@ -51,6 +69,7 @@ object Validator {
   }
 
   // contains the relative path to our "main" query JsonSchema file
+  //FIXME: does not work apparently
```
BadrTad, line 72: Fixed in PR#583 commit '66d492f28a788fe6954c5230de778c5209221370'

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
```diff
@@ -47,7 +52,63 @@ case object RollCallHandler extends MessageHandler {
...
+            rpcMessage.id
+          ))
+
+        }
+      case _ => Right(PipelineError(
+        ErrorCodes.SERVER_ERROR.id,
+        s"Unable to handle lao message $rpcMessage. Not a Publish/Broadcast message",
+        rpcMessage.id
+      ))
+    }
```
BadrTad, line 112: Feels good to see this come to life 😃 after some tears.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
```diff
@@ -47,7 +52,63 @@ case object RollCallHandler extends MessageHandler {
...
+        val data: CloseRollCall = message.decodedData.get.asInstanceOf[CloseRollCall]
+        val unknownAnswer: String = "Database actor returned an unknown answer"
+
+        //Creates a channel for each attendee (of name /root/lao_id/social/PublicKeyAttendee), returns a GraphMessage
+        def createAttendeeChannels(attendees: List[PublicKey], rpcMessage: JsonRpcRequest): GraphMessage = {
+          attendees match {
+            case Nil => Left(rpcMessage)
+            case head::tail => 
+              //a closeRollCall is always sent in the lao's main channel so we are allowed to do this
+              val socialChannel: String = rpcMessage.getParamsChannel + "social/" + head.toString
```
BadrTad, line 66: Yep we'll definitely need some helper to create the keys for database.

BadrTad (commented):
Great overall. Have some suggestions to make the interaction with `DBactor `even better.
Such as providing helper methods to generate the different keys.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
tuomas-p: Indeed, I have a problem with getting the MessageData out of the Message in order to get the MessageType and haven't been able to solve it yet.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
tuomas-p: Oh well I think you found the issue I somehow missed...

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/ElectionValidator.scala
tuomas-p: I had a problem with the id calculation, so I changed some stuff around and visibly left the check there twice, once commented

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
tuomas-p: This was a temporary solution, as my getter for the ObjectType of the message was wrong

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
tuomas-p: There actually was, as I had some trouble sending one somewhere in DbActor, which was weird, but would it be better to change it back or is it fine if we keep it as a class?

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
tuomas-p: I'll remove this indeed

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
tuomas-p: Indeed, I totally agree with you

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
tuomas-p: I was just thinking about this, I believe the best way would be to simply add some kind of "getObjectString()" method in the supertype of LaoData, ElectionData, etc. once I change this method and implement these types so that it can also return ElectionData (for example) and other data so the log sees which type of data was being accessed

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
tuomas-p: good catch

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
tuomas-p: :)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/Validator.scala
tuomas-p: indeed, I forgot to remove the fixme after I pulled

tuomas-p:
I'll correct the errors and see what would be the best way to create a function for the key computation, thank you @BadrTad !

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/ElectionValidator.scala
tuomas-p: I still haven't been able to solve the id calculation error (which was coded before this semester), as the message examples we received didn't pass the validator even though they were generated by a frontend, so I left it commented for now

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
```diff
@@ -137,157 +181,249 @@ object DbActor extends AskPatternConstants {
...
-
-        }
+    DbActor(mediatorRef, db, DATABASE_FOLDER)
+  }
 
-      // store the channel name and its database in the map
-      initialChannelsMap += (Channel(
-        channelName.toString.stripPrefix(DATABASE_FOLDER)
-      ) -> channelDb)
+  // In the arguments, we might remove the initialChannelsMap and replace the channelNamesDb by the general Db??
```
pierluca, line 202: I think you've done this already?

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
```diff
@@ -137,157 +181,249 @@ object DbActor extends AskPatternConstants {
...
+  // FIXME: find way to keep count of channel nb. Idea: store key "channelCount" with value int.getBytes to db and increment it with each channel created
+  sealed case class DbActor(mediatorRef: ActorRef, db: DB, DATABASE_FOLDER: String) extends Actor with ActorLogging {
+    
+    //helper function to determine if a channel should be created during write operations
+    private def writeCreateNewChannel(channel: Channel, message: Message): DbActorMessage = {
```
pierluca, line 207: Doesn't this implementation mean that I can create a channel by casting a vote for a non-existent election?

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
```diff
@@ -137,157 +181,249 @@ object DbActor extends AskPatternConstants {
...
-        s"Actor $self (db) was initialised with a total of ${initialChannelsMap.size} recovered channels"
-      )
-      if (initialChannelsMap.size > DATABASE_MAX_CHANNELS) {
-        log.warning(
-          s"Actor $self (db) has surpassed a large number of active lao channels (${initialChannelsMap.size} > $DATABASE_MAX_CHANNELS)"
-        )
+    //may return null if the extractLaoId fails
+    private def generateLaoDataKey(channel: Channel): String = {
+      channel.decodeSubChannel match {
+        case Some(data) => new String(data, StandardCharsets.UTF_8) + "/data"
```
pierluca, line 239: What happens if we have a new channel in the format`laoId/data` ? 

Here, in `generateMessageKey`, in storing channels, in storing everything, I would suggest you use prefixes for all of your keys. Otherwise you might bump into key conflicts that will be extremely hard to debug.
It also makes it more obvious what types you're dealing with if (in the future) we need to look into the database to debug.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
```diff
@@ -61,11 +87,11 @@ object DbActor extends AskPatternConstants {
   final case class WriteAndPropagate(channel: Channel, message: Message) extends Event
```
pierluca, line 87: This is outside the scope of this PR, but I'm wondering, should it be the `DbActor` responsibility to propagate messages? 🤔 (any opinion @RaulinN ?)

pierluca (commented):
I did a quick pass on the code to have a look, this doesn't replace in any way @RaulinN 's expert look on the Scala codebase.

First of all, wow ! Massive changes, and I think we're getting to a somewhat cleaner implementation. Good job !

However, parts of the code still feels a bit "hacky": commented out bits, lots of FIXMEs, comments not up to date, etc. It's fine to leave some parts not implemented until a subsequent PR, but try to commit clean code.

Ideally, try to program "defensively": don't assume anything, don't trust user inputs unless they've been validated, etc. ...when in doubt, DON'T ASSUME ANYTHING :-) (cf. google "defensive programming" if you've never heard the term)
In my limited, broken understanding of Scala code, it seems your code makes a lot of assumptions about inputs (not just the user's), which makes it more brittle and you can end up fighting unforeseen bugs if any of the assumptions break.

Since you've made some pretty awesome work on the DbActor, I'd also try to document how things are stored for the reference of future developers on this codebase. I believe it'd be pretty helpful even for yourselves 6 weeks down the line.

Overalls, and despite all my comments asking you to change stuff, I believe we're already in a much better place. It's expected that such a major piece of work would have some rough edges at the beginning :)

be2-scala/src/main/scala/ch/epfl/pop/model/network/method/message/data/dataObject/LaoData.scala
```diff
@@ -0,0 +1,30 @@
+package ch.epfl.pop.model.network.method.message.data.dataObject
+
+import ch.epfl.pop.json.HighLevelProtocol._
+import ch.epfl.pop.json.MessageDataProtocol._
+import ch.epfl.pop.model.network.Parsable
+import ch.epfl.pop.model.objects._
+import spray.json._
+
+case class LaoData(
```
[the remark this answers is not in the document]
tuomas-p: It should indeed be moved somewhere else now that you say it, I originally left it there as in my first draft I considered the new classes like some kind of messages, however, I'll find a better place for them

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/Validator.scala
```diff
@@ -17,6 +18,23 @@ import ch.epfl.pop.pubsub.graph.validators.ParamsValidator._
...
+import com.networknt.schema.ValidationMessage
+import com.networknt.schema.SpecVersion
+
+import scala.io.{BufferedSource, Source}
+import scala.util.{Failure, Success, Try}
+
+import spray.json._
+
+import java.io.File
+
```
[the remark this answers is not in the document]
tuomas-p: typo/mistake from my part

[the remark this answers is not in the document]
tuomas-p: typo/mistake from my part

be2-scala/src/main/scala/ch/epfl/pop/model/network/method/message/data/dataObject/ChannelData.scala
```diff
@@ -0,0 +1,38 @@
...
+import ch.epfl.pop.json.HighLevelProtocol._
+import ch.epfl.pop.json.MessageDataProtocol._
+import ch.epfl.pop.model.network.Parsable
+import ch.epfl.pop.model.network.method.message.data.{ObjectType}
+import ch.epfl.pop.model.objects._
+import spray.json._
+
+
+//the general ObjectType will be LAO (for all but election and chirp channels for now)
+case class ChannelData(
```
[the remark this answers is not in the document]
tuomas-p: same as above, I'll move both to a better place

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/SocialMediaHandler.scala
```diff
@@ -0,0 +1,43 @@
...
+      ))
+    }
+    case graphMessage@_ => graphMessage
+  }
+  
+  def handleAddChirp(rpcMessage: JsonRpcRequest): GraphMessage = {
+    //broadcast chirp too? On general channel?
+    //create message (Broadcast one)
+    //we can use PubSubMediator.Propagate(channel (fixed), message (created here))
+    // mediatorRef ! PubSubMediator.Propagate(channel, message)
```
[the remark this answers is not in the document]
tuomas-p: Yes indeed, it is currently being treated in my other branch about social media

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
```diff
@@ -137,157 +181,249 @@ object DbActor extends AskPatternConstants {
...
-        }
+    DbActor(mediatorRef, db, DATABASE_FOLDER)
+  }
 
-      // store the channel name and its database in the map
-      initialChannelsMap += (Channel(
-        channelName.toString.stripPrefix(DATABASE_FOLDER)
-      ) -> channelDb)
+  // In the arguments, we might remove the initialChannelsMap and replace the channelNamesDb by the general Db??
+  // FIXME: find way to keep count of channel nb. Idea: store key "channelCount" with value int.getBytes to db and increment it with each channel created
```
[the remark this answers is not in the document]
tuomas-p: As originally there was a channel limit of 1 << 10 and a message at server start logging how many channels were found in existing databases, however I'm not sure about the need for that

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
tuomas-p: my mistake for leaving it there!

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
tuomas-p: The thing is that in the original DbActorSuite, it was a requirement for write to create a new channel when the current one did not exist. However, I indeed find this a little bit weird. However, in the new CastVoteElectionValidator, it will be made sure that only messages to valid election channels will go through, but it certainly does not feel much like defensive programming. We could maybe discuss this a bit more on Slack (I'll send a message there right away), but the current implementation is the same as the previous one.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
tuomas-p: This could be problematic indeed, however with our current implementation, we'll have to find some working prefix and incorporate it in the protocol. The documentation you were talking about in your general comment should be very useful for things like that. I will add it in the prefixes in Channel.scala

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/SocialMediaValidator.scala
```diff
@@ -0,0 +1,66 @@
...
+
+                val listAttendees = Await.result(ask, duration)*/
+
+                if (!validateTimestampStaleness(data.timestamp)) {
+                    Right(validationError(s"stale timestamp (${data.timestamp})"))
+                }
+                /*else if(!listAttendees.contains(actualSender)){
+                    Right(validationError(s"invalid sender (${actualSender})"))
+                }*/
+                // FIXME: validate parent ID: need to store all the channel's tweet IDs somewhere to compare?
```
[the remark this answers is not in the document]
tuomas-p: I left this here by mistake I believe, it will be removed

be2-scala/src/test/scala/ch/epfl/pop/pubsub/graph/DbActorSuite.scala
```diff
@@ -152,4 +196,108 @@ class DbActorSuite() extends TestKit(ActorSystem("myTestActorSystem"))
...
+    answer2.asInstanceOf[DbActor.DbActorReadLaoDataAck].laoData should equal(Some(LaoData(PublicKey(Base64Data("a")), List.empty)))
+
+  }
+
+  test("DbActor stores and reads two distinct LaoData objects, to simulate two LAOs in the same database"){
+    val channel: Channel = Channel(TEST_CHANNEL_WRITELAO_2)
+    val channel2: Channel = Channel(TEST_CHANNEL_WRITELAO_3)
+    val message: Message = MessageExample.MESSAGE
+    val laoData: LaoData = LaoData(PublicKey(Base64Data("a")), List.empty)
+    val laoData2: LaoData = LaoData(PublicKey(Base64Data("a")), List(PublicKey(Base64Data("b"))))
```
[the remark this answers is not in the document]
tuomas-p: No, maybe I have understood things wrong, but the pk which is stored is the organizer's public key and the same organizer can create multiple LAOs right?

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
```diff
@@ -47,7 +52,64 @@ case object RollCallHandler extends MessageHandler {
...
+        def createAttendeeChannels(attendees: List[PublicKey], rpcMessage: JsonRpcRequest): GraphMessage = {
+          attendees match {
+            case Nil => Left(rpcMessage)
+            case head::tail => 
+              //a closeRollCall is always sent in the lao's main channel so we are allowed to do this
+              val socialChannel: String = generateSocialChannel(rpcMessage.getParamsChannel, head)
+              val ask: Future[GraphMessage] = (dbActor ? DbActor.CreateChannel(Channel(socialChannel), ObjectType.CHIRP)).map {
+                case DbActorWriteAck() => createAttendeeChannels(tail, rpcMessage)
+                //FIXME: Once the real PoP tokens are implemented, the same person will not have access to the same channel name twice in a row, so we can remove this line
+                //however, this is to prevent errors for now within a LAO for multiple Roll Calls and the same participants since the public key will probably not change
```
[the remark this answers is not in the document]
tuomas-p: Since for now, the personal social channels are identified by public keys and not ever-changing PoP tokens (which could be the case later on), we can't create new channels at each Roll Call for each member, because with the way channel creation works (and worked before I changed database implementation), it will fail also if the channel already exists and make things way more complicated to handle.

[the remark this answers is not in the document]
tuomas-p: Therefore, I left a FIXME for later on, as we could simply create new channels each time if everybody gets a different PoP token each time

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
```diff
@@ -33,6 +36,15 @@ object DbActor extends AskPatternConstants {
...
   final case class Write(channel: Channel, message: Message) extends Event
 
+  /**
+   * Request to write a message and a LaoData object in the database
+   *
+   * @param channel the channel where the message should be published
+   * @param message the message to write in the database
+   * @param obj the LaoData we write
+   */
+  final case class WriteLaoData(channel: Channel, message: Message, obj: LaoData) extends Event
```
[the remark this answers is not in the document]
tuomas-p: My initial idea was that since every object (LaoData and the future ElectionData) will always be accompagnied by a message, to keep the database consistent and achieve atomic writing, we should therefore not allow one of the two (message or object) to be lost and the other written in the database, but both should fail at once or be written at once, therefore I write both.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/LaoHandler.scala
```diff
@@ -39,9 +41,17 @@ case object LaoHandler extends MessageHandler {
         val data: CreateLao = message.decodedData.get.asInstanceOf[CreateLao]
         // we are using the lao id instead of the message_id at lao creation
         val channel: Channel = Channel(s"${Channel.rootChannelPrefix}${data.id}")
-
-        val ask: Future[GraphMessage] = (dbActor ? DbActor.Write(channel, message)).map {
-          case DbActorWriteAck => Left(rpcMessage)
+        val laoData: LaoData = LaoData(message.sender, List(message.sender)) //so that the operations the owner does next are authorized
```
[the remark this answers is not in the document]
tuomas-p: These lines basically add the one sending CreateLao as the owner and as a Roll Call participant (this could maybe be removed, or do we consider that the owner should automatically be present at all Roll Calls?), in order to later on, when I implement the verifiers to check the message sender's identity, allow the owner to directly send messages to the server, but I'll update the comment since it is confusing

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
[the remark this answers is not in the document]
pierluca: Ensuring that the changes happens atomically makes a lot of sense.

My concern is that:
- the API name "WriteLaoData" is not aligned with the parameters (channel, message)
- the API doesn't seem to restrict which messages are allowed to change LaoData
- the modification of the LaoData could be entirely unrelated to the message itself.

Effectively, a buggy user of this API could write a message (say, a chirp) and change the LaoData to remove all attendees, and nothing would stop it.

I'm not saying the following is a good idea, as a matter of fact it might be very bad practice in Scala (I wouldn't know), but imagine the following solution, in which Write is used instead of WriteLaoData. 
Pseudocode:
```
Write(channel: Channel, message: Message) {
   // check if lao data would be modified by message, through a static function
   if ( ! LaoData.IsAffectedBy(message) ) {
     save(message) // store in db, directly
   }
   else { 
     startBatch()
     laoData = retrieveLaoData(channel) // get current lao data from db
     laoData.updateWith(message) // updates LaoData as a function of message
     save(laoData) // store in db
     save(message) // store in db
     executeBatch() // finish transaction
   }
}
```

This would make the DbActor a bit "smarter" (not just a plain storage layer), but it makes sure that no amount of buggy code can mess up the database.

On a side note, you'll realize that the code above can be a source of concurrency issues: if multiple threads change the LaoData in parallel, they might end up overwriting each other. It might be worthwhile to make sure that the LaoData read-update-write cycle doesn't happen concurrently.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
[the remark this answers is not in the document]
pierluca: That was to avoid an explosion of the number of folders and DB size.
You could easily use another metric at startup, it should be easier now ;)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
pierluca: I'm not sure Channel.scala is the best place. Prefixes would really be an internal implementation detail of the storage layer :)

be2-scala/src/test/scala/ch/epfl/pop/pubsub/graph/DbActorSuite.scala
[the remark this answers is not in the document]
pierluca: oh right, sorry, I mixed things up, the first parameter is the organizer's PK, I thought it was the LAO id, but that's embedded in the channel.

For clarity in this test, I'd still try to have different organizer's PK :)

Whether an organizer can have 2 LAOs could be its own test.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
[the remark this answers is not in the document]
pierluca: I'm still confused a bit, because I thought that at the moment PoP tokens were being used for the social media application, but I don't know what you discussed within the social media meetings 🤔
In any case your latest message makes sense, so go ahead :)

pierluca (commented):
Replied to a few of our comments :)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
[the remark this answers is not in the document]
tuomas-p: Good point, thank you!

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
tuomas-p: I suggested Channel.scala since the other prefixes were there but I will move it directly to DbActor since it effectively makes more sense

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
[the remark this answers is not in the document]
tuomas-p: This seems like a good method. Normally, only a CloseRollCall will change LaoData and it should not happen in parallel with another one, however, we should probably not assume such things, so I believe some kind of compare-and-swap instruction could be useful. However, the biggest concurrency issue I see here for now is having the entire LaoData overwritten by one thread, which with the current functionalities should not create inconsistencies, since there shouldn't be any modifications done on the existing LaoData (the attendee list is reset every time and the owner should not change)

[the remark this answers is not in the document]
tuomas-p: However, I'm still a little hesitant about adding the additional check to write, since it will complexify the method even more and take away some utility from the handlers, however having a single write method would be easier and more secure so I'll go with that.

be2-scala/src/test/scala/ch/epfl/pop/model/network/method/message/data/dataObject/ChannelDataSuite.scala
```diff
@@ -0,0 +1,40 @@
...
+        channelData2.messages.size should equal (1)
+    }
+
+    test("Json conversions work for ChannelData") {
+        val channelData: ChannelData = ChannelData(ObjectType.LAO, List.empty)
+
+        val channelData2: ChannelData = ChannelData.buildFromJson(channelData.toJsonString)
+
+        channelData2.messages should equal (List.empty)
+        channelData2.channel_type should equal (ObjectType.LAO)
```
RaulinN, line 37: Same as LaoDataSuite

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/SocialMediaValidator.scala
```diff
@@ -0,0 +1,66 @@
...
+import ch.epfl.pop.model.network.method.message.data.{ActionType, ObjectType}
+import ch.epfl.pop.model.network.method.message.data.rollCall.CloseRollCall
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.{Await, Future}
+
+
+
+case object SocialMediaValidator extends MessageDataContentValidator with EventValidator {
+    override def EVENT_HASH_PREFIX: String = "SocialMedia" //to check later on
```
RaulinN, line 17: Is this not defined yet? If not, I'd suggest creating an issue to discuss it with other members. Moreover, we should (I hope) have a list of these prefixes somewhere in the protocol

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/SocialMediaValidator.scala
```diff
@@ -0,0 +1,66 @@
+package ch.epfl.pop.pubsub.graph.validators
+
+import ch.epfl.pop.model.network.JsonRpcRequest
```
RaulinN, line 3: Does not look very clean. A lot of artifacts in this file. Do you still need all the commented code?

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/SocialMediaValidator.scala
```diff
@@ -0,0 +1,66 @@
...
+
+case object SocialMediaValidator extends MessageDataContentValidator with EventValidator {
+    override def EVENT_HASH_PREFIX: String = "SocialMedia" //to check later on
+
+    def validateAddChirp(rpcMessage: JsonRpcRequest): GraphMessage = {
+        def validationError(reason: String): PipelineError = super.validationError(reason, "AddChirp", rpcMessage.id)
+
+        // FIXME: create method for any message instance as 2nd arg
+        // method not functional, find some other way
+        //def isInstanceOfCloseRollCall(msg: Message): Boolean = msg.decodedData.get._object == ObjectType.ROLL_CALL && msg.decodedData.get.action == ActionType.CLOSE
```
RaulinN, line 24: I don't understand why you need that

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
```diff
@@ -47,7 +52,64 @@ case object RollCallHandler extends MessageHandler {
...
+                case DbActorNAck(code, description) if description == "Database cannot create an already existing channel (" + socialChannel + ")" => createAttendeeChannels(tail, rpcMessage)
```
RaulinN, line 71: This is dangerous. What happens if we change the message that the database returns (e.g. "can't" instead of "cannot")?

Minor: you are not using "code", you can thus change it into `case DbActorNAck(_, description) if ...`

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
```diff
@@ -47,7 +52,64 @@ case object RollCallHandler extends MessageHandler {
...
-    val ask: Future[GraphMessage] = dbAskWritePropagate(rpcMessage)
-    Await.result(ask, duration)
+    rpcMessage.getParamsMessage match {
+      case Some(message: Message) =>
+        val data: CloseRollCall = message.decodedData.get.asInstanceOf[CloseRollCall]
+        val unknownAnswer: String = "Database actor returned an unknown answer"
+
+        //Creates a channel for each attendee (of name /root/lao_id/social/PublicKeyAttendee), returns a GraphMessage
+        def createAttendeeChannels(attendees: List[PublicKey], rpcMessage: JsonRpcRequest): GraphMessage = {
+          attendees match {
```
RaulinN, line 62: This function looks strange/messy/out of place. Is there a better way to do it?

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
```diff
@@ -47,7 +52,64 @@ case object RollCallHandler extends MessageHandler {
...
+        }
+
+        val laoChannel: Option[Array[Byte]] = rpcMessage.getParamsChannel.decodeSubChannel
+        laoChannel match {
+          case None => Right(PipelineError(
+            ErrorCodes.SERVER_ERROR.id,
+            s"There is an issue with the data of the LAO",
+            rpcMessage.id
+          ))
+          case Some(channel) =>
```
RaulinN, line 86: Unused "channel" => switch to `case Some(_) =>`. Not sure, but I think Scala 3 doesn't like this kind of unused variables. So might as well change them now :)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
```diff
@@ -47,7 +52,64 @@ case object RollCallHandler extends MessageHandler {
...
+            val askOldData = dbActor ? DbActor.ReadLaoData(rpcMessage.getParamsChannel)
+            
+            Await.result(askOldData, duration) match {
```
RaulinN, line 89: Would be slightly cleaner to use `Await.ready` instead since you are pattern matching on the result (if any)

be2-scala/src/test/scala/ch/epfl/pop/pubsub/graph/DbActorSuite.scala
```diff
@@ -152,4 +196,108 @@ class DbActorSuite() extends TestKit(ActorSystem("myTestActorSystem"))
...
+
+    answer1 shouldBe a [DbActor.DbActorAck]
+
+    val ask2 = dbActorRef ? DbActor.CreateChannel(channel, ObjectType.LAO)
+    val answer2 = Await.result(ask2, duration)
+
+    answer2 shouldBe a [DbActor.DbActorNAck]
+
+    val ask3 = dbActorRef ? DbActor.Write(channel, message)
+    val answer3 = Await.result(ask3, duration)
```
RaulinN, line 215: Answer3 is not tested

RaulinN (changes requested):
Alright! Here we go ^^' I apologize for the time it took, but doing a good review of half a semester worth of code takes a looot of time! Here are a few pointers for next time:

- Create shorter pull requests. You have multiple different topics covered in this pull request. You could have had for example 4 PR (e.g. Add chirp pr, election pr, roll call pr, db pr)
- Only create meaningful commits. A lot of commits are one-liners

Overall, some important changes to do in my opinion. Tests are good!

One more thing, @BadrTad @tuomas-p: I would **strongly** suggest performing an automatic import cleanup on the whole project. We have **a lot**, and I mean **A LOT** of unused imports in basically every single file. You could do it in this PR (as a separate commit please :) ) or create a PR for that specifically after this one is merged

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
tuomas-p: Yes, it does that

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/Validator.scala
```diff
@@ -17,6 +18,23 @@ import ch.epfl.pop.pubsub.graph.validators.ParamsValidator._
...
+import com.networknt.schema.JsonSchemaFactory
+import com.networknt.schema.ValidationMessage
+import com.networknt.schema.SpecVersion
+
+import scala.io.{BufferedSource, Source}
+import scala.util.{Failure, Success, Try}
+
+import spray.json._
+
+import java.io.File
```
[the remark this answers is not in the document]
tuomas-p: corrected (from Pierluca's comment)

be2-scala/src/main/scala/ch/epfl/pop/model/network/method/message/data/dataObject/ChannelData.scala
```diff
@@ -0,0 +1,38 @@
...
+    channel_type: ObjectType.ObjectType,
```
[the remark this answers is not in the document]
tuomas-p: I'll correct this

be2-scala/src/main/scala/ch/epfl/pop/model/network/method/message/data/dataObject/ChannelData.scala
```diff
@@ -0,0 +1,38 @@
...
+  def apply(
+             channel_type: ObjectType.ObjectType,
+             messages: List[Hash]
+           ): ChannelData = {
+    new ChannelData(channel_type, messages)
+  }
+
+  override def buildFromJson(payload: String): ChannelData = payload.parseJson.asJsObject.convertTo[ChannelData] // doesn't decode data
+
+  def getName: String = "ChannelData"
```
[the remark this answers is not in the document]
tuomas-p: the buildfromJson yes, since we need it to convert it back from the database, the getName could be, for the logs.

be2-scala/src/main/scala/ch/epfl/pop/model/network/method/message/data/dataObject/LaoData.scala
```diff
@@ -0,0 +1,30 @@
...
+  def apply(
+             owner: PublicKey,
+             attendees: List[PublicKey]
+           ): LaoData = {
+    new LaoData(owner, attendees)
+  }
+
+  override def buildFromJson(payload: String): LaoData = payload.parseJson.asJsObject.convertTo[LaoData] // doesn't decode data
+
+  def getName: String = "LaoData"
```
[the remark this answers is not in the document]
tuomas-p: could also be useful for the logs (if we want to print the type of the stored object)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandler.scala
```diff
@@ -34,13 +35,14 @@ object ElectionHandler extends MessageHandler {
   }
 
   def handleSetupElection(rpcMessage: JsonRpcRequestSetupElection): GraphMessage = {
+    //FIXME: add election info to election channel/electionData
```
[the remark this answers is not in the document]
tuomas-p: Actually, for validating the votes, we will need to create a new object, Electiondata (similar to LaoData for the use), to check the amount of questions, choices, etc. This will be the subject of another PR.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
```diff
@@ -61,11 +87,11 @@ object DbActor extends AskPatternConstants {
   final case class WriteAndPropagate(channel: Channel, message: Message) extends Event
 
   /**
-   * Request to create channel <channel> in the db
+   * Request to create channel <channel> in the db with a type
    *
    * @param channel channel to create
    */
-  final case class CreateChannel(channel: Channel) extends Event
+  final case class CreateChannel(channel: Channel, objectType: ObjectType.ObjectType) extends Event
```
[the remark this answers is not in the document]
tuomas-p: thanks!

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
```diff
@@ -137,157 +181,249 @@ object DbActor extends AskPatternConstants {
...
+  // In the arguments, we might remove the initialChannelsMap and replace the channelNamesDb by the general Db??
+  // FIXME: find way to keep count of channel nb. Idea: store key "channelCount" with value int.getBytes to db and increment it with each channel created
+  sealed case class DbActor(mediatorRef: ActorRef, db: DB, DATABASE_FOLDER: String) extends Actor with ActorLogging {
+    
+    //helper function to determine if a channel should be created during write operations
+    private def writeCreateNewChannel(channel: Channel, message: Message): DbActorMessage = {
+      Try(db.get(channel.toString.getBytes)) match {
+        case Success(bytes) if bytes != null =>
+          log.info(s"Channel '$channel' already exists.")
+          DbActorAck()
```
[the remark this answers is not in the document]
tuomas-p: Normally yes, but I'll go through them again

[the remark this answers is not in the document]
tuomas-p: Okay so it looks a bit confusing, but it is not an issue. The real CreateChannel function returns a NAck, as it should. The WriteCreateNewChannel tests whether a channel exists for a write, if not, it creates it. There might be a write-through problem (as discussed with @pierluca and @BadrTad), but the functionality can be removed easily if needed. If the write is on an existing channel, the function will return an Ack and the code will run without doing anything new. If the new channel is created, there will be an Ack too and the write-through functionality will be used to then write the message to the newly created channel.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
```diff
@@ -137,157 +181,249 @@ object DbActor extends AskPatternConstants {
...
+          val objectType = objectTypeOption match {
+            case Some(data) => data._object
+            case _ => ObjectType.LAO
+          }
+          // for now, we don't have meetup or roll call channels, so we just create Lao channels instead, easy to change if needed
+          val actualObjectType: ObjectType.ObjectType = if (objectType == ObjectType.ELECTION || objectType == ObjectType.CHIRP){
+            objectType
+          } else {
+            ObjectType.LAO
+          }
```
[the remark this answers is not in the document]
tuomas-p: Cleaned this up a bit, but basically at the moment all of the messages excluding chirps and messages concerning elections are sent to a channel of "type lao" (the main channel), which I chose as the default ObjectType. This is to enforce that no message is written to a wrong type of channel (e.g. a chirp on the main lao channel)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
```diff
@@ -137,157 +181,249 @@ object DbActor extends AskPatternConstants {
...
+              }
+            }
+            case Failure(exception) =>
+              log.error(s"Actor $self (db) encountered a problem while writing message_id '$messageId' on channel '$channel'")
+              DbActorNAck(ErrorCodes.SERVER_ERROR.id, exception.getMessage)
+            case _ =>
+              log.error(s"Actor $self (db) encountered a problem while writing message_id '$messageId' on channel '$channel', as the channel does not exist")
+              DbActorNAck(ErrorCodes.SERVER_ERROR.id, "Could not decode channel data")
+          }
+        }
```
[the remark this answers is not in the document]
tuomas-p: I need to change the error message to "channel does not exist" indeed

be2-scala/src/main/scala/ch/epfl/pop/model/objects/Channel.scala
```diff
@@ -28,6 +28,15 @@ final case class Channel(channel: String) {
...
   def extractChildChannel: Hash = Hash(Base64Data(channel.split(Channel.SEPARATOR).last))
 
+  def extractLaoId: String = {
+    if (isRootChannel){
+      null
+    }
+    else{
+      channel.split(Channel.SEPARATOR)(2) //since the channel starts with '/'
+    }
+  }
```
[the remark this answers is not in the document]
tuomas-p: This method is now useless actually, I'm deleting it (decodeSubChannel works very well)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
```diff
@@ -137,157 +181,249 @@ object DbActor extends AskPatternConstants {
...
+              case Failure(exception) =>
+                log.error(s"Actor $self (db) encountered a problem while writing message_id '$messageId' on channel '$channel'")
+                DbActorNAck(ErrorCodes.SERVER_ERROR.id, exception.getMessage)
+              case _ =>
+                log.error(s"Actor $self (db) encountered a problem while writing message_id '$messageId' on channel '$channel', as the channel does not exist")
+                DbActorNAck(ErrorCodes.SERVER_ERROR.id, "Could not decode channel data")
+            }
+          }
+        }
       }
```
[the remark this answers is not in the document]
tuomas-p: I originally did not want to use returns (because well, Scala), but if it is recommended, then I'll go ahead, but I'll remove the redundant braces

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
```diff
@@ -137,157 +181,249 @@ object DbActor extends AskPatternConstants {
...
+                    val json = new String(bytes, StandardCharsets.UTF_8)
+                    batch.put(dataKey.getBytes, obj.toJsonString.getBytes)
+                    batch.put(generateMessageKey(channel, messageId).getBytes, message.toJsonString.getBytes)
+                    batch.put(channel.toString.getBytes, ChannelData.buildFromJson(json).addMessage(messageId).toJsonString.getBytes)
+                    //allows writing all data atomically
+                    writeBatch(channel, messageId, batch)
+                  }
+                  case Failure(exception) =>
+                    log.error(s"Actor $self (db) encountered a problem while writing message_id '$messageId' and objects " + ChannelData.getName + " and " + LaoData.getName + s" on channel '$channel' because of batch creation")
+                    DbActorNAck(ErrorCodes.SERVER_ERROR.id, exception.getMessage)
```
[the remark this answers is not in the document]
tuomas-p: This function will be removed with the next commit (as everything is done in write now)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
```diff
@@ -137,157 +181,249 @@ object DbActor extends AskPatternConstants {
...
-      if (channelsMap.contains(channel)) {
-        val iterator: DBIterator = channelsMap(channel).iterator
-
-        iterator.seekToFirst()
-        val result: List[Message] = buildCatchupList(iterator, List.empty).reverse
-        iterator.close()
+    private def readLaoData(channel: Channel): DbActorMessage = {
+      val dataKey: String = generateLaoDataKey(channel)
+      dataKey match {
+        case null => DbActorReadLaoDataAck(None)
```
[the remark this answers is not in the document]
tuomas-p: A NAck could be cleaner I guess, I'll add that

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
```diff
@@ -296,10 +432,24 @@ object DbActor extends AskPatternConstants {
       case Write(channel, message) =>
         log.info(s"Actor $self (db) received a WRITE request on channel '$channel'")
         sender ! write(channel, message)
+      
+      
+      case WriteLaoData(channel, message, laoData) =>
+        log.info(s"Actor $self (db) received a WRITELAODATA request on channel '$channel'")
+        sender ! writeLaoData(channel, message, laoData)
```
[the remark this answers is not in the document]
tuomas-p: I'd prefer the snake case, since the other logs are written with uppercase letters, it would seem more logical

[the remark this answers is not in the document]
tuomas-p: to remain consistent with Scala, I'll use camelCase

be2-scala/src/test/scala/ch/epfl/pop/pubsub/graph/DbActorSuite.scala
```diff
@@ -10,6 +10,8 @@ import ch.epfl.pop.pubsub.{AskPatternConstants, PubSubMediator}
 import org.scalatest.{BeforeAndAfterAll, FunSuiteLike, Matchers}
 import akka.pattern.ask
```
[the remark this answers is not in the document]
tuomas-p: Thank you, I'll pay attention when merging

be2-scala/src/main/scala/ch/epfl/pop/model/network/method/message/data/dataObject/LaoData.scala
```diff
@@ -0,0 +1,30 @@
...
+import ch.epfl.pop.json.HighLevelProtocol._
+import ch.epfl.pop.json.MessageDataProtocol._
+import ch.epfl.pop.model.network.Parsable
+import ch.epfl.pop.model.objects._
+import spray.json._
+
+case class LaoData(
+    owner: PublicKey,
+    attendees: List[PublicKey]
+){
```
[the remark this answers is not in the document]
tuomas-p: good point

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/SocialMediaValidator.scala
tuomas-p: No, as it was created this semester. Indeed, it should be discussed

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/LaoHandler.scala
```diff
@@ -39,9 +41,17 @@ case object LaoHandler extends MessageHandler {
...
+        val ask: Future[GraphMessage] = (dbActor ? DbActor.WriteLaoData(channel, message, laoData)).map {
+          case DbActorWriteAck() => {
+            val socialChannel: Channel = Channel(channel + "/posts")
```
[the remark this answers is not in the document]
tuomas-p: corrected

be2-scala/src/test/scala/ch/epfl/pop/model/network/method/message/data/dataObject/LaoDataSuite.scala
```diff
@@ -0,0 +1,28 @@
...
+        laoData2.attendees should equal(List(PublicKey(Base64Data("b"))))
+    }
+
+    test("Json conversions work for LaoData") {
+        val laoData: LaoData = LaoData(PublicKey(Base64Data("a")), List.empty)
+
+        val laoData2: LaoData = LaoData.buildFromJson(laoData.toJsonString)
+
+        laoData2.attendees should equal (List.empty)
+        laoData2.owner should equal (PublicKey(Base64Data("a")))
```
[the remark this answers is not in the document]
tuomas-p: indeed

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/SocialMediaValidator.scala
tuomas-p: No, as it was created this semester. Indeed, it should be discussed

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/SocialMediaValidator.scala
tuomas-p: me neither to be honest, removing this

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/SocialMediaValidator.scala
tuomas-p: nope, removing useless stuff

be2-scala/src/main/scala/ch/epfl/pop/model/network/method/message/data/dataObject/ChannelData.scala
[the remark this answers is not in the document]
RaulinN: Yes github does not like when you highlight a single line of code :/ I was talking about line 37. I noticed that you use `getName` only 1-2 times in the code. I absolutely don't mind keeping it if you think it's better/you might reuse it later. Just wanted to make sure :)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
```diff
@@ -137,157 +181,249 @@ object DbActor extends AskPatternConstants {
...
+
+              @scala.annotation.tailrec
+              def buildCatchupList(msgIds: List[Hash], acc: List[Message]): List[Message] = {
+                msgIds match {
+                  case Nil => acc
+                  case head::tail => {
+                    buildCatchupList(tail, Message.buildFromJson(new String(db.get(generateMessageKey(channel, head).getBytes), StandardCharsets.UTF_8)) :: acc)
+                  }
+                }
+              }
```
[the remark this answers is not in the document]
tuomas-p: thank you for the advice!

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/MessageHandler.scala
```diff
@@ -44,11 +44,13 @@ trait MessageHandler extends AskPatternConstants {
...
-      case DbActor.DbActorWriteAck => Left(rpcMessage)
+      case DbActor.DbActorWriteAck() => Left(rpcMessage)
       case DbActor.DbActorNAck(code, description) => Right(PipelineError(code, description, rpcMessage.id))
       case _ => Right(PipelineError(ErrorCodes.SERVER_ERROR.id, "Database actor returned an unknown answer", rpcMessage.id))
     }
 
     ask
   }
+
+  //FIXME: add method/modify existing methods in order to use the new database structure, for both with and without objects
```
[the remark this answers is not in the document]
tuomas-p: Actually, with the new implementation of the write object and the write in the same function @pierluca suggested, these will not be needed!

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
tuomas-p: actually we can use the error code, as it changes between the two different outputs!

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/SocialMediaValidator.scala
tuomas-p: Actually, there is no need for this, at least for the current social media messages

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
tuomas-p: I've been thinking about this and will talk about it during our meeting, the problem being that each channel needs to be created separately and we should take into account the failure directly, in order to prevent problems later on. The "easiest" way I found for now was therefore to directly return a GraphMessage, however, the function is still a bit messy I agree with you

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/SocialMediaValidator.scala
tuomas-p: We just need to override it with something (for now) because of what it extends

tuomas-p: Or maybe I should remove the eventvalidator trait for now

RaulinN (approved):
Alright look good! 

I will not block the PR longer don't worry, but what happened to the file `LaoData`? I see that it has new methods inside that you only use in tests. Do you plan on using them in the actual codebase as well or not? 
- if yes, then everything is fine
- if not, then I'd suggest using a RichClass in test to enhance the functionaliies of LaoData. This way you don't have to keep test-only methods in the real codebase

tuomas-p:
@RaulinN , I'm planning to use (at least probably) all the functions present in LaoData in the future (I think for now DbActor uses most of them), thank you for the review!

PR #594, opened 2021-11-16
Store and display chirps

Xelowak:
Here is what I did for the social media project:

- Add a message handler
- Add a reducer
- Display sent chirps by the organizer in user feed of every attendee (and of the organizer himself) using @jiabaow 's ChirpCard component
- Social page is only accessible if we are currently connected to a LAO

Tests will be in another PR.


fe1-web/store/reducers/SocialReducer.ts
```diff
@@ -0,0 +1,87 @@
...
+const socialReducerPath = 'social';
+
+const socialSlice = createSlice({
+  name: socialReducerPath,
+  initialState,
+  reducers: {
+
+    // Add a chirp to the list of chirps
+    addChirp: {
+      prepare(laoId: Hash | string, chirp: ChirpState): any {
```
jiabaow, line 41: why Hash | string?

jiabaow (dismissed):

fe1-web/ingestion/handlers/SocialHandler.ts
```diff
@@ -0,0 +1,74 @@
...
+    console.warn('handleAddChirp was called to process an unsupported message');
+    return false;
+  }
+
+  const makeErr = (err: string) => `chirp/add was not processed: ${err}`;
+
+  const storeState = getStore().getState();
+  const lao = getCurrentLao(storeState);
+  if (!lao) {
+    console.warn(makeErr('no LAO is currently active'));
```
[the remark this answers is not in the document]
Xelowak: I don't think it should make the application crash if it happens.

fe1-web/ingestion/handlers/index.ts
```diff
@@ -3,3 +3,4 @@ export * from './Lao';
 export * from './Meeting';
 export * from './RollCall';
 export * from './Witness';
+export * from './SocialHandler';
```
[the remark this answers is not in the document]
Xelowak: Actually it was named 'Social' first, but I don't like having files that have the same name in different folders. In my opinion, every other handler should be renamed to <purpose>Handler, but I don't know what you think of it.

fe1-web/components/ChirpCard.tsx
```diff
@@ -0,0 +1,79 @@
...
+        </View>
+      </View>
+    </View>
+  );
+};
+
+const propTypes = {
+  sender: PropTypes.string.isRequired,
+  text: PropTypes.string.isRequired,
+  time: PropTypes.instanceOf(Date).isRequired,
```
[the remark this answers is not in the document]
Xelowak: Yes I just need to update it.

fe1-web/store/reducers/SocialReducer.ts
Xelowak: laoId.valueOf()

Xelowak: Actually, it is directly converted to a value as you can see on the line just below this one. Then, it is just for convenience, so that we can call this function with an id as a Hash or as a string. In the codebase, the Hash is pretty much everywhere but in states, so we can omit any conversion when calling addChirp.

fe1-web/ingestion/handlers/index.ts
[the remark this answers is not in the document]
jiabaow: I agree that we can rename others to xxxHandler, so it's clear about their functionality. We can keep the SocialHandler and rename the others later.

fe1-web/ingestion/handlers/RollCall.ts
```diff
@@ -132,6 +135,14 @@ function handleRollCallCloseMessage(msg: ExtendedMessage): boolean {
     }
   }));
 
+  // For now, everyone is automatically subscribed to the organizer's social channel at the end of
+  // the roll call
+  subscribeToChannel(getUserSocialChannel(lao.id, lao.organizer.valueOf())).then(() => {})
```
pierluca, line 140: you can omit the empty `then`, it'll also make Sonar happier

fe1-web/ingestion/handlers/SocialHandler.ts
```diff
@@ -0,0 +1,75 @@
+import { ExtendedMessage } from 'model/network/method/message';
+import { ActionType, AddChirp, ObjectType } from 'model/network/method/message/data';
+import {
+  addChirp,
+  dispatch,
+  getStore,
+  KeyPairStore,
```
pierluca, line 7: ehm ehm...

pierluca (dismissed):
Tiny remark, you can merge but I'd suggest fixing those minor issues.

fe1-web/store/reducers/SocialReducer.ts
```diff
@@ -53,7 +53,7 @@ const socialSlice = createSlice({
           };
         }
 
-        state.byLaoId[laoId].allChirps.push(chirp);
+        state.byLaoId[laoId].allChirps.unshift(chirp);
```
pierluca, line 56: no need to address it in this PR, but just be aware that this will be buggy as you could receive multiple times the same chirp.

pierluca (approved):
Approved as -is, coverage will come as part of a separate PR, as discussed.

fe1-web/store/reducers/SocialReducer.ts
Xelowak: Is it the fact that I use unshift the problem or is it about the fact that this is a list ?
Anyway, this list is only temporary and will be replaced by something more suitable.

PR #600, opened 2021-11-17
Work fe1 jwen social home navigation 

jiabaow:
Screenshot 2021-11-17 at 22 14 42
Add a navigation bar for social media.

Same as the following one but solve conflicts by merging this to the master (on this new branch)
https://github.com/dedis/student_21_pop/pull/559


jiabaow:
Cannot test the navigation for some weird reason.

fe1-web/parts/lao/socialMedia/SocialFollows.tsx
```diff
@@ -0,0 +1,11 @@
+import * as React from 'react';
+import { Text, View } from 'react-native';
+import STRINGS from 'res/strings';
+
+const SocialFollows = () => (
+  <View style={{ flex: 1, justifyContent: 'center', alignItems: 'center' }}>
```
Xelowak, line 6: Wouldn't it be better to store this style in a Stylesheet ? I know this page is temporary but it is cleaner this way in my opinion.

fe1-web/parts/lao/socialMedia/SocialMediaNavigation.tsx
```diff
@@ -0,0 +1,51 @@
...
+            iconName = 'search';
+            break;
+          case STRINGS.social_media_navigation_tab_follows:
+            iconName = 'people';
+            break;
+          case STRINGS.social_media_navigation_tab_profile:
+            iconName = 'person';
+            break;
+          default:
+            console.log('wrong route.');
```
Xelowak, line 34: Is console.log the most appropriate there or is it better to throw an error ? I'm not sure to be honest, but it may be a good idea to think about it.

fe1-web/parts/lao/socialMedia/SocialProfile.tsx
```diff
@@ -0,0 +1,11 @@
+import * as React from 'react';
+import { Text, View } from 'react-native';
+import STRINGS from 'res/strings';
+
+const SocialProfile = () => (
+  <View style={{ flex: 1, justifyContent: 'center', alignItems: 'center' }}>
```
Xelowak, line 6: Same comment about style sheet. By the way since it is the same everywhere, maybe you can declare it somewhere and export it to other files, to avoid duplication.

fe1-web/parts/lao/socialMedia/SocialSearch.tsx
```diff
@@ -0,0 +1,11 @@
+import * as React from 'react';
+import { Text, View } from 'react-native';
+import STRINGS from 'res/strings';
+
+const SocialSearch = () => (
+  <View style={{ flex: 1, justifyContent: 'center', alignItems: 'center' }}>
```
Xelowak, line 6: Same comment.

fe1-web/tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
   "compilerOptions": {
+    "experimentalDecorators": true,
```
Xelowak, line 3: Is this needed, and for what ?

Xelowak (approved):
Nice job overall ! I made some comments you need to consider, but nothing really important.

fe1-web/tsconfig.json
jiabaow: I don't remember where it came from, but the code works well without it. I will delete it.

fe1-web/parts/lao/socialMedia/SocialProfile.tsx
jiabaow: Thanks for the advice! But I will delete them for simplicity, we don't need them here.

fe1-web/parts/lao/socialMedia/SocialMediaNavigation.tsx
jiabaow: I will change it to console.error but I am nearly sure that we will not enter the default case.

Xelowak:
I've just remarked that you have created a new folder for all social pages in parts/lao/socialMedia. I would suggest using parts/social, as it is more consistent with the rest. The first level of parts seems to contain all the pages on the navigation bar  (which is what we want) and parts/lao is more about who you are on the current lao.

fe1-web/package.json
```diff
@@ -58,6 +58,7 @@
     "react-navigation": "^4.4.1",
     "react-qr-reader": "^2.2.1",
     "react-redux": "^7.2.4",
+    "react-test-renderer": "^17.0.2",
```
pierluca, line 61: I assume this is a devDependency too :)

pierluca (commented):
Minor package.json issue, otherwise LGTM

fe1-web/tsconfig.json
pierluca: I think WebStorm was possibly complaining without it 😅 I remember having a warning like that.

jiabaow:
> I've just remarked that you have created a new folder for all social pages in parts/lao/socialMedia. I would suggest using parts/social, as it is more consistent with the rest. The first level of parts seems to contain all the pages on the navigation bar (which is what we want) and parts/lao is more about who you are on the current lao.

I put this folder inside /lao because we only want to show them after connecting to a lao. Maybe we can have a discussion about the arrangements later.

Xelowak (approved):
LGTM

pierluca (approved):

PR #604, opened 2021-11-18
Fix consensus UI and subscription

arnauds5:
- When proposer received all elect-accept, now it send directly a learn message instead of asking to confirm it.
- Change how subscription to consensus are done : only witness and organizer subscribe now.
- Always send a catchup after receiving a subscribe response.


GabrielFleischer (approved):
Great work !

Lugiasc (approved):

jbsv (approved):

PR #567, opened 2021-11-06
show a notification after scanned a participant

jiabaow:
I use https://www.npmjs.com/package/react-native-toast-notifications to display notifications after successfully scanning a participant.
Screenshot 2021-11-10 at 17 24 46

PR #609, opened 2021-11-20
Work fe1 jwen roll call scanned notification copy

jiabaow:
same as #567


jiabaow:
CI is not happy with my import order, but if I change it, the mocks won't work :(

jiabaow:
I don't know how to test duplicate scans, checking if a list contains the data in the argument made the mock reader too specific.
Also testing onCloseRollCall is called with scanned Pop tokens, CloseRollCall data gets LaoID from OpenedLaoStore, I don't know if it's worth it for mocking a bunch of things just for this small improvement.

fe1-web/parts/lao/organizer/RollCallOpened.tsx
```diff
@@ -67,11 +71,11 @@ const RollCallOpened = () => {
         <TextBlock text={STRINGS.roll_call_scan_description} />
         <QrReader
           delay={300}
-          onScan={handleScan}
+          onScan={handleScan} // {()=>handleScan} doesn't work
```
Xelowak, line 74: Is it underlined in red ? If yes, try this:
```suggestion
          onScan={() => handleScan()}
```
You didn't type "handleScan()" in your comment, that's why I'm asking

fe1-web/parts/lao/organizer/RollCallOpened.tsx
```diff
@@ -29,20 +30,23 @@ const RollCallOpened = () => {
...
-      setQrWasScanned(true);
-      updateAttendeesSet((prev) => new Set<string>(prev.add(data)));
-      // TODO: use toast to display the scanned message, instead of console.log
-      console.log(STRINGS.roll_call_scan_participant);
+      if (!attendees.includes(data)) {
+        updateAttendees((arr) => [...arr, data]);
+        toast.show(STRINGS.roll_call_scan_participant, {
+          type: 'success',
+          placement: 'top',
+          duration: 4000,
```
Xelowak, line 47: Is this 4 seconds ? I suggest storing it in a constant to make it clearer.

fe1-web/parts/lao/organizer/RollCallOpened.tsx
```diff
@@ -29,20 +30,23 @@ const RollCallOpened = () => {
   const route = useRoute();
   const { rollCallID, time } = route.params;
   const navigation = useNavigation();
-  const [, setQrWasScanned] = useState(false);
-  const [attendeesSet, updateAttendeesSet] = useState(new Set<string>());
-  const attendees = Array.from(attendeesSet);
+  const [attendees, updateAttendees] = useState<string[]>([]);
```
Xelowak, line 33: Why did we switch to a list ? Isn't it problematic if someone is scanned twice ?

fe1-web/parts/lao/organizer/__tests__/RollCallOpened.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+  Hash, Lao, LaoState, PublicKey, Timestamp,
+} from 'model/objects';
+import STRINGS from 'res/strings';
+import keyPair from 'test_data/keypair.json';
+import { requestCloseRollCall as mockRequestCloseRollCall } from 'network/MessageApi';
+import { OpenedLaoStore } from 'store';
+// @ts-ignore
+import { fireScan as fakeQrReaderScan } from 'react-qr-reader';
+
+import RollCallOpened from '../RollCallOpened'; // Don't change the order, otherwise the tests will fail
```
Xelowak, line 14: Is the import order still a problem ? I thought you fixed it. If yes, you can safely remove this comment.

fe1-web/parts/lao/organizer/__tests__/RollCallOpened.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+jest.mock('@react-navigation/core');
+
+export const mockPublicKey = new PublicKey(keyPair.publicKey);
+
+const org = mockPublicKey;
+const time = new Timestamp(1609455600).toString(); // 1st january 2021
```
Xelowak, line 21: I'd suggest storing the number 1609455600 in a constant, since you use it multiple times.

Xelowak (approved):
Thanks a lot for the work ! I've done some comments you might need to address.

fe1-web/parts/lao/organizer/RollCallOpened.tsx
jiabaow: I use a list here because later we need a list of attendees for requestCloseRollCall, I verify newly scanned data is not included in the attendees' list in handleScan to make sure there will be no duplicate.

Xelowak: Okay, it seems reasonable

fe1-web/package.json
```diff
@@ -29,6 +29,7 @@
     "@react-navigation/routers": "^6.0.1",
     "@react-navigation/stack": "^6.0.7",
     "@reduxjs/toolkit": "^1.5.0",
+    "@types/react-qr-reader": "^2.1.4",
     "@types/react-timeago": "^4.1.3",
```
pierluca, line 33: shouldn't these `@types/*` be devDependencies ?

fe1-web/package.json
```diff
@@ -39,7 +40,9 @@
     "expo": "^42.0.0",
     "expo-status-bar": "~1.0.4",
     "idb-keyval": "^5.0.4",
+    "jest-expo": "^43.0.1",
     "js-sha256": "^0.9.0",
+    "madge": "^5.0.1",
```
pierluca, line 45: I believe these (except `js-sha256`) should also be devDependencies.
By the way, I didn't know about madge - that's a cool tool !

fe1-web/parts/lao/organizer/RollCallOpened.tsx
pierluca: I won't bother you with this given the application, but keep in mind:
1. Effectively, this means you're artificially forcing an array to behave as a set - it might not be super intuitive when reading the code
2. You could always convert from a set to an array later
3. From an algorithmic perspective, this changed the creation of a list of attendees from a complexity of O(n*log(n)) to O(n^2) algorithm

Again, these are no issues in this context, but the change is a bit counter-intuitive to me :) 
At the opposite end of this solution, you could also imagine changing the `requestCloseRollCall` to take a set.

fe1-web/parts/lao/organizer/__tests__/RollCallOpened.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+  id: '1234',
+  name: 'MyLao',
+  creation: TIMESTAMP,
+  last_modified: TIMESTAMP,
+  organizer: '1234',
+  witnesses: [],
+};
+
+jest.mock('react-qr-reader');
+jest.mock('network/MessageApi');
```
pierluca, line 36: why not group all the mocks together? (with the `@react-navigation/core`)

pierluca (commented):
Minor remarks, this looks very good to me !
I'd really try to fix the package.json, everything else is cosmetic :)

fe1-web/package.json
jiabaow: I forgot to remove madge after checking circular dependency because of the problem with the import order.

fe1-web/parts/lao/organizer/RollCallOpened.tsx
jiabaow: Thanks for the comment, I didn't think about the complexity. I will change it back to a set and convert the set to an array for requestCloseRollCall, I won't change requestCloseRollCall for this PR because this will also change JSON.

pierluca (approved):
LGTM !

fe1-web/package.json
pierluca: It's a cool tool though !
Do you think it'd be useful to add it as a check in the CI ?

jiabaow: It might not be useful to add it in the CI, because if there's a circular dependency, the tests will fail, we won't do a PR. But it will be useful to add it somewhere else like useful tools, to tell others what they can do in case of a similar problem.

PR #611, opened 2021-11-21
Add json validation

arnauds5:
Some infos
- The validation of json is done when serializing/deserializing, using the schemas from the protocol
- All schemas/examples are copied into 'src/main/resources/protocol' and ignored in the .gitignore
- Fix PublicKeySignaturePair serialization (fields need to be base64 string, not bytes array)
- Fix some tests that was using invalid data
- Add some json tests
- Add equals/hashcode to immutable classes
- Currently the schemas are loaded asynchronously in HomeActivity, but in Android 11 it seems that AsyncTask is depreacted ?
- If a test use the json validation, it could need to load schemas in @Before (it could take up to 1 seconds the first time to load)
- Once the json schemas are loaded, it doesn't seems to slow the app

### DIFF
--- a/fe2-android/app/build.gradle
+++ b/fe2-android/app/build.gradle
@@ -132,6 +132,7 @@ dependencies {
     implementation 'com.google.android.gms:play-services-vision:20.1.3'
     implementation 'com.google.code.gson:gson:2.8.6'
     implementation 'com.google.crypto.tink:tink-android:1.5.0'
+    implementation 'com.networknt:json-schema-validator:1.0.46'
     implementation 'com.squareup.okhttp3:logging-interceptor:4.9.1'
     implementation 'com.tinder.scarlet:scarlet:0.1.11'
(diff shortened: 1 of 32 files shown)


fe2-android/app/build.gradle
```diff
@@ -160,7 +160,7 @@ dependencies {
 
     testImplementation 'org.slf4j:slf4j-api:1.7.30'
     testImplementation 'org.slf4j:slf4j-simple:1.7.30'
-    testImplementation 'com.networknt:json-schema-validator:1.0.46'
+    implementation 'com.networknt:json-schema-validator:1.0.46'
```
GabrielFleischer, line 163: Move this with the other implementation libraries

GabrielFleischer (dismissed):
LGTM

Lugiasc (approved):

GabrielFleischer (approved):

jbsv (approved):

PR #612, opened 2021-11-21
Redesign social page

Xelowak:
I changed the design of the social page and chirps, for it to be more visually appealing. 
Don't hesitate to make comments/suggestions. Note the screenshots have been taken on a 1440p screen, that's why it may appear quite small.
The "time ago" text needs to be in the same font as the rest, but since it is an external library, it's harder to do than I thought, so I'll probably do it later.

Here are screenshots of the new design:

Capture d’écran 2021-11-21 à 18 38 40

Capture d’écran 2021-11-21 à 18 31 14

In the future, these "person" icons will be replaced by images generated by each token.


jiabaow (dismissed):
Thanks for the improvement!
LGTM but don't forget to update the snapshot as tests failed

pierluca:
Do you want to resolve the merge conflicts? :)

Xelowak:
Yes I will !

fe1-web/components/ChirpCard.tsx
```diff
@@ -63,7 +105,7 @@ const ChirpCard = (props: IPropTypes) => {
 const propTypes = {
   sender: PropTypes.string.isRequired,
   text: PropTypes.string.isRequired,
-  time: PropTypes.instanceOf(Timestamp).isRequired,
+  time: PropTypes.number.isRequired,
```
pierluca, line 108: Why not directly taking a chirp object as property?

Xelowak: That's a really good remark, thanks a lot for the suggestion. I didn't think about it but it is much more practical.

pierluca (approved):
LGTM !

PR #617, opened 2021-11-24
Merge the two command line interface

rjaccard:


be1-go/cli/mod.go
```diff
@@ -1,6 +1,6 @@
-// Package organizer contains the entry point for starting the organizer
+// Package cli contains the entry point for starting the organizer
 // server.
```
Anghul, line 2: You don't mention the witness in this comment

be1-go/cli/mod.go
```diff
@@ -19,23 +19,30 @@ import (
...
+
+	// get command line args which specify public key, port to use for clients
+	// and witnesses, witness' address
 	clientPort := cliCtx.Int("client-port")
 	witnessPort := cliCtx.Int("witness-port")
 	if clientPort == witnessPort {
 		return xerrors.Errorf("client and witness ports must be different")
 	}
-	witness := cliCtx.StringSlice("other-witness")
+	otherWitness := cliCtx.StringSlice("other-witness")
```
Anghul, line 41: I think we would need a variable to add one (or more) organizer too

Anghul (changes requested):
Good job. I think that you could remove the package cli/utility and put back its content in the cli as I created it only because I had to use a function in both the witness and the organizer's cli

be1-go/cli/mod.go
rjaccard: I will leave it for some future work as the system does not support multiple organisers yet

nkcr (approved):
Good for me 👍 
Sonarcloud is not happy because there is less than 20% coverage on new code :/

Anghul (approved):

PR #619, opened 2021-11-24
Rename wallet reducer for tests

Xelowak:
I modified the name of the exported wallet reducer constant, to avoid any name collision during tests of other reducers.


pierluca:
CI complains:
Using exported name 'walletReducer' as identifier for default export  import/no-named-as-default

pierluca (approved):

PR #621, opened 2021-11-24
Check if the start time is in the past, and add tests

arnauds5:
- computeTimesInSeconds now return false if there was a problem with start/end.
- if start is in the past (when we select or if we wait and then click on confirm)
  - more than 5 min => invalidate time
  - less thant 5 min => replace start by now
- if the end is before the start => invalidate time
  for example : set start to 16:00, end to 16:02, wait 4 min (start is updated to 16:04 > 16:02)
- If the start/end time are in the future, they are truncated to the minute
  16h05m25s => 16h05m00s

### DIFF
--- a/fe2-android/app/src/androidTest/java/com/github/dedis/popstellar/ui/detail/event/fragments/ElectionSetupFragmentTest.java
+++ b/fe2-android/app/src/androidTest/java/com/github/dedis/popstellar/ui/detail/event/fragments/ElectionSetupFragmentTest.java
@@ -233,9 +233,9 @@ public class ElectionSetupFragmentTest {
   }
 
   @Test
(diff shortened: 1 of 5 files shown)


Lugiasc (approved):
Seems good

GabrielFleischer (commented):
We discussed it in the global meeting and we agreed on :
Do not block event creation in the past, set the start time to now (and check that the end time is later)

When creating the event, make sure the start time is not too far in the past. (5 mins) and set it to now if it is in the past.

GabrielFleischer (approved):
LGTM

Lugiasc (approved):

jbsv (approved):

PR #622, opened 2021-11-25
Wallet tokens update only on rollcall change

pierluca:
This breaks a loop that would happen when setting up the wallet.

Effectively, we would update the state every time we render the component.
Every time we update the state, the component gets rendered again.

Using `useEffect`, bound to the changes to the `rollCalls` state variable, ensures that the rendering happens again ONLY if new roll calls are detected.


Xelowak (approved):
LGTM, thanks a lot for the bug fix !

jiabaow (approved):
LGTM

PR #623, opened 2021-11-25
Added build-web command

pierluca:


jiabaow (dismissed):
🤩

Xelowak (approved):
Thanks for the addition ! Actually, I think it is always recommended to run "npx expo-optimize" before building the website. Maybe you can create another command in package.json, or add it directly in the build.

Xelowak (approved):
Perfect, thank you !

fe1-web/package.json
```diff
@@ -5,6 +5,8 @@
     "android": "expo start --android",
     "ios": "expo start --ios",
     "web": "expo start --web",
+    "prebuild-web": "node_modules/expo-optimize/build/index.js",
+    "build-web": "expo build:web",
```
pierluca, line 9: @Xelowak @jiabaow 
FYI, "pre" and "post" are prefixes that can be used to execute some operation before or after a given command.
So `npm run build-web` will actually execute:
- prebuild-web
- build-web
- postbuild-web ( which doesn't exist, so it won't be run )

Xelowak: Okay, I didn't know that

PR #632, opened 2021-12-01
Work be2 tuomas p social media chirp

tuomas-p:
need to test broadcast functionality with front-end, but should be good otherwise


be2-scala/src/main/scala/ch/epfl/pop/model/objects/LaoData.scala
```diff
@@ -9,9 +9,15 @@ import ch.epfl.pop.model.network.method.message.data.lao.CreateLao
 import ch.epfl.pop.model.network.method.message.data.rollCall.CloseRollCall
 import spray.json._
 
+//Ed25519 is used to verify the signatures, therefore I also use it to sign data
+import com.google.crypto.tink.subtle.Ed25519Sign
+
+//a pair of byte arrays works better as keypair than the actual keypair object, since Ed25519Sign.Keypair isn't directly convertible to json with spray
```
BadrTad, line 15: For readability I would suggest defining a type like this for example.
`type KeyPaire = (Array[Bytes], Array[Bytes])`.
You can then pattern match on the keyPaire(tuple2) or even get public/private key array by using `._1 or ._2`

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/SocialMediaHandler.scala
```diff
@@ -30,14 +36,57 @@ case object SocialMediaHandler extends MessageHandler {
...
+                val timestamp: Timestamp = params.decodedData.get.asInstanceOf[AddChirp].timestamp
+                val addBroadcastChirp: AddBroadcastChirp = AddBroadcastChirp(chirpId, channelChirp.channel, timestamp)
+                val broadcastData: Base64Data = Base64Data.encode(addBroadcastChirp.toJson.toString) //FIXME: check json/string conversion
+                
+                //FIXME: however, should we have a package-private getter? I don't know whether it is secure enough.
+                val askLaoData = (dbActor ? DbActor.ReadLaoData(rpcMessage.getParamsChannel))
+                Await.result(askLaoData, duration) match {
+                  case DbActor.DbActorReadLaoDataAck(Some(laoData)) => {
+                    val pk: PublicKey = PublicKey(Base64Data.encode(laoData.publicKey))
+                    val ed: Ed25519Sign = new Ed25519Sign(laoData.privateKey)
```
BadrTad, line 59: I would recommend doing the signing in a helper method.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/SocialMediaHandler.scala
```diff
@@ -30,14 +36,57 @@ case object SocialMediaHandler extends MessageHandler {
...
+        channelChirp.decodeSubChannel match {
+          case(Some(lao_id)) => {
+            val broadcastChannel: Channel = Channel(Channel.rootChannelPrefix + Base64Data.encode(lao_id) + LaoHandler.SOCIALMEDIAPOSTSPREFIX)
+            rpcMessage.getParamsMessage match {
+              case Some(params) => {
+                // we can't get the message_id as a Base64Data, it is a Hash
+                val chirpId: Hash = params.message_id
+                val timestamp: Timestamp = params.decodedData.get.asInstanceOf[AddChirp].timestamp
+                val addBroadcastChirp: AddBroadcastChirp = AddBroadcastChirp(chirpId, channelChirp.channel, timestamp)
+                val broadcastData: Base64Data = Base64Data.encode(addBroadcastChirp.toJson.toString) //FIXME: check json/string conversion
```
BadrTad, line 52: This is no longer a valid FIXME or is it ?

BadrTad (dismissed):
Great overall ! 
+ The tests look great to me 👍 
+ Few suggestions when it comes to handling the  add of a chirp
and the key pairs for the LaoData

be2-scala/src/main/scala/ch/epfl/pop/model/objects/LaoData.scala
tuomas-p: I considered this, however for the practical use, I thought getting the public/private keys with their names instead of ._1 or ._2 (which could lead to confusion) or with getters would be less intuitive than with .publicKey/.privateKey

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/SocialMediaHandler.scala
tuomas-p: Indeed this comment was forgotten here

tuomas-p: going to do this

BadrTad (commented):
Looks good !

BadrTad (approved):

be2-scala/src/test/scala/ch/epfl/pop/model/network/ResultObjectSuite.scala
```diff
@@ -0,0 +1,27 @@
...
+        obj.resultMessages should equal (Some(List.empty))
+    }
+
+    test("isIntResult returns right result"){
+        val obj: ResultObject = new ResultObject(1)
+        val obj2: ResultObject = new ResultObject(List.empty)
+
+        obj.isIntResult should equal(true)
+        obj2.isIntResult should equal (false)
+    }
```
BadrTad, line 26: I have added a method `equals `to this class can you make some tests for it please since you already have it in your PR.
Commit where I add the equals  : `3f2204b909bfd9c1086642b7f1f6e56bd92d1eff`

BadrTad (approved):
Test equals method of ResultObject !

pierluca:
I see you added me as a reviewer in addition to @RaulinN .
The PR isn't exactly small, do you need a review from me on something specific ?

BadrTad:
@tuomas-p don't mean to bother you, this PR is big enough, but since my PR was merged yesterday (includes the `equals `method for `ResultObject`) is there any chance you cover it in your test suite ? Thanks !

tuomas-p:
@BadrTad yes of course, now that the PR was merged, I'll do it right away as it will be easier

tuomas-p:
@pierluca sorry for the bother, I believe I just added the same reviewers as last time by reflex, I don't believe that anything specific for this is needed, but thank you!

be2-scala/src/test/scala/ch/epfl/pop/model/network/JsonRpcResponseSuite.scala
```diff
@@ -0,0 +1,31 @@
...
+        val rpc: String = "rpc"
+        val res: Option[ResultObject] = Some(new ResultObject(2))
+        val err: Option[ErrorObject] = Some(new ErrorObject(1, "Error"))
+        val id: Option[Int] = Some(0)
+
+        val response: JsonRpcResponse = JsonRpcResponse(rpc, res, err, id)
+        response.jsonrpc should equal(rpc)
+        response.result should equal(res)
+        response.error should equal(err)
+        response.id should equal(id)
```
RaulinN, line 16: Interestingly enough, having an error field and a result field should not be legal. It is either one or the other. A JsonRpcResponse like this would not be accepted by front-ends. Could you add a simple check that prevents having both fields in the apply?

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/SocialMediaValidator.scala
```diff
@@ -33,4 +33,6 @@ case object SocialMediaValidator extends MessageDataContentValidator {
             case _ => Right(validationErrorNoMessage(rpcMessage.id))
         }
     }
-}
\ No newline at end of file
+
+    //no need for a addBroadcastChirpValidator, since the server will never receive one in theory
```
RaulinN, line 37: I am not very sure about that. I do not know exactly how the consensus project works. Maybe servers validate message they receive before trying to reach consensus with other servers. Could @pierluca tell us more about this (i.e. should we validate thoroughly such messages, or not)

In the general case, I would still have a validator for every type of messages (even if it does nothing or returns an error).

be2-scala/src/main/scala/ch/epfl/pop/model/network/method/message/data/socialMedia/AddBroadcastChirp.scala
```diff
@@ -0,0 +1,28 @@
...
+import ch.epfl.pop.json.MessageDataProtocol._
+import ch.epfl.pop.model.network.Parsable
+import ch.epfl.pop.model.network.method.message.data.ActionType.ActionType
+import ch.epfl.pop.model.network.method.message.data.ObjectType.ObjectType
+import ch.epfl.pop.model.network.method.message.data.{ActionType, MessageData, ObjectType}
+import ch.epfl.pop.model.objects.{Hash, Timestamp}
+import spray.json._
+
+case class AddBroadcastChirp(
+                          chirpId: Hash,
```
RaulinN, line 12: Careful here! this should read `chirp_id`. These fields are the **only** exceptions where we use snake_case in nthe project. The reason behind it is that the library we use for parsing/encoding will look for the fields you defined here in the json payload. In this case, spray-json will not find any `chirpId` field and reject the message because every message you receive will be using the syntax `chrip_id`

be2-scala/src/main/scala/ch/epfl/pop/model/network/method/message/data/socialMedia/AddBroadcastChirp.scala
```diff
@@ -0,0 +1,28 @@
...
+                          channel: String,
```
RaulinN, line 13: Is there a reason why you did not define it as a `Channel`?

be2-scala/src/main/scala/ch/epfl/pop/model/objects/LaoData.scala
```diff
@@ -9,9 +9,15 @@ import ch.epfl.pop.model.network.method.message.data.lao.CreateLao
...
 case class LaoData(
     owner: PublicKey,
     attendees: List[PublicKey],
+    privateKey: Array[Byte], 
+    publicKey: Array[Byte],
```
RaulinN, line 20: Mhmmm I would still use the objects we define here. Otherwise, we use all the benefits of OOP for keys. You can use PublicKey and a new PrivateKey objects here. PrivateKey does not exist (yet) because we had no reason (up until now) to treat private keys in the server. You could then add the function `signData` you defined in SocialMediaHandler as a method of private key

And finally, to address the comment `//a pair of byte arrays works better as keypair than the actual keypair object, since Ed25519Sign.Keypair isn't directly convertible to json with spray`. If, at keypair creation, you convert them to a PublicKey and PrivateKey, you'll have no issue. The reason behind it is that we tell spray-json how to encode/decode PublicKeys (see json/ObjectProtocol.scala:35):
```scala
  implicit object PublicKeyFormat extends JsonFormat[PublicKey] {
    override def read(json: JsValue): PublicKey = json match {
      case dataJs@JsString(_) => PublicKey(dataJs.convertTo[Base64Data])
      case _ => throw new IllegalArgumentException(s"Can't parse json value $json to a PublicKey object")
    }

    override def write(obj: PublicKey): JsValue = obj.base64Data.toJson
  }
```

Of course you'd have to copy-paste this with minor adjustements for a PrivateKey

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/SocialMediaHandler.scala
```diff
@@ -28,16 +34,64 @@ case object SocialMediaHandler extends MessageHandler {
     }
     case graphMessage@_ => graphMessage
   }
+
+  // helper function to sign data (can be used later elsewhere if we remove the private attribute if needed)
+  private def signData(data: Base64Data, privateKey: Array[Byte]): Signature = {
+    val ed: Ed25519Sign = new Ed25519Sign(privateKey)
+    Signature(Base64Data.encode(ed.sign(data.decode)))
```
RaulinN, line 41: `data.decode()`

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/SocialMediaHandler.scala
```diff
@@ -28,16 +34,64 @@ case object SocialMediaHandler extends MessageHandler {
...
+            val broadcastChannel: Channel = Channel(Channel.rootChannelPrefix + Base64Data.encode(lao_id) + LaoHandler.SOCIALMEDIAPOSTSPREFIX)
+            rpcMessage.getParamsMessage match {
+              case Some(params) => {
+                // we can't get the message_id as a Base64Data, it is a Hash
+                val chirpId: Hash = params.message_id
+                val timestamp: Timestamp = params.decodedData.get.asInstanceOf[AddChirp].timestamp
+                val addBroadcastChirp: AddBroadcastChirp = AddBroadcastChirp(chirpId, channelChirp.channel, timestamp)
+                val broadcastData: Base64Data = Base64Data.encode(addBroadcastChirp.toJson.toString)
+                
+                //FIXME: however, should we have a package-private getter? I don't know whether it is secure enough.
```
RaulinN, line 60: getter for what?

RaulinN (changes requested):
Nice work! I left you a few comments :)

One more thing that I could not highlight: it seems we have something... strange on this branch (even master?) in the `Channnel` file. The line 56 does not compile (`isBlank()`)

```
Error:(56, 16) value isBlank is not a member of String
    if(channel.isBlank() || !channel.matches(channelRegex)){
```

Could you check what happened/if it compiles on your machine?

be2-scala/src/main/scala/ch/epfl/pop/model/network/method/message/data/ActionType.scala
```diff
@@ -22,6 +22,7 @@ object ActionType extends Enumeration {
   val CAST_VOTE: Value = MatchingValue("cast_vote")
   // social media actions:
   val ADD: Value = MatchingValue("add")
+  val ADD_BROADCAST: Value = MatchingValue("addBroadcast")
```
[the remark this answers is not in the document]
tuomas-p: Yes, I'll do that. I will change that to snake_case, it was merged when the first Social Media part at the same time as my other PR.

tuomas-p:
Concerning the isBlank() issue, I don't really know what's wrong. It compiles locally (and should have compiled if it passed the CI) and I haven't changed anything inside the apply function.

be2-scala/src/main/scala/ch/epfl/pop/model/objects/LaoData.scala
```diff
@@ -9,9 +9,15 @@ import ch.epfl.pop.model.network.method.message.data.lao.CreateLao
...
 import spray.json._
 
+//Ed25519 is used to verify the signatures, therefore I also use it to sign data
+import com.google.crypto.tink.subtle.Ed25519Sign
+
+//a pair of byte arrays works better as keypair than the actual keypair object, since Ed25519Sign.Keypair isn't directly convertible to json with spray
 case class LaoData(
     owner: PublicKey,
     attendees: List[PublicKey],
+    privateKey: Array[Byte],
```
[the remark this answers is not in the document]
pierluca: Actually, if there are security implications, I would even put them in the system-level documentation.

be2-scala/src/main/scala/ch/epfl/pop/model/network/method/message/data/ActionType.scala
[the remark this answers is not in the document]
tuomas-p: The issue has been raised and discussed with the Social Media team, the change will be made as soon as the protocol is updated.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/LaoHandler.scala
```diff
@@ -34,7 +34,7 @@ case object LaoHandler extends MessageHandler {
     case graphMessage@_ => graphMessage
   }
 
-  private final val SOCIALMEDIAPOSTSPREFIX: String = Channel.SEPARATOR + "posts"
+  final val SOCIALMEDIAPOSTSPREFIX: String = Channel.SEPARATOR + "posts"
```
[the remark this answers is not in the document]
tuomas-p: I'll do that, however wouldn't storing it in Channel.scala be also possible?

be2-scala/src/main/scala/ch/epfl/pop/model/objects/LaoData.scala
```diff
@@ -55,6 +64,9 @@ object LaoData extends Parsable {
...
 
-  //to simplify the use of updateWith during a CreateLao process
-  def emptyLaoData: LaoData = LaoData(null, List.empty, List.empty)
-}
\ No newline at end of file
+  //to simplify the use of updateWith during a CreateLao process, the keypair is generated here in the same way as it would be elsewhere
+  def emptyLaoData: LaoData = {
+    val keyPair: Ed25519Sign.KeyPair = Ed25519Sign.KeyPair.newKeyPair
+    LaoData(null, List.empty, keyPair.getPrivateKey, keyPair.getPublicKey, List.empty)
+  }
```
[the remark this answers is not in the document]
tuomas-p: When we create an empty LAO data, the owner is null, however, in the updateWith function, in the case CreateLao, the owner is modified.

be2-scala/src/test/scala/ch/epfl/pop/model/network/JsonRpcResponseSuite.scala
tuomas-p: of course

be2-scala/src/test/scala/ch/epfl/pop/model/network/JsonRpcRequestSuite.scala
```diff
@@ -0,0 +1,183 @@
...
+        val rpcReq3: JsonRpcRequest = JsonRpcRequest(rpc, methodType, paramsWithMessageAndDecoded, id)
+
+        rpcReq3.getDecodedDataHeader should equal (Some((ObjectType.LAO, ActionType.CREATE)))
+    }
+
+    test("setDecodedData sets data as intended"){
+        val params: Params = new Params(channelEx)
+        val messageToModify: Message = Message(messageEx.data, messageEx.sender, messageEx.signature, messageEx.message_id, messageEx.witness_signatures)
+        val paramsWithMessage: ParamsWithMessage = new ParamsWithMessage(channelEx, messageToModify)
+        val paramsWithMessageAndDecoded: ParamsWithMessage = new ParamsWithMessage(channelEx, messageLao)
```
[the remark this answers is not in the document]
tuomas-p: removed the unused val

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/SocialMediaValidator.scala
tuomas-p: I'll add an "useless" validator for now

be2-scala/src/main/scala/ch/epfl/pop/model/network/method/message/data/socialMedia/AddBroadcastChirp.scala
tuomas-p: Actually no, I'll change that

tuomas-p: Actually yes, to make it easier, as for now there is no converter from Channel to Json

RaulinN (changes requested):
Checked all previous comments, looks like everything got resolved :) Very nice!

⚠️  One more thing, I "unresolved" a comment that Pierluca commented. Might be useful to simply add a comment in the doc :)

RaulinN (approved):

PR #634, opened 2021-12-02
Work fe1 jwen store chirps

jiabaow:


fe1-web/store/reducers/SocialReducer.ts
```diff
@@ -64,6 +126,8 @@ export const {
   addChirp,
 } = socialSlice.actions;
 
+export const socialReducer = socialSlice.reducer;
```
Xelowak, line 129: Doesn't it create issues ? Because socialReducer is already defined in RootReducer.
```suggestion
export const socialReduce = socialSlice.reducer;
```

fe1-web/store/reducers/__tests__/SocialReducer.test.ts
```diff
@@ -0,0 +1,146 @@
+import 'jest-extended';
+import { AnyAction } from 'redux';
+import keyPair from 'test_data/keypair.json';
+import { Hash, PublicKey, Timestamp } from 'model/objects';
+import { Chirp } from 'model/objects/Chirp';
+import { socialReducer, addChirp } from '../SocialReducer';
+
+export const mockPublicKey = new PublicKey(keyPair.publicKey);
```
Xelowak, line 8: Do you use it anywhere else ? Otherwise you can remove the export.

fe1-web/store/reducers/__tests__/SocialReducer.test.ts
```diff
@@ -0,0 +1,146 @@
...
+const org = mockPublicKey;
+const TIMESTAMP = 1609455600;
```
Xelowak, line 10: You use this constant only once, I'd suggest removing it.

Xelowak (approved):
Nice job for the database and tests. 
I made comments about unclear things for me, and for small typos or details.

fe1-web/store/reducers/.eslintrc
```diff
@@ -2,6 +2,7 @@
   "root": false,
   "rules": {
     /* All reducers use the createSlice function, which requires the user to param-reassign */
-    "no-param-reassign": "off"
+    "no-param-reassign": "off",
+    "import/no-named-as-default": 0
```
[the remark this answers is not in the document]
jiabaow: Otherwise, the CI fails because of `Using exported name 'socialReducer' as identifier for default export  import/no-named-as-default`, I found this solution here https://stackoverflow.com/questions/44437203/how-do-i-resolve-eslint-import-no-named-as-default

fe1-web/model/objects/Chirp.ts
```diff
@@ -7,12 +7,12 @@ import { PublicKey } from './PublicKey';
...
-  text: string;
-  time: number;
-  likes: number;
-  parentId: string;
+  readonly id: string;
+  readonly sender: string;
+  readonly text: string;
+  readonly time: number;
+  readonly likes: number;
+  readonly parentId: string;
```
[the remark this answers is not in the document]
jiabaow: To be honest I don't know if they are necessary. There were some weird behaviours then I saw in model/objects/LaoEvent they add 'readonly' for state interface so I add them here. In the end, the problem wasn't because of this, and I forgot to remove them.

fe1-web/store/reducers/.eslintrc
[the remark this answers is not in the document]
jiabaow: Ah no, actually this is because of export socialReducer, I will remove both of them.

fe1-web/store/reducers/SocialReducer.ts
jiabaow: Indeed it creates an issue but I need it for testing, this is also the reason I added "import/no-named-as-default": 0 in eslintrc. I don't understand why the wallet reducer is doing the same thing but works without any problem.

fe1-web/store/reducers/SocialReducer.ts
```diff
@@ -49,12 +94,29 @@ const socialSlice = createSlice({
...
+
+        if (!store.byId[chirp.id]) {
+          store.byId[chirp.id] = chirp;
+
+          // eslint-disable-next-line max-len
+          const insertIdxInAll = findInsertIdx(store.allIdsInOrder, store.byId, chirp.time);
+          store.allIdsInOrder.splice(insertIdxInAll, 0, chirp.id);
+
+          if (!state.byLaoId[laoId].byUser[chirp.sender]) {
+            store.byUser[chirp.sender] = [chirp.id];
```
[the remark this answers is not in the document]
jiabaow: Push doesn't work because here we don't have the sender in byUser yet, I create a new entry for the sender and add the chirp.id to it.

fe1-web/store/reducers/SocialReducer.ts
```diff
@@ -24,13 +28,54 @@ interface SocialLaoReducerState {
...
+      if (byId[array[mid]].time > element) {
+        left = mid;
+      } else {
+        right = mid;
+      }
+    }
+  }
+  return index;
+}
+
```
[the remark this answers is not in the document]
jiabaow: I didn't find a built-in function, but I found a cleaner way to implement it.

fe1-web/store/reducers/.eslintrc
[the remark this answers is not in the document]
jiabaow: Things fix with the socialReduce, I will remove it

fe1-web/store/reducers/SocialReducer.ts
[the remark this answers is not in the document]
Xelowak: Okay my bad

fe1-web/store/reducers/SocialReducer.ts
```diff
@@ -80,7 +124,11 @@ export const makeChirpsList = () => createSelector(
       return [];
     }
     if (chirpList.byLaoId[laoId]) {
-      return chirpList.byLaoId[laoId].allChirps;
+      const allChirps: ChirpState[] = [];
+      chirpList.byLaoId[laoId].allIdsInOrder.forEach(
+        (id) => allChirps.push(chirpList.byLaoId[laoId].byId[id]),
```
pierluca, line 129: Do you want the chirp state or the chirp itself out of the selector?

fe1-web/store/reducers/SocialReducer.ts
```diff
@@ -24,13 +28,33 @@ interface SocialLaoReducerState {
...
+      byUser: {},
     },
   },
 };
 
 const socialReducerPath = 'social';
 
+// helper function to find where to insert the new chirp in ascending order
+function findInsertIdx(
+  array: string[], byId: Record<string, ChirpState>, element: number,
```
pierluca, line 42: Good job on the efficient implementation.
I might have been lazier :laughing:

fe1-web/store/reducers/__tests__/SocialReducer.test.ts
```diff
@@ -0,0 +1,146 @@
...
+
+test('should add the newer chirp after the second chirp', () => {
+  expect(socialReduce(filledState2, addChirp(mockLaoId, chirp3)))
+    .toEqual(filledState3);
+});
+
+test('should add the newer chirp on top', () => {
+  expect(socialReduce(filledState3, addChirp(mockLaoId, chirp4)))
+    .toEqual(filledState4);
+});
```
pierluca, line 146: Would it make sense to test the selector as well?

pierluca (approved):
Nice job !
I have a few questions but it looks mergeable to me :)

fe1-web/store/reducers/SocialReducer.ts
jiabaow: We let the selector output chirp state, and later in SocialHome, we convert from state to the chirp object for rendering. I think we can also do the conversion inside the selector, but will it make a big difference?

fe1-web/store/reducers/__tests__/SocialReducer.test.ts
jiabaow: That's a good idea, I will add it.

fe1-web/store/reducers/SocialReducer.ts
jiabaow: 🤩 I would love to be lazier, can I have a hint of that, please?

fe1-web/store/reducers/__tests__/SocialReducer.test.ts
jiabaow: done

fe1-web/store/reducers/SocialReducer.ts
pierluca: Lazier as in: inefficient. I might have done a linear search and optimized later "if needed" 😄 
But really, your implementation is much better 👍

pierluca (approved):
You can merge as-is.

General remark: a good practice is to make unit tests as focused and independent as possible.
I.e., if a unit test fails, there's only one possible reason for it. This makes it a lot easier to identify immediately a problem and fix it.
In this case, your unit tests could fail because of either the selector or the reducer. This requires a tiny bit more analysis, but I don't think it's an issue in this case.

jiabaow:
> You can merge as-is.
> 
> General remark: a good practice is to make unit tests as focused and independent as possible. I.e., if a unit test fails, there's only one possible reason for it. This makes it a lot easier to identify immediately a problem and fix it. In this case, your unit tests could fail because of either the selector or the reducer. This requires a tiny bit more analysis, but I don't think it's an issue in this case.

OK, I'll separate them in the test then merge it.

pierluca (approved):
So shiny ! 😄

PR #636, opened 2021-12-02
Add Hilt dependency injection

GabrielFleischer:
Add [Hilt ](https://developer.android.com/training/dependency-injection/hilt-android) to the project.

I am sorry for the length of the PR and I strongly encourage you to review the code commit by commit.

Two things that might be interesting to notice :
1. I had to create the `FragmentScenario` which work basically the same way as `androidx.fragment.app.testing.FragmentScenario` but allows custom activities to be provided (Which is very useful as the activity used by the androidx scenario is not Hilt ready...)
2. The reason why the `EmptyHiltActivity` is in the `debug` folder and not in the `androidtest` folder, is to be recognized by the compiler as an activity without polluting the production application.


fe2-android/app/src/test/java/com/github/dedis/popstellar/utility/handler/ConsensusHandlerTest.java
```diff
@@ -74,7 +74,7 @@
   private static final ConsensusElect elect =
       new ConsensusElect(CREATION_TIME, KEY_ID, TYPE, PROPERTY, VALUE);
 
-  private static final Gson GSON = Injection.provideGson();
+  private static final Gson GSON = JsonModule.provideGson();
```
arnauds5, line 77: Is there a reason that the Gson is not injected but directly use the provideGson ?

fe2-android/app/src/androidTest/java/com/github/dedis/popstellar/testutils/BundleBuilder.java
```diff
@@ -0,0 +1,31 @@
+package com.github.dedis.popstellar.testutils;
+
+import android.os.Bundle;
+
+import java.io.Serializable;
+
+/** Helper class to build {@link Bundle} */
+public class BundleBuilder {
```
arnauds5, line 8: Is it normal that this builder in testutils is not used yet ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/LaoDetailActivity.java
```diff
@@ -35,19 +32,20 @@
...
+@AndroidEntryPoint
 public class LaoDetailActivity extends AppCompatActivity {
 
   private static final String TAG = LaoDetailActivity.class.getSimpleName();
   private LaoDetailViewModel mViewModel;
 
   public static LaoDetailViewModel obtainViewModel(FragmentActivity activity) {
-    ViewModelFactory factory = Injection.provideViewModelFactory(activity.getApplication());
-    return new ViewModelProvider(activity, factory).get(LaoDetailViewModel.class);
+    return new ViewModelProvider(activity).get(LaoDetailViewModel.class);
```
arnauds5, line 48: Is it possible (and wouldn't it be better) to directly inject viewmodel and remove the function obtainViewModel ?

fe2-android/app/src/test/java/com/github/dedis/popstellar/utility/handler/ConsensusHandlerTest.java
GabrielFleischer: Yes, as it is a Unit test, Hilt injection does not work for the tests. (It can probably be setup, but I have no idea how)

So I just called the provider directly.

fe2-android/app/src/androidTest/java/com/github/dedis/popstellar/testutils/BundleBuilder.java
GabrielFleischer: When I wrote the test framework, I needed it at some point. But I then changed the way the test was made and didn't need it anymore.

I left it because I am confident it will be useful in the future.

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/LaoDetailActivity.java
GabrielFleischer: No, this is an android limitation. The ViewModel must be created using a provider.
And Hilt knows that. That's why the ViewModels are injected when provided by a provider.

fe2-android/docs/README.md
```diff
@@ -165,13 +165,61 @@ This is the current organization, but as this project evolves you should feel fr
...
+  public abstract Database bindDatabase(DatabaseImpl impl);
+
+  @Provides
+  public static DatabaseImpl provideDatabaseImpl(NetworkService networkService) {
+    return new DatabaseImpl(networkService);
+  }
+}
+```
+
+If a module needs Database, Hilt will find a the Binds rule that specifies it should look for a DatabaseImpl. Then it finds the Provides rule and can provide the actual instance of the Database object. Here, the `networkService` argument is provided by Hilt.
```
celinecamacho, line 201: Typo, "will find a the Binds rule"

fe2-android/docs/README.md
```diff
@@ -165,13 +165,61 @@ This is the current organization, but as this project evolves you should feel fr
 
 For more information on linking up together the user interface with the application state, please make sure you have a solid understanding of the [MVVM pattern](#application-design) and look at the `ViewModel` classes and the `LAORepository`.
 
+### Dependency Injection
 
-## Testing
+The project uses dependency injection to provide the different dependencies instances. If you don't know what it is, android provides a [simple guide](https://developer.android.com/training/dependency-injection)
```
celinecamacho, line 170: There is a missing point after the link. Same thing for line 181.

fe2-android/app/src/androidTest/java/com/github/dedis/popstellar/pages/detail/LaoDetailPageObject.java
```diff
@@ -0,0 +1,25 @@
+package com.github.dedis.popstellar.pages.detail;
+
+import androidx.annotation.IdRes;
+
+import com.github.dedis.popstellar.R;
+
+public class LaoDetailPageObject {
```
celinecamacho, line 7: Why is this class neded?

fe2-android/app/src/main/java/com/github/dedis/popstellar/di/LAOModule.java
```diff
@@ -0,0 +1,49 @@
...
+
+import dagger.Binds;
+import dagger.Module;
+import dagger.Provides;
+import dagger.hilt.InstallIn;
+import dagger.hilt.components.SingletonComponent;
+
+@Module
+@InstallIn(SingletonComponent.class)
+public abstract class LAOModule {
```
celinecamacho, line 24: I think a better name would be `RepositoryModule` or `DataSourceModule`

fe2-android/app/src/test/java/com/github/dedis/popstellar/utility/handler/ConsensusHandlerTest.java
celinecamacho: There is no need to use Hilt injection in unit tests.

celinecamacho (commented):
Thank you for this! It's good to see the Injection classes are not needed anymore. I have a few comments and suggestions.

fe2-android/app/src/androidTest/java/com/github/dedis/popstellar/ui/detail/event/consensus/ElectionStartFragmentTest.java
```diff
@@ -67,36 +65,42 @@
...
-                LAORepository.getInstance(
+                new LAORepository(
                     remoteDataSource,
                     localDataSource,
-                    Injection.provideAndroidKeysetManager(
-                        ApplicationProvider.getApplicationContext()),
-                    Injection.provideGson(),
+                    keysetManager,
+                    gson,
                     new ProdSchedulerProvider());
```
arnauds5, line 104: The test fails because the LAORepository of the ElectionStartFragment is no longer replaced by this one before the test.
Is it possible to uninstall/reinject this laoRepository inside the ElectionStartFragment before the test run ?

fe2-android/app/src/androidTest/java/com/github/dedis/popstellar/pages/detail/LaoDetailPageObject.java
GabrielFleischer: I tried to write the ElectionStartFragmentTest to work on the LaoDetailActivity, so started writing the LaoDetailActivity page object.

But it was a lot of work and I decided to forget it. I kept the class as it will be useful in the future.

fe2-android/app/src/main/java/com/github/dedis/popstellar/di/LAOModule.java
GabrielFleischer: True, I will change that

fe2-android/docs/README.md
GabrielFleischer: Thanks, I will change that

fe2-android/docs/README.md
GabrielFleischer: Thanks, I will change that

fe2-android/app/src/main/java/com/github/dedis/popstellar/di/NetworkModule.java
```diff
@@ -0,0 +1,64 @@
...
+    return new LAORequestFactory(DEFAULT_SERVER_URL);
+  }
+
+  @Provides
+  @Singleton
+  public static Scarlet provideScarlet(
+      Application application, OkHttpClient okHttpClient, Gson gson) {
+    return new Scarlet.Builder()
+        .webSocketFactory(
+            OkHttpClientUtils.newWebSocketFactory(okHttpClient, provideRequestFactory()))
```
arnauds5, line 51: The settings for the url seems to not work anymore.
Maybe it's because the LAORequestFactory is not used as a singleton if the provideScarlet call provideRequestFactory() directly ?

arnauds5 (changes requested):

fe2-android/app/src/main/java/com/github/dedis/popstellar/di/NetworkModule.java
GabrielFleischer: I missed that, I will fix this soon.

fe2-android/app/src/androidTest/java/com/github/dedis/popstellar/ui/detail/event/consensus/ElectionStartFragmentTest.java
GabrielFleischer: It was working before I made some fixes...
I will have to inject LAORepository manually for the test.

fe2-android/app/src/main/java/com/github/dedis/popstellar/di/NetworkModule.java
GabrielFleischer: Fixed

fe2-android/app/src/androidTest/java/com/github/dedis/popstellar/ui/detail/event/consensus/ElectionStartFragmentTest.java
GabrielFleischer: Fixed

fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/LAORepository.java
```diff
@@ -104,7 +105,8 @@
   // Disposable of with the lifetime of an LAORepository instance
   private final Disposable disposables;
 
-  private LAORepository(
+  @Inject
+  public LAORepository(
```
arnauds5, line 109: If we create a lao, then try to click on it, it will crash the app.
It's maybe because multiple LAORepository are used instead of a singleton.

arnauds5 (changes requested):

fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/LAORepository.java
GabrielFleischer: That's right, I forgot to add the @Singleton at the top of the class.

This should be fixed now

arnauds5 (approved):

fe2-android/docs/README.md
```diff
@@ -165,13 +165,61 @@ This is the current organization, but as this project evolves you should feel fr
...
+  public abstract Database bindDatabase(DatabaseImpl impl);
+
+  @Provides
+  public static DatabaseImpl provideDatabaseImpl(NetworkService networkService) {
+    return new DatabaseImpl(networkService);
+  }
+}
+```
+
+If a module needs Database, Hilt will find the Binds rule that specifies it should look for a DatabaseImpl. Then it finds the Provides rule and can provide the actual instance of the Database object. Here, the `networkService` argument is provided by Hilt.
```
pierluca, line 201: Where does this `NetworkService` come from ? What does it refer to? Is it a Hilt resource or something you assume gets injected ? This needs to be clarified, ideally ahead of the example, otherwise the example will be even more confusing

fe2-android/docs/README.md
```diff
@@ -36,7 +37,9 @@ app/src
 │    │                           
 │    ├── res                    # resources (layouts, strings, dimensions, etc.)
 │    │    
-│    └── java/P                 
+│    └── java/P       
+│         ├── di                # module containing the hilt modules used for dependecy injection
```
pierluca, line 41: I'm wondering, does it make sense to have all the injected dependencies in one package, or should they be stored in semantic packages ? (e.g., a `utility/camera` package for instance)
What's your opinion on this?

pierluca (commented):
Wow, this is very nice work and it makes the code so much cleaner.
Good job @GabrielFleischer !

You'll find a few comments below, essentially I think the documentation still needs a bit of improvement, both in the README and in the classes.

In particular, dead code should be removed unless you're going to use it (soon), and in any case you should document very clearly that it's dead code introduced at a certain time with a specific purpose.

fe2-android/app/src/androidTest/java/com/github/dedis/popstellar/HiltTestRunner.java
```diff
@@ -0,0 +1,24 @@
+package com.github.dedis.popstellar;
```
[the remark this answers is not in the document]
GabrielFleischer: This is some dark magic yes. 
I followed this official tutorial : https://developer.android.com/training/dependency-injection/hilt-testing.

This is probably the java equivalent of the Kotlin reference implementation.

I will add the link and explain a bit more its purpose.

fe2-android/docs/README.md
GabrielFleischer: It only stores the modules that contain the explicit injection rules. (Opposed to the implicit defined directly on the dependencies with @Inject on the constructor)

Currently, it only has 5 classes.

I might need to be more precise on the description, but I wanted to stay concise with the line length.

fe2-android/docs/README.md
GabrielFleischer: I made some changes regarding this, I hope it is more clear.

fe2-android/docs/README.md
```diff
@@ -165,13 +165,61 @@ This is the current organization, but as this project evolves you should feel fr
...
 For more information on linking up together the user interface with the application state, please make sure you have a solid understanding of the [MVVM pattern](#application-design) and look at the `ViewModel` classes and the `LAORepository`.
 
+### Dependency Injection
 
-## Testing
+The project uses dependency injection to provide the different dependencies instances. If you don't know what it is, android provides a [simple guide](https://developer.android.com/training/dependency-injection).
+
+Hilt is used to do the injection. It is an annotation based frameworks. It is especially useful when writing tests because it allows the user to choose specifically what is injected and thus what module is mocked and how it is mocked.
+
+For detailed information, take a look at those :
```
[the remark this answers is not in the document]
GabrielFleischer: Done

fe2-android/docs/README.md
```diff
@@ -165,13 +165,61 @@ This is the current organization, but as this project evolves you should feel fr
 
 For more information on linking up together the user interface with the application state, please make sure you have a solid understanding of the [MVVM pattern](#application-design) and look at the `ViewModel` classes and the `LAORepository`.
 
+### Dependency Injection
 
-## Testing
+The project uses dependency injection to provide the different dependencies instances. If you don't know what it is, android provides a [simple guide](https://developer.android.com/training/dependency-injection).
+
+Hilt is used to do the injection. It is an annotation based frameworks. It is especially useful when writing tests because it allows the user to choose specifically what is injected and thus what module is mocked and how it is mocked.
```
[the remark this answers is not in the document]
GabrielFleischer: Done

fe2-android/docs/README.md
```diff
@@ -165,13 +165,61 @@ This is the current organization, but as this project evolves you should feel fr
 
 For more information on linking up together the user interface with the application state, please make sure you have a solid understanding of the [MVVM pattern](#application-design) and look at the `ViewModel` classes and the `LAORepository`.
 
+### Dependency Injection
 
-## Testing
+The project uses dependency injection to provide the different dependencies instances. If you don't know what it is, android provides a [simple guide](https://developer.android.com/training/dependency-injection).
```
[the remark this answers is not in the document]
GabrielFleischer: Done

fe2-android/docs/README.md
pierluca: just to clarify, you're saying it only contains the mappings, and not the injected classes themselves?

celinecamacho (dismissed):
LGTM

fe2-android/docs/README.md
GabrielFleischer: Yes, that's basically it.

The classes holding those mappings are called Module in the Hilt terminology.

fe2-android/app/src/androidTest/java/com/github/dedis/popstellar/testutils/BundleBuilder.java
jbsv: Useful but unused (TM)
=> see above: prove it useful

jbsv (changes requested):

jbsv (approved):

PR #638, opened 2021-12-04
Work rjaccard protocol json reaction

rjaccard:
I removed "add topic" for now. I will add it back when we reach part 3.


jiabaow (approved):
LGTM

Lugiasc:
why don't we need the chirp id for the delete protocol ?

rjaccard:
> why don't we need the chirp id for the delete protocol ?

We have the message_id of the add reaction message and this message contains the chirp ID. It is enough I believe.

    "reaction_id": {
      "description": "message_id of the add reaction message",
      "type": "string",
      "contentEncoding": "base64"
    }

Lugiasc:
Then it means the reaction_id from add isn't the same as delete ? Where the first one is the unicode and the second one the message_id generated by its corresponding add reaction message ?

rjaccard:
> Then it means the reaction_id from add isn't the same as delete ? Where the first one is the unicode and the second one the message_id generated by its corresponding add reaction message ?

yes that's right. I imagine that we could update the names as it is a bit confusing like this.

jiabaow (approved):

lhmerino (approved):
Looks good to me but indeed based on the comments, maybe its best rename the reaction_id under delete as to not confuse this reaction_id with the one in add reaction or vice versa.

protocol/query/method/message/data/dataAddReaction.json
```diff
@@ -0,0 +1,31 @@
...
+      "const": "reaction"
+    },
+    "action": {
+      "const": "add"
+    },
+    "reaction_id": {
+      "description": "Code indicating a reaction",
+      "type": "string",
+      "$comment": "unicode of the reaction (ex: \"U+1F44D\" for like)"
+    },
```
pierluca, line 17: As per other comments, based on the description and Unicode terminology, I'd suggest renaming the field `reaction_codepoint`.

You might also want to clarify how you want the unicode codepoint to be expressed.
JSON strings support Unicode, you can natively express Unicode code points with `\uXXXX` escaping or the representation of the character itself (e.g., `😄`).
I'd advise against using a `U+XXXX` representation in a format (JSON) which already has a standard way of representing unicode.
It's best if the JSON-Schema $comment explicitly states which of the two standard JSON representations of Unicode is preferred.

fe2-android/app/build.gradle
```diff
@@ -186,6 +186,11 @@ copy {
     include 'examples/**'
     include 'query/**'
     include 'jsonRPC.json'
+
+    // Replace the remote part of the base URL ($id) to the local resource path
+    filter {
+        String line -> line.replaceAll("https://raw\\.githubusercontent\\.com/dedis/[a-zA-Z0-9_-]+/master/", "resource:/")
+    }
```
pierluca, line 193: out of curiosity, how come you need this?

JSON Schema defines the `$id` field as having to be a canonical URI, I believe this replacement violates the standard.

I'm very open to a pragmatic violation though if (1) there's a good enough reason, (2) it's well documented.

pierluca (changes requested):
Sorry for being very annoying here with my change request, but I think there are a couple of things worth discussing further :-)

fe2-android/app/build.gradle
arnauds5: The problem was that when the json validation in the fe2 is done, the schemas was correctly loaded from the local file but then the sub-schemas (from the $ref) were loaded from the online github master. (issue#630)
If a schemas was modified in a branch, it wouldn't work because it's the one from github master that would be loaded.
The library used (com.networknt:json-schema-validator) provide some ways to only use local file but it needs a mapping for each URI.
I found that it was easier to just modify the $id in the local copy, but maybe there are better solution.
The local files with theses modifications are only used to validate json.

rjaccard: I was failing the CI for FE2 and I noticed that Arnaud had the same failing test 2 weeks ago so I asked him if he found a solution. He answered that I could cherry-pick the following commit: https://github.com/dedis/student_21_pop/pull/574/commits/4e86313492eb920a485ed520c467036dbc6ef418. The problem is well explained by Arnaud in this issue: https://github.com/dedis/student_21_pop/issues/630. 

In the go-backend, we have the same behaviour so I am very curious about a solution. However, I think it would be better to move this thread to PR #574.

rjaccard: Oups, didn't see your comment before writing mine @arnauds5. Thanks for the explanation!

protocol/query/method/message/data/dataAddReaction.json
rjaccard: Thank you for the clarification. Is it better now?

fe2-android/app/build.gradle
pierluca: Understood ! Thanks for the explanation.

protocol/query/method/message/data/dataAddReaction.json
pierluca: It's clearer. I'm not sure it's simpler for you to implement though 😅 
Do all the front-end / back-end teams agree that this is easier for them (collectively) to implement ? 
I'd have guessed that the character itself (e.g., 😄) would be a lot simpler, especially as escaping is not necessary in JSON and JSON serializers might not do that by default.

In any case, I'll approve the PR and let you discuss between yourselves as needed.

pierluca (dismissed):

protocol/query/method/message/data/dataAddReaction.json
rjaccard: I agree, as soon as I submitted the change I wanted to change it back for the character itself. I'll discuss this with the social media team. Thank you for the review.

haoqianzhang (approved):

PR #643, opened 2021-12-09
Work fe1 jwen remove chirp merge with be1

jiabaow:
Screenshot 2021-12-13 at 08 30 35
The delete button will only show for the sender of this chirp.


pierluca (changes requested):
I'd love to move forward with this, but:
1. There are merge conflicts that need to be resolved
2. Awaiting @Xelowak review

PR #644, opened 2021-12-09
Organizer's pop token creation

Xelowak:
The organizer now adds his own pop token when requesting to close a roll call.
I also updated tests accordingly.


Xelowak:
> LGTM You can also add the organizer's PopToken in the beginning when he opens the roll call for the first time. The problem with adding it before the close roll call is later when we implement the re-open roll call, we need to be careful not to add it twice.

I thought about doing it at the beginning, but it seemed more logical for me to do it at the end. Actually, I haven't about trying to close, re-open and close again to see if the public key appears 2 times. I'll have to figure it out, thanks for pointing this out !

pierluca (approved):
LGTM ! Nice job, clean and simple

Issue #357, opened 2021-06-01
Attendees can create an Lao and an election


flmichel:
It is now possible to create a Lao and an election with any public key so any attendee can create lao/elections. This shouldn't be the case. Only the organizer client should such events for now.

rjaccard:
I have started to work on this issue.

rjaccard:
I have submitted a PR for this issue (https://github.com/dedis/student_21_pop/pull/650). I will close it when it is merged.

PR #650, opened 2021-12-10
fix for issue #357

rjaccard:
This solves issue #357

### DIFF
--- a/be1-go/channel/lao/mod.go
+++ b/be1-go/channel/lao/mod.go
@@ -283,6 +283,11 @@ func (c *Channel) Publish(publish method.Publish, socket socket.Socket) error {
 
 // processLaoObject processes a LAO object.
 func (c *Channel) processLaoObject(action string, msg message.Message) error {
+	err := verifySenderIsOrganizer(msg, c.organizerPubKey)
+	if err != nil {
+		return err
+	}
+
 	switch action {
 	case messagedata.LAOActionUpdate:
 	case messagedata.LAOActionState:
@@ -454,22 +459,9 @@ func (c *Channel) processMessageObject(action string, msg message.Message) error
 
 // processRollCallObject handles a roll call object.
 func (c *Channel) processRollCallObject(action string, msg message.Message, socket socket.Socket) error {
-	sender := msg.Sender
-
-	senderBuf, err := base64.URLEncoding.DecodeString(sender)
+	err := verifySenderIsOrganizer(msg, c.organizerPubKey)
 	if err != nil {
-		return xerrors.Errorf(keyDecodeError, err)
-	}
-
-	// Check if the sender of the roll call message is the organizer
-	senderPoint := crypto.Suite.Point()
-	err = senderPoint.UnmarshalBinary(senderBuf)
-	if err != nil {
-		return answer.NewErrorf(-4, keyUnmarshalError, err)
-	}
-
-	if !c.organizerPubKey.Equal(senderPoint) {
-		return answer.NewErrorf(-5, "sender's public key %q does not match the organizer's", msg.Sender)
+		return err
 	}
 
 	switch action {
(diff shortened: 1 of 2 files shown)


Anghul (changes requested):
I don't think this fix work when there is multiple servers in the system, as only one server will know the organizer key

rjaccard:
> I don't think this fix work when there is multiple servers in the system, as only one server will know the organizer key

I'm not very familiar with the multiple servers scheme. Only the witnesses and organizers can have a server. So if I add a check for the type of the hub it would be enough, wouldn't be?

be1-go/channel/lao/mod.go
```diff
@@ -283,6 +280,11 @@ func (c *Channel) Publish(publish method.Publish, socket socket.Socket) error {
 
 // processLaoObject processes a LAO object.
 func (c *Channel) processLaoObject(action string, msg message.Message) error {
+	err := verifySenderIsOrganizer(msg, c.organizerPubKey)
+	if err != nil {
+		return err
```
nkcr, line 285: We always wrap errors ; )

be1-go/channel/lao/mod.go
```diff
@@ -454,22 +456,9 @@ func (c *Channel) processMessageObject(action string, msg message.Message) error
...
-	// Check if the sender of the roll call message is the organizer
-	senderPoint := crypto.Suite.Point()
-	err = senderPoint.UnmarshalBinary(senderBuf)
-	if err != nil {
-		return answer.NewErrorf(-4, keyUnmarshalError, err)
-	}
-
-	if !c.organizerPubKey.Equal(senderPoint) {
-		return answer.NewErrorf(-5, "sender's public key %q does not match the organizer's", msg.Sender)
+		return err
```
nkcr, line 461: error wrap

be1-go/channel/lao/mod.go
```diff
@@ -530,27 +519,15 @@ func (c *Channel) createChirpingChannel(publicKey string, socket socket.Socket)
...
-
-	if action != expectedAction {
-		return answer.NewErrorf(-4, "invalid action: %s != %s)", action, expectedAction)
-	}
-
-	senderBuf, err := base64.URLEncoding.DecodeString(msg.Sender)
+	err := verifySenderIsOrganizer(msg, c.organizerPubKey)
 	if err != nil {
-		return xerrors.Errorf(keyDecodeError, err)
+		return err
```
nkcr, line 524: error wrap

be1-go/channel/lao/mod.go
```diff
@@ -711,6 +688,25 @@ func (c *Channel) processRollCallClose(msg messagedata.RollCallClose, socket soc
 	return nil
 }
 
+func verifySenderIsOrganizer(msg message.Message, org kyber.Point) error {
```
nkcr, line 691: this function could arguably be part of the Channel. Then you already have access to the organizerPukey and must only pass the message.

be1-go/channel/lao/mod.go
```diff
@@ -711,6 +688,25 @@ func (c *Channel) processRollCallClose(msg messagedata.RollCallClose, socket soc
...
+	senderPoint := crypto.Suite.Point()
+	err = senderPoint.UnmarshalBinary(senderBuf)
+	if err != nil {
+		return xerrors.Errorf("failed to unmarshal public key of the sender: %v", msg.Sender)
+	}
+
+	if !senderPoint.Equal(org) {
+		return answer.NewErrorf(-5, "sender's public key %q does not match the organizer's", msg.Sender)
+	}
+	return nil
```
nkcr, line 707: I would add a newline above

be1-go/hub/standard_hub/mod.go
```diff
@@ -478,13 +478,18 @@ func (h *Hub) createLao(publish method.Publish, laoCreate messagedata.LaoCreate,
...
+	// Check if the sender of create lao message is the organizer
 	senderPubKey := crypto.Suite.Point()
 	err = senderPubKey.UnmarshalBinary(senderBuf)
 	if err != nil {
 		return answer.NewErrorf(-4, "failed to unmarshal public key of the sender: %v", err)
 	}
 
+	if !senderPubKey.Equal(h.pubKeyOrg) {
+		return answer.NewErrorf(-5, "sender's public key %q does not match the organizer's",
+			publish.Params.Message.Sender)
```
nkcr, line 490:
```suggestion
		return answer.NewErrorf(-5, "sender's public key does not match the organizer's: %q != %q",
			senderPubKey, h.pubKeyOrg)
```

Anghul:
> > I don't think this fix work when there is multiple servers in the system, as only one server will know the organizer key
> 
> I'm not very familiar with the multiple servers scheme. Only the witnesses and organizers can have a server. So if I add a check for the type of the hub it would be enough, wouldn't be?

The biggest problem is that for createLAO, you are using the key known by the hub, but the hub know the key of its owner and not of the organizer. Therefore the witness won't be able to create the copy of the lao needed to keep the necessary state to communicate with the clients connecting to the lao using the witness (I don't know if I'm clear).
Adding a check to let the witness ignore this check isn't the best I think, because it would allow anyone to create a lao on the witnesses.
A possibility would be to make sure that the socket sending the message is an organizer socket, and trust that every message sent by the organizer is correct, but there is still a problem as for now the message are sent to all servers, then treated, so the organizer won't do the check before sending the message further.
I don't have more ideas of how that could be done.

PR #652, opened 2021-12-13
Go add registry

Anghul:


be1-go/channel/register/mod.go
```diff
@@ -0,0 +1,43 @@
+package register
```
nkcr, line 1: `registry` would be better, what do you think ?

be1-go/channel/register/mod.go
```diff
@@ -0,0 +1,43 @@
...
+//   }
+//
+//   // register the "elect" type
+//   registry.register("elect", execElect, messagedata.ConsensusElect{})
+//
+//   // when we need to process a message we call "processMsg"
+//   err = registry.processMsg(msg)
+//
+func (m MessageRegistry) Register(action string, f func(message.Message, interface{}) error, c interface{}) {
+	m.Registry[action] = CallbackData{
```
nkcr, line 39: In fact we could simplify it and take the "action" from `c` directly. You do that by having a map of `reflect.Type`:

```go
map[reflect.Type]
```

Then you can use:

```go
key := reflect.TypeOf(c)
map[key] = CallbackData{...}
```

nkcr: And add a function:

```go
func (m MessageRegistry) Process(c interface{}) error {
  key := reflect.TypeOf(c)

  data, found := m.Registry[key]
  if !found {...}

  concreteType := reflect.New(reflect.ValueOf(data.ConcreteType).Type()).Interface()

	err := msg.UnmarshalData(&concreteType)
	if err != nil {
		return xerrors.Errorf("failed to unmarshal data: %v", err)
	}

	err = data.Callback(c, concreteType)
	if err != nil {
		return xerrors.Errorf("failed to call message '%s': %v", key, err)
	}

  return nil
}

Comment (nkcr) [reply be1-go/channel/register/mod.go:39]: Then the attribute of `MessageRegistry` (`Registry`) can be private.

File: be1-go/channel/consensus/mod.go [hunk 3]
```diff
@@ -103,6 +107,25 @@ func NewChannel(channelID string, hub channel.HubFunctionalities, log zerolog.Lo
 		consensusInstances: make(map[string]*ConsensusInstance),
 		messageStates:      make(map[string]*MessageState),
 	}
+
+	newChannel.register = newChannel.NewConsensusRegistry()
```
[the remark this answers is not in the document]
Anghul: We need the channel for the registry, as all of the functions processing the messages are functions of the channel, and it therefore need to be created before the registry.

nkcr (dismissed):
LGTM 🚀

nkcr (approved):

PR #654, opened 2021-12-13
Follow attendees of a Roll Call

Xelowak:
- Resolved an issue about the organizer's token not being generated at the correct time. It used an updated ID of the roll call, which is not the one we store in Redux.
- Resolved the issue about the last tokenized roll call id of the lao that was not set at the end of a roll call.
- Changed the documentation and the description of `getUserSocialChannel` in `Channel.ts`, so that it accepts a Public Key instead of a string.
- When closing a roll call, the user will be able to find all attendees in the search page of the social media:
Capture d’écran 2021-12-13 à 16 14 21
When clicking on follow, a subscribe request is sent to the backend for the corresponding user's channel. The profile button doesn't do anything for now, but becomes usable only after following (since we need to subscribe to someone to see his chirps).

Work that still needs to be done later: 

- The user sees himself in the list of attendees, and has to follow himself to see his chirps. This will be solved with the change of message signature's PR.
- Follows are not counted or stored anywhere.


fe1-web/parts/lao/socialMedia/SocialSearch.tsx
```diff
@@ -1,11 +1,67 @@
...
+
+const SocialSearch = () => {
+  const laoSelect = makeCurrentLao();
+  const currentLao = useSelector(laoSelect);
+
+  if (!currentLao) {
+    throw new Error('Impossible to open Social media Search if you are not connected to a LAO');
+  }
+
+  const rollCallId = currentLao.last_tokenized_roll_call_id;
```
jiabaow, line 43: Do we need to change to rollCallID?

fe1-web/store/reducers/EventsReducer.ts
```diff
@@ -218,9 +218,9 @@ export const makeEventsMap = (laoId: string | undefined = undefined) => createSe
   },
 );
 
-export const makeEventGetter = (laoId: Hash | string, eventId: Hash | string) => {
+export const makeEventGetter = (laoId: Hash | string, eventId: Hash | string | undefined) => {
   const id = laoId.valueOf();
-  const evtId = eventId.valueOf();
+  const evtId = eventId?.valueOf();
```
jiabaow, line 223: Why do we change the eventId to be optional?

jiabaow (dismissed):
LGTM!
You can also add test for EventsReducer but I understand it's not the most important thing.

fe1-web/store/reducers/EventsReducer.ts
Xelowak: Because this function will be called using the last_tokenized_roll_call_id of an lao, which can be undefined. This way, the hook can be called without any conditions directly in the React component, so that the Rules of Hooks are not broken.

fe1-web/parts/lao/socialMedia/SocialSearch.tsx
Xelowak: I am not changing the roll call id, just retrieving it to be able to make a list of attendees from it (using `makeLastRollCallAttendeesList`).

fe1-web/model/objects/RollCall.ts
```diff
@@ -135,7 +135,7 @@ export class RollCall implements LaoEvent {
     }
 
     return this.attendees.some(
-      (attendee: PublicKey) => attendee.equals(token.publicKey),
+      (attendee: PublicKey) => attendee.valueOf() === token.publicKey.valueOf(),
```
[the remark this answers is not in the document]
Xelowak: I'm not 100% sure but this condition was broken on my side at some point. Using this comparison, I'm sure it works fine.

fe1-web/parts/lao/socialMedia/SocialSearch.tsx
jiabaow: I see, I saw you changed all rollCallIds to rollCallIDs in the previous file. But yes, since it's a const here it makes more sense to name it rollCallId.

fe1-web/store/reducers/EventsReducer.ts
jiabaow: I see, thanks for the explanation.

fe1-web/parts/lao/socialMedia/SocialSearch.tsx
Xelowak: Oh ok I didn't understand that, sorry. Yeah actually that was because of the name being used in the route. `rollCallId `wasn't recognized in tests because it was not the name of the id in the route (which was `rollCallID`)

fe1-web/model/objects/RollCall.ts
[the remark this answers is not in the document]
Xelowak: Actually we overwrite it in `Base64Url.ts`, I'll revert this.

fe1-web/parts/lao/organizer/RollCallOpened.tsx
```diff
@@ -34,6 +37,12 @@ const RollCallOpened = () => {
   const navigation = useNavigation();
   const [attendees, updateAttendees] = useState(new Set<string>());
```
pierluca, line 38: it seems like you could now directly store PublicKey's here.

fe1-web/parts/lao/organizer/RollCallOpened.tsx
```diff
@@ -57,14 +66,18 @@ const RollCallOpened = () => {
       EventTags.ROLL_CALL, OpenedLaoStore.get().id.toString(),
       rollCallID, time,
     );
-    requestCloseRollCall(updateId, Array.from(attendees).map((key: string) => new PublicKey(key)))
-      .then(() => {
+
+    // Add the token of the organizer before closing the roll call
+    Wallet.generateToken(lao.id, new Hash(rollCallID)).then((token) => {
+      const attendeesList = Array.from(attendees).map((key: string) => new PublicKey(key));
+      attendeesList.push(token.publicKey);
```
pierluca, line 73: ... which means you wouldn't need to convert them here.
You'd also be able to add the token from the very creation of the `attendees` state (possibly in a `useEffect` hook)

This way: 
- the default view would show that you have `1` attendee
- the number at the bottom of this screen would match the number of public keys listed once the roll call is closed
- you're not going to have `string` items representing something else :) 

Is it feasible or am I missing an obvious problem?

fe1-web/parts/lao/organizer/RollCallOpened.tsx
```diff
@@ -57,14 +66,18 @@ const RollCallOpened = () => {
...
+      requestCloseRollCall(updateId, attendeesList).then(() => {
```
pierluca, line 74:
```suggestion
      return requestCloseRollCall(updateId, attendeesList).then(() => {
```

if you chain `Promise`s (without the `async`/`await` syntax), you need to `return` the inner `Promise` for this to work as you expect it. Otherwise the outer promise will consider that the asynchronous work is finished before the inner promise has finished executing, i.e., they wouldn't be chained.

fe1-web/store/reducers/EventsReducer.ts
```diff
@@ -268,6 +268,32 @@ export const makeEventByTypeSelector = <T extends LaoEvent>(eventType: string) =
...
+      }
+      if (!evtId || !eventMap.byLaoId[id].byId[evtId]) {
+        return [];
+      }
+
+      const rollCall = eventFromState(eventMap.byLaoId[id].byId[evtId]) as RollCall;
+      if (!rollCall) {
+        return [];
+      }
+      return rollCall.attendees ? rollCall.attendees : [];
```
pierluca, line 292:
```suggestion
      return rollCall.attendees || [];
```

simpler :)

fe1-web/components/UserListItem.tsx
```diff
@@ -0,0 +1,106 @@
...
+import WideButtonView from './WideButtonView';
+
+/**
+ * Component that shows a user's profile picture, his public key and two buttons:
+ * - Follow, which subscribes to the user's social channel.
+ * - Profile, that navigates to the user's profile page.
+ *
+ * @remarks
+ * For now, it is not possible to unfollow users, and you have to follow someone to access to
+ * his profile.
```
pierluca, line 20:
```suggestion
 * For now, it is not possible to unfollow users, and you have to follow someone to access
 * their profile.
```
Inclusive 😜 

Also, in general, I would always recommend indicating a date (and possibly a username) whenever there's a comment with a notion of time.
This makes it easy to revisit / challenge the assumption if a lot of time has passed.
E.g., if you see code that says "We probably need to fix this soon", you won't react and interpret it in the same way if the full comment says "... fix this soon (2016-01-01, xyz)" or "... fix this soon (2021-11-19, jiabaow)".

fe1-web/components/UserListItem.tsx
```diff
@@ -0,0 +1,106 @@
...
+});
+
+const UserListItem = (props: IPropTypes) => {
+  const [isFollowing, setIsFollowing] = useState(false);
+  const { laoId, publicKey } = props;
+
+  const followUser = () => {
+    subscribeToChannel(getUserSocialChannel(laoId, publicKey))
+      .catch((error) => {
+        console.error(`Could not subscribe to channel of user ${publicKey.valueOf()}, error: ${error}`);
```
pierluca, line 56: why use (only) the console.error now that you have a nice toast library? :)

fe1-web/components/UserListItem.tsx
```diff
@@ -0,0 +1,106 @@
...
+    console.error('Profile navigation is not implemented yet.');
+  };
+
+  return (
+    <View style={styles.container}>
+      <View style={styles.leftView}>
+        <Ionicons name="person" size={40} color="black" />
+      </View>
+      <View style={styles.rightView}>
+        <Text style={styles.publicKeyText}>{publicKey.valueOf()}</Text>
```
pierluca, line 72: Out of scope for this PR:
At some point, we might want to have a graphical representation of public keys.
If he hasn't done so already, @nkcr might be able to point you to examples.

fe1-web/parts/lao/socialMedia/SocialSearch.tsx
```diff
@@ -1,11 +1,67 @@
...
+    borderColor: gray,
+  } as ViewStyle,
+});
+
+const SocialSearch = () => {
+  const laoSelect = makeCurrentLao();
+  const currentLao = useSelector(laoSelect);
+
+  if (!currentLao) {
+    throw new Error('Impossible to open Social media Search if you are not connected to a LAO');
```
pierluca, line 40: Out of scope for this PR:
Would it make sense to have a more "robust" behavior to this kind of errors ?

pierluca (commented):
Very nice job overalls ! 
I've added a bunch of comments concerning (mostly) minor things, feel free to disregard some if you think they'll hold back the progress too much.

fe1-web/components/UserListItem.tsx
Xelowak: Yes, we already have a library ready for that. We just need to implement it.

fe1-web/parts/lao/socialMedia/SocialSearch.tsx
Xelowak: I think so, since I need to put it pretty much everywhere since `makeCurrentLao()` is used a lot. What would you suggest ? It can be done in a separated PR.

fe1-web/components/UserListItem.tsx
Xelowak: Okay, I'll do it ! It will make things clearer

fe1-web/parts/lao/organizer/RollCallOpened.tsx
Xelowak: I think it is. 
I've tried it, but the problem is that the organizer's public key keeps being added to the list (because the code inside the component is executed again and again). I don't know how to solve this, I've tried many ways but neither of them worked. Since it might take some time to fix, I suggest we keep it as it is because it works fine. We can tackle this later or in another PR if that's ok for you.
We're now storing Public Keys instead of strings though.

fe1-web/parts/lao/organizer/RollCallOpened.tsx
Xelowak: True, I changed it.

fe1-web/components/UserListItem.tsx
Xelowak: That's true. We might need to replace them everywhere. For now, I prefer keeping it as it is for consistency.

fe1-web/parts/lao/organizer/RollCallOpened.tsx
Xelowak: Done

fe1-web/parts/lao/organizer/RollCallOpened.tsx
Xelowak: Actually it's not. Having PublicKeys instead of strings generate a problem because it seems that you can add them multiple times (neither of them are equal). So we'll keep strings.
But I was able to add the organizer's token directly when we open a roll call as you described.

pierluca (approved):
LGTM, but it seems the CI is broken. I've restarted the jobs.

DO NOT merge if the CI is broken, unless you have another PR ready to merge that fixes this.

pierluca (approved):

PR #657, opened 2021-12-14
Add OpenElection transition

arnauds5:
For the moment, this draft is for testing if adding an OpenElection transition could be usefull or not.
If accepted, it will require to be implemented in both backend/frontend.

- Add OpenElection to the protocol
- Add the OpenElection data
- Add ElectionHandler for OpenElection
- Modify the setupElectionElement in the event list
- Send an OpenElection message after a consensus to start is reached


pierluca:
@arnauds5 I only became aware of this now, I don't remember seeing this PR mentioned anywhere.
I had a quick look. This seems like jolly good work, do you think you could bring it forward ?

arnauds5:
> @arnauds5 I only became aware of this now, I don't remember seeing this PR mentioned anywhere. I had a quick look. This seems like jolly good work, do you think you could bring it forward ?

I wrote a proposition to add a new state transition for Election here : Issue #657 
I wrote this draft for testing if this would be possible to do. I tried to do it with Android+Go at first but I didn't continue because of others tasks and because I waited the merge of another PR.
And now, it's maybe a bit too late before the semester's end because it will need to be implemented in all backend/frontend. Otherwise half of the 4 systems would break.
I have the go backend and some android code in local (not pushed). So Android + Go could be finished maybe.

Tyratox:
Closed this in favor of #850, #886 and other PRs that add this feature

PR #658, opened 2021-12-15
add Cache in DbActor

tuomas-p:


be2-scala/src/main/scala/ch/epfl/pop/model/objects/Channel.scala
```diff
@@ -51,6 +51,7 @@ object Channel {
   final val rootChannel: Channel = Channel(s"${SEPARATOR}root")
   final val rootChannelPrefix: String = s"${SEPARATOR}root${SEPARATOR}"
   private final def channelRegex: String = "^/root(/[^/]+)*$"
+  final val LAODATALOCATION: String = Channel.SEPARATOR + "data"
```
BadrTad, line 54: Any particular reason why this path is here, it's kind of odd to (to me at least) to see this in a Channel class.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
```diff
@@ -253,9 +254,13 @@ object DbActor extends AskPatternConstants {
...
             if(bt != null){
               val laoJson = new String(bt, StandardCharsets.UTF_8)
-              batch.put(laoDataKey.getBytes, LaoData.buildFromJson(laoJson).updateWith(message).toJsonString.getBytes)
+              val laoDataNew: LaoData = LaoData.buildFromJson(laoJson).updateWith(message)
+              batch.put(laoDataKey.getBytes, laoDataNew.toJsonString.getBytes)
+              laoDataCache.put(laoDataKey, laoDataNew)
             } else if (bt == null) {
-              batch.put(laoDataKey.getBytes, LaoData.emptyLaoData.updateWith(message).toJsonString.getBytes)
+              val laoDataNew: LaoData = LaoData.emptyLaoData.updateWith(message)
+              batch.put(laoDataKey.getBytes, laoDataNew.toJsonString.getBytes)
```
BadrTad, line 262: Could `laoDataKey.getBytes` be null and hence has to be treated as an edge case. ?

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
```diff
@@ -329,12 +334,20 @@ object DbActor extends AskPatternConstants {
...
-              val json = new String(bytes, StandardCharsets.UTF_8)
-              DbActorReadLaoDataAck(Some(LaoData.buildFromJson(json)))
-            case _ =>
-              DbActorNAck(ErrorCodes.SERVER_ERROR.id, s"Lao data not found for channel $channel")
+          if(laoDataCache.keySet.contains(dataKey)){
+            //with our implementation, if a key is in the map, there is always a LaoData value attached to it
+            DbActorReadLaoDataAck(Some(laoDataCache.get(dataKey).get))
+          } else{
+            Try(db.get(dataKey.getBytes)) match {
+              case Success(bytes) if bytes != null =>
```
BadrTad, line 342: Great check !

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
BadrTad: And it would be great to have some modularization for future caching of other objects , don't you think so ?

BadrTad (dismissed):
Great work !

be2-scala/src/main/scala/ch/epfl/pop/model/objects/Channel.scala
tuomas-p: Yes, we discussed this with @RaulinN in my other PR, the Channel path values should be in channel instead of being defined in "random" places in the code, it's better to group everything.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
tuomas-p: It could be useful, however, for now, the object we need the most (by far) is LaoData, whether it's for "access rights" (is a person allowed to send a message) or for server-side broadcasts. The other structures will be used less, however, caching them in the future could be an option.

tuomas-p: In the current implementation, I don't see it being possible, however for the code to be more robust for the future, I'll add extra checks

be2-scala/src/main/scala/ch/epfl/pop/model/objects/Channel.scala
RaulinN, line 54: AAAAAAAAAAAH ^^'. Let's keep our styling conventions :) I'd suggest going for `LAO_DATA_LOCATION` (all caps since constant). Moreover, you could also apply the same treatmet to rootChannel and rootChannelPrefix. At the beginning, I don't believe they were supposed to be used elsewhere (other than this file), thus probably why they look like simple private variables

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
```diff
@@ -192,8 +192,9 @@ object DbActor extends AskPatternConstants {
   // FIXME: find way to keep track of db size
   sealed case class DbActor(mediatorRef: ActorRef, db: DB, DATABASE_FOLDER: String) extends Actor with ActorLogging {
```
RaulinN, line 194: This file is absolutely correct. However, what do you think about creating a class `Cache`? This way, you could hide the dirty things from DbActor (e.g. line 337 or 262). Moreover, this could be useful if we want to add a cache for something other than LaoData in the future

Open for discussion/suggestion! What do you think @tuomas-p ?

RaulinN (commented):
Just a few liiiiittle changes and you are good to go :)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
tuomas-p: This could be a good idea indeed, however, it should be defined as a typed class then (like Map), so I don't know how useful it would be in practice, as all the operations are very basic Map operations, but it could be cleaner

be2-scala/src/main/scala/ch/epfl/pop/model/objects/Channel.scala
tuomas-p: Can I do a separate PR directly after this one to refactor these, as the rootChannel etc. are also used in my other PR and it would be confusing to change things in both at the same time?

tuomas-p: (for rootChannel and rootChannelPrefix, I'll change LAO_DATA_LOCATION here of course)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
tuomas-p: To me, for now, the best option would maybe be to add a helper method in DbActor (as it's the only class directly in contact with the database) which verifies the validity of the data in the map instead of having to do all the checks manually (taking a map and a key as arguments in order to be valid for other "caches" too)

RaulinN (approved):

PR #659, opened 2021-12-15
Make ChirpCard tests use fake time

Xelowak:
Since ChirpCard tests depended on the actual time, I modified them by setting the system time directly.

### DIFF
--- a/fe1-web/components/__tests__/ChirpCard.test.tsx
+++ b/fe1-web/components/__tests__/ChirpCard.test.tsx
@@ -8,8 +8,13 @@ const chirp = new Chirp({
   id: new Hash('1234'),
   text: 'Don\'t panic.',
   sender: new PublicKey('Douglas Adams'),
-  time: new Timestamp(1609455600),
+  time: new Timestamp(1609455600), // 31 December 2020
   likes: 42,
+});
+
+beforeAll(() => {
+  jest.useFakeTimers('modern');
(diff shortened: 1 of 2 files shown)


jiabaow (approved):
LGTM!

fe1-web/components/__tests__/ChirpCard.test.tsx
```diff
@@ -8,10 +8,15 @@ const chirp = new Chirp({
...
   text: 'Don\'t panic.',
   sender: new PublicKey('Douglas Adams'),
-  time: new Timestamp(1609455600),
+  time: new Timestamp(1609455600), // 31 December 2020
   likes: 42,
 });
 
+beforeAll(() => {
+  jest.useFakeTimers('modern');
+  jest.setSystemTime(new Date(1620255600000)); // 5 May 2021
```
pierluca, line 17: would it make sense to update the `Timestamp` mock as well, so that they provide the same absolute time? Or even replace that mock with this fake timer everywhere? 😅

pierluca (approved):
LGTM and thank you ! I learned something :-)
I have a tiny suggestion to improve the rest of the codebase (now or later).

fe1-web/components/__tests__/ChirpCard.test.tsx
Xelowak: I think this is cleaner, in the way that the system time is set directly. I think I'll replace that mock with the timer.

PR #662, opened 2021-12-16
Add front-end testing to the Karate project

GabrielFleischer:
Add testing for the two front-end subprojects.

I greatly encourage you to apply the process described in the README to make sure nothing is missing.

This PR also add common utility classes with the MessageBuffer that allows multiple messages to be received and to take only the ones we want to check with filters.


tests/karate/src/test/java/fe/create_lao.feature
```diff
@@ -0,0 +1,25 @@
...
+    * replace page_object.env = karate.env
+    * call read(page_object)
+
+  Scenario: Create a LAO send the right messages to the backend
+    When click(tab_launch_selector)
+    And input(tab_launch_lao_name_selector, 'Lao Name')
+    And click(tab_launch_create_lao_selector)
+
+    # Retrieving sent messages
+    * json create_lao = backend.takeTimeout(1000)
```
BadrTad, line 15: For the timeouts i 'll suggest configuring those in the Background 
`configure timeout 1000`

tests/karate/src/test/java/fe/net/mockbackend.feature
```diff
@@ -0,0 +1,20 @@
+@ignore @report=false
+Feature: Mock Backend
+
+  Scenario: Setup Mock-backend
+    * def getBackend =
+            """
+              function() {
+                var Backend = Java.type("fe.net.MockBackend")
+                return new Backend(9005)
```
BadrTad, line 9: Same remark as above the port parameter can be placed in karte-config.js and later 
configured  with `* configure port 9005`

tests/karate/src/test/java/fe/net/mockbackend.feature
```diff
@@ -0,0 +1,20 @@
...
+              }
+            """
+
+    * def backend = call getBackend
+    * karate.log('Backend started at ', backend.getPort())
+
+    * def stopBackend = function() { backend.stop() }
+
+    # Shutdown backend automatically after the end of a scenario and  feature
+    * configure afterScenario = stopBackend
```
BadrTad, line 19: Those should typically be in the Background 😄

tests/karate/src/test/java/fe/net/mockbackend.feature
```diff
@@ -0,0 +1,20 @@
...
+    * configure afterFeature = stopBackend
```
BadrTad, line 20: the configuration of after scenario/feature should be done in the Background don't you think so ?

tests/karate/src/test/java/logback-test.xml
```diff
@@ -14,7 +14,7 @@
     </encoder>
   </appender>
 
-  <logger name="com.intuit" level="DEBUG"/>
+  <logger name="com.intuit" level="debug"/>
```
BadrTad, line 17: Just wondering how long it took to figure that this was an issue 😂

tests/karate/src/test/java/fe/create_lao.feature
```diff
@@ -0,0 +1,25 @@
+@env=android,web
```
BadrTad, line 1: Test looks clean

tests/karate/src/test/java/karate-config.js
```diff
@@ -14,18 +14,32 @@ function fn() {
     timeout: 5000, //Timeout for websocket response
     args: [],
   }
-  if (env == 'go') {
+  if (env === 'go') {
```
BadrTad, line 17: Good to know 👍

tests/karate/src/test/java/fe/utils/web.feature
```diff
@@ -0,0 +1,43 @@
...
+    * def tab_launch_lao_name_selector = "input[placeholder='Organization name']"
+    * def tab_launch_address_selector = "input[placeholder='Address']"
+    * def tab_launch_create_lao_selector = '{}Launch -- Connect, Create LAO & Open UI'
+
+  Scenario: Setup connection to the backend and complete on the home page
+    Given driver driverOptions
+
+    # The default input function is not consistent and does not work every time.
+    # This replaces the input function with one that just tries again until it works.
+    * def input =
```
BadrTad, line 31: What if it does not work do we get stuck infinitely in the loop ?

BadrTad (commented):
This is a huge work , well done. Testing other features now should be straightforward.

tests/karate/src/test/java/fe/utils/web.feature
GabrielFleischer: Hmmm... it goes into an infinit loop...

I'll add a maximum number of tries :D

tests/karate/src/test/java/logback-test.xml
GabrielFleischer: I was just annoyed by the number of information in the console, so I changed that part. And afterward when resetting it to debug, I wrote it in lowercase, that wasn't an issue

tests/karate/src/test/java/fe/utils/web.feature
GabrielFleischer: Done

tests/karate/src/test/java/common/net/MessageFilters.java
```diff
@@ -0,0 +1,29 @@
...
+
+
+  /**
+   * Only accept the messages with the given method
+   *
+   * @param method to accept
+   * @return true if the message is accepted
+   */
+  public static Predicate<String> withMethod(String method) {
+    return msg -> msg.contains("\"method\":\"" + method + "\"");
```
pierluca, line 27: careful. This can fail if the JSON is formatted with a space between the `method` field and its value (it'd still be valid JSON and respecting to have `"method":  "aaa"`)
As such, consider parsing it as JSON.

tests/karate/src/test/java/common/net/MessageFilters.java
```diff
@@ -0,0 +1,29 @@
...
+ *  - {@link MessageBuffer#takeAll(Predicate)}
+ *  - {@link MessageBuffer#takeTimeout(Predicate, long)}
+ */
+public final class MessageFilters {
+
+  private MessageFilters() {
+    // Static lass, this should never be called
+  }
+
+
```
pierluca, line 19: extra blank lines? why? :'(

tests/karate/src/test/java/common/net/MultiMsgWebSocketClient.java
```diff
@@ -0,0 +1,106 @@
...
+import com.intuit.karate.http.WebSocketClient;
+import com.intuit.karate.http.WebSocketOptions;
+
+import java.util.List;
+import java.util.function.Predicate;
+
+/**
+ * A WebSocketClient that can handle multiple received messages
+ */
+public class MultiMsgWebSocketClient extends WebSocketClient implements MessageBuffer {
```
pierluca, line 13: semantically, does it make sense to say that the client "is" a MessageBuffer as opposed to "has" a MessageBuffer? 
Do we really need the delegate functions as opposed to having a getter for a MessageBuffer ?

tests/karate/src/test/java/fe/net/ReplyMethods.java
```diff
@@ -0,0 +1,17 @@
...
+import java.util.function.Function;
+
+public class ReplyMethods {
+
+  private static final String VALID_REPLY_TEMPLATE = "{\"jsonrpc\":\"2.0\",\"id\":%ID%,\"result\":0}";
+  private static final String VALID_CATCHUP_REPLY_TEMPLATE = "{\"jsonrpc\":\"2.0\",\"id\":%ID%,\"result\":[]}";
+
+  public static Function<String, String> ALWAYS_VALID = msg -> {
+    int id = Json.of(msg).get("id");
+    String template = msg.contains("\"method\":\"catchup\"") ? VALID_CATCHUP_REPLY_TEMPLATE : VALID_REPLY_TEMPLATE;
```
pierluca, line 14: again, don't rely on a fixed formatting of the JSON. It'll make the tests very brittle.

tests/karate/src/test/java/common/net/MessageQueue.java
```diff
@@ -0,0 +1,124 @@
...
+    String date = dataSupplier.get();
+
+    try {
+      while (date == null && System.currentTimeMillis() - start < timeout) {
+        wait(timeout);
+        date = dataSupplier.get();
+      }
+    } catch (InterruptedException e) {
+      return null;
+    }
```
pierluca, line 120: I'm not sure I fully understand this implementation. it seems that we're always waiting all the way until timeout unless the data was retrieved immediately, is that the desired behavior?

tests/karate/src/test/java/common/net/MessageQueue.java
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  @Override
+  public synchronized void clear() {
+    queue.clear();
+  }
+
+  private synchronized String retrieveWithTimeout(Supplier<String> dataSupplier, long timeout) {
+    long start = System.currentTimeMillis();
+    String date = dataSupplier.get();
```
pierluca, line 111: I'm sure we could find a better variable name ;)

pierluca (commented):
I haven't ye thad time to go through the README, but here are a few remarks concerning the infrastructure itself.

tests/karate/src/test/java/common/net/MessageFilters.java
GabrielFleischer: I wanted to avoid this for some reason, but I will have to do that

tests/karate/src/test/java/common/net/MessageFilters.java
GabrielFleischer: :'(

tests/karate/src/test/java/common/net/MultiMsgWebSocketClient.java
GabrielFleischer: The Multi-Message Client stores incoming messages, so I don't see it being a Buffer as a problem.

Also, it is for the karate implementation.
I like the idea to have a test that looks like this :

```
Then def msg1 = client.take()
And def msg2 = client.take()
And def msg3 = client.take()
```

rather than 

```
Then def msg1 = client.getBuffer().take()
And def msg2 = client.getBuffer().take()
And def msg3 = client.getBuffer().take()
```


But, now that I think about it, we could also write 
```
Then def buffer = client.getBuffer()
And def msg1 = buffer.take()
And def msg2 = buffer.take()
And def msg3 = buffer.take()
```

tests/karate/src/test/java/common/net/MessageQueue.java
GabrielFleischer: In java, wait is different from a sleep.

A wait can be notified by notify() or notifyAll() and it will unlock.
The notifyAll is called when a new message is received and added to the buffer.

I was not sure myself if this was working and thus I wrote the Tests (MessageQueueTest).

tests/karate/src/test/java/common/net/MessageQueue.java
GabrielFleischer: Oh, a Typo.

I could rename it to msg and msgSupplier too

tests/karate/src/test/java/common/net/MessageQueue.java
pierluca: Thanks. I learned something

tests/karate/src/test/java/fe/net/MockBackend.java
```diff
@@ -0,0 +1,102 @@
...
+  private final Logger logger = new Logger(getClass().getSimpleName());
+
+  private final MessageQueue queue;
+  private final WebSocketServerBase server;
+  // Will be set to true once the connection is established
+  private final CompletableFuture<Boolean> connected = new CompletableFuture<>();
+
+  // Defines the rule to apply on incoming messages to produce its reply.
+  // Can be null if no reply should be sent back.
+  private Function<String, String> replyProducer = ReplyMethods.ALWAYS_VALID;
```
BadrTad, line 28: Don't you think this should be initialized to null, and the tester should always set a replyProducer manual for each message he/she eants to handle.

tests/karate/src/test/java/fe/net/MockBackend.java
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  @Override
+  protected void channelRead0(
+      ChannelHandlerContext channelHandlerContext, TextWebSocketFrame frame) {
+    logger.info("message received : {}", frame.text());
+    queue.onNewMsg(frame.text());
+
+    // Send back the reply
+    if (replyProducer != null) send(replyProducer.apply(frame.text()));
```
BadrTad, line 62: So following above comment here the replyProducer would be basically set back to null

tests/karate/src/test/java/fe/net/ReplyMethods.java
```diff
@@ -0,0 +1,31 @@
...
+ * This class contains useful message replies that can be used when using {@link
+ * MockBackend#setReplyProducer(Function)}
+ */
+public class ReplyMethods {
+
+  private static final String VALID_REPLY_TEMPLATE =
+      "{\"jsonrpc\":\"2.0\",\"id\":%ID%,\"result\":0}";
+  private static final String VALID_CATCHUP_REPLY_TEMPLATE =
+      "{\"jsonrpc\":\"2.0\",\"id\":%ID%,\"result\":[]}";
+
```
BadrTad, line 19: What about error templates ?

tests/karate/src/test/java/fe/create_lao.feature
```diff
@@ -0,0 +1,25 @@
...
+Feature: Create LAO
+
+  Background: Driver basic setup
+    * def page_object = 'classpath:fe/utils/<env>.feature@name=basic_setup'
```
BadrTad, line 5: A comment here maybe explaining what is going about the karate tags system

tests/karate/src/test/java/common/net/MessageQueue.java
```diff
@@ -0,0 +1,128 @@
+package common.net;
+
+import java.util.*;
+import java.util.function.Predicate;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+public class MessageQueue implements MessageBuffer {
+
+  private final List<String> queue = new LinkedList<>();
```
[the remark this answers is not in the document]
GabrielFleischer: I don't think it is possible to dequeue the first element matching a filter with a Dequeu.

And as I want to provide a way to retrieve the first message with a certain type with taking the order into account, the List is needed

tests/karate/src/test/java/fe/net/MockBackend.java
GabrielFleischer: I like thinking of the default behavior as the one that will be used in the most common case.

And for me the most common case is to always return a Result.

tests/karate/src/test/java/fe/net/MockBackend.java
GabrielFleischer: I don't agree.

If we want to do a message by message response, we can use backend.send() directly inside the test.

It will be easier to understand.

tests/karate/src/test/java/fe/net/ReplyMethods.java
GabrielFleischer: No test need it yet so I did not implement the error template yet. But when a new test needs it, we can create the template easily.

tests/karate/src/test/java/fe/create_lao.feature
GabrielFleischer: If you think so, I can add a link and a short explaination yes.

pierluca (commented):
I went through everything this time. I can't speak much for the Java code (not a language I'm an absolute expert in), and I'll let others review those aspects (@celinecamacho ?), but the overall design seems clean and I'd say very close to a mergeable state 👍 

I'm happy to approve once a few typos / documentation / comments are fixed, and consider other improvements in separate PRs.

tests/karate/src/test/java/fe/net/MockBackend.java
GabrielFleischer: Ok

tests/karate/README.md
```diff
@@ -1,10 +1,74 @@
...
+
+Start the Android Emulator and Appium.
+
+The emulator window name should match : `Android Emulator - avd:id` \
+Ex: `Android Emulator - Pixel_2_API_30:5554`
+
+Make sure the karate-config is correct, more precisely :
+
+- The `deviceName` is set to `emulator-<id>` where id is the one found on the emulator window
+- avd is set the avd name found on the emulator window
```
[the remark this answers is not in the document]
GabrielFleischer: Yes, I tried to clarify this part with an example

tests/karate/src/test/java/fe/create_lao.feature
GabrielFleischer: This should be clearer

tests/karate/README.md
```diff
@@ -1,10 +1,74 @@
...
+### Backend
+
+To test a backend, you don't need more setup that what is needed to build that backend. Use the resources provided by
+those projects.
+
+### Android Front-end
+
+To test the Android Frontend, you need to have an Android emulator installed. The easiest way to achieve it is to
+install it through [Android Studio](https://developer.android.com/studio) :
+Go to `Tools -> ADV Manager` and create an emulator.
```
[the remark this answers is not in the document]
GabrielFleischer: Done

tests/karate/src/test/java/common/net/MessageFilters.java
```diff
@@ -0,0 +1,30 @@
...
+ * <li>{@link MessageBuffer#peek(Predicate)}
+ * <li>{@link MessageBuffer#peekAll(Predicate)}
+ * <li>{@link MessageBuffer#take(Predicate)}
+ * <li>{@link MessageBuffer#takeAll(Predicate)}
+ * <li>{@link MessageBuffer#takeTimeout(Predicate, long)}
+ */
+public final class MessageFilters {
+
+  private MessageFilters() {
+    // Static lass, this should never be called
```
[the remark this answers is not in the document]
GabrielFleischer: Done

tests/karate/src/test/java/fe/net/MockBackend.java
```diff
@@ -0,0 +1,102 @@
...
+
+  @Override
+  public void channelActive(ChannelHandlerContext ctx) {
+    channel = ctx.channel();
+    connected.complete(true);
+    logger.trace("Client connected from the server side");
+  }
+
+  @Override
+  protected void channelRead0(
```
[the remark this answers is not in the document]
GabrielFleischer: I tried to explain it

pierluca (approved):
LGTM !

PR #668, opened 2021-12-18
Fix issue where the built web-page is blank

GabrielFleischer:
The issue was caused by the path of the js script that are absolute.

Use a small trick to make it relative


pierluca (approved):
I'm going to trust you blindly on this one

PR #669, opened 2021-12-19
Force the encoding to UTF-8 for local files copy

arnauds5:
Fix a bug where unicode characters were corrupted by the local files copy because of the default local encoding.


GabrielFleischer (approved):
Ok, great !

jbsv (approved):

PR #670, opened 2021-12-19
Add the architecture of the refactored key system

GabrielFleischer:
Implement and test the architecture of the new key system.

The idea is to use the created objects everywhere is the rest of the project.
As I did not implement the rest of the project yet, I do not know if every feature is present, but I am confident everything is there.

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/security/Base64URLData.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/security/Base64URLData.java
@@ -1,0 +1,48 @@
+package com.github.dedis.popstellar.model.objects.security;
+
+import java.util.Arrays;
+import java.util.Base64;
+
+/** Represents a data that can be encoded into a Base64 form */
+public class Base64URLData {
+
+  protected final byte[] data;
+
+  public Base64URLData(byte[] data) {
+    this.data = data;
+  }
+
+  public Base64URLData(String data) {
+    this(decode(data));
+  }
+
+  public byte[] getData() {
+    return Arrays.copyOf(data, data.length);
+  }
+
+  /** @return the Base64 - encoded string representation of the data */
+  public String getEncoded() {
+    return encode(data);
+  }
+
+  private static byte[] decode(String data) {
+    return Base64.getUrlDecoder().decode(data);
+  }
+
+  private static String encode(byte[] data) {
+    return Base64.getUrlEncoder().encodeToString(data);
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) return true;
+    if (!(o instanceof Base64URLData)) return false;
+    Base64URLData that = (Base64URLData) o;
+    return Arrays.equals(data, that.data);
+  }
+
+  @Override
+  public int hashCode() {
+    return Arrays.hashCode(data);
+  }
+}
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/security/Base64URLData.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/security/Base64URLData.java
@@ -1,5 +1,7 @@
 package com.github.dedis.popstellar.model.objects.security;
 
+import androidx.annotation.NonNull;
+
 import java.util.Arrays;
 import java.util.Base64;
 
@@ -31,6 +33,12 @@ public class Base64URLData {
 
   private static String encode(byte[] data) {
     return Base64.getUrlEncoder().encodeToString(data);
+  }
+
+  @NonNull
+  @Override
+  public String toString() {
+    return getClass().getSimpleName() + "(" + getEncoded() + ")";
   }
 
   @Override
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/security/Base64URLData.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/security/Base64URLData.java
@@ -36,7 +36,8 @@ public class Base64URLData {
   @Override
   public boolean equals(Object o) {
     if (this == o) return true;
-    if (!(o instanceof Base64URLData)) return false;
+    if (o == null || getClass() != o.getClass()) return false;
+
     Base64URLData that = (Base64URLData) o;
     return Arrays.equals(data, that.data);
   }
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/security/KeyPair.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/security/KeyPair.java
@@ -1,0 +1,55 @@
+package com.github.dedis.popstellar.model.objects.security;
+
+import com.google.crypto.tink.PublicKeySign;
+
+import java.security.GeneralSecurityException;
+import java.util.Objects;
+
+/**
+ * Represent a private/public key pair used for signing data
+ *
+ * <p>This object does not actually store the private key as is it safely stored by another module.
+ */
+public class KeyPair {
+
+  private final PrivateKey privateKey;
+  private final PublicKey publicKey;
+
+  public KeyPair(PrivateKey privateKey, PublicKey publicKey) {
+    this.privateKey = privateKey;
+    this.publicKey = publicKey;
+  }
+
+  /**
+   * Signs the given data with the private key of the pair {@link PublicKeySign#sign(byte[])}
+   *
+   * @param data to sign
+   * @return the signature
+   * @throws GeneralSecurityException if an error occurs
+   */
+  public Signature sign(Base64URLData data) throws GeneralSecurityException {
+    return privateKey.sign(data);
+  }
+
+  public boolean verify(Signature signature, Base64URLData data) {
+    return publicKey.verify(signature, data);
+  }
+
+  public PublicKey getPublicKey() {
+    return publicKey;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) return true;
+    if (!(o instanceof KeyPair)) return false;
+    KeyPair keyPair = (KeyPair) o;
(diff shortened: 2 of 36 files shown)


fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/security/Base64URLData.java
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  private static String encode(byte[] data) {
+    return Base64.getUrlEncoder().encodeToString(data);
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) return true;
+    if (!(o instanceof Base64URLData)) return false;
```
arnauds5, line 39: There are warnings on sonarcloud about subclasses don't override the equals.
And the equals of Base64URLData use "instanceof Base64URLData" instead of the default "getClass() != o.getClass()", is it intentional so that all subclass equality are only based on the byte[] data ?
For example this code will return that they are equals.
```
Base64URLData base64URLData = new Base64URLData(VALID_PRIVATE_KEY);
PlainPrivateKey plainPrivateKey = new PlainPrivateKey(VALID_PRIVATE_KEY);
assertEquals(base64URLData, plainPrivateKey);
```

GabrielFleischer: Yes, this is intended.

The sonarcloud issue is a false positive because the fields are not supposed to be tested for equality.

And I designed this so that it compares the data even if the types are not exactly the same.

If you think this is a bad idea, just let me know.

fe2-android/app/src/test/java/com/github/dedis/popstellar/model/objects/security/PrivateKeyTest.java
```diff
@@ -0,0 +1,48 @@
...
+  @Test
+  public void signGivesSameValueForBothKeyType() throws GeneralSecurityException {
+    PrivateKey key1 = new PlainPrivateKey(VALID_PRIVATE_KEY);
+
+    KeysetHandle keyset = mock(KeysetHandle.class);
+    when(keyset.getPrimitive(PublicKeySign.class)).thenReturn(new Ed25519Sign(VALID_PRIVATE_KEY));
+    PrivateKey key2 = new ProtectedPrivateKey(keyset);
+
+    Signature sign1 = key1.sign(DATA);
+    Signature sign2 = key2.sign(DATA);
```
pierluca, line 35: aren't you effectively comparing `Ed25519Sign` with itself ? 🤔

pierluca (commented):
Had a quick look, dropped some comments here and there.
Overalls it seems pretty clean.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/security/PublicKey.java
```diff
@@ -0,0 +1,36 @@
...
+  private final PublicKeyVerify verifier;
+
+  public PublicKey(byte[] data) {
+    super(data);
+    verifier = new Ed25519Verify(data);
+  }
+
+  public PublicKey(String data) {
+    super(data);
+    verifier = new Ed25519Verify(this.data);
```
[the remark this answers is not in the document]
GabrielFleischer: This is an optimisation that I do without thinking of it when there is an object that is created from a immutable value, I just create that object once and store it.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/security/Signature.java
```diff
@@ -0,0 +1,17 @@
...
+public class Signature extends Base64URLData {
+
+  public Signature(byte[] data) {
+    super(data);
+  }
+
+  public Signature(String data) {
+    super(data);
+  }
+}
```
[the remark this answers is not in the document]
GabrielFleischer: For me, giving the "responsibility" of the validation to the Key makes more sense.

It feels a bit weird to "ask" the signature to validate itself.

fe2-android/app/src/test/java/com/github/dedis/popstellar/model/objects/security/PoPTokenTest.java
```diff
@@ -0,0 +1,65 @@
...
+  private static final byte[] VALID_PRIVATE_KEY2 =
+      Utils.hexToBytes("cf74d353042400806ee94c3e77eef983d9a1434d21c0a7568f203f5b091dde1d");
+  private static final byte[] VALID_PUBLIC_KEY2 =
+      Utils.hexToBytes("6015ae4d770294f94e651a9fd6ba9c6a11e5c80803c63ee472ad525f4c3523a6");
+
+  @Test
+  public void signAndVerifyWorks() throws GeneralSecurityException {
+    PoPToken pair = new PoPToken(VALID_PRIVATE_KEY, VALID_PUBLIC_KEY);
+    Signature signing = pair.sign(DATA);
+    assertTrue(pair.verify(signing, DATA));
```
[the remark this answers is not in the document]
GabrielFleischer: Ok, that is a good think to add yes.

fe2-android/app/src/test/java/com/github/dedis/popstellar/model/objects/security/PrivateKeyTest.java
GabrielFleischer: What do you mean, this is a very dumb test that make sure to different PrivateKey with the same data generate the same Signature.

I'm not sure I understand what you mean.

GabrielFleischer:
> How are you going to test for (type-aware) equality in tests ?

For some reason I cannot reply in thread, but what do you mean by "type-aware" ?

I made it this way so that a signature/ private key/public key and a Base64URLData could be compared.

But it might be a bad design choice and we probably want to keep equality to the same type.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/security/KeyPair.java
```diff
@@ -39,10 +41,22 @@ public PublicKey getPublicKey() {
     return publicKey;
   }
 
+  @NonNull
+  @Override
+  public String toString() {
+    return getClass().getSimpleName()
+        + "{privateKey="
+        + privateKey.toString()
```
pierluca, line 49: I'm not sure it's a great idea to print the private key _anywhere_

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/security/PublicKey.java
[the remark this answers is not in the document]
pierluca: It's okay here, but be careful about "optimisation without thinking", as you're effectively making a "space vs. speed" optimization which in a different context could be very detrimental 😉

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/security/Signature.java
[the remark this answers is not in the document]
pierluca: Arguable, but fair enough 😄

fe2-android/app/src/test/java/com/github/dedis/popstellar/model/objects/security/PrivateKeyTest.java
pierluca: I meant that effectively you've mocked the dependencies of `ProtectedPrivateKey` to the point where it behaves exactly like `PlainPrivateKey` and you're really just testing that `Ed25519Sign` behaves in the same way in two different places.
Thinking about it, though, it's at the very least a good non-regression test 👍

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/security/Base64URLData.java
pierluca: By type-aware equality I meant that in a test you wouldn't have been able to assert that `PublicKey(x) != PlainPrivateKey(x)`

PS: you needed to go on the outdated commit to be able to respond directly 😄

pierluca (changes requested):
Looks perfect to me, *except* for the fact that you are including a private key in `toString`.
That seems like a very easy way to leak secrets. Besides, the PublicKey should be enough to unique identify the KeyPair.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/security/Base64URLData.java
GabrielFleischer: OK, I intended that at first, but it doesn't make sense now that I think about it more. I fixed it in the latest commits.

fe2-android/app/src/test/java/com/github/dedis/popstellar/model/objects/security/PrivateKeyTest.java
GabrielFleischer: In the best case scenario, I would want to give a real KeyHandler. But it is not generated in the Unit tests, so I mocked its behavior.

In my mind, this is a very simple test that makes sure the two types provide the same effective functionality.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/security/KeyPair.java
GabrielFleischer: Good point

Now, for debug purpose, I think it should be great to see whether it is a plain or a protected key.
I think I will keep this but redeclare the `toString` of `PlainPrivateKey` to hide it if it is ever printed.

GabrielFleischer: I pushed a change on PlainPrivateKey, it should fix this isssue.

jbsv (approved):
LGTM

Issue #635, opened 2021-12-02
Public key/PoP token data messages


celinecamacho:
There is no documentation specifying what messages should be sent only by the organizer or witness. For example, a close election should be sent by the organizer and a cast vote can be sent by anyone.

The following table shows all the data messages of the protocol specifications and which should be sent using the **public key** of the organizer, witness, or server. The empty slots mean that **anyone** can send the messages using their **PoP token**. 

The idea is to edit this table until we agree and then add this information to the proto-specs documentation.

| messages/data | public key 
|---|---|
| dataAddChirp | 
| dataAddChirpBroadcast | server |
| dataAddReaction | 
| dataCastVote | 
| dataCloseRollCall | organizer |
| dataCreateLao | organizer |
| dataCreateMeeting | organizer | 
| dataCreateRollCall | organizer |
| dataDeleteReaction |
| dataDeleteChirp | 
| dataDeleteChirpBroadcast | server |
| dataElect | organizer/witness
| dataElectAccept | organizer/witness
| dataPrepare | server
| dataPromise | server
| dataPropose | server
| dataAccept | server
| dataLearn | server
| dataEndElection | organizer
| dataOpenRollCall | organizer
| dataResultElection | server | 
| dataSetupElection | organizer
| dataStateLao | organizer
| dataStateMeeting | organizer
| dataUpdateLao | organizer
| dataWitnessMessage | organizer/witness

TLDR: only the add chirp, add reaction, cast vote, delete reaction and delete chirp can be sent using a PoP token. 

This is up to discussion, what do you think?

BadrTad:
👍

PR #673, opened 2021-12-20
Change message signatures

Xelowak:
I changed message signatures according to issue #635, and added tests for it.
There was an issue with pop token's private key generation that is now fixed.
I added another correct key pair to be able to test the system with 2 key pairs (one for an attendee and one for an organizer for example)


fe1-web/model/network/method/message/data/MessageData.ts
```diff
@@ -35,3 +35,18 @@ export interface MessageData {
...
+ * @remarks
+ * This function has to be updated for each new kind of message that is going to be signed using
+ * the pop token.
+ *
+ * @param data - The message data we want to know how to sign
+ */
+export function isSignedWithToken(data: MessageData): Boolean {
+  return (data.object === ObjectType.ELECTION && data.action === ActionType.CAST_VOTE)
+    || (data.object === ObjectType.CHIRP && data.action === ActionType.ADD);
+}
```
pierluca, line 52: to be discussed: would it be simpler to maintain if this were a property within the `data` itself?

fe1-web/model/network/method/message/__tests__/Message.test.ts
```diff
@@ -0,0 +1,85 @@
...
+    const messageData = new AddChirp({
+      text: 'text',
+      timestamp: new Timestamp(1607277600),
+    });
+    const encodedDataJson: Base64UrlData = encodeMessageData(messageData);
+    const signature = mockPopToken.privateKey.sign(encodedDataJson);
+    Message.fromData(messageData).then((m) => {
+      expect(m.sender).toEqual(mockPopToken.publicKey);
+      expect(m.signature).toEqual(signature);
+    });
```
pierluca, line 62: `then` might not execute within the scope of the test if you don't return the promise.
Jest requires asnyc tests to return the Promise, otherwise it's not aware of their async nature.

pierluca (commented):
I would encourage you to use the `async`/`await` pattern in TS when possible / convenient.
I suspect a lot of the code will actually look simpler, the `publish(channel, msgData)` function being a prime candidate.

fe1-web/model/network/method/message/data/MessageData.ts
Xelowak: I think it would, because having to update this function every time we add a new type of message that needs to be signed with a token doesn't seem really practical. But to be honest, I don't know how to do it.

fe1-web/model/network/method/message/__tests__/Message.test.ts
Xelowak: Okay thanks a lot, it might be the problem of the CI.

fe1-web/model/network/method/message/Message.ts
```diff
@@ -111,68 +112,57 @@ export class Message {
...
-  }
-}
 
-/*
-public static async fromData(
-    data: MessageData, witnessSignatures?: WitnessSignature[],
-  ): Promise<Message> {
-    const encodedDataJson: Base64UrlData = encodeMessageData(data);
+    const isSignedWithToken = messageObjectMap.get(data.action);
+    if (isSignedWithToken === undefined) {
```
jiabaow, line 138: maybe change to !isSignedWithToken so it's more consistent with !messageObjectMap

jiabaow (approved):
Thanks for the changes!
LGTM

fe1-web/model/network/method/message/Message.ts
Xelowak: It's a boolean | undefined. I made this mistake first, but then every message signed with the public key threw an error, because the condition was true. It's correct that way

fe1-web/model/network/method/message/Message.ts
```diff
@@ -121,42 +121,52 @@ export class Message {
...
-          signature,
-          message_id: Hash.fromStringArray(encodedDataJson.toString(), signature.toString()),
-          witness_signatures: (witnessSignatures === undefined) ? [] : witnessSignatures,
-        });
+    const messageObjectMap = isMessageSignedWithToken.get(data.object);
+    if (!messageObjectMap) {
+      throw new Error(`Message signature for object ${data.object} is unsupported.`);
+    }
+
+    const isSignedWithToken = messageObjectMap.get(data.action);
```
pierluca, line 137: Couldn't you make your life easier by doing `x.get(data.object)?.get(data.action)` and skipping one level of error ?

fe1-web/model/network/method/message/data/MessageData.ts
```diff
@@ -37,16 +37,39 @@ export interface MessageData {
...
- * the pop token.
- *
- * @param data - The message data we want to know how to sign
+ * This map has to be updated for each new kind of message.
  */
-export function isSignedWithToken(data: MessageData): Boolean {
-  return (data.object === ObjectType.ELECTION && data.action === ActionType.CAST_VOTE)
-    || (data.object === ObjectType.CHIRP && data.action === ActionType.ADD);
-}
+export const isMessageSignedWithToken: Map<string, Map<string, boolean>> = new Map([
```
pierluca, line 46: as discussed on slack, I'd suggest using an interface rather than a `boolean` as the content of this map of maps.
it'd make your life easier when it comes to adding more elements.

fe1-web/parts/lao/organizer/RollCallOpened.tsx
```diff
@@ -66,17 +74,13 @@ const RollCallOpened = () => {
...
-        // @ts-ignore
-        navigation.navigate(STRINGS.organizer_navigation_tab_home);
-      }).catch((err) => {
-        console.error('Could not close roll call, error: ', err);
-      });
+    return requestCloseRollCall(updateId, attendeesList).then(() => {
+      // @ts-ignore
+      navigation.navigate(STRINGS.organizer_navigation_tab_home);
+    }).catch((err) => {
+      console.error('Could not close roll call, error: ', err);
```
pierluca, line 83: some day this could be a toast 😇

Xelowak: It is now one !

fe1-web/model/network/method/message/Message.ts
Xelowak: True, but I'll use an interface anyway since this solution is preferable.

fe1-web/model/network/method/message/data/MessageData.ts
Xelowak: I tried to do it, but I'm not sure about what I've done.

fe1-web/model/network/method/message/data/MessageProperties.ts
```diff
@@ -0,0 +1,60 @@
...
+ * Map to know the properties of the type of message we're going to send, according to the object
+ * and the action.
+ *
+ * @remarks
+ * This map has to be updated for each new kind of message, and each new properties we would like
+ * to add.
+ */
+export const messagePropertiesMap: Map<ObjectType, Map<ActionType, MessageProperties>> = new Map([
+  [ObjectType.CHIRP, new Map<ActionType, MessageProperties>([
+    [ActionType.ADD, new PopTokenSignedMessage()],
```
pierluca, line 33: I'm not going to annoy you further, but this could be simply :

```suggestion
    [ActionType.ADD, { isPopTokenSigned: true }],
```

pierluca (approved):
I'll let you see if you want to simplify this further or merge as-is.
Approved in the meanwhile :)

fe1-web/model/network/method/message/data/MessageProperties.ts
pierluca: Effectively this means you wouldn't need classes at all, just this gigantic table which we'll be able to expand in the future :)

Xelowak: Thanks, I did it ! I think it's now ready to merge

pierluca (approved):
Approving blindly

PR #674, opened 2021-12-20
Work rjaccard be1 reaction

rjaccard:
For now, there is no check that the messages that we add a reaction to truly exist. I would like to add this check in another PR to simplify the review process. I plan to do the following. The reaction channel will have a generalChannel as attribute and can call a function from the generalChannel to see if this messageID exists. To make the check, generalChannel will have a set will all the messageID posted until now in which we remove the messageID of the deleted chirps. Feel free to give feedbacks if you think there is a better way.


Anghul (dismissed):
Good job, just a few comments about some choices in your tests

nkcr (commented):
Very nice and pleasant to read! 👍 🙂
I just have a single comment, and I agree with what @Anghul said.

nkcr (approved):
LGTM 🚀

Issue #663, opened 2021-12-16
[BUG] data race during Test_Send_And_Handle_Message


rjaccard:
### Description (Actual behavior)
A data race occurs during Test_Send_And_Handle_Message

### Expected behavior
We would expect the test to succeed without a data race

### How to reproduce
Run "make check" on the go backend

### Version & Environment
This bug was reproduced on:
- Commit: 1d764c9a5ba788ddd825f172d51ee71fa3bede10
- 
##### Front-ends:
- [ ] Fe1-Web (please include browser's names & version)
- [ ] Fe2-Android (please specify if phone or emulation, and Android version)
- [x] Not applicable

##### Back-ends:
- [x] Be1-Go
- [ ] Be2-Scala
- [ ] Not applicable

### Impact
The tests fail

### Possible root cause
Multiple goroutines acceding the same variable.

### Stacktrace

==================
WARNING: DATA RACE
Read at 0x00c000125c30 by goroutine 25:
  popstellar/hub/standard_hub.Test_Send_And_Handle_Message()
      /Users/robinjaccard/student_21_pop/be1-go/hub/standard_hub/mod_test.go:712 +0x789
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1259 +0x22f
  testing.(*T).Run·dwrap·21()
      /usr/local/go/src/testing/testing.go:1306 +0x47

Previous write at 0x00c000125c30 by goroutine 26:
  popstellar/hub/standard_hub.(*fakeSocket).Send()
      /Users/robinjaccard/student_21_pop/be1-go/hub/standard_hub/mod_test.go:829 +0x4a
  popstellar/channel.(*Sockets).SendToAll()
      /Users/robinjaccard/student_21_pop/be1-go/channel/mod.go:64 +0x169
  popstellar/hub/standard_hub.(*Hub).broadcastToServers()
      /Users/robinjaccard/student_21_pop/be1-go/hub/standard_hub/mod.go:497 +0x153
  popstellar/hub/standard_hub.(*Hub).handlePublish()
      /Users/robinjaccard/student_21_pop/be1-go/hub/standard_hub/message_handling.go:211 +0x264
  popstellar/hub/standard_hub.(*Hub).SendAndHandleMessage.func1()
      /Users/robinjaccard/student_21_pop/be1-go/hub/standard_hub/mod.go:276 +0x6b

Goroutine 25 (running) created at:
  testing.(*T).Run()
[...]

PR #675, opened 2021-12-20
Fix for issue #663

rjaccard:
fix for issue #663. I don't think the go-routine is really necessary here. I removed it and now there is no more a race condition during the tests.

### DIFF
--- a/be1-go/hub/standard_hub/mod.go
+++ b/be1-go/hub/standard_hub/mod.go
@@ -272,12 +272,10 @@ func (h *Hub) SendAndHandleMessage(publishMsg method.Publish) error {
 
 	h.serverSockets.SendToAll(byteMsg)
 
-	go func() {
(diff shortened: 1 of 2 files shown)


nkcr (commented):
I'm ok with removing the concurency, but you should check with @Anghul if this is still OK for the consensus.

rjaccard:
> I'm ok with removing the concurency, but you should check with @Anghul if this is still OK for the consensus.

Message from @Anghul :
@Robin Jaccard about your pull request for issue#663 (https://github.com/dedis/student_21_pop/pull/675/files), I don't think the goroutine should be removed.
Yes, the program still work without the goroutine, but the problem is that it makes one goroutine treat multiple messages, which doesn't really make sense.
It also introduces a problem when a message is sent in reaction to another, but throws an error, in that either the error depends on both messages or the error of the second message will be shown before the success of the first.


So I'm closing this PR and I gave it another try in PR #679

PR #676, opened 2021-12-21
Add a registry of data handlers

arnauds5:
## Content of this PR
- Add a DataRegistry that holds the registry of classes and handlers for the Data
- Add an interface for DataHandler with a ContextHandler for arguments
- Move the creation of the DataRegistry outside the Data class (to a DI module)
- Add a DataRegistry.Builder
- Remove an unused and deprecated duplicated function "createGson"
- Use real key/Id in ElectionStartFragmentTest (because of some refactoring)
- Refactor the MessageHandler/GenericMessage from static functions into a single class object (done by Gabriel)


GabrielFleischer (commented):
Damn you're fast, I expected most of what I added in the `To Do` not to be addressed before the end of the semester, it was more ideas of what we should probably be done next.

But it is starting very well !

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/DataConsumer.java
```diff
@@ -0,0 +1,20 @@
...
+import com.github.dedis.popstellar.model.network.method.message.data.Data;
+import com.github.dedis.popstellar.repository.LAORepository;
+import com.github.dedis.popstellar.utility.error.DataHandlingException;
+
+@FunctionalInterface
+public interface DataConsumer<T> {
+
+  void accept(
+      LAORepository laoRepository, String channel, T data, String messageId, String senderPk)
+      throws DataHandlingException;
```
GabrielFleischer, line 12: I feel like it would be great to have a function that takes only two arguments :
The data and a context that holds everything else.

This would make it a lot more maintainable for the future, as adding fields would not impact all handlers. And it will also make sure that a handler only use what it needs.

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/DataConsumer.java
```diff
@@ -0,0 +1,20 @@
+package com.github.dedis.popstellar.utility.handler;
+
+import com.github.dedis.popstellar.model.network.method.message.data.Data;
+import com.github.dedis.popstellar.repository.LAORepository;
+import com.github.dedis.popstellar.utility.error.DataHandlingException;
+
+@FunctionalInterface
+public interface DataConsumer<T> {
```
GabrielFleischer, line 8: Maybe DataHandler would be a more intuitive name. What do you think ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/DataConsumer.java
arnauds5: Something like that for example ?
```
public final class HandlerContext {

  private final LAORepository laoRepository;
  private final String channel;
  private final MessageGeneral message;

  public HandlerContext(LAORepository laoRepository, String channel, MessageGeneral message) {
    this.laoRepository = laoRepository;
    this.channel = channel;
    this.message = message;
  }
  public LAORepository getLaoRepository() { return laoRepository; }
  public String getChannel() { return channel; }
  public MessageGeneral getMessage() { return message; }
  public String getMessageId() { return message.getMessageId(); }
  public String getSenderPk() { return message.getSender(); }
}
```
```
public static void handleCreateLao(HandlerContext context, CreateLao createLao) {
  Log.d(TAG, "handleCreateLao: channel " + context.getChannel() + ", msg=" + createLao);
  Lao lao = context.getLaoRepository().getLaoByChannel(context.getChannel());

  lao.setName(createLao.getName());
  lao.setCreation(createLao.getCreation());
  lao.setLastModified(createLao.getCreation());
  lao.setOrganizer(createLao.getOrganizer());
  lao.setId(createLao.getId());
  lao.setWitnesses(new HashSet<>(createLao.getWitnesses()));

  String publicKey = context.getLaoRepository().getPublicKey();
  if (lao.getOrganizer().equals(publicKey) || lao.getWitnesses().contains(publicKey)) {
    context.getLaoRepository().sendSubscribe(lao.getChannel() + "/consensus");
  }
  context.getLaoRepository().updateNodes(context.getChannel());
}
```

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/DataConsumer.java
arnauds5: If we use somthing that only has 2 arguments (context + data), we can use a BiConsumer instead of a custom interface for > 2 arguments.
And change DataConsumer<? extends Data> to BiConsumer<HandlerContext, ? extends Data>

fe2-android/app/src/main/java/com/github/dedis/popstellar/di/DataRegistryModule.java
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  private static <T extends Data> void add(
+      Map<DataRegistry.EntryPair, Pair<Class<? extends Data>, DataHandler<? extends Data>>> map,
+      Objects obj,
+      Action action,
+      Class<T> dataClass,
+      DataHandler<T> dataHandler) {
+    map.put(DataRegistry.pair(obj, action), Pair.create(dataClass, dataHandler));
+  }
```
GabrielFleischer, line 140: Very good idea to register the datas in the Registry module !

But that that point, I think we should create a Builder for the registry rather than passing it a Map.
This Map signature is starting to be quite long, and it won't get better in the future.

I will do in during the day if I am tired of my wallet refactor.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/DataRegistry.java
```diff
@@ -0,0 +1,93 @@
...
+
+import java.util.Collections;
+import java.util.Map;
+import java.util.Optional;
+
+/** A registry of Data classes and handlers */
+public final class DataRegistry {
+
+  /** A mapping of (object, action) -> (class, handler) */
+  private final Map<EntryPair, Pair<Class<? extends Data>, DataHandler<? extends Data>>> mapping;
```
GabrielFleischer, line 17: I am not a Registry specialist, but usually what I do is to create an inner Entry class that contains everything :
 - Method
- Action
- DataClass
- DataHandler

And then at creation generate multiple mapping given our needs.

Here I would probably create a mapping `EntryPair -> Entry` and probably a mapping `DataClass -> Entry` will also be useful in the near future.

What do you think about it ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/DataHandler.java
```diff
@@ -0,0 +1,16 @@
...
+@FunctionalInterface
+public interface DataHandler<T> {
+
+  void accept(HandlerContext context, T data) throws DataHandlingException;
+
+  @SuppressWarnings("unchecked")
+  default void accept(HandlerContext context, Data data)
+      throws DataHandlingException, ClassCastException {
+    accept(context, (T) data);
+  }
```
GabrielFleischer, line 15: This should be removed. I would rather give the Registry the responsibility to cast a data.

Maybe add a `handle()` function in the Registry rather than a `getHandler()`.

GabrielFleischer (commented):
This makes the code so much cleaner !

As for the issue with the injection, we probably need to move everything into Hilt and inject dependencies through the constructor.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/DataRegistry.java
```diff
@@ -90,4 +89,64 @@ public int hashCode() {
       return java.util.Objects.hash(object, action);
     }
   }
+
+  public static final class Entry<T extends Data> {
```
GabrielFleischer, line 93: Should this be private ?

GabrielFleischer (commented):
Is this already complete ?

We just need a bit more documentation and inject everything through HIlt and it is ready for reviews !

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/DataRegistry.java
arnauds5: I changed the visibility to private because it was only used inside the DataRegistry

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/MessageHandler.java
```diff
@@ -37,45 +197,23 @@ private MessageHandler() {
...
-        handleChirpMessage(laoRepository, channel, data, message.getMessageId(), senderPk);
-        break;
-      case MESSAGE:
-        handleWitnessMessage(laoRepository, channel, senderPk, data);
-        break;
-      default:
-        throw new UnhandledDataTypeException(data, dataObj.getObject());
-    }
+    Action dataAction = Action.find(data.getAction());
+    if (dataAction == null) throw new UnknownDataActionException(data);
```
GabrielFleischer, line 214: Actually, this should never occur, as the JsonDeserialization will fail if the pair is not registered. 
So the packet execution will not go to this point.

GabrielFleischer (approved):
Great job !

We could just maybe remove an unnecessary check.

GabrielFleischer (approved):
LGTM !

fe2-android/app/src/main/java/com/github/dedis/popstellar/di/DataRegistryModule.java
```diff
@@ -0,0 +1,127 @@
+package com.github.dedis.popstellar.di;
```
pierluca, line 1: this whole file:😍

pierluca (approved):
As discussed, we can't do in-depth review over the holidays, but this looks (at least superficially) like REALLY. GOOD. WORK. 👍 
I'll approve and let you merge if you think it's ready, or ask another student's review.

PR #678, opened 2021-12-22
Adds a common interface for message data

nkcr:
To simplify the registry implementation.


Anghul (approved):
Thanks for the big modification. I didn't see anything strange while reviewing.

PR #679, opened 2021-12-22
fix for issue 663

rjaccard:
I have added a mutex in the fake implementation of channel and socket because the race condition was coming from there.


Anghul (approved):
Thanks for the fix

nkcr (approved):

PR #680, opened 2021-12-22
Add consensus frontend stage 3

arnauds5:
A Failure in consensus is only triggered by the backend. When this happens, a node will be able to retry again and every nodes can see that this node failed to start. 
- Add ConsensusFailure Data class 
- Add ConsensusFailure handler
- Refactor ElectionStartFragmentTest and add tests (instead of a big single test, now it's multiple small tests)
- The code for displaying a failed consensus proposal, was already present


GabrielFleischer (approved):
Nice ! Great tests !

pierluca (approved):
Approving (blindly) to unblock the PR :)

PR #683, opened 2021-12-23
Enable attendees to send chirps

Xelowak:
Made the attendees able to send chirps. The organizer can also send them using his pop token, and not his public key.
I also removed a function from `Channel.ts` that is not used anymore.


jiabaow (approved):
LGTM

fe1-web/parts/lao/socialMedia/SocialHome.tsx
```diff
@@ -45,12 +48,38 @@ const styles = StyleSheet.create({
...
+    });
+  }, [lao.last_tokenized_roll_call_id]);
+
   const publishChirp = () => {
-    requestAddChirp(inputChirp)
-      .catch((err) => {
-        console.error('Could not add chirp, error:', err);
-      });
-    setInputChirp('');
+    if (userPublicKey) {
```
pierluca, line 74: This will fail until useEffect + generateToken runs.
It'd be more robust to disable publishing until userPublicKey is set.
I also suspect that the userPublicKey should be a state for this to work correctly 🤔

pierluca (commented):
Had a super quick look. Let me know if I should approve blindly.

fe1-web/parts/lao/socialMedia/SocialHome.tsx
Xelowak: I did it !

pierluca (approved):

PR #685, opened 2021-12-23
Update mod_test.go

rjaccard:


be1-go/channel/reaction/mod_test.go
```diff
@@ -392,6 +392,10 @@ func Test_DeleteReaction(t *testing.T) {
 	pub.Params.Message = m
 	pub.Params.Channel = reactionChannelName
 
+	// Wait before storing a new message to be able to have a unique
+	// timestamp for each message
+	time.Sleep(time.Millisecond)
```
pierluca, line 397: can this problem happen in production?

pierluca (approved):
Blind approve. As discussed.
But you may want to be sure that that problem doesn't show up in production code :-)

PR #689, opened 2021-12-23
Add the new architecture to the Wallet and create a KeyManager Module that will help retrieving key pairs

GabrielFleischer:
This PR is mostly type changes.

To avoid the propagation of the type changes, I stopped it in unclean ways and added FIXME to resolve them later

The private key was shown in plain text of the RollCall Token Fragment. This should not be done, so it was removed from the activity.

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Wallet.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Wallet.java
@@ -29,6 +29,9 @@ import java.util.List;
 import java.util.Map;
 import java.util.StringJoiner;
 
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
 import io.github.novacrypto.bip39.MnemonicGenerator;
 import io.github.novacrypto.bip39.MnemonicValidator;
 import io.github.novacrypto.bip39.SeedCalculator;
@@ -39,6 +42,7 @@ import io.github.novacrypto.bip39.wordlists.English;
  * This class represent a wallet that will enable users to store their PoP tokens with reasonable,
  * realistic security and usability.
  */
+@Singleton
 public class Wallet {
 
   private static final String TAG = Wallet.class.getSimpleName();
@@ -48,13 +52,8 @@ public class Wallet {
   private Aead aead;
   private boolean isSetup = false;
 
-  private static final Wallet instance = new Wallet();
-
-  public static Wallet getInstance() {
-    return instance;
-  }
-
   /** Class constructor, initialize the wallet with a new random seed. */
+  @Inject
   public Wallet() {
     setRandomSeed();
   }
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Wallet.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Wallet.java
@@ -91,7 +91,7 @@ public class Wallet {
   }
 
   /**
-   * Method that allow generate a different key for each path that you give.
+   * Generate a PoPToken (i.e. a key pair) from a given path.
    *
    * @param path a String path of the form: m/i/j/k/... where i,j,k,.. are 31-bit integer.
    * @return the generated PoP Token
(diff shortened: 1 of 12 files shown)


fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Wallet.java
```diff
@@ -93,11 +94,10 @@ public void initKeysManager(Context applicationContext)
    * Method that allow generate a different key for each path that you give.
```
jbsv, line 94:
```suggestion
   * Generate a PoPToken (i.e. a key pair) from a given path.
```

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/LaoDetailViewModel.java
```diff
@@ -229,9 +232,10 @@ private boolean attendedOrOrganized(Lao lao, RollCall rollcall) {
     String pk;
     try {
       pk =
-          Base64.getUrlEncoder()
-              .encodeToString(
-                  Wallet.getInstance().findKeyPair(firstLaoId, rollcall.getPersistentId()).second);
+          wallet
```
jbsv, line 235: please rename pk so that we don't need to dig into the code to find if its public or private

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/LaoDetailViewModel.java
```diff
@@ -207,11 +208,13 @@ public LaoDetailViewModel(
       @NonNull Application application,
       LAORepository laoRepository,
       Gson gson,
+      Wallet wallet, // FIXME Use KeyManager instead
```
jbsv, line 211: FIXME... when ? :-)

GabrielFleischer: I don't want to change everything at once, so I add FIXME to do it on later commit.

I plan to fix them all before the end fo the feature.
You can be sure that it won't go on master.

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/LaoDetailViewModel.java
GabrielFleischer: Done !

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Wallet.java
GabrielFleischer: Applied

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/security/PoPToken.java
```diff
@@ -8,4 +8,9 @@
   public PoPToken(byte[] privateKey, byte[] publicKey) {
     super(new PlainPrivateKey(privateKey), new PublicKey(publicKey));
   }
+
+  @Override
+  public PlainPrivateKey getPrivateKey() {
+    return (PlainPrivateKey) super.getPrivateKey();
+  }
```
arnauds5, line 15: Is there a difference for PoPToken to return a PrivateKey or a PlainPrivateKey ? (is it only used for sign ?)
If it return a PlainPrivateKey, the private key can be accessed using the getData. Could this be a problem ?

GabrielFleischer: It is not the best, but it does not pose that much threat.
The private key might be needed for some math later or for storage.

It allows the wallet to be tested easily.

And most of all, removing it and the getter of KeyPair does not resolve the issue.
With this piece of code, if it is executed from the right scope, you can access any private field value as it is stored in memory :

```java
PoPToken pair = new PoPToken(VALID_PRIVATE_KEY, VALID_PUBLIC_KEY);
try {
  Field key = pair.getClass().getDeclaredField("privateKey");
  key.setAccessible(true);
  PlainPrivateKey privateKey = (PlainPrivateKey) key.get(pair);
  System.out.println(privateKey.getEncoded());
} catch (NoSuchFieldException | IllegalAccessException e) {
  e.printStackTrace();
}
```

If we really wanted to be secure, we should store the encoded key in memory. But this is way too advanced for the current PR

arnauds5 (approved):

PR #691, opened 2021-12-23
Fix election id generation on the web front-end

GabrielFleischer:
The web front-end generates election ids based on the creation time of the LAO instead of the election's.

### DIFF
--- a/fe1-web/network/MessageApi.ts
+++ b/fe1-web/network/MessageApi.ts
@@ -243,7 +243,7 @@ export function requestCreateElection(
   const message = new SetupElection({
     lao: currentLao.id,
     id: Hash.fromStringArray(
-      EventTags.ELECTION, currentLao.id.toString(), currentLao.creation.toString(), name,
+      EventTags.ELECTION, currentLao.id.toString(), time.toString(), name,
     ),
     name: name,
     version: version,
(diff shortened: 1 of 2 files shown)


fe1-web/network/MessageApi.ts
```diff
@@ -243,7 +243,7 @@ export function requestCreateElection(
   const message = new SetupElection({
     lao: currentLao.id,
     id: Hash.fromStringArray(
-      EventTags.ELECTION, currentLao.id.toString(), currentLao.creation.toString(), name,
+      EventTags.ELECTION, currentLao.id.toString(), time.toString(), name,
```
jbsv, line 246: Was that a confusion between current Lao time and election creation time ?

Would it be possible to catch that in a test ?

jbsv:
LGTM, but as it is fe1, I'll let @pierluca have an eye on it :-)

fe1-web/network/MessageApi.ts
GabrielFleischer: I am not very proficient with front-end tests. But it should be yes.

GabrielFleischer:
I had not changed the id verification, it is now done too.

pierluca (approved):

PR #692, opened 2021-12-23
Update fakeHub with new methods in election tests

rjaccard:


pierluca (approved):
Blind approve. As discussed.

PR #694, opened 2021-12-23
Fix the inbox of the consensus

Anghul:


rjaccard (approved):

rjaccard:
It is not related to this PR but do you have any idea why this test would fail? I can't reproduce it locally but it regularly happens when running the CI a lot of times.

Anghul:
> It is not related to this PR but do you have any idea why this test would fail? I can't reproduce it locally but it regularly happens when running the CI a lot of times.

It comes from a timeout during TestConnectToSocket, but I don't know why

pierluca (approved):

PR #695, opened 2021-12-23
Work fe1 jwen remove chirp

jiabaow:
#643 but change to use the pop token instead of the public key.
Implement remove chirp
Screen Shot 2022-01-02 at 15 39 35


fe1-web/model/objects/Chirp.ts
```diff
@@ -33,6 +34,8 @@ export class Chirp {
   // The id of the parent chirp (if it is a reply)
   public readonly parentId?: Hash;
 
+  public readonly isDeleted?: boolean;
```
Xelowak, line 37: Why can it be undefined ?

fe1-web/network/MessageApi.ts
```diff
@@ -2,11 +2,12 @@ import {
   EventTags, Hash, Lao, PublicKey, Timestamp,
 } from 'model/objects';
 import {
+  AddChirp,
   CastVote,
   CloseRollCall,
   CreateLao,
   CreateMeeting,
-  CreateRollCall,
+  CreateRollCall, DeleteChirp,
```
Xelowak, line 10: Maybe put DeleteChirp on a new line, otherwise it looks weird.

fe1-web/res/strings.ts
```diff
@@ -60,6 +60,7 @@ const STRINGS = {
   /* --- Social Media Strings --- */
   button_publish: 'Publish',
   your_chirp: 'Your chirp',
+  deleted_chirp: 'This chirp was deleted',
```
Xelowak, line 63:
```suggestion
  deleted_chirp: 'This chirp has been deleted',
```

protocol/query/method/message/data/dataAddReaction.json
```diff
@@ -29,4 +29,4 @@
   },
   "additionalProperties": false,
   "required": ["object", "action", "reaction_codepoint", "chirp_id", "timestamp"]
-}
\ No newline at end of file
+}
```
Xelowak: Is this necessary ?

Xelowak (changes requested):
Thanks for the work done. I've made some comments and had some question regarding what you've done. Especially, I think it's easier after all to not make a DeletedChirpCard component, since the ChirpCard is nearly the same.

fe1-web/components/ChirpCard.tsx
```diff
@@ -1,69 +1,64 @@
...
-  } as ViewStyle,
-});
-
 const ChirpCard = (props: IPropTypes) => {
   const { chirp } = props;
   const likesText = `  ${chirp.likes}`;
 
+  const laoSelect = makeCurrentLao();
+  const lao = useSelector(laoSelect);
+  const [userPublicKey, setUserPublicKey] = useState(new PublicKey(''));
```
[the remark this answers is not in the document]
jiabaow: I didn't think about that. But that's true and it's going to make my life much easier on testing. Thanks!

fe1-web/res/strings.ts
jiabaow: I will keep the 'was' as this is the word used in Slack.

protocol/query/method/message/data/dataAddReaction.json
jiabaow: No this may because of the merge.

fe1-web/model/network/method/message/data/MessageData.ts
```diff
@@ -28,10 +28,28 @@ export enum ActionType {
   RESULT = 'result',
   CLOSE = 'close',
   ADD = 'add',
-  ADD_BROADCAST = 'add_broadcast',
+  NOTIFY_ADD = 'add_broadcast',
+  DELETE = 'delete',
+  NOTIFY_DELETE = 'delete_broadcast',
```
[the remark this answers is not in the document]
jiabaow: I suggest we keep them for now and change the protocol together with other systems in a separate PR, so we will not break the CI for the others.

Xelowak (approved):
LGTM !

fe1-web/store/reducers/SocialReducer.ts
```diff
@@ -99,11 +99,29 @@ const socialSlice = createSlice({
...
+      reducer(state, action: PayloadAction<{
+        laoId: string,
+        chirp: ChirpState,
+      }>) {
+        const { laoId, chirp } = action.payload;
+        const store = state.byLaoId[laoId];
+
+        if (store.byId[chirp.id]) {
+          store.byId[chirp.id] = chirp;
+        }
```
pierluca, line 117: This seems to mean that if I receive the messages out-of-order (i.e., delete first, add later), only the added version will be kept.
You might want to test (and fix) this scenario 😉

pierluca (commented):
A few issues to fix, but nice job overalls.
You'll also have to fix the conflicts with the master branch.

fe1-web/model/objects/Chirp.ts
```diff
@@ -56,6 +60,11 @@ export class Chirp {
     } else {
       this.likes = obj.likes;
```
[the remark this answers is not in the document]
Xelowak: No we're not, you're right. I can do it in a separated PR.

fe1-web/components/ChirpCard.tsx
```diff
@@ -59,11 +70,20 @@ const styles = StyleSheet.create({
 
 const ChirpCard = (props: IPropTypes) => {
   const { chirp } = props;
-  const likesText = `  ${chirp.likes}`;
+  const { userPublicKey } = props;
 
   // This is temporary for now
   const zero = '  0';
 
+  const isSender = userPublicKey.valueOf() === chirp.sender.valueOf();
```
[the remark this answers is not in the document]
Xelowak: I don't think we will be able to handle this case, since we have no way to store the token of a past roll call yet.

fe1-web/model/objects/Chirp.ts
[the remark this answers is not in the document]
Xelowak: Actually Jiabao has already done it in PR #720

fe1-web/components/ChirpCard.tsx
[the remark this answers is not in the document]
pierluca: ... and I don't think it's necessarily a priority to fix it 🙂 

However, be aware that you don't need to store it, you can re-compute it from the lao_id and roll_call_id, and you do know of all the roll calls in a given LAO 😉

fe1-web/model/network/validation/DataSchema.ts
```diff
@@ -12,9 +12,9 @@ import dataCastVote from 'protocol/query/method/message/data/dataCastVote.json';
 import dataEndElection from 'protocol/query/method/message/data/dataEndElection.json';
 import dataResultElection from 'protocol/query/method/message/data/dataResultElection.json';
 import dataAddChirp from 'protocol/query/method/message/data/dataAddChirp.json';
-import dataAddChirpBroadcast from 'protocol/query/method/message/data/dataAddChirpBroadcast.json';
+import dataNotifyAdd from 'protocol/query/method/message/data/dataNotifyAddChirp.json';
```
pierluca, line 15:
```suggestion
import dataNotifyAddChirp from 'protocol/query/method/message/data/dataNotifyAddChirp.json';
```

fe1-web/model/network/validation/DataSchema.ts
```diff
@@ -36,7 +36,7 @@ const dataSchemas = [
   dataWitnessMessage,
 
   dataAddChirp,
-  dataAddChirpBroadcast,
+  dataNotifyAdd,
```
pierluca, line 39:
```suggestion
  dataNotifyAddChirp,
```

fe1-web/store/reducers/SocialReducer.ts
pierluca: This still looks like it needs fixing.

pierluca (commented):
Looks much better, still a few issues (+ conflict)

fe1-web/store/reducers/SocialReducer.ts
```diff
@@ -99,11 +99,31 @@ const socialSlice = createSlice({
...
+        chirp: ChirpState,
+      }>) {
+        const { laoId, chirp } = action.payload;
+        const store = state.byLaoId[laoId];
+
+        if (store.byId[chirp.id]) {
+          if (!store.byId[chirp.id].isDeleted) {
+            store.byId[chirp.id] = chirp;
+          }
+        }
```
pierluca, line 119:
```suggestion
        store.byId[chirp.id] = chirp;
```
wouldn't this be more robust / sufficient ?

pierluca: Also, thinking about it, should we actually consider setting the `isDeleted` and `text` fields here? 
It might make sense not to rely on the `chirp` parameter being configured correctly from the calling code.

fe1-web/store/reducers/__tests__/SocialReducer.test.ts
```diff
@@ -145,6 +184,21 @@ test('reducer should add the newest chirp on top', () => {
...
+
+test('delete a non-stored chirp should do nothing', () => {
+  expect(socialReduce(filledState4, deleteChirp(mockLaoId, chirp0)))
+    .toEqual(filledState4);
+});
+
+test('reducer should not re-add a chirp if it has already been deleted', () => {
+  expect(socialReduce(filledState5, addChirp(mockLaoId, chirp1Deleted)))
+    .toEqual(filledState5);
+});
```
pierluca, line 200:
```suggestion
test('reducer should not re-add a chirp if it has already been deleted', () => {
  var stateDeleted = socialReduce(filledState3, deleteChirp(mockLaoId, chirp4));
  expect(socialReduce(stateDeleted, addChirp(mockLaoId, chirp4)))
    .toEqual(filledState4);
});
```

would the following work ? I guess not, but it's entirely possible that in the future we get messages out of order (e.g., if we're connected to multiple servers)

fe1-web/store/reducers/SocialReducer.ts
pierluca: Alternatively, we could check in the ingestion that we "fail fast" on processing if we haven't received the chirp yet. 
In this case, your current approach makes a lot of sense, but then the ingestion should really fail so that the message remains marked as unprocessed.

pierluca: Final question, concerning the protocol, should a different sender (than the original chirp) be allowed to delete a chirp? And if not, shouldn't we check it here?

pierluca (commented):
LGTM, but I'm not convinced that the handling of chirp deletion is robust yet.

fe1-web/store/reducers/__tests__/SocialReducer.test.ts
jiabaow: With the current implementation, this test passes (first delete a non-stored chirp, then add it).
So, if I understand correctly, when we receive out-of-order messages (it should be: add then delete, but we get: delete then add), the expected state should mark the chirp as deleted (toEqual(filledState5))?

fe1-web/store/reducers/SocialReducer.ts
jiabaow: > wouldn't this be more robust / sufficient ?
This enables marking a non-stored chirp as deleted to avoid reordering situation, then later if the reducer receives addChirp, since the chirp id already presents in byId, the store will not update its state. But it still wants to add it to the allIdsInOrder list and byUser map to display the deleted chirp. So, along with this, I will also need to move the insertion to allIdsInOrder and byUser outside the if(!store.byId[chirp.id]) in addChirp, or I still miss something?

jiabaow: > Also, thinking about it, should we actually consider setting the `isDeleted` and `text` fields here? It might make sense not to rely on the `chirp` parameter being configured correctly from the calling code.

Can you elaborate on why we cannot trust it from the calling code (does it refer to deleteChirp(lao.id, chirp.toState()) in Socialhandler), please? I don't see what makes a difference in setting them in SocialHandler and in SocialReducer.

jiabaow: > Final question, concerning the protocol, should a different sender (than the original chirp) be allowed to delete a chirp? And if not, shouldn't we check it here?

No, only the sender of the original chirp can delete a chirp. With the current UI, only the sender can send a delete request, but it's always safer to have a check on it. So if the chirp is present in `byId`, I compare the sender of the two. But how can I handle the case if the chirp is not yet stored in `byId`, in an out-of-order situation? And, if we cannot rely on the chirp parameter from the calling code as the same reason for i`sDeleted` and `text`, can I rely on the `sender`?

fe1-web/store/reducers/SocialReducer.ts
```diff
@@ -86,21 +86,26 @@ const socialSlice = createSlice({
...
+          store.allIdsInOrder, store.byId, chirp.time,
+        );
+        store.allIdsInOrder.splice(insertIdxInAll, 0, chirp.id);
+
+        if (!state.byLaoId[laoId].byUser[chirp.sender]) {
+          store.byUser[chirp.sender] = [chirp.id];
+        } else {
+          const senderChirps = store.byUser[chirp.sender];
+          const insertIdxInUser = findInsertIdx(senderChirps, store.byId, chirp.time);
+          senderChirps.splice(insertIdxInUser, 0, chirp.id);
```
pierluca, line 108: Good job on handling the out-of-order messages.
I believe you've changed the adding logic here though. You're *always* going to modify `allIdsInOrder`, whereas before you were doing it only when adding. Is this on purpose ?

fe1-web/store/reducers/SocialReducer.ts
```diff
@@ -115,20 +120,68 @@ const socialSlice = createSlice({
...
+        }
+
         const store = state.byLaoId[laoId];
 
-        if (store.byId[chirp.id]) {
-          if (!store.byId[chirp.id].isDeleted) {
-            store.byId[chirp.id] = chirp;
-          }
+        // store the deleted chirp
+        store.byId[chirp.id] = new Chirp({
```
pierluca, line 136: and I believe here you're not doing the corresponding check on the sender, to make sure that only the original sender can delete ?

pierluca (commented):
There seems to be a tiny incoherence between addChirp and deleteChirp, but we're almost there !

fe1-web/store/reducers/SocialReducer.ts
jiabaow: Yes, because we use `allIdsInOrder` to display all chirps (including the deleted ones) in the right order. 
For out-of-message messages, there're two possible cases: 
1. same senders: we keep the state in `byId` as deleted, but still add it to `allIdsInOrder` to show it as a deleted chirp using the `addChirp` time.
2. different senders: the sender doesn't want to delete the chirp, we do the same things as for a normal add.
I will add a comment here to make it clear.

pierluca (approved):
Great !

PR #699, opened 2021-12-26
Fix roll call handling

tuomas-p:


be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
```diff
@@ -66,7 +66,7 @@ case object RollCallHandler extends MessageHandler {
               //a closeRollCall is always sent in the lao's main channel so we are allowed to do this
               val socialChannel: String = generateSocialChannel(rpcMessage.getParamsChannel, head)
               val ask: Future[GraphMessage] = (dbActor ? DbActor.CreateChannel(Channel(socialChannel), ObjectType.CHIRP)).map {
-                case DbActorWriteAck() => createAttendeeChannels(tail, rpcMessage)
+                case DbActorAck() => createAttendeeChannels(tail, rpcMessage)
```
BadrTad, line 69: I suppose that this case was not handled in the test suite, Would you please cover it ?

BadrTad (commented):
Would be great to cover this in the tests

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
tuomas-p: (starting to be done in PR#701)

pierluca (approved):
Approving blindly.

PR #700, opened 2021-12-26
change question's ID in electionSetup

rjaccard:
The ID of the questions in the election ID setup is currently wrong. It should be the election ID and not the lao ID.


pierluca (commented):
I'll let you test this. I hadn't seen that both Election and ElectionState use those interfaces. 
It might work as-is, but this seems like quite a few changes with no corresponding tests.
If you confirm you've tested it, I'm happy to approve it.

rjaccard:
> I'll let you test this. I hadn't seen that both Election and ElectionState use those interfaces. It might work as-is, but this seems like quite a few changes with no corresponding tests. If you confirm you've tested it, I'm happy to approve it. @jiabaow @rjaccard

Fine for me! Thank you Jiabaow! We can now have an election :)

jiabaow:
> I'll let you test this. I hadn't seen that both Election and ElectionState use those interfaces. It might work as-is, but this seems like quite a few changes with no corresponding tests. If you confirm you've tested it, I'm happy to approve it. @jiabaow @rjaccard

The tests already exit, I updated them with the correct id.

pierluca (approved):

PR #704, opened 2021-12-27
Work-be2-taddist-closeDB

BadrTad:
Made this to a separate PR since it's urgent.
With this PR the database is closed and its resources are released upon stopping the DbActor


be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/DbActor.scala
```diff
@@ -192,6 +192,11 @@ object DbActor extends AskPatternConstants {
   // FIXME: find way to keep track of db size
   sealed case class DbActor(mediatorRef: ActorRef, db: DB, DATABASE_FOLDER: String) extends Actor with ActorLogging {
 
+    //Close the db and release the resources right before stopping the actor
+    override def postStop() = {
+      db.close();
+    }
+
```
tuomas-p, line 199: :+1:

tuomas-p (approved):
looks good to me also

pierluca (approved):

PR #705, opened 2021-12-27
Additional checks for  base64 and refactor tests

BadrTad:
This PR  adds additional checks for the Base64Data constructor and refactor the tests that are affected by  this new changes

### DIFF
--- a/be2-scala/src/main/scala/ch/epfl/pop/model/objects/Base64Data.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/model/objects/Base64Data.scala
@@ -3,7 +3,7 @@ package ch.epfl.pop.model.objects
 import java.nio.charset.StandardCharsets
 import java.util.Base64
 
-case class Base64Data(data: String) {
+sealed case class Base64Data(data: String) {
   val DECODER: Base64.Decoder = Base64.getUrlDecoder
 
   def decodeToString(): String = new String(this.decode(), StandardCharsets.UTF_8)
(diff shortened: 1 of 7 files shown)


be2-scala/src/main/scala/ch/epfl/pop/model/objects/Base64Data.scala
```diff
@@ -15,6 +15,12 @@ case class Base64Data(data: String) {
   def equals(that: Base64Data): Boolean = data == that.data
 
   override def toString: String = data.toString
+
+  try {
```
tuomas-p, line 19: good check!

tuomas-p (approved):
Not much to say since most of the PR is about Base64Data refactoring, the Base64Data check is god for code robustness

be2-scala/src/main/scala/ch/epfl/pop/model/objects/Base64Data.scala
```diff
@@ -15,6 +15,12 @@ case class Base64Data(data: String) {
...
+    DECODER.decode(data);
+  } catch {
+    case _: IllegalArgumentException => throw new IllegalArgumentException(s"String $data is not Base64")
```
pierluca, line 22: I don't see a test that checks this case, it might be worthwhile to have one.

pierluca (approved):
Approving semi-blindly. Just an advice for (maybe) a future PR.

be2-scala/src/main/scala/ch/epfl/pop/model/objects/Base64Data.scala
BadrTad: Sure I'll add to one of  my pending PRs

PR #710, opened 2021-12-27
skip TestConnectToSocket

rjaccard:


pierluca (approved):
In a normal flow I wouldn't approve this, but I understand this is needed to unblock other teams.
Do try to come back to this by the end of the project if possible.

PR #720, opened 2022-01-03
Work fe1 jwen reaction merge be1

jiabaow:


fe1-web/model/objects/__tests__/Channel.test.ts
```diff
@@ -4,7 +4,7 @@ import { Hash } from '../Hash';
 import {
   channelFromIds,
   getGeneralChirpsChannel,
-  getLastPartOfChannel,
+  getLastPartOfChannel, getReactionChannel,
```
Xelowak, line 7:
```suggestion
  getLastPartOfChannel,
  getReactionChannel,
```

fe1-web/store/reducers/SocialReducer.ts
```diff
@@ -99,11 +104,46 @@ const socialSlice = createSlice({
...
+
+        if (!store.reactionsByChirp[reaction.chirp_id]) {
+          store.reactionsByChirp[reaction.chirp_id] = { [reaction.codepoint]: [reaction.sender] };
+        } else if (!store.reactionsByChirp[reaction.chirp_id][reaction.codepoint]) {
+          store.reactionsByChirp[reaction.chirp_id][reaction.codepoint] = [reaction.sender];
+        } else if (!store.reactionsByChirp[reaction.chirp_id][reaction.codepoint]
+          .includes(reaction.sender)) {
+          store.reactionsByChirp[reaction.chirp_id][reaction.codepoint].push(reaction.sender);
+        } else {
+          console.warn('You already reacted to this reaction');
```
Xelowak, line 138: I don't think it's necessary. What we could do to ensure that the user is informed of this is changing the style of the icon or of the counter. This way, the user knows it whiteout having a console message (which shows up only in the web console anyway).
I can do it in a separated PR though.

Xelowak (changes requested):
Nice work ! I made some comments for you.

fe1-web/store/reducers/SocialReducer.ts
pierluca: Also, technically, it's "The sender already reacted to this reaction", not necessarily the user who parses this message🙂 
I wouldn't mind keeping the logging: it's useful to know that the message was processed.
Do however make it a "debug" message, this is definitely not a serious issue that one needs to be `warn`'d about

fe1-web/components/ChirpCard.tsx
```diff
@@ -59,10 +62,19 @@ const styles = StyleSheet.create({
 
 const ChirpCard = (props: IPropTypes) => {
   const { chirp } = props;
-  const likesText = `  ${chirp.likes}`;
+  const reactions = useSelector(makeReactionsList())[chirp.id.toString()];
 
-  // This is temporary for now
   const zero = '  0';
```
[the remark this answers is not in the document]
jiabaow: The zero is kept for the reply

fe1-web/ingestion/handlers/RollCall.ts
```diff
@@ -144,6 +144,12 @@ function handleRollCallCloseMessage(msg: ExtendedMessage): boolean {
     }
   });
 
+  // everyone is automatically subscribed to the reaction channel after the roll call
+  subscribeToChannel(getReactionChannel(lao.id))
+    .catch((err) => {
+      console.error('Could not subscribe to reaction channel, error:', err);
+    });
```
pierluca, line 151: Though not strictly needed, I'd suggest putting this and `await`ing it in the `dispatch` call above, so that all async operations are being performed in one go, in a predictable order.

fe1-web/store/reducers/SocialReducer.ts
```diff
@@ -124,12 +164,44 @@ export const makeChirpsList = () => createSelector(
...
+    }
+    if (list.byLaoId[laoId]) {
+      const store = list.byLaoId[laoId];
+      const reactions: Record<string, Record<string, number>> = {};
+      store.allIdsInOrder.forEach(
+        (id) => {
+          const reactionByUser = store.reactionsByChirp[id];
+          if (reactionByUser) {
+            reactions[id] = createReactionsEntry(reactionByUser);
+          }
```
pierluca, line 200: I'm a bit surprised by the naming. Shouldn't this be ? 
```suggestion
          const chirpReactions = store.reactionsByChirp[id];
          if (chirpReactions) {
            reactions[id] = createReactionsEntry(chirpReactions);
          }
```

fe1-web/store/reducers/__tests__/SocialReducer.test.ts
```diff
@@ -116,66 +165,312 @@ const filledState4 = {
...
-  expect(makeChirpsList().resultFunc(filledState3, mockLaoId))
-    .toEqual([chirp2, chirp3, chirp1]);
+describe('chirp selector', () => {
+  test('selector should return an empty list of chirpState when no lao is opened', () => {
+    expect(makeChirpsList().resultFunc(emptyState, undefined))
+      .toEqual([]);
+  });
+
+  test('selector should return an empty list', () => {
+    socialReduce(undefined, {} as AnyAction);
```
pierluca, line 396: In this and all the following tests, what is the call to `socialReduce` being used for ? 
My understanding is that we're not saving its return value and not using it in subsequent calls 🤔

pierluca (commented):
A few minor issues, but this looks very good overalls 👍

pierluca (approved):
LGTM, but I'll let @Xelowak re-review the changes :)

Xelowak (approved):
LGTM !

PR #724, opened 2022-01-04
Fix BE2 build

BadrTad:
The following changes are required in order to run properly the BE2 server with the new sbt version since some utilities were deprecated and default tasks were upgraded.


tuomas-p (approved):

RaulinN (approved):
I will have to trust you on that one :p

PR #682, opened 2021-12-22
Work taddist be2 data schema validator

BadrTad:
For now the schema validation is done for the object_type: LAO,
as discussed in the meeting, validation of the schema is done along with the decoding and parsing of the payload, meaning the work is within the same graph component MessageDecoder.scala

PR #725, opened 2022-01-05
refactor add_broadcast to notify_add to respect the protocol change in Scala

tuomas-p:
rename files, classes and variables to respect protocol


BadrTad:
This does not seem to include `DataRegistryModule`, and `DataSchemaValidator `?
I just saw that you are merging this in `work-be1-rjaccard-notify_add`,  be careful it will definitely encounter merge conflicts with master. 😅

tuomas-p:
yeah I need to be careful about that, but I had issued the request to merge with Robin's branch before #682 was online

tuomas-p:
I believe I have corrected the issues now

PR #727, opened 2022-01-05
Modified add_broadcast to notify_add to fit with the new protocol

Lugiasc:
We changed the action name from add_broadcast to notify_add but the protocol hasn't changed otherwise


fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/socialmedia/NotifyAddChirp.java
```diff
@@ -19,7 +19,7 @@
    * @param channel channel where the post is located
    * @param timestamp UNIX timestamp in UTC
    */
-  public AddChirpBroadcast(String postId, String channel, long timestamp) {
+  public NotifyAddChirp(String postId, String channel, long timestamp) {
```
arnauds5, line 22: The toString of NotifyAddChirp should be updated too, it's still using "AddChirpBroadcast"

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/socialmedia/NotifyAddChirp.java
```diff
@@ -6,7 +6,7 @@
 import com.google.gson.annotations.SerializedName;
 
 /** Data sent to broadcast AddChirp to the general channel */
```
arnauds5, line 8: Is the documentation still correct ?

Lugiasc: Yes the protocol in itself still does the same thing, we changed the action name because we thought it was more intuitive. Hence why I changed the name of the class as well

PR #730, opened 2022-01-06
Work be2 tuomas p refactor notify

tuomas-p:
Update the protocol in fe2, be1 and be2 to replace add_broadcast and delete_broadcast by notify_add and notify_delete as well as fix merge conflict in Scala. The three languages have already merged their modifications without conflicts


pierluca (changes requested):
The typescript changes are inconsistent with the naming convention, as the object (chirp) has been completely removed from the class name.
E.g. `notifyAddChirp` has become `notifyAdd` , which could be in conflict with `notifyAddSomethingElse` in the future.

The fe1 CI also seems to be broken for good reasons. 
If you take the time to read fully the errors, they're legitimate (that JSON schema doesn't exist) and they still show inconsistencies between broadcast and notify in the test naming.
E.g.
```
FAIL model/network/method/message/data/__tests__/NotifyAdd.test.ts
  ● AddBroadcastChirp › should be parsed correctly from JSON

    no schema with key or ref "https://raw.githubusercontent.com/dedis/student_21_pop/master/protocol/query/method/message/data/dataNotifyAdd.json"
```
As such, this doesn't seem mergeable to me.

fe1-web/model/network/method/message/data/__tests__/NotifyAddChirp.test.ts
```diff
@@ -29,20 +29,20 @@ const dataNotifyAdd = `{
 
 describe('AddBroadcastChirp', () => {
```
pierluca, line 30: You might want to change the describe too :-)

pierluca (approved):

PR #735, opened 2022-01-09
Enter attendees manually

Xelowak:
For testing purposes, we are in the need of adding attendees without having to scan their token.
I added a way to do it. On the opened roll call page, you can now press on the button "Add attendee manually". A modal will pop up on screen, asking for an attendee token.
Capture d’écran 2022-01-10 à 14 03 01

Then, a regex will be used to check that it is a token. It is a simple check that there are 43 chars that are letters, numbers, - or _ and followed by '=' at the end. If is passes, there is a new toast message that is displayed:
Capture d’écran 2022-01-10 à 14 03 46

If not, the organized will be informed:
Capture d’écran 2022-01-10 à 14 02 50

Attendees are correctly added to the list when closing the roll call:
Capture d’écran 2022-01-09 à 17 23 25

### DIFF
--- a/fe1-web/parts/lao/organizer/RollCallOpened.tsx
+++ b/fe1-web/parts/lao/organizer/RollCallOpened.tsx
@@ -19,6 +19,7 @@ import {
 import { makeCurrentLao, OpenedLaoStore } from 'store';
 import { useToast } from 'react-native-toast-notifications';
 import { useSelector } from 'react-redux';
+import ConfirmModal from '../../../components/ConfirmModal';
 
 const styles = StyleSheet.create({
   viewCenter: {
@@ -36,6 +37,7 @@ const RollCallOpened = () => {
   const { rollCallID, time } = route.params;
   const navigation = useNavigation();
   const [attendees, updateAttendees] = useState(new Set<string>());
+  const [inputModalIsVisible, setInputModalIsVisible] = useState(false);
   const toast = useToast();
   const laoSelect = makeCurrentLao();
   const lao = useSelector(laoSelect);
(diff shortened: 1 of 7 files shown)


fe1-web/parts/lao/organizer/RollCallOpened.tsx
```diff
@@ -69,6 +77,25 @@ const RollCallOpened = () => {
     }
   };
 
+  const handleEnterManually = (input: string) => {
+    if (base64Matcher.test(input)) {
+      if (!attendees.has(input)) {
+        updateAttendees((prev) => new Set<string>(prev.add(input)));
+        toast.show(STRINGS.roll_call_participant_added, {
```
jiabaow, line 84: Since the toast is the same as handleScan, you can factorize them into a single function.

fe1-web/parts/lao/organizer/RollCallOpened.tsx
```diff
@@ -69,6 +77,25 @@ const RollCallOpened = () => {
     }
   };
 
+  const handleEnterManually = (input: string) => {
```
jiabaow, line 80: It will be nice if this function has its test too

jiabaow (approved):
LGTM

GabrielFleischer (commented):
That's insane ! Thank you so much !

From the screens I see, it is very good.

But I'm not proficient in TypeScript, so I'll let other review this

fe1-web/parts/lao/organizer/RollCallOpened.tsx
Xelowak: Actually it's not the same toast, the text is not the same. I'll try to find a way to have less duplication though.

fe1-web/parts/lao/organizer/RollCallOpened.tsx
Xelowak: It's now done

PR #736, opened 2022-01-10
Quick fix in NotifyAddChirp and changes in Chirp

Lugiasc:
As pointed out by Arnaud, one of the variable was still called post_id despite being named chirp_id in the protocol.
He also recommended to add test for data validation.
I deleted a variable to count the number of likes in the Chirp object because it won't be counted that way, then I added a flag isDeleted to know whether a chirp is deleted or not.


GabrielFleischer (approved):
LGTM outside of Arnaud's requested changes

jbsv (approved):
LGTM

PR #738, opened 2022-01-10
Enter attendees manually

arnauds5:
Add a button for the organizer during a roll call for adding an attendee manually (image 1).
When clicked, it will open a popup asking the attendee token (image 2).
When confirming the token (image 3), if the key is invalid (for example if not base64 or decoded length != 32) or if it was already added, it will display a warning (image 5), else it will add the token to the attendees list (image 4).

### TODO fix the tests (work locally but not in the CI) or drop this PR if the QRCodeScanningFragment will be redo from scratch

image
image
image
image
image

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/LaoDetailViewModel.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/LaoDetailViewModel.java
@@ -1360,6 +1360,18 @@ public class LaoDetailViewModel extends AndroidViewModel
     mOpenAttendeesListEvent.postValue(new SingleEvent<>(rollCallId));
   }
 
+  public void addAttendee(PublicKey attendee) {
+    if (attendees.contains(attendee)) {
+      mScanWarningEvent.postValue(
+          new SingleEvent<>("This QR code has already been scanned. Please try again."));
+      return;
+    }
+
+    attendees.add(attendee);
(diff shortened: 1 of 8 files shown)


fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/LaoDetailViewModel.java
```diff
@@ -1385,20 +1398,19 @@ public void onQRCodeDetected(Barcode barcode) {
...
+      mScanWarningEvent.postValue(
+          new SingleEvent<>(getApplication().getString(R.string.invalid_qrcode)));
       return;
     }
 
     if (attendees.contains(attendee)
         || Objects.requireNonNull(mWitnesses.getValue()).contains(attendee)) {
       mScanWarningEvent.postValue(
-          new SingleEvent<>("This QR code has already been scanned. Please try again."));
+          new SingleEvent<>(getApplication().getString(R.string.qrcode_already_scanned)));
```
GabrielFleischer, line 1409: Nice side refactor !

GabrielFleischer (commented):
Good job ! we still need some tests and it's perfect !

GabrielFleischer:
In fact, now that I think about it, it is probably simpler to allow manual input all the time.
Not only for attendee.
And use the listener to give the value.

GabrielFleischer:
The current implementation of ElectionFragment is extremely complex for no reason.
For now, we will just add a text input to enter data as-if it was scanned.

This is the easiest and smallest modification to do while keeping the wanted feature.
This PR will be closed for now.

PR #671, opened 2021-12-19
Work be2 tuomas p social media delete chirp

tuomas-p:

PR #739, opened 2022-01-10
Work be2 tuomas p fix git history social media delete chirp

tuomas-p:
implement the delete chirp functionality but without the broken git history


pierluca (approved):
As discussed, approving this in lieu of #671

PR #737, opened 2022-01-10
Add blockies

Xelowak:
I added images generated by the public keys of users, to be able to represent them visually.
They are now on the Home Screen (in the text input and in chirps) and on the Search screen (list of attendees). These images' colors have been taken from the Popstellar icon you can find in the be1 folder of GitHub. If you don't have participated in a roll call yet, it is still a person icon. Here are screenshots:

social_home

attendees_list

I also mocked them in tests, since this external library that generates images doesn't work in them.

PR #740, opened 2022-01-11
Add current user public key in social media navigation

Xelowak:
Since we use it everywhere in the social media, I realized that it would be much simpler if the PoP token's public key was given as a props to the social screens. Then, I added it directly in SocialMediaNavigation and passed it to SocialHome and SocialSearch.
With this, I could resolve an issue where you would see yourself in the list of attendees in search. It is now no longer the case.
I also addressed Jiabao's comments in my PR #737.

### DIFF
--- a/fe1-web/parts/lao/socialMedia/SocialMediaNavigation.tsx
+++ b/fe1-web/parts/lao/socialMedia/SocialMediaNavigation.tsx
@@ -16,7 +16,7 @@ import SocialHome from './SocialHome';
 const Tab = createMaterialTopTabNavigator();
 
 const SocialMediaNavigation = () => {
-  const [userPublicKey, setUserPublicKey] = useState(new PublicKey(''));
+  const [currentUserPublicKey, setCurrentUserPublicKey] = useState(new PublicKey(''));
 
   const laoSelect = makeCurrentLao();
   const lao = useSelector(laoSelect);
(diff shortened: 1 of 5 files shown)


fe1-web/parts/lao/socialMedia/SocialMediaNavigation.tsx
```diff
@@ -3,49 +3,88 @@ import { Ionicons } from '@expo/vector-icons';
...
+  }
+
+  // Get the pop token of the user using the last tokenized roll call
+  const rollCallId = lao.last_tokenized_roll_call_id;
+  const eventSelect = makeEventGetter(lao.id, rollCallId);
+  const rollCall: RollCall = useSelector(eventSelect) as RollCall;
+
+  // This will be run again each time the lao.last_tokenized_roll_call_id changes
+  useEffect(() => {
+    generateToken(lao.id, rollCallId).then((token) => {
```
pierluca, line 35: what if rollCallId is undefined ?

Xelowak: As you can see [here](https://github.com/dedis/student_21_pop/blob/work-fe1-kschneiter-user-public-key-in-nav/fe1-web/model/objects/wallet/Token.ts), the promise resolves to undefined.

jiabaow (approved):
LGTM

pierluca (approved):
As discussed, approving and bypassing the code coverage check.

PR #742, opened 2022-01-12
Add functionalities to KeyManager to simplify expressions and give access to PoP Tokens

GabrielFleischer:
Generate the device key object only once, as it should never change during the whole application lifecycle
With this change, there is no need to try for errors each time a key is retrieved.

Also add the functionalities to retrieve a valid PoP Token from an LAO.

Use that new functionality to finally fix the issue with the CastVote sender that was the device key and not the public part of the PoP Token.

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Wallet.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Wallet.java
@@ -8,6 +8,7 @@ import androidx.annotation.NonNull;
 import com.github.dedis.popstellar.model.objects.security.PoPToken;
 import com.github.dedis.popstellar.model.objects.security.PublicKey;
 import com.github.dedis.popstellar.ui.wallet.stellar.SLIP10;
+import com.github.dedis.popstellar.utility.error.keys.InvalidPoPTokenException;
 import com.github.dedis.popstellar.utility.error.keys.KeyGenerationException;
 import com.github.dedis.popstellar.utility.error.keys.SeedValidationException;
 import com.github.dedis.popstellar.utility.error.keys.UninitializedWalletException;
@@ -28,7 +29,6 @@ import java.util.Arrays;
 import java.util.Base64;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.Optional;
 import java.util.Set;
 import java.util.StringJoiner;
 
@@ -105,15 +105,16 @@ public class Wallet {
    * @param rollCallTokens a {@link Set} containing the public keys of all attendees present on
    *     roll-call’s results.
    * @return the PoP Token if the user participated in that roll-call or else empty.
-   * @throws KeyGenerationException if an error occurs
+   * @throws KeyGenerationException if an error occurs during key generation
    * @throws UninitializedWalletException if the wallet is not initialized with a seed
+   * @throws InvalidPoPTokenException if the token is not a valid attendee
    */
-  public Optional<PoPToken> recoverKey(
+  public PoPToken recoverKey(
       @NonNull String laoID, @NonNull String rollCallID, @NonNull Set<PublicKey> rollCallTokens)
-      throws KeyGenerationException, UninitializedWalletException {
+      throws KeyGenerationException, UninitializedWalletException, InvalidPoPTokenException {
     PoPToken token = generatePoPToken(laoID, rollCallID);
-    if (rollCallTokens.contains(token.getPublicKey())) return Optional.of(token);
-    else return Optional.empty();
+    if (rollCallTokens.contains(token.getPublicKey())) return token;
+    else throw new InvalidPoPTokenException(token);
   }
 
   /**
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Wallet.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Wallet.java
@@ -104,7 +104,7 @@ public class Wallet {
    * @param rollCallID a String.
    * @param rollCallTokens a {@link Set} containing the public keys of all attendees present on
    *     roll-call’s results.
-   * @return the PoP Token if the user participated in that roll-call or else empty.
+   * @return the PoP Token if the user participated in that roll-call.
    * @throws KeyGenerationException if an error occurs during key generation
    * @throws UninitializedWalletException if the wallet is not initialized with a seed
    * @throws InvalidPoPTokenException if the token is not a valid attendee
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/LaoDetailViewModel.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/LaoDetailViewModel.java
@@ -247,12 +247,7 @@ public class LaoDetailViewModel extends AndroidViewModel
    * @return the public key
    */
   public PublicKey getPublicKey() {
-    try {
-      return mKeyManager.getMainPublicKey();
-    } catch (GeneralSecurityException | IOException e) {
-      Log.d(TAG, PK_FAILURE_MESSAGE, e);
-      return null;
-    }
+    return mKeyManager.getMainPublicKey();
   }
 
   @Override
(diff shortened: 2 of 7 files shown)


fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/security/KeyManager.java
```diff
@@ -22,46 +26,125 @@
...
-    return getPublicKey(keysetManager.getKeysetHandle());
+  /**
+   * This will regenerate the cached device KeyPair.
+   *
+   * <p>Use this only if you know what you are doing
+   *
+   * @throws IOException when the key cannot be retrieved due to IO errors
+   * @throws GeneralSecurityException when the retrieved key is not valid
+   */
+  public void regenerateMainKey() throws GeneralSecurityException, IOException {
```
arnauds5, line 67: Except from the constructor, when/where will this method be called ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/LaoDetailViewModel.java
```diff
@@ -360,8 +351,11 @@ public void sendVote(List<ElectionVote> votes) {
                   throwable -> Log.d(TAG, "timed out waiting for result on cast_vote", throwable));
 
       disposables.add(disposable);
-    } catch (GeneralSecurityException | IOException e) {
-      Log.d(TAG, PK_FAILURE_MESSAGE, e);
+    } else {
+      Log.i(
+          TAG,
+          "User tried to send a message that expected proof-of-personhood but it could not be achieved.");
+      Toast.makeText(getApplication(), "your PoPToken is invalid", Toast.LENGTH_LONG).show();
```
pierluca, line 356: is it invalid, or unavailable ? 🤔

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/security/KeyManager.java
pierluca: and isn't it a problem that we're regenerating at every initialization?

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/security/KeyManager.java
```diff
@@ -22,46 +26,136 @@
...
 import java.util.Base64;
+import java.util.Comparator;
+import java.util.Optional;
 
 import javax.inject.Inject;
 import javax.inject.Singleton;
 
+/** Service manging keys and providing easy access to the main device key and PoP Tokens */
 @Singleton
 public class KeyManager {
```
pierluca, line 37: the API of this class confuses me a bit, between `retrieveValidPoPToken`, `getPoPToken` and `requirePoPToken`, they all sound sort of similar.

1. Do we need both `requirePoPToken` and `getPoPToken`, or could you drop one ? The only (likely) failure scenario is the wallet not being initialized. I also don't see them being used in the code.

2. Is there a distinction between `retrieve` and `get` in this context? It might be easier to quickly distinguish between `getPoPToken` and `getValidPoPToken`

3. In my understanding, both `get` and `require` are actually generating a PoP token, which would simplify the API to just `getValidPoPToken` and `generatePoPToken`.

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/security/KeyManager.java
```diff
@@ -22,46 +26,136 @@
...
+   *
+   * <p>Exceptions are thrown if the process fails
+   *
+   * @param laoID id of the lao
+   * @param rollCallID id of the lao's roll call
+   * @return the generated PoPToken
+   * @throws GeneralSecurityException if something goes wrong during the generation
+   */
+  public PoPToken requirePoPToken(String laoID, String rollCallID) throws GeneralSecurityException {
     return wallet.findKeyPair(laoID, rollCallID);
```
pierluca, line 92: This looks like a more accurate name would be 

```suggestion
    return wallet.generatePoPToken(laoID, rollCallID);
```

though that would entail changes on the wallet side.

pierluca (commented):
Not a thorough review, just a few comments :-)

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/socialmedia/SocialMediaViewModel.java
```diff
@@ -204,34 +201,26 @@ public void subscribeToGeneralChannel(String laoId) {
...
-
-      disposables.add(disposable);
-
-    } catch (GeneralSecurityException | IOException e) {
-      Log.d(TAG, PK_FAILURE_MESSAGE, e);
-      Toast.makeText(
-              getApplication().getApplicationContext(), PK_FAILURE_MESSAGE, Toast.LENGTH_SHORT)
-          .show();
-    }
+    String channel = ROOT + laoId + "/social/" + mKeyManager.getMainPublicKey().getEncoded();
```
Lugiasc, line 204: the social media uses the PoP token for the channel

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/socialmedia/SocialMediaViewModel.java
```diff
@@ -252,38 +241,31 @@ public void sendChirp(String text, @Nullable MessageID parentId, long timestamp)
...
-    } catch (GeneralSecurityException | IOException e) {
-      Log.d(TAG, PK_FAILURE_MESSAGE, e);
-      Toast.makeText(
-              getApplication().getApplicationContext(), PK_FAILURE_MESSAGE, Toast.LENGTH_SHORT)
-          .show();
-    }
+    KeyPair mainKey = mKeyManager.getMainKeyPair();
+    String channel = laoChannel + "/social/" + mainKey.getPublicKey().getEncoded();
+    Log.d(TAG, PUBLISH_MESSAGE);
+    MessageGeneral msg = new MessageGeneral(mainKey, addChirp, mGson);
```
Lugiasc, line 247: here it will use the PoP token as well

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/security/KeyManager.java
GabrielFleischer: It is not really generated. More retrieved from the library.

I will rename the function to `cacheMainKey()` and set it private as there is currently no reason to call it from outside the service.

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/LaoDetailViewModel.java
GabrielFleischer: It depends.

I will rethink the interface because it might be good to differentiate between an uninitialized Wallet, an error in the generation and a token that is not present as attendees.

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/security/KeyManager.java
GabrielFleischer: I think I will keep `require` and `get`.
Require throws Exceptions when something goes wrong as the PoPToken is required where Get optionally return the token.

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/socialmedia/SocialMediaViewModel.java
GabrielFleischer: I want yo uto merge your PR before applying this change as it might create conflicts. But as you want

GabrielFleischer:
Pending on #755 to use the same error system.

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/security/KeyManager.java
```diff
@@ -22,46 +31,92 @@
...
+  /** @return the deice keypair */
+  public KeyPair getMainKeyPair() {
+    return keyPair;
+  }
+
+  /**
+   * Try to retrieve the user's PoPToken for the given Lao.
+   *
+   * @param lao we want to retrieve the PoP Token from
+   * @return the PoP Token if it was retrieve, empty if no RollCall exist in the LAO
```
Lugiasc, line 89:
```suggestion
   * @return the PoP Token if it was retrieved, empty if no RollCall exist in the LAO
```

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/security/KeyManager.java
```diff
@@ -22,46 +31,92 @@
...
+        lao.getRollCalls().values().stream().max(Comparator.comparing(RollCall::getEnd))
+            .orElseThrow(() -> new NoRollCallException(lao));
+
+    return getValidPoPToken(lao, rollCall);
   }
 
-  public PoPToken getPoPToken(String laoID, String rollCallID) throws GeneralSecurityException {
-    return wallet.findKeyPair(laoID, rollCallID);
+  /**
+   * Try to retrieve the user's PoPToken for the given Lao. It will fail is the user did not attend
```
Lugiasc, line 105:
```suggestion
   * Try to retrieve the user's PoPToken for the given Lao and RollCall. It will fail is the user did not attend
```

Lugiasc (dismissed):

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Wallet.java
```diff
@@ -105,15 +105,16 @@ public PoPToken generatePoPToken(@NonNull String laoID, @NonNull String rollCall
    * @param rollCallTokens a {@link Set} containing the public keys of all attendees present on
    *     roll-call’s results.
    * @return the PoP Token if the user participated in that roll-call or else empty.
```
arnauds5, line 107: the @return comment was not updated (exception instead of empty)

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/security/KeyManager.java
```diff
@@ -23,46 +31,92 @@
...
+            .orElseThrow(() -> new NoRollCallException(lao));
+
+    return getValidPoPToken(lao, rollCall);
   }
 
-  public PoPToken getPoPToken(String laoID, String rollCallID) throws KeyException {
-    return wallet.generatePoPToken(laoID, rollCallID);
+  /**
+   * Try to retrieve the user's PoPToken for the given Lao and RollCall. It will fail is the user did not attend
+   * the roll call of if the token cannot be generated
```
arnauds5, line 106:
```suggestion
   * Try to retrieve the user's PoPToken for the given Lao and RollCall. It will fail if the user did not attend
   * the roll call or if the token cannot be generated
```

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/security/KeyManager.java
```diff
@@ -23,46 +31,92 @@
...
+  /** @return the deice keypair */
+  public KeyPair getMainKeyPair() {
+    return keyPair;
+  }
+
+  /**
+   * Try to retrieve the user's PoPToken for the given Lao.
+   *
+   * @param lao we want to retrieve the PoP Token from
+   * @return the PoP Token if it was retrieved, empty if no RollCall exist in the LAO
```
arnauds5, line 89: the @return comment was not updated (exception instead of empty)

jbsv (approved):

PR #743, opened 2022-01-13
Work be2 tuomas p add reactions fix

tuomas-p:
Add reaction addition and deletion functionalities, fix roll call closing broadcast, add extensive tests for SocialMediaHandler, fix lao id resolution in Channel.scala and use the new function where it is needed


be2-scala/src/test/scala/ch/epfl/pop/model/network/method/message/data/socialMedia/AddChirpSuite.scala
```diff
@@ -7,13 +7,13 @@ import ch.epfl.pop.model.network.method.message.data.{ActionType, ObjectType}
 import spray.json._
 
 class AddChirpSuite extends FunSuite with Matchers {
-    test("Constructor/apply works as intended"){
-        val text: String = "text"
-        val parent_id: Option[String] = None
-        val timestamp = Timestamp(0)
+    private final val text: String = "text"
```
BadrTad, line 10: Test suite would look better if you put `AddChirp message creation in a separate scala object (same file should be fine)

be2-scala/src/test/scala/ch/epfl/pop/model/network/method/message/data/socialMedia/AddReactionSuite.scala
```diff
@@ -0,0 +1,35 @@
+package ch.epfl.pop.model.network.method.message.data.socialMedia
+
+import org.scalatest.{FunSuite, Matchers}
+import ch.epfl.pop.model.objects.Timestamp
+import ch.epfl.pop.json.MessageDataProtocol._
+import ch.epfl.pop.model.network.method.message.data.{ActionType, ObjectType}
+import spray.json._
+
+class AddReactionSuite extends FunSuite with Matchers {
```
BadrTad, line 9: Same remark:
Test suite would look better if you put `AddReaction` messages creation in a separate scala object (same file should be fine)

BadrTad (changes requested):
Great work :+1: just a few remarks.

be2-scala/src/test/scala/util/examples/data/traits/SocialMediaChirpMessagesTrait.scala
```diff
@@ -0,0 +1,19 @@
...
+
+import ch.epfl.pop.model.network.{JsonRpcRequest,MethodType}
+import util.examples.data.builders.HighLevelMessageGenerator
+import util.examples.data.traits.ExampleMessagesTrait
+
+import java.nio.file.Path
+import java.nio.file.Files
+
+/**
+  * Trait to be implemented by RollCallMessages examples
```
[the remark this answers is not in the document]
tuomas-p: yeah forgot to change that!

be2-scala/src/test/scala/util/examples/data/traits/SocialMediaReactionMessagesTrait.scala
```diff
@@ -0,0 +1,19 @@
...
+
+import ch.epfl.pop.model.network.{JsonRpcRequest,MethodType}
+import util.examples.data.builders.HighLevelMessageGenerator
+import util.examples.data.traits.ExampleMessagesTrait
+
+import java.nio.file.Path
+import java.nio.file.Files
+
+/**
+  * Trait to be implemented by RollCallMessages examples
```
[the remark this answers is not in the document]
tuomas-p: yeah

be2-scala/src/test/scala/util/examples/data/SocialMediaMessages.scala
```diff
@@ -0,0 +1,72 @@
+package util.examples.data
+
+import ch.epfl.pop.model.network.method.message.data.{ObjectType, ActionType}
+import ch.epfl.pop.model.network.JsonRpcRequest
+import ch.epfl.pop.model.network.MethodType
+import ch.epfl.pop.model.network.method.ParamsWithMessage
+import ch.epfl.pop.model.network.method.message.Message
+import ch.epfl.pop.model.network.method.message.data.rollCall.CreateRollCall
```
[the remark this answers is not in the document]
tuomas-p: indeed

be2-scala/src/main/scala/ch/epfl/pop/model/objects/Channel.scala
```diff
@@ -20,6 +20,20 @@ final case class Channel(channel: String) {
     case _ => None
   }
 
+  /**
+   * Extract the laoId from a channel name (even though it might be in the middle)
+   *
+   * @return an array of bytes corresponding to the decoded laoId or None if an error occurred
+   */
+  def decodeChannelLaoId: Option[Array[Byte]] = channel match {
```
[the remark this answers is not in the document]
tuomas-p: I believe that it does not really change much in this case, apart from saving us some .getBytes in other parts of the code since it's directly a Byte array, since the return values are mostly used for database accesses anyway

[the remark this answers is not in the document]
tuomas-p: But if you really believe that it would be a relevant change I'll make the modifications in the code

be2-scala/src/test/scala/ch/epfl/pop/model/network/method/message/data/socialMedia/AddChirpSuite.scala
tuomas-p: done, as we discussed

be2-scala/src/test/scala/ch/epfl/pop/model/network/method/message/data/socialMedia/AddReactionSuite.scala
tuomas-p: same as above

be2-scala/src/test/scala/ch/epfl/pop/model/network/method/message/data/socialMedia/AddChirpSuite.scala
```diff
@@ -6,30 +6,30 @@ import ch.epfl.pop.json.MessageDataProtocol._
...
+        ADDCHIRP_MESSAGE.timestamp should equal(TIMESTAMP)
+        ADDCHIRP_MESSAGE._object should equal(ObjectType.CHIRP)
+        ADDCHIRP_MESSAGE.action should equal(ActionType.ADD)
     }
 
     test("json conversions work back and forth"){
-        val text: String = "text"
-        val parent_id: Option[String] = None
-        val timestamp = Timestamp(0)
+        val msg2: AddChirp = AddChirp.buildFromJson(AddChirpExample.ADDCHIRP_MESSAGE.toJson.toString)
```
BadrTad, line 23: Non need for the `AddChirpExample.`

tuomas-p: corrected, thanks

BadrTad (approved):
Looks nice :+1:

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/LaoHandler.scala
```diff
@@ -48,16 +50,24 @@ case object LaoHandler extends MessageHandler {
           case DbActorAck() => {
             val ask: Future[GraphMessage] = (dbActor ? DbActor.Write(channel, message)).map {
               case DbActorWriteAck() => {
-                val socialChannel: Channel = Channel(channel + Channel.SOCIAL_MEDIA_POSTS_PREFIX)
+                val socialChannel: Channel = Channel(channel + Channel.SOCIAL_MEDIA_CHIRPS_PREFIX)
```
pierluca, line 53: the nesting in this function has become crazy and the readability is somewhat bad. 
It seems that we're creating a ton of nested futures and awaiting for them.
How about moving some of the logic to the DbActor so that multiple channels can be created with a single command ? 

something akin to (pseudocode)
```
val ask: Future[GraphMessage] = (dbActor ? DbActor.CreateChannels([
   (laoChan, ObjectType.LAO),
   (socChan, ObjectType.CHIRP),
   (reacChan, ObjectType.REACTION),
]))
```

be2-scala/src/test/scala/ch/epfl/pop/model/objects/ChannelSuite.scala
```diff
@@ -73,8 +73,26 @@ class ChannelSuite extends FunSuite with Matchers {
...
+    val laoId = "mEKXWFCMwb";
+    def channel = Channel(Channel.ROOT_CHANNEL_PREFIX + Base64Data.encode(laoId) + Channel.SEPARATOR + Base64Data.encode("social"))
+    val expected = laoId.getBytes()
 
+    noException shouldBe thrownBy(channel)
+    channel.decodeChannelLaoId.get should equal(expected)
   }
+
+  test("Real LaoId extraction fails with wrong channel structure (not base64Data)") {
+    val laoId = "base64_lao_id"; // Not encoded in BASE64
```
pierluca, line 88: The beauty of fake variables, is that the comment can be the variable 😄 
```suggestion
    val laoId = "not_a_base64_lao_id";
```

be2-scala/src/test/scala/util/examples/data/SocialMediaMessages.scala
```diff
@@ -0,0 +1,70 @@
...
+  final val deleteChirp: JsonRpcRequest =  getJsonRPCRequestFromFile("chirp_delete_publish/chirp_delete_publish.json")()
+
+  //TODO: Generate other messages
+}
+
+/*
+object SocialMediaCustomBuilders {
+  //TODO: Add here custom highlevel message builders to forge data
+}*/
+
```
pierluca, line 70: This seems like altogether dead code, you may want to remove it.

be2-scala/src/main/scala/ch/epfl/pop/model/network/method/message/data/socialMedia/AddReaction.scala
```diff
@@ -0,0 +1,28 @@
...
+import ch.epfl.pop.model.network.Parsable
+import ch.epfl.pop.model.network.method.message.data.ActionType.ActionType
+import ch.epfl.pop.model.network.method.message.data.ObjectType.ObjectType
+import ch.epfl.pop.model.network.method.message.data.{ActionType, MessageData, ObjectType}
+import ch.epfl.pop.model.objects.Timestamp
+import spray.json._
+
+case class AddReaction(
+                          reaction_codepoint: String,
+                          chirp_id: String,
```
pierluca, line 13: This should be a `Hash` , no?

be2-scala/src/main/scala/ch/epfl/pop/model/network/method/message/data/socialMedia/AddReaction.scala
```diff
@@ -0,0 +1,28 @@
...
+                          timestamp: Timestamp
+                        ) extends MessageData {
+  override val _object: ObjectType = ObjectType.REACTION
+  override val action: ActionType = ActionType.ADD
+}
+
+object AddReaction extends Parsable {
+  def apply(
+            reaction_codepoint: String,
+            chirp_id: String,
```
pierluca, line 23: This should be a `Hash` , no?

be2-scala/src/main/scala/ch/epfl/pop/model/network/method/message/data/socialMedia/DeleteReaction.scala
```diff
@@ -0,0 +1,26 @@
...
+                          reaction_id: String,
+                          timestamp: Timestamp
+                        ) extends MessageData {
+  override val _object: ObjectType = ObjectType.REACTION
+  override val action: ActionType = ActionType.DELETE
+}
+
+object DeleteReaction extends Parsable {
+  def apply(
+            reaction_id: String,
```
pierluca, line 21: This should be a `Hash` , no?

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
```diff
@@ -70,62 +72,68 @@ sealed class RollCallHandler(dbRef: => AskableActorRef) extends MessageHandler {
...
+            def createAttendeeChannels(attendees: List[PublicKey], rpcMessage: JsonRpcRequest): GraphMessage = {
+              attendees match {
+                case Nil => Left(rpcMessage)
+                case head::tail =>
+                  //a closeRollCall is always sent in the lao's main channel so we are allowed to do this
+                  val socialChannel: String = generateSocialChannel(rpcMessage.getParamsChannel, head)
+                  val ask: Future[GraphMessage] = (dbActor ? DbActor.CreateChannel(Channel(socialChannel), ObjectType.CHIRP)).map {
+                    case DbActorAck() => createAttendeeChannels(tail, rpcMessage)
+                    //the distinction between the NAck cases this is to prevent errors for now within a LAO for successive Roll Calls with the same participants since the public key should not change, as the channel already exists
+                    case DbActorNAck(code, description) if code == ErrorCodes.INVALID_RESOURCE.id => createAttendeeChannels(tail, rpcMessage)
```
pierluca, line 92: The attendee's pop token should change at every roll call, so I'm it sure you need this 🤔

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
```diff
@@ -70,62 +72,68 @@ sealed class RollCallHandler(dbRef: => AskableActorRef) extends MessageHandler {
...
+      case Left(msg) => {
+        rpcMessage.getParamsMessage match {
+          case Some(message: Message) =>
+            val data: CloseRollCall = message.decodedData.get.asInstanceOf[CloseRollCall]
+
+            //Creates a channel for each attendee (of name /root/lao_id/social/PublicKeyAttendee), returns a GraphMessage
+            def createAttendeeChannels(attendees: List[PublicKey], rpcMessage: JsonRpcRequest): GraphMessage = {
+              attendees match {
+                case Nil => Left(rpcMessage)
+                case head::tail =>
```
pierluca, line 86: you may not need all this complexity if you have, as for the LAO, a way of creating multiple channels at once.
Then the code can become much simpler:  (pseudocode)
```
list = List[PublicKey] -map-> List[Channel] -map-> List[(Channel, ObjectType.CHIRP)]
DbActor.CreateChannels( list ) 
```

be2-scala/src/main/scala/ch/epfl/pop/model/objects/Channel.scala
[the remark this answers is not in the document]
pierluca: I strongly second @BadrTad 's remark.

If I look at the usage of this function, it seems that the use of a `Array[Byte]` has already led to mistakes.
The bytes are being written out directly as a string (which is most likely wrong) , sometimes re-encoded as Base64Data,  etc. 
If you use the `Base64Data` abstraction, you'll probably save yourself from some bugs 😄 

Essentially, it's more or less the difference between a weak-typed and strongly-typed language. You can do the same thing in both, and you can do the same things with `Byte` and `Base64Data`, but the former requires you to be A LOT more careful whereas the latter keeps a convenient abstraction which prevents you from making mistakes.

This is also why we introduced these types, because at the beginning everything was a "string", and you wouldn't notice if swapping a base64 field with, say, a channel. 😉

pierluca (changes requested):
I think, functionality-wise, that there's a lot of good in this PR, but the code really needs quite a bit of clean-up.

Do not hesitate to revisit the APIs of the difference abstractions, if they're forcing you to write very verbose or ugly code. There's maybe something that can be improved elsewhere 😄

be2-scala/src/main/scala/ch/epfl/pop/model/network/method/message/data/socialMedia/DeleteReaction.scala
```diff
@@ -0,0 +1,26 @@
...
+import ch.epfl.pop.json.MessageDataProtocol._
+import ch.epfl.pop.model.network.Parsable
+import ch.epfl.pop.model.network.method.message.data.ActionType.ActionType
+import ch.epfl.pop.model.network.method.message.data.ObjectType.ObjectType
+import ch.epfl.pop.model.network.method.message.data.{ActionType, MessageData, ObjectType}
+import ch.epfl.pop.model.objects.Timestamp
+import spray.json._
+
+case class DeleteReaction(
+                          reaction_id: String,
```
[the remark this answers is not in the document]
tuomas-p: yes indeed

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/SocialMediaHandler.scala
```diff
@@ -16,25 +17,36 @@ import scala.concurrent.{Await, Future}
...
+    * Overrides default DbActor with provided parameter
+    */
+  override final val dbActor: AskableActorRef = dbRef
+
+  override val handler: Flow[GraphMessage, GraphMessage, NotUsed] = Flow[GraphMessage].map {
+  case Left(jsonRpcMessage) => jsonRpcMessage match {
+    case message@(_: JsonRpcRequestAddChirp) => handleAddChirp(message)
+    case message@(_: JsonRpcRequestDeleteChirp) => handleDeleteChirp(message)
+    case message@(_: JsonRpcRequestNotifyAddChirp) => handleAddChirp(message)
+    case message@(_: JsonRpcRequestNotifyDeleteChirp) => handleDeleteChirp(message)
```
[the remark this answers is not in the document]
tuomas-p: This is of course wrong, I put the wrong handler, we are not supposed to receive this message and therefore will propagate an error if it happens using the handleNotifyAddChirp and handleNotifyDeleteChirp functions

be2-scala/src/test/scala/ch/epfl/pop/model/objects/ChannelSuite.scala
tuomas-p: fair enough thank you!

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
tuomas-p: Yeah, this was originally left here in the case the front-end was using the PublicKeys and not the tokens yet (before the in-meeting discussion) but now that it was cleared it can be removed

be2-scala/src/main/scala/ch/epfl/pop/model/network/method/message/data/socialMedia/AddReaction.scala
tuomas-p: yes indeed

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/SocialMediaHandler.scala
```diff
@@ -98,9 +110,9 @@ case object SocialMediaHandler extends MessageHandler {
     Await.result(ask, duration) match {
       case Left(msg) => {
         val channelChirp: Channel = rpcMessage.getParamsChannel
-        channelChirp.decodeSubChannel match {
+        channelChirp.decodeChannelLaoId match {
           case(Some(lao_id)) => {
-            val broadcastChannel: Channel = Channel(Channel.ROOT_CHANNEL_PREFIX + Base64Data.encode(lao_id) + Channel.SOCIAL_MEDIA_POSTS_PREFIX)
+            val broadcastChannel: Channel = Channel(Channel.ROOT_CHANNEL_PREFIX + Base64Data.encode(lao_id) + Channel.SOCIAL_MEDIA_CHIRPS_PREFIX)
```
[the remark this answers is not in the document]
tuomas-p: yes, I did that

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/LaoHandler.scala
tuomas-p: this is a good reason for refactoring, adding this

be2-scala/src/main/scala/ch/epfl/pop/model/objects/Channel.scala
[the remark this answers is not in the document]
tuomas-p: Added this!

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
tuomas-p: it becomes easier now indeed

be2-scala/src/main/scala/ch/epfl/pop/model/objects/Channel.scala
```diff
@@ -11,10 +11,10 @@ final case class Channel(channel: String) {
    *
    * @return an array of bytes corresponding to the decoded sub-channel name or None if an error occurred
```
pierluca, line 12: the comments here and below might need updating

be2-scala/src/test/scala/ch/epfl/pop/model/objects/ChannelSuite.scala
```diff
@@ -61,15 +61,15 @@ class ChannelSuite extends FunSuite with Matchers {
...
+    val expected = Base64Data.encode(laoId)
     noException shouldBe thrownBy(channel)
     channel.decodeSubChannel.get should equal(expected)
   }
 
   test("Real LaoId extraction channel test") {
     val laoId = "mEKXWFCMwb";
     def channel = Channel(Channel.ROOT_CHANNEL_PREFIX + Base64Data.encode(laoId))
-    val expected = laoId.getBytes()
+    val expected = Base64Data.encode(laoId)
```
pierluca, line 72: careful, here you're taking the string `mEKXWFCMwb` and transforming it into base64.
You're not _interpreting_ that string as Base64Data. Is that what you want ? 

Alternatively, you should just be able to do this:
```suggestion
    val laoId = "mEKXWFCMwb";
    def channel = Channel(Channel.ROOT_CHANNEL_PREFIX + Base64Data(laoId))
    val expected = Base64Data(laoId)
```

be2-scala/src/test/scala/ch/epfl/pop/model/objects/ChannelSuite.scala
```diff
@@ -61,15 +61,15 @@ class ChannelSuite extends FunSuite with Matchers {
   test("LaoId extraction channel test") {
     val laoId = "base64_lao_id";
     def channel = Channel(Channel.ROOT_CHANNEL_PREFIX + Base64Data.encode(laoId))
-    val expected = laoId.getBytes()
+    val expected = Base64Data.encode(laoId)
```
pierluca, line 64: what's the difference between this test and `Real LaoId extraction fails with wrong channel structure (not base64Data)` ?

be2-scala/src/main/scala/ch/epfl/pop/model/objects/Channel.scala
```diff
@@ -11,10 +11,10 @@ final case class Channel(channel: String) {
...
    */
-  def decodeSubChannel: Option[Array[Byte]] = channel match {
+  def decodeSubChannel: Option[Base64Data] = channel match {
     case _ if channel.startsWith(Channel.ROOT_CHANNEL_PREFIX) =>
       Try(Base64.getUrlDecoder.decode(channel.substring(Channel.ROOT_CHANNEL_PREFIX.length).getBytes)) match {
-        case Success(value) => Some(value)
+        case Success(value) => Some(Base64Data.encode(value))
```
pierluca, line 17: are you sure this is not just the following ? 
```suggestion
        case Success(value) => Some(Base64Data(value))
```

Also, this function only works for a channel in the format `/root/myLaoId`, right ?
If I understand correctly, `/root/myLaoId/social` will most likely fail.
How is then not the same as `decodeChannelLaoId` ?

be2-scala/src/main/scala/ch/epfl/pop/model/objects/Channel.scala
```diff
@@ -25,10 +25,10 @@ final case class Channel(channel: String) {
    *
    * @return an array of bytes corresponding to the decoded laoId or None if an error occurred
    */
-  def decodeChannelLaoId: Option[Array[Byte]] = channel match {
+  def decodeChannelLaoId: Option[Base64Data] = channel match {
     case _ if channel.startsWith(Channel.ROOT_CHANNEL_PREFIX) =>
       Try(Base64.getUrlDecoder.decode(channel.substring(Channel.ROOT_CHANNEL_PREFIX.length).split(Channel.SEPARATOR)(0).getBytes)) match {
-        case Success(value) => Some(value)
+        case Success(value) => Some(Base64Data.encode(value))
```
pierluca, line 31: I believe the LaoID you're getting here as "value" is already the base64-representation of the channel.
Why do you need to encode it (i.e. get the base64 representation of that string), as opposed to just doing the following ? 
```suggestion
        case Success(value) => Some(Base64Data(value))
```

pierluca (commented):
This looks much much better !

I may be mistaken, but I think there might still be a few issues with the usage of Base64Data though.

be2-scala/src/main/scala/ch/epfl/pop/model/objects/Channel.scala
tuomas-p: thank you, I had missed that

be2-scala/src/test/scala/ch/epfl/pop/model/objects/ChannelSuite.scala
tuomas-p: Here we do not forget to encode it

be2-scala/src/main/scala/ch/epfl/pop/model/objects/Channel.scala
tuomas-p: indeed, however, this function could be used for subchannel extraction if necessary, but this is why I added the decodeChannelLaoId function.

be2-scala/src/main/scala/ch/epfl/pop/model/objects/Channel.scala
tuomas-p: It is a mistake, I fixed it, thank you!

tuomas-p: However, I had to modify the Try code above for that

be2-scala/src/main/scala/ch/epfl/pop/model/objects/Channel.scala
tuomas-p: The subchannel extraction will indeed fail, as I looked into it in detail. This function is therefore not useful and will be removed.

pierluca (approved):
Seems much better, good job ! 

Feel free to merge if it's blocking other work, or if you have time, do have a look at Sonar's "major" code smells.

pierluca (approved):

PR #748, opened 2022-01-16
add ElectionSetupFragment tests

arnauds5:
- Add tests for the election setup
- Disable the election submit button if there is a duplicate ballot (avoid crash if it's the case)
- Fix usage of Lao constructor (it was using channel instead of id)
- Add the Lao channel correctly in constructor (it was using this.channel = id instead of this.channel = "/root/"+id)

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Lao.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Lao.java
@@ -42,7 +42,7 @@ public final class Lao {
     } else if (id.isEmpty()) {
       throw new IllegalArgumentException(" The id of the Lao is empty");
     }
-    this.channel = id;
+    this.channel = "/root/" + id;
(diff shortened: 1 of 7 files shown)


fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Lao.java
```diff
@@ -42,7 +42,7 @@ public Lao(String id) {
     } else if (id.isEmpty()) {
       throw new IllegalArgumentException(" The id of the Lao is empty");
     }
-    this.channel = id;
+    this.channel = "/root/" + id;
```
GabrielFleischer, line 45: Wait, how could this go unseen for so long ?

GabrielFleischer (approved):
Great job !

Lugiasc (approved):

jbsv (approved):
LGTM

PR #753, opened 2022-01-17
Fix crash cause of undefined public key

Xelowak:
The public key of the current user was not passed to chirp cards in `SocialUserProfile`. It happened because Jiabao and I merged our work and I didn't realize she added a prop to `ChirpCard`.


jiabaow (approved):
LGTM

pierluca (approved):
This seems a bit hacky, but I'll approve it to fix the crash

PR #754, opened 2022-01-17
show publish time even for a deleted chirp

jiabaow:
when displaying a deleted chirp, keep the time when it's been published instead of the delete time.


Xelowak (approved):
LGTM

pierluca (approved):

PR #763, opened 2022-01-19
Acceptor signature in consensus

Anghul:


pierluca:
@Anghul there seems to be a conflict here :)

nkcr (approved):

PR #765, opened 2022-01-20
Add JSON Schema for the json contained in the connect QRCode

GabrielFleischer:
This PR aims to add a new side of the protocol : the QR Code.

The goal is to use Json strings in the QR Code to convoy data.
This PR aims to define the data contained in the QR Code used to connect to an LAO.

I am not so sure about the URL regex pattern, if anyone has more knowledge than me on that point, I would appreciate any review.


protocol/qrcode/connect_to_lao.json
```diff
@@ -0,0 +1,22 @@
...
+    "$id": "https://raw.githubusercontent.com/dedis/student_21_pop/master/protocol/jsonRPC.json",
+    "title": "QRCode data",
+    "description": "Match a data contained in a qrcode used to connect to an LAO.",
+    "type": "object",
+    "additionalProperties": false,
+    "properties": {
+        "server": {
+            "type": "string",
+            "format": "uri-reference",
+            "pattern": "(http|ws)s?://.*",
```
pierluca, line 12: to ensure this pattern happens only at the beginning of the string:

```suggestion
            "pattern": "^(http|ws)s?://.*",
```

pierluca (approved):
Minor improvement suggested

PR #768, opened 2022-01-22
Add the server URL directly in the LAO QRCode

GabrielFleischer:
Add the server URL directly in the LAO QRCode for Android

Generate the right QRCode and retrieve the data correctly following the Json-Schema.

The new code isn't tested, as the current implementation is almost impossible to test.
It would need camera mocking, which is very complex to do inside a CI.

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/qrcode/ConnectToLao.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/qrcode/ConnectToLao.java
@@ -9,7 +9,7 @@ public class ConnectToLao {
   public final String server;
   public final String lao;
 
-  private ConnectToLao(String server, String lao) {
+  public ConnectToLao(String server, String lao) {
     this.server = server;
     this.lao = lao;
   }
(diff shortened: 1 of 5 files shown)


fe2-android/app/src/main/java/com/github/dedis/popstellar/model/qrcode/ConnectToLao.java
```diff
@@ -0,0 +1,21 @@
...
+
+  public final String server;
+  public final String lao;
+
+  public ConnectToLao(String server, String lao) {
+    this.server = server;
+    this.lao = lao;
+  }
+
+  public static ConnectToLao extractFrom(Gson mGson, Barcode barcode) {
```
pierluca, line 17: do we need the `Barcode` dependency here, or could we just pass in a string containing `barcode.rawValue` ? 
it seems like we could avoid coupling `ConnectToLao` with `Barcode`

pierluca (changes requested):
@GabrielFleischer "new code isn't tested" = no automatic tests, or not tested even locally? 😅

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/qrcode/ConnectToLao.java
GabrielFleischer: Yes, but it could also be changed in `onQRCodeDetected`. I will change that.

And I tested it locally from android to android. I was just not able to to it under automated tests.

pierluca (approved):

PR #771, opened 2022-01-25
rename connect_to_lao $id

jiabaow:
rename the $id of connect_to_lao to avoid having the same $id as jsonRPC

### DIFF
--- a/protocol/qrcode/connect_to_lao.json
+++ b/protocol/qrcode/connect_to_lao.json
@@ -1,6 +1,6 @@
 {
     "$schema": "http://json-schema.org/draft-07/schema#",
-    "$id": "https://raw.githubusercontent.com/dedis/student_21_pop/master/protocol/jsonRPC.json",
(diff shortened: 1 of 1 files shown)


protocol/qrcode/connect_to_lao.json
```diff
@@ -1,6 +1,6 @@
 {
     "$schema": "http://json-schema.org/draft-07/schema#",
-    "$id": "https://raw.githubusercontent.com/dedis/student_21_pop/master/protocol/jsonRPC.json",
+    "$id": "https://raw.githubusercontent.com/dedis/student_21_pop/master/protocol/connect_to_lao.json",
```
GabrielFleischer, line 3: Ok, I now see what this is.

It should be the path to the file which is actaully

```suggestion
    "$id": "https://raw.githubusercontent.com/dedis/student_21_pop/master/protocol/qrcode/connect_to_lao.json",
```

GabrielFleischer (changes requested):

pierluca (approved):

PR #772, opened 2022-01-27
add files to index

jiabaow:


Xelowak (approved):
Nice ! Are you sure these were the only places you could do this though ?

jiabaow:
> Nice ! Are you sure these were the only places you could do this though ?

Yes, they seem to be the only places I can do it in this way. It doesn't work the same for files in /components and /parts.

pierluca (approved):

PR #776, opened 2022-01-28
add toast message for console.error

jiabaow:


Xelowak (approved):
Nice job ! I think it's smart to keep console.errors along with toasts, so that the user is informed and it is easier to debug.

pierluca (approved):
Approving blindly

PR #777, opened 2022-01-28
Remove code smells

Xelowak:
I tried to remove as much code smells as I could.

Unfortunately I don't think there's a way to know how may of them I've removed before this is merged.


jiabaow (approved):
LGTM
Thanks for all the changes!

pierluca (approved):
Approved more or less blindly :)

PR #780, opened 2022-02-02
Remove warning when QRcode shows on screen

Xelowak:
When a QR code showed on screen, a warning appeared:
Capture d’écran 2022-02-01 à 16 12 48

Actually, the package we use for QR codes is not regularly updated. [This package](https://www.npmjs.com/package/react-qr-code) is better maintained, so I replaced our old library by this one and the warning is now gone. ([source](https://github.com/zpao/qrcode.react/issues/134)).

I also updated some documentation in navigation.


jiabaow (approved):
LGTM

pierluca (approved):
Great initiative. Thanks !

PR #788, opened 2022-02-04
Add tests for the LaoDetail Activity

GabrielFleischer:
Add some tests to the LAO Activity.

This also shows how to use the BundleBuilder in a test rule.

### DIFF
--- a/fe2-android/app/src/androidTest/java/com/github/dedis/popstellar/ui/detail/LaoDetailActivityTest.java
+++ b/fe2-android/app/src/androidTest/java/com/github/dedis/popstellar/ui/detail/LaoDetailActivityTest.java
@@ -27,6 +27,7 @@ import android.graphics.drawable.Drawable;
 import android.view.View;
 import android.widget.ImageView;
 
+import androidx.test.espresso.matcher.BoundedMatcher;
 import androidx.test.ext.junit.rules.ActivityScenarioRule;
 
 import com.github.dedis.popstellar.model.objects.Lao;
@@ -49,7 +50,6 @@ import com.google.zxing.qrcode.QRCodeReader;
 
 import org.hamcrest.Description;
 import org.hamcrest.Matcher;
-import org.hamcrest.TypeSafeMatcher;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExternalResource;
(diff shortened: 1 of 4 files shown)


fe2-android/app/src/androidTest/java/com/github/dedis/popstellar/ui/detail/LaoDetailActivityTest.java
```diff
@@ -0,0 +1,182 @@
...
+
+          if (drawable instanceof BitmapDrawable) {
+            Bitmap qrcode = ((BitmapDrawable) drawable).getBitmap();
+            // Convert the QRCode to something zxing understands
+            int[] buffer = new int[qrcode.getWidth() * qrcode.getHeight()];
+            qrcode.getPixels(
+                buffer, 0, qrcode.getWidth(), 0, 0, qrcode.getWidth(), qrcode.getHeight());
+            LuminanceSource source =
+                new RGBLuminanceSource(qrcode.getWidth(), qrcode.getHeight(), buffer);
+            BinaryBitmap bitmap = new BinaryBitmap(new HybridBinarizer(source));
```
pierluca, line 164: this looks like a good candidate for a separate conversion function.
you could also just throw an exception if the drawable is not of the right type.

fe2-android/app/src/androidTest/java/com/github/dedis/popstellar/ui/detail/LaoDetailActivityTest.java
```diff
@@ -0,0 +1,182 @@
...
+            try {
+              // Parse the bitmap and check it against expected value
+              return expectedContent.equals(new QRCodeReader().decode(bitmap).getText());
+            } catch (NotFoundException | ChecksumException | FormatException e) {
+              throw new IllegalArgumentException("The provided image is not a valid QRCode", e);
+            }
+          }
+        }
+        return false;
```
pierluca, line 173: for my own understanding, what's the semantic difference between throwing and returning false, in a `TypeSafeMatcher` ?

fe2-android/app/src/androidTest/java/com/github/dedis/popstellar/ui/detail/LaoDetailActivityTest.java
```diff
@@ -0,0 +1,182 @@
...
+    witnessButton().perform(click());
+    fragmentContainer().check(matches(withChild(withId(witnessFragmentId()))));
+  }
+
+  // Matches an ImageView containing a QRCode with expected content
+  private Matcher<? super View> withQrCode(String expectedContent) {
+    return new TypeSafeMatcher<View>() {
+      @Override
+      protected boolean matchesSafely(View item) {
+        if (item instanceof ImageView) {
```
pierluca, line 152: could we do the following ? 
```suggestion
    return new TypeSafeMatcher<ImageView>() {
      @Override
      protected boolean matchesSafely(ImageView item) {
```

fe2-android/app/src/androidTest/java/com/github/dedis/popstellar/ui/detail/LaoDetailActivityTest.java
```diff
@@ -0,0 +1,182 @@
...
+
+  @Test
+  public void witnessButtonShowsWitnessTab() {
+    witnessButton().perform(click());
+    fragmentContainer().check(matches(withChild(withId(witnessFragmentId()))));
+  }
+
+  // Matches an ImageView containing a QRCode with expected content
+  private Matcher<? super View> withQrCode(String expectedContent) {
+    return new TypeSafeMatcher<View>() {
```
pierluca, line 149: it seems that the `TypeSafeMatcher` encourages to also override `describeMismatchSafely`.
Would it be helpful to explain what failed ? 
There seems to be a number of cases that return `false` with no further information.

fe2-android/app/src/androidTest/java/com/github/dedis/popstellar/ui/detail/LaoDetailActivityTest.java
```diff
@@ -0,0 +1,182 @@
...
+      new ExternalResource() {
+        @Override
+        protected void before() {
+          hiltAndroidRule.inject();
+
+          when(laoRepository.getLaoObservable(anyString()))
+              .thenReturn(BehaviorSubject.createDefault(LAO));
+        }
+      };
+  // Activity scenario rule that starts the activity.
```
pierluca, line 93:
```suggestion
      };
      
  // Activity scenario rule that starts the activity.
```

pierluca (commented):
Nice work on testing QR codes !
There are a couple of things that make me (slighlty) uncomfortable here, is it realistic to fix them?

fe2-android/app/src/androidTest/java/com/github/dedis/popstellar/ui/detail/LaoDetailActivityTest.java
GabrielFleischer: This is not possible because TypeSafeMatcher<ImageView> is not a Matcher<? super View>
But I found another way :  BoundedMatcher.

fe2-android/app/src/androidTest/java/com/github/dedis/popstellar/ui/detail/LaoDetailActivityTest.java
GabrielFleischer: Yes, it's better

fe2-android/app/src/androidTest/java/com/github/dedis/popstellar/ui/detail/LaoDetailActivityTest.java
GabrielFleischer: From my understanding, a boolean is returned if the data could be extracted. 
An exception is thrown when the data cannot be retrieved.

I made some changes to make it clearer.

fe2-android/app/src/androidTest/java/com/github/dedis/popstellar/ui/detail/LaoDetailActivityTest.java
GabrielFleischer: The error should now be more pertinent.

pierluca (approved):
So much cleaner 👍

PR #790, opened 2022-02-07
Changed Android application id

pierluca:
I've updated all the documentation, but this seems to impact the Android release.
@GabrielFleischer is it safe to make this change ?


GabrielFleischer (approved):
I don't think this should be changed anywhere else.
The release contains the ID in its metadata, but it isn't a configuration.

We could generate a new release too.

PR #793, opened 2022-02-08
Initial code cleanup

RaulinN:
**A lot** of small changes that should have been done at the end of the last semester... Includes:
- unused imports removal / optimization
- application of the coding convention w/ "Code Cleanup..." tool
- typos / unused code / ... removal
- manual compiler warnings removal
- make the program work with java 8


RaulinN:
@pierluca For the first time in the history of be2, the project compilation is **entirely** warning-less. Do you think we should activate a compiler flag transforming warnings into errors like when one programs in C or leave it as is?

pierluca:
@RaulinN Yes please ! Do turn any warning into an error.

pierluca:
Overalls, looks good. I've reviewed by commits to make it manageable.
It all seems pretty fine, and I'd argue we should probably merge this ASAP.
Maybe for a subsequent PR, do we have no way of avoiding `@unchecked`s ?

be2-scala/build.sbt
```diff
@@ -32,10 +36,10 @@ copyProtocolTask := {
     }else{
         val source = new File("../protocol")
         val dest   = new File("./src/main/resources/protocol")
-        Try(IO.copyDirectory(source,dest, true)) match {
+        Try(IO.copyDirectory(source,dest, overwrite = true)) match {
```
pierluca, line 39:
```suggestion
        Try(IO.copyDirectory(source, dest, overwrite = true)) match {
```
shouldn't the spacing be something like this?

be2-scala/src/test/scala/ch/epfl/pop/json/MessageDataProtocolSuite.scala
```diff
@@ -15,16 +15,16 @@ class MessageDataProtocolSuite extends FunSuite with Matchers {
     def shouldEqualTo(o: CastVoteElection): Unit = {
 
       @scala.annotation.tailrec
-      def checkVoteElectionLists(l1: List[VoteElection], l2: List[VoteElection]) {
-        l1.length should equal (l2.length)
+      def checkVoteElectionLists(l1: List[VoteElection], l2: List[VoteElection]): Unit = {
+        l1.length should equal(l2.length)
 
-        (l1, l2) match {
+        ((l1, l2): @unchecked) match {
```
pierluca, line 21: Do we need an `@unchecked` here? Is there no better way?

pierluca (commented):
Given the massive impact, I'd argue for a quick merge. Let me know how you want to go about it.

be2-scala/src/test/scala/ch/epfl/pop/json/MessageDataProtocolSuite.scala
RaulinN: The unchecked is to avoid the "non-exhaustive match warning" when I know the compiler is wrong. We can definitely modify it but the code will be very slightly less efficient/clean. However I think you are definitely right (showing a clean implementation to students > gaining a few clock cycles in the tests ^^)! I'll change it

be2-scala/build.sbt
RaulinN: Strange that it was not modified by the automatic linter... 🤔. I couldn't find any other examples of such commas. There might be some, but probably insanely scattered/rare

RaulinN:
> @RaulinN Yes please ! Do turn any warning into an error.

Done! I also added a list of recommended Scala compiler flags. The code passes all new compiler checks as well :)

RaulinN:
> Overalls, looks good. I've reviewed by commits to make it manageable. It all seems pretty fine, and I'd argue we should probably merge this ASAP. Maybe for a subsequent PR, do we have no way of avoiding `@unchecked`s ?

Removed the unchecked. It was the only occurence in the entire codebase. In this case, the compiler could not infer that he was wrong and that the match was indeed exhaustive. I modified it slightly to calm the compiler

be2-scala/build.sbt
RaulinN: EDIT: turns out IntelliJ didnt change anything because build.sbt is not a scala file

PR #794, opened 2022-02-08
Update build file to make it faster and cleaner

GabrielFleischer:
Redo the build file to fix some issues and reorganize the dependencies

### DIFF
--- a/fe2-android/app/build.gradle
+++ b/fe2-android/app/build.gradle
@@ -16,15 +16,30 @@ android {
 
         testInstrumentationRunner "com.github.dedis.popstellar.HiltTestRunner"
 
+        // The following feature is not working properly with the coverage on the current build version
+        // As it's purpose is to clear physical data between each tests, its is not used yet.
+        // Let's hope the feature will be fixed before we actually need it.
+
+        // The following argument makes the Android Test Orchestrator run its
+        // "pm clear" command after each test invocation. This command ensures
+        // that the app's state is completely cleared between tests.
+        // testInstrumentationRunnerArguments clearPackageData: 'true'
+
+        buildTypes {
+            debug {
+                isDefault true
+                testCoverageEnabled true
+            }
+
+            release {
+                minifyEnabled false
+                proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
+            }
+        }
+
         compileOptions {
             sourceCompatibility JavaVersion.VERSION_1_8
             targetCompatibility JavaVersion.VERSION_1_8
-        }
-
-        testOptions {
-            unitTests.returnDefaultValues = true
-            // Improve tests performances by disabling fancy animations
-            animationsDisabled = true
         }
 
         javaCompileOptions {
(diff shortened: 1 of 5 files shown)


fe2-android/app/build.gradle
```diff
@@ -16,179 +16,147 @@ android {
 
         testInstrumentationRunner "com.github.dedis.popstellar.HiltTestRunner"
 
+        // The following feature is not working properly with the coverage on the current build version
```
pierluca, line 19: which following feature?

fe2-android/app/build.gradle
```diff
@@ -16,179 +16,147 @@ android {
...
+        // As it's purpose is to clear physical data between each tests, its is not used yet.
```
pierluca, line 20: `its is`?

pierluca (commented):
A lot of this feels like magic to me, as I'm very ignorant on these topics, but I think
1) the documentation improvements are great !
2) more documentation and references to external docs would be good :)

fe2-android/app/build.gradle
GabrielFleischer: the 'clearPackage: true'

GabrielFleischer: I don't really see how to improve the comments, but I want a comment to explain its purpose and another to explain why it is disabled

fe2-android/app/build.gradle
jbsv, line 19:
```suggestion
```

fe2-android/app/build.gradle
jbsv, line 20:
```suggestion
        // The purpose of 'clearPackageData: true' is to clear physical data between each tests, but it is not used yet, because it is not working properly with the coverage on the current build version.
```

jbsv (changes requested):
I was able to start the app in debug mode on my PC. So Just tidy up a bit the comments please, and we're good to go.

fe2-android/app/build.gradle
pierluca: I should have been clearer. `its is` looks like a typo 😄 
```suggestion
        // Its purpose is to clear physical data between each tests, but it is not used yet.
```

fe2-android/app/build.gradle
```diff
@@ -16,179 +16,147 @@ android {
...
+        // Let's hope the feature will be fixed before we actually need it.
+
```
pierluca, line 22: I would add an empty comment here, to signal that it's a continuous comment.
```suggestion
        // 
```

fe2-android/app/build.gradle
pierluca: I think this will be clearer with the changes below.
I would also reference https://issuetracker.google.com/issues/123987001 explicitly in the comment, so that we can easily check on the status of the resolution.

They seem to be indicating a sort of temporary fix there, by the way, is it worth looking into?

fe2-android/app/build.gradle
pierluca: And indeed the comment is clear.

pierluca, line 20: On a side note, isn't the Android crypto key manager stuff considered "physical data" ?

GabrielFleischer: I am not sure, but I don't think so as reinstalling the app does not seem to affect the device public key.

jbsv (approved):

PR #795, opened 2022-02-09
Message registry for handlers

Xelowak:
I added a new way to handle incoming messages. Instead of using a lot of switch/case, I created a MessageRegistry that contains all types of messages and their corresponding handling function.
I also renamed all handlers to <purpose>Handler to prevent having same file names.


fe1-web/model/network/method/message/data/MessageRegistry.ts
```diff
@@ -0,0 +1,35 @@
...
+const k = (object: ObjectType, action: ActionType): string => `${object}, ${action}`;
+
+/**
+ * Registry of message data classes and their corresponding properties.
+ */
+export class MessageRegistry {
+  private readonly mapping: Map<string, MessageEntry> = new Map<string, MessageEntry>();
+
+  addHandler(object: ObjectType, action: ActionType, handleFunc: HandleFunction) {
+    this.mapping.set(k(object, action), { ingestion: handleFunc });
```
pierluca, line 23: as the `MessageEntry` interface grows, we may need to just update that specific field, i.e. do a `this.mapping.get` first and update the entry.

fe1-web/model/network/method/message/data/MessageRegistry.ts
```diff
@@ -0,0 +1,35 @@
+import { ActionType, ObjectType } from './MessageData';
+import { ExtendedMessage } from '../ExtendedMessage';
+
+type HandleFunction = (msg: ExtendedMessage) => boolean;
+
+/**
+ * Represents an entry of the MessageRegistry.
+ */
+interface MessageEntry {
+  ingestion: HandleFunction; // Function to handle this type of message
```
pierluca, line 10: I'm not a big fan of the name `ingestion` here, but I don't have a better idea right now 😅

fe1-web/ingestion/Handler.ts
```diff
@@ -6,7 +6,7 @@ import { Channel } from 'model/objects';
 import {
   addMessages, dispatch, OpenedLaoStore,
 } from 'store';
-import { handleLaoMessage } from './handlers';
+import { handleMessage } from './handlers/Default';
```
pierluca, line 9: In a future step, we can probably delete the whole `./handlers/Default` file 😄

fe1-web/ingestion/Handler.ts
```diff
@@ -19,7 +19,7 @@ function handleLaoCreateMessages(msg: ExtendedMessage) : boolean {
   // processing the lao/create message:
   // - we either connect to the LAO (if there's no active connection) OR
   // - we simply add it to the list of known LAOs
-  handleLaoMessage(msg);
+  handleMessage(msg);
```
pierluca, line 22: ... and just call `messageRegistry.handleMessage` from here.
We would need the `messageRegistry` to be a variable we can set from outside this module, though

fe1-web/ingestion/handlers/Default.ts
```diff
@@ -1,12 +1,11 @@
...
-import { handleRollCallMessage } from './RollCall';
-import { handleElectionMessage } from './Election';
-import { handleChirpMessage } from './ChirpHandler';
-import { handleReactionMessage } from './ReactionHandler';
+import { MessageRegistry } from 'model/network/method/message/data';
+import { configure } from './index';
+
+// Create a registry and configure it
+const registry = new MessageRegistry();
+configure(registry);
```
pierluca, line 7: later, I believe we'll be able to just shift the configuration higher in 
https://github.com/dedis/popstellar/blob/master/fe1-web/ingestion/Configure.ts

fe1-web/ingestion/handlers/Default.ts
```diff
@@ -1,12 +1,11 @@
...
-import { handleMeetingMessage } from './Meeting';
-import { handleRollCallMessage } from './RollCall';
-import { handleElectionMessage } from './Election';
-import { handleChirpMessage } from './ChirpHandler';
-import { handleReactionMessage } from './ReactionHandler';
+import { MessageRegistry } from 'model/network/method/message/data';
+import { configure } from './index';
+
+// Create a registry and configure it
+const registry = new MessageRegistry();
```
pierluca, line 6: later, I believe the creation could end up being part of the setup here:
https://github.com/dedis/popstellar/blob/master/fe1-web/App.tsx

pierluca (commented):
Good progress 👍 keep going !
Just a few comments mostly concerning future evolutions

fe1-web/model/network/method/message/data/MessageRegistry.ts
Xelowak: That's true, I realized it while trying to add signatures. I'll change it now.

fe1-web/ingestion/Configure.ts
```diff
@@ -1,12 +1,17 @@
...
 import { handleRpcRequests } from './Handler';
 import { makeMessageStoreWatcher } from './Watcher';
+import { configure } from './handlers';
 
 export function configureIngestion() {
   // setup the handler for incoming messages
   getNetworkManager().setRpcHandler(handleRpcRequests);
 
+  // configure the message registry
+  configure(messageRegistry);
```
pierluca, line 13:
```suggestion
export function configureIngestion(messageRegistry: MessageRegistry) {
  // configure the message handlers
  configure(messageRegistry);

  // setup the handler for incoming messages
  getNetworkManager().setRpcHandler(handleRpcRequests);
```

This way, we don't accept messages from the network _before_ we have configured the handlers.
It's a very theoretical thing, but it's good to keep it clean :-)

fe1-web/ingestion/Watcher.ts
```diff
@@ -3,7 +3,7 @@ import {
   getLaoMessagesState, getLaosState, processMessages,
 } from 'store';
 import { ExtendedMessage } from 'model/network/method/message';
-import { handleMessage } from './handlers/Default';
+import { messageRegistry } from 'App';
 
 export function makeMessageStoreWatcher(store: Store) {
```
pierluca, line 8: this will, unfortunately, create an import loop again, as `App` is the entrypoint and therefore imports (indirectly) the whole application.

We may need to have something like

```suggestion
export function makeMessageStoreWatcher(store: Store, messageRegistry: MessageRegistry) {
```

fe1-web/ingestion/Handler.ts
```diff
@@ -6,7 +6,7 @@ import { Channel } from 'model/objects';
 import {
   addMessages, dispatch, OpenedLaoStore,
 } from 'store';
-import { handleMessage } from './handlers/Default';
+import { messageRegistry } from 'App';
```
pierluca, line 9: this will, unfortunately, create an import loop again, as App is the entrypoint and therefore imports (indirectly) the whole application.

we may be able to avoid the import here as well, by applying the good old principle of dependency injection :-)
Something along the lines of : 

```suggestion
let messageRegistry: MessageRegistry;

export function setMessageRegistry(r: MessageRegistry) {
  messageRegistry = r;
}
```

Then `ingestion/Configure.ts` can simply use the setter, no `import` involved 😄

fe1-web/model/network/method/message/data/MessageRegistry.ts
```diff
@@ -7,29 +7,82 @@ type HandleFunction = (msg: ExtendedMessage) => boolean;
...
 
 // Generates a string key of the form "object, action" for the MessageRegistry mapping
 const k = (object: ObjectType, action: ActionType): string => `${object}, ${action}`;
 
 /**
- * Registry of message data classes and their corresponding properties.
+ * Registry of message data classes and their corresponding properties. By default, it already
+ * contains all types of MessageData as keys.
  */
 export class MessageRegistry {
```
pierluca, line 20: Looks pretty good !

pierluca (commented):
Sorry for being annoying on a few things, but I hope my comments will make sense.
The MessageRegistry is looking pretty good at this point.
Next step would be extending the Registry to replace the Builder.ts :)

fe1-web/ingestion/Configure.ts
Xelowak: It's a nice remark, I didn't think about it. Thanks !

pierluca (approved):
Good job so far !

PR #798, opened 2022-02-10
Try new Ci setup for android

GabrielFleischer:


.github/workflows/ci.yaml
```diff
@@ -163,60 +163,78 @@ jobs:
...
+          distribution: 'zulu'
+          java-version: '11'  
 
-    - name: Build
-      shell: bash
-      run: |
-        # use grep to remove log spam due to jetifier transformation, c.f. https://stackoverflow.com/a/52353446
-        # use set -o pipefail to keep error code from gradle, c.f. https://stackoverflow.com/a/37128767
-        ( set -o pipefail; ./gradlew build --parallel --build-cache --info | grep -vE "^Transforming (.*) with (.*)Transform" )
+      - name: Gradle cache
```
pierluca, line 175: I'd still try to keep verbs rather than nouns here (and below), e.g. `Cache X packages` or `Cache AVD` rather than `X cache`

.github/workflows/ci.yaml
```diff
@@ -163,60 +163,78 @@ jobs:
         working-directory: ./fe2-android
 
     steps:
-    - name: Setup repo
-      uses: actions/checkout@v2
-      with:
-        submodules: recursive
-        fetch-depth: 0  # Shallow clones should be disabled for a better relevancy of Sonar analysis
```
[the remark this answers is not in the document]
GabrielFleischer: I applied those changes

pierluca (approved):

pierluca (changes requested):
Performance seems worse than before, and CI fails. I'd missed that 😅

.github/workflows/ci.yaml
```diff
@@ -162,61 +162,83 @@ jobs:
...
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        SONAR_TOKEN: ${{ secrets.SONAR_TOKEN_FE2 }}
-      run: |
-        ./gradlew sonarqube
+      - name: Run Tests
+        uses: reactivecircus/android-emulator-runner@v2
+        with:
+          api-level: 29
+          force-avd-creation: false
+          emulator-options: -no-snapshot-save -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
```
jbsv, line 230: "no window" but "gpu" seems weird.

.github/workflows/ci.yaml
```diff
@@ -162,61 +162,83 @@ jobs:
...
+          chmod +x ./gradlew
+
+      - name: Build
+        shell: bash
+        run: |
+          # use grep to remove log spam due to jetifier transformation, c.f. https://stackoverflow.com/a/52353446
+          # use set -o pipefail to keep error code from gradle, c.f. https://stackoverflow.com/a/37128767
+          ( set -o pipefail; ./gradlew build --parallel --build-cache --info | grep -vE "^Transforming (.*) with (.*)Transform" )
+
+      - name: Cache AVD
```
jbsv, line 198: Can we add some documentation on this AVD casing: at least why it is needed ? It doesn't seem to run in the check steps.

.github/workflows/ci.yaml
jbsv, line 230: not sure why we need this emulator-options as the AVD is created in the previous step ?

pierluca: I suspect that it means "run this in headless mode" (no window) , and use "swiftshader_indirect" as GPU ( e.g. https://developer.android.com/studio/run/emulator-acceleration )

pierluca: I believe AVD (e.g., the Android device) != startup options (e.g., boot params). Can you confirm @GabrielFleischer  ?

pierluca (dismissed):

.github/workflows/ci.yaml
GabrielFleischer: `Cache AVD` is executed, but `Create AVD and generate snapshot for caching` is only executed the first time when there is no cache available to generate it. That is why it is skipped on later runs.

.github/workflows/ci.yaml
GabrielFleischer: The emulator is started on and closed on for each action. And it is actually started only once in a normal run as the `Create AVD and generate snapshot for caching` is executed only when the cache is not generated.

pierluca (approved):

PR #799, opened 2022-02-10
refactor the handlers to have a single Await.result instead of nested ones

tuomas-p:
Change RollCallHandler (for CloseRollCall), LaoHandler (for CreateLao) and ElectionHandler (for SetupElection) in order to have a single Future as a result instead of separate ones in order to only need a single Await.result.
The error message handling should be discussed.


be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
```diff
@@ -72,8 +72,59 @@ sealed class RollCallHandler(dbRef: => AskableActorRef) extends MessageHandler {
...
+
+        // We create a new Future (tuple of Futures) to remove the obligatory sequential execution
+        // TODO: A possible addition would be to handle the error messages better
+        val resFuture: Future[GraphMessage] = (for{
+          f1 <- ask
+          f2 <- askOldData
+          f3 <- askCreateChannels
+        } yield(f1, f2, f3)).map{
+          case (Left(_), DbActorReadLaoDataAck(Some(_)), Left(_)) => Left(rpcMessage)
+          case (Left(_), DbActorNAck(code, description), Left(_)) => Right(PipelineError(code, description, rpcMessage.id))
```
pierluca, line 104: In this case, wouldn't we have created the channels already, even though the LAO doesn't exist? 
My understanding is that the `flatMap` / comprehension would go through each step, as `dbActorNAck` is not a failure.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
```diff
@@ -72,8 +72,59 @@ sealed class RollCallHandler(dbRef: => AskableActorRef) extends MessageHandler {
...
+
+    val askOldData = dbActor ? DbActor.ReadLaoData(rpcMessage.getParamsChannel)
+
+    rpcMessage.getParamsMessage match {
+      case Some(message: Message) =>
+        val data: CloseRollCall = message.decodedData.get.asInstanceOf[CloseRollCall]
+        val askCreateChannels: Future[GraphMessage] = createAttendeeChannels(data, rpcMessage)
+
+        // We create a new Future (tuple of Futures) to remove the obligatory sequential execution
+        // TODO: A possible addition would be to handle the error messages better
```
pierluca, line 97: how so? what's wrong with the current code?

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
```diff
@@ -72,8 +72,59 @@ sealed class RollCallHandler(dbRef: => AskableActorRef) extends MessageHandler {
...
+        }
+
+        val laoChannel: Option[Base64Data] = rpcMessage.getParamsChannel.decodeChannelLaoId
+        laoChannel match {
+          case None => Right(PipelineError(
+            ErrorCodes.SERVER_ERROR.id,
+            s"There is an issue with the data of the LAO",
+            rpcMessage.id
+          ))
+          case Some(_) => Await.result(resFuture, duration)
```
pierluca, line 117: doesn't this mean that `resFuture` will always be executed, even if the `rpcMessage` contains an invalid LAO channel ?

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
```diff
@@ -72,8 +72,59 @@ sealed class RollCallHandler(dbRef: => AskableActorRef) extends MessageHandler {
   }
 
   def handleCloseRollCall(rpcMessage: JsonRpcRequest): GraphMessage = {
+    // We first create the three (independent) futures
```
pierluca, line 75: This comment is essentially repeating the code, which is rarely useful. (cf. code complete 2nd edition, section 32.4 "self-documenting code" / "keys to effective comments")

I don't think this code suffers heavily from lack of documentation, so don't stress it, but if we wanted to improve documentation a lot here, I would recommend: 

(A) having function API documentation, which explains what `handleCloseRollCall` is going to do (at API-abstraction levels)

(B) within the function, use comments when the code (even rewritten and clean) isn't good enough to convey your ideas and intents. Code Complete (32.5) provides a ton of good examples, but I'd roughly summarize it as:
- explaining the intent behind the code, "why" are we doing this. The code should be already clean & understandable, but explaining the programmer's intent in writing it often helps a lot when maintaining it. 
- providing "summaries" of the code base. "what" are we going to do in the following few lines (not how!)
- clarifying anything "surprising" to the reader, i.e. optimizations, odd bug workarounds (possibly with a link to related resources), justified disrespect of best practices, etc.

pierluca (commented):
Great to see much cleaner code, though I suspect we may have introduced a few bugs.

One thing of concern to me, that's absolutely not your responsibility, is whether we should make sure that `DbActorNAck`s are considered failures (Future-wise). It seems to me that it would make your code a lot more correct, as far as I understand it. 🤔

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
```diff
@@ -72,8 +72,59 @@ sealed class RollCallHandler(dbRef: => AskableActorRef) extends MessageHandler {
...
+    // Creates a channel for each attendee (of name /root/lao_id/social/PublicKeyAttendee), returns a Future[GraphMessage]
+    def createAttendeeChannels(data: CloseRollCall, rpcMessage: JsonRpcRequest): Future[GraphMessage] = {
+      val listAttendeeChannels: List[(Channel, ObjectType.ObjectType)] = data.attendees.map(attendee => (generateSocialChannel(rpcMessage.getParamsChannel, attendee), ObjectType.CHIRP))
+      val askCreateChannels: Future[GraphMessage] = (dbActor ? DbActor.CreateChannelsFromList(listAttendeeChannels)).map{
+        case DbActorAck() => Left(rpcMessage)
+        case DbActorNAck(code, description) => Right(PipelineError(code, description, rpcMessage.id))
+        case _ => Right(PipelineError(ErrorCodes.SERVER_ERROR.id, unknownAnswer, rpcMessage.id))
+      }
+      askCreateChannels
+    }
```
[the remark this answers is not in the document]
tuomas-p: indeed

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
```diff
@@ -72,8 +72,59 @@ sealed class RollCallHandler(dbRef: => AskableActorRef) extends MessageHandler {
...
+        }
+      
+      case _ => Right(PipelineError(
+            ErrorCodes.SERVER_ERROR.id,
+            s"Unable to handle lao message $rpcMessage. Not a Publish/Broadcast message",
+            rpcMessage.id
+          ))
+    }
+
+    /*Await.result(ask, duration) match {
```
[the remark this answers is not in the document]
tuomas-p: Doing it now!

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
```diff
@@ -72,8 +72,59 @@ sealed class RollCallHandler(dbRef: => AskableActorRef) extends MessageHandler {
...
+      val listAttendeeChannels: List[(Channel, ObjectType.ObjectType)] = data.attendees.map(attendee => (generateSocialChannel(rpcMessage.getParamsChannel, attendee), ObjectType.CHIRP))
+      val askCreateChannels: Future[GraphMessage] = (dbActor ? DbActor.CreateChannelsFromList(listAttendeeChannels)).map{
+        case DbActorAck() => Left(rpcMessage)
+        case DbActorNAck(code, description) => Right(PipelineError(code, description, rpcMessage.id))
+        case _ => Right(PipelineError(ErrorCodes.SERVER_ERROR.id, unknownAnswer, rpcMessage.id))
+      }
+      askCreateChannels
+    }
+
+    val askOldData = dbActor ? DbActor.ReadLaoData(rpcMessage.getParamsChannel)
```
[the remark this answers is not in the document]
tuomas-p: Seems better

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
```diff
@@ -72,8 +72,59 @@ sealed class RollCallHandler(dbRef: => AskableActorRef) extends MessageHandler {
   }
 
   def handleCloseRollCall(rpcMessage: JsonRpcRequest): GraphMessage = {
+    // We first create the three (independent) futures
     val ask: Future[GraphMessage] = dbAskWritePropagate(rpcMessage)
```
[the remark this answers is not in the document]
tuomas-p: changed this in all 3 classes, it is way clearer

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
tuomas-p: Nothing is actually wrong, however, if there are two different kinds of errors at the same time, we do not combine the two but just pick one

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandler.scala
```diff
@@ -40,7 +40,22 @@ object ElectionHandler extends MessageHandler {
...
     val electionChannel: Channel = Channel(s"${rpcMessage.getParamsChannel.channel}${Channel.SEPARATOR}$electionId")
 
-    val ask: Future[GraphMessage] = (dbActor ? DbActor.Write(rpcMessage.getParamsChannel, message)).map {
+    // We create the two (independent) Futures
+    val ask = (dbActor ? DbActor.Write(rpcMessage.getParamsChannel, message))
+    val askCreateChannel = (dbActor ? DbActor.CreateChannel(electionChannel, ObjectType.ELECTION))
+
+    val resFuture: Future[GraphMessage] = (for{
+      f1 <- ask
+      f2 <- askCreateChannel
```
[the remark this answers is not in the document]
tuomas-p: Do you mean that the front-end should clearly receive the ElectionSetup message before the channel for the election is created? If so, we can change it, but we will have to use two Awaits as far as I know.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
tuomas-p: Since the decodeChannelLaoId should return a None as soon as things go wrong, it should not be executed when the channel is invalid, but I can change the case order if it looks better that way (first Some(_) then None)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandler.scala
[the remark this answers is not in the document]
tuomas-p: Since the ElectionSetup message goes in the main LAO channel and the newly created channel is the one concerning only the election

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
```diff
@@ -72,8 +72,59 @@ sealed class RollCallHandler(dbRef: => AskableActorRef) extends MessageHandler {
...
+        val data: CloseRollCall = message.decodedData.get.asInstanceOf[CloseRollCall]
+        val askCreateChannels: Future[GraphMessage] = createAttendeeChannels(data, rpcMessage)
+
+        // We create a new Future (tuple of Futures) to remove the obligatory sequential execution
+        // TODO: A possible addition would be to handle the error messages better
+        val resFuture: Future[GraphMessage] = (for{
+          f1 <- ask
+          f2 <- askOldData
+          f3 <- askCreateChannels
+        } yield(f1, f2, f3)).map{
```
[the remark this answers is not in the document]
tuomas-p: changed it :+1:

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
RaulinN: Pierluca is correct in this case. I want to also bring another point... In this case, assuming that the future called `ask` fails (i.e. a true exception, not a Success(DbActorNack)), then **the 2nd features (i.e. askOldData) will still be executed**. Let me show you one last example

Assuming our database is `dbState` (you can imagine adding something to `dbState` as "writing something to be2's database"), then watch what happens when I define ff1 (that fails) and ff2 (that succeeds). No matter what ff1 does, ff2 will still accomplish his task. 

Screenshot 2022-02-11 at 05 06 03



In the project, if ff1 creates an LAO (and fails) and ff2 updates its information, then ff2 will update information on a LAO that does not exist

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
RaulinN: You are both talking about something different I believe ^^

- @tuomas-p : changing the order of the cases will not change anything (except aesthetically/code comprehension) :)
- @pierluca (and @tuomas-p) : yes that is correct. At line 109-110, the `resFuture` will **already** be "in progress" but not necessarily finished. We distinguish two cases:
  - assuming we take the branch at line 117 (Some case), everything is gucci 😎 
  - assuming we take the branch at line 112 (None case), then:
    - if resFuture finishes (i.e. either Successes or Fails), my guess is that it will be "garbage collected" (i.e. no problem there)
    - if resFuture stays pending, then the future will keep running/calculating in the background until completion (which could take forever in some cases)

RaulinN (changes requested):

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
RaulinN: PS: check my message that starts with "I would argue it's already good:" for a solution (screenshot)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
tuomas-p: all right thank you!

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandler.scala
```diff
@@ -40,7 +40,22 @@ object ElectionHandler extends MessageHandler {
...
+    } yield(f1, f2)).map{
+      case (DbActor.DbActorWriteAck(), DbActor.DbActorAck()) => Left(rpcMessage)
+      case (DbActor.DbActorNAck(code, description), _) => Right(PipelineError(code, description, rpcMessage.id))
+      case (_, DbActor.DbActorNAck(code, description)) => Right(PipelineError(code, description, rpcMessage.id))
```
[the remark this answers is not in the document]
tuomas-p: Thank you for the precise explanation and tests on your side @RaulinN !

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
tuomas-p: Oh right I was a bit confused on what the issue was and I get it now. However, updating a non-existing LAO should yield an error and therefore not break the code, or am I wrong and something should be changed?

pierluca:
@tuomas-p @RaulinN is this still a draft or should it be re-reviewed and merged ?

tuomas-p:
It should now be re-reviewed as discussed during the call

PR #800, opened 2022-02-11
Extend message registry

Xelowak:
I extended the message registry by adding the signature type of the messages along with their build function.


fe1-web/model/network/method/message/Message.ts
```diff
@@ -130,14 +155,11 @@ export class Message {
...
-    const messagesProperties = messagePropertiesMap.get(data.object)?.get(data.action);
-    if (messagesProperties === undefined) {
-      throw new Error(`Message signature for object ${data.object} and action ${data.action} is unsupported.`);
-    }
+    // Get the signature type of the type of message we want to sign
+    const signatureType = messageRegistry.getSignatureType(data);
 
     // If the message is signed with the pop token, get it from the store and sign the message
-    if (messagesProperties.isPopTokenSigned) {
+    if (signatureType === SignatureType.POP_TOKEN) {
```
pierluca, line 162: not for this PR, but I believe we could refactor the code from here to the end of the function. It's a bit repetitive and the `console.error` (line 168) doesn't abort/throw/return even though no meaningful work can be done after that error 😅

fe1-web/model/network/method/message/data/MessageData.ts
```diff
@@ -34,6 +34,12 @@ export enum ActionType {
   NOTIFY_DELETE = 'notify_delete',
 }
 
+/** Enumeration of all possible signatures of a message */
+export enum SignatureType {
+  KEYPAIR,
+  POP_TOKEN,
```
pierluca, line 40: would it maybe help debugging to make it a string-valued enum?

fe1-web/model/network/method/message/data/MessageRegistry.ts
```diff
@@ -1,54 +1,88 @@
...
 type HandleFunction = (msg: ExtendedMessage) => boolean;
 
+const {
+  LAO, MEETING, ROLL_CALL, ELECTION, MESSAGE, CHIRP, REACTION,
+} = ObjectType;
+const {
+  CREATE, STATE, UPDATE_PROPERTIES, OPEN, CLOSE, REOPEN, SETUP, CAST_VOTE, END, RESULT, WITNESS,
+  ADD, NOTIFY_ADD, DELETE, NOTIFY_DELETE,
+} = ActionType;
+const { KEYPAIR, POP_TOKEN } = SignatureType;
```
pierluca, line 28: I didn't know you could do this. It looks convenient 😅

fe1-web/model/network/method/message/data/MessageRegistry.ts
```diff
@@ -80,9 +115,45 @@ export class MessageRegistry {
     }
     const { handle } = messageEntry;
     if (handle === undefined) {
-      console.warn(`Message ${data.object} ${data.action} does not have a 'handle' property in MessageRegistry`);
+      console.warn(`Message ${data.object} ${data.action} does not have an 'handle' property in MessageRegistry`);
```
pierluca, line 118: <grammar-nazi>
It was correct before. 
The "h" in "handle" is not silent, hence "a" 😄 
</grammar-nazi>

fe1-web/model/network/method/message/data/MessageRegistry.ts
```diff
@@ -80,9 +115,45 @@ export class MessageRegistry {
...
+  getSignatureType(data: MessageData): SignatureType {
+    const messageEntry = this.mapping.get(k(data.object, data.action));
+    if (messageEntry === undefined) {
+      throw new Error(`Message ${data.object} ${data.action} is not contained in MessageRegistry`);
+    }
+    const { signature } = messageEntry;
+    if (signature === undefined) {
+      throw new Error(`Message ${data.object} ${data.action} does not have a 'signature' property in MessageRegistry`);
+    }
+    return signature;
```
pierluca, line 157: as we discussed, we can probably have a single verification at the application startup, to fail fast / fail hard

pierluca (commented):
I'm being annoying here, but it's pretty good already 👍

fe1-web/model/network/method/message/Message.ts
Xelowak: That's true, I can do it.

PR #804, opened 2022-02-11
Added clean server startup option '-Dclean'

RaulinN:
By adding the clean flag at compilation, server starts in a fresh state


be2-scala/src/main/scala/ch/epfl/pop/config/RuntimeEnvironment.scala
```diff
@@ -19,19 +20,28 @@ object RuntimeEnvironment {
   private lazy val sp = new SystemProperties()
 
   private def getConfDir: String = {
+    /* Get config directory path form JVM */
```
pierluca, line 23:
```suggestion
    /* Get config directory path from JVM */
```

pierluca (approved):
Minor typo, but LGTM

tuomas-p (approved):

PR #808, opened 2022-02-14
Implementation of the – brand new – DbActor

RaulinN:


be2-scala/src/main/scala/ch/epfl/pop/storage/DiskStorage.scala
```diff
@@ -0,0 +1,62 @@
...
+        throw e
+    }
+  }
+
+  def close(): Unit = db.close()
+
+
+  @throws [org.iq80.leveldb.DBException]
+  def read(key: String): String = {
+    new String(db.get(key.getBytes(StandardCharsets.UTF_8)), StandardCharsets.UTF_8)
```
pierluca, line 33: I think it wouldn't hurt readability to split this in more lines , to avoid 3 nested function calls😉

be2-scala/src/main/scala/ch/epfl/pop/storage/DiskStorage.scala
```diff
@@ -0,0 +1,62 @@
...
+        logger.error("Could not open database folder '{}': {}", databaseFolder, e.getMessage)
+        throw e
+    }
+  }
+
+  def close(): Unit = db.close()
+
+
+  @throws [org.iq80.leveldb.DBException]
+  def read(key: String): String = {
```
pierluca, line 32: would it make sense to log & return an Option instead of throwing ? The higher layers probably won't be able to deal with the error anyways.

be2-scala/src/main/scala/ch/epfl/pop/storage/DiskStorage.scala
```diff
@@ -0,0 +1,62 @@
+package ch.epfl.pop.storage
+
+import java.io.{File, IOException}
+import java.nio.charset.StandardCharsets
+
+import org.iq80.leveldb.{DB, Options, WriteBatch}
+import org.iq80.leveldb.impl.Iq80DBFactory.factory
+import org.slf4j.{Logger, LoggerFactory}
+
+class DiskStorage(val databaseFolder: String = DiskStorage.DATABASE_FOLDER) {
```
pierluca, line 10: To keep the API clean, I would make sure that the API doesn't expose `org.iq80.leveldb.*` types, such as the DBException.

Otherwise the mocks will need to do that too.

It may make sense to see what API design would keep the usage simpler.
-> simple success / failure return values (or `Option`s) + internal logging ?
-> a dedicated exception that wraps the DBException ? 
-> something else?

be2-scala/src/main/scala/ch/epfl/pop/storage/DiskStorage.scala
RaulinN: No the whole new actor 10x shorter than the previous one (in part) because of this ^^. Let me show you how it works and we can then discuss if it's a good or if a change would be better :)

Screenshot 2022-02-14 at 15 18 23

Look at the `write` method. We spare 2 levels of indentation (match on Option & Try) using the fact that IF the readChannelData fails, then the rest of the method (the storage.write call) will not be execute and an exception will be thrown.

This alone is not enough to justify the use of Exceptions I agree. However here's the catch! For Tuomas to use his futures correctly, the dbActor needs to send back Futures (either successful containing an ACK, or failure containing an exception). HOWEVER, it turns out that an actor does *not* send back a failed future on exception throw (didn't know that ^^, thank you Tuomas for the enlightenment). This is why we have to have a Try block in the actor before sending back to client.

It is not there yet because it is being implemented in Tuomas PR, but you can see the screenshot above for an example (line 23 and 24)

be2-scala/src/main/scala/ch/epfl/pop/storage/DiskStorage.scala
RaulinN: That's a good point! If my logic from previous message makes sense, then the wrapped exception is definitely the best in our case

be2-scala/src/main/scala/ch/epfl/pop/storage/DiskStorage.scala
RaulinN: (Already corrected in #809 :) )

be2-scala/src/main/scala/ch/epfl/pop/storage/DiskStorage.scala
pierluca: It does make sense 😄

be2-scala/src/main/scala/ch/epfl/pop/storage/DiskStorage.scala
tuomas-p: :+1:

be2-scala/src/test/scala/ch/epfl/pop/storage/DiskStorageSuite.scala
```diff
@@ -0,0 +1,108 @@
...
+  test("write correctly adds tuples one-by-one in db") {
+    val storage: DiskStorage = generateDiskStorage()
+
+    storage.write(KEY -> VALUE)
+    storage.read(KEY) should equal (Some(VALUE))
+
+    storage.write(key(1) -> value(1))
+
+    storage.read(KEY) should equal (Some(VALUE))
+    storage.read(key(1)) should equal (Some(value(1)))
```
pierluca, line 65: why are we doing the same operations & assertions twice?

be2-scala/src/test/scala/ch/epfl/pop/storage/DiskStorageSuite.scala
```diff
@@ -0,0 +1,108 @@
...
+
+    storage.write(BATCH : _*)
+
+    for ((k, v) <- BATCH) storage.read(k) should equal (Some(v))
+  }
+
+  test("write correctly does nothing on no tuple add in db") {
+    val storage: DiskStorage = generateDiskStorage()
+
+    storage.write()
```
pierluca, line 79: we could check if there's any failure associated with the operation, it might be more meaningful 🤔

be2-scala/src/test/scala/ch/epfl/pop/storage/DbActorNewSuite.scala
```diff
@@ -1,9 +1,277 @@
...
+
+    storage.size should equal (0)
+
+    dbActor ! DbActorNew.CreateChannel(Channel(CHANNEL_NAME), ObjectType.LAO); sleep()
+
+    expectMsg(DbActorNew.DbActorAck())
+    storage.size should equal (1)
+
+    // adding 2 new channels (one of which already exists)
+    dbActor ! DbActorNew.CreateChannelsFromList((Channel(CHANNEL_NAME), ObjectType.ELECTION) :: (Channel(s"${CHANNEL_NAME}2"), ObjectType.ROLL_CALL) :: Nil); sleep()
```
pierluca, line 205: this seems precisely a case that should fail.
if `CHANNEL_NAME` was already added as `ObjectType.LAO`, how can we allow it to be created as `ObjectType.ELECTION` without failure ?

tuomas-p (approved):
Very clean (especially as most of it had already been reviewed, left a few really minor suggestions

tuomas-p (approved):

PR #819, opened 2022-02-17
Reorganize Wallet and use barrel imports among features

Xelowak:
I moved all wallet files to where they should be, and updated all features to use barrel imports.


fe1-web/features/wallet/store/WalletStore.ts
```diff
@@ -26,26 +30,28 @@ async function decrypt(cipher: string): Promise<Uint8Array> {
...
+   * @returns a Promise which completes after storage
    */
   export async function store(mnemonic: string, seed: Uint8Array) {
     const binaryMnemonic: Uint8Array = new TextEncoder().encode(mnemonic);
 
     await getStore().dispatch(async (dispatch: AsyncDispatch): Promise<void> => {
       const encryptedSeed = await encrypt(seed);
       const encryptedMnemonic = await encrypt(binaryMnemonic);
-      await dispatch(setWallet({
+      dispatch(setWallet({
```
pierluca, line 45: nice catch!

fe1-web/res/strings.ts
```diff
@@ -29,7 +29,7 @@ const STRINGS = {
 
   /* --- Witness Navigation Strings --- */
   witness_navigation_tab_home: 'Witness',
-  witness_navigation_tab_video: 'Witness Video',
+  witness_navigation_tab_video: 'WitnessScreen Video',
```
pierluca, line 32: Was this by design? I believe this string does show up in the user interface.

fe1-web/features/rollCall/components/EventRollCall.tsx
```diff
@@ -5,10 +5,11 @@ import { useSelector } from 'react-redux';
 import { useNavigation } from '@react-navigation/native';
 
 import { makeCurrentLao } from 'store';
-import { Timestamp, Wallet } from 'model/objects';
+import { Timestamp } from 'model/objects';
 import QRCode from 'components/QRCode';
 import WideButtonView from 'components/WideButtonView';
 import STRINGS from 'res/strings';
+import * as Wallet from 'features/wallet/objects';
```
pierluca, line 12: it seems to me that in the future (not this PR), we could simply export from the wallet two custom hooks, along the lines of `useCurrentPopToken(): PopToken | undefined` and a `usePopTokenGenerator(laoId, rollCallId): PopToken | undefined` to cover virtually all the needs of import the wallet 😄 

Maybe we wouldn't even need to export `usePopTokenGenerator` and we could instead have a component which simply displays the QR code for a generated token, and it'd be enough. Just thinking ... 🤔

pierluca (approved):
Thanks for the clean-up, good job

fe1-web/res/strings.ts
pierluca: If relevant, I'll let you fix this in a subsequent PR, merging this one for now 👍

PR #820, opened 2022-02-17
Add DbActor Read functionalities as well as WriteLaoData and unit tests

tuomas-p:
Add and test DbActor Read functionalities (only successful cases, as failing cases are not ready yet), add WriteLaoData feature and tests (easier to manage than edge cases in Write)

### DIFF
--- a/be2-scala/src/main/scala/ch/epfl/pop/model/objects/Channel.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/model/objects/Channel.scala
@@ -46,12 +46,13 @@ final case class Channel(channel: String) {
 
 object Channel {
   final val SEPARATOR: Char = '/'
+  final val SPECIAL_SEPARATOR: Char = ':'
   final val ROOT_CHANNEL: Channel = Channel(s"${SEPARATOR}root")
   final val ROOT_CHANNEL_PREFIX: String = s"${SEPARATOR}root$SEPARATOR"
 
   private final def channelRegex: String = "^/root(/[^/]+)*$"
 
-  final val LAO_DATA_LOCATION: String = s"${SEPARATOR}data"
+  final val LAO_DATA_LOCATION: String = s"${SPECIAL_SEPARATOR}laodata"
 
   final val SOCIAL_CHANNEL_PREFIX: String = s"${SEPARATOR}social$SEPARATOR"
   final val SOCIAL_MEDIA_CHIRPS_PREFIX: String = s"${SOCIAL_CHANNEL_PREFIX}chirps"
--- a/be2-scala/src/main/scala/ch/epfl/pop/model/objects/Channel.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/model/objects/Channel.scala
@@ -11,7 +11,7 @@ final case class Channel(channel: String) {
    */
   def decodeChannelLaoId: Option[Base64Data] = channel match {
     case _ if channel.startsWith(Channel.ROOT_CHANNEL_PREFIX) =>
-      Try(Base64Data(channel.substring(Channel.ROOT_CHANNEL_PREFIX.length).split(Channel.SEPARATOR)(0))) match {
+      Try(Base64Data(channel.substring(Channel.ROOT_CHANNEL_PREFIX.length).split(Channel.CHANNEL_SEPARATOR)(0))) match {
         case Success(value) => Some(value)
         case _ => None
       }
@@ -27,7 +27,7 @@ final case class Channel(channel: String) {
   def extractChildChannel: Hash = {
     //After successful channel creation
     //c cannot be empty
-    val c = channel.split(Channel.SEPARATOR)
+    val c = channel.split(Channel.CHANNEL_SEPARATOR)
     assert(!c.isEmpty)
     Hash(Base64Data(c.last))
   }
@@ -45,16 +45,16 @@ final case class Channel(channel: String) {
 }
 
 object Channel {
-  final val SEPARATOR: Char = '/'
-  final val SPECIAL_SEPARATOR: Char = ':'
-  final val ROOT_CHANNEL: Channel = Channel(s"${SEPARATOR}root")
-  final val ROOT_CHANNEL_PREFIX: String = s"${SEPARATOR}root$SEPARATOR"
+  final val CHANNEL_SEPARATOR: Char = '/'
+  final val DATA_SEPARATOR: Char = '#'
+  final val ROOT_CHANNEL: Channel = Channel(s"${CHANNEL_SEPARATOR}root")
+  final val ROOT_CHANNEL_PREFIX: String = s"${CHANNEL_SEPARATOR}root${CHANNEL_SEPARATOR}"
 
   private final def channelRegex: String = "^/root(/[^/]+)*$"
 
-  final val LAO_DATA_LOCATION: String = s"${SPECIAL_SEPARATOR}laodata"
+  final val LAO_DATA_LOCATION: String = s"${DATA_SEPARATOR}laodata"
 
-  final val SOCIAL_CHANNEL_PREFIX: String = s"${SEPARATOR}social$SEPARATOR"
+  final val SOCIAL_CHANNEL_PREFIX: String = s"${CHANNEL_SEPARATOR}social${CHANNEL_SEPARATOR}"
   final val SOCIAL_MEDIA_CHIRPS_PREFIX: String = s"${SOCIAL_CHANNEL_PREFIX}chirps"
   final val REACTIONS_CHANNEL_PREFIX: String = s"${SOCIAL_CHANNEL_PREFIX}reactions"
 
--- a/be2-scala/src/main/scala/ch/epfl/pop/storage/DbActorNew.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/storage/DbActorNew.scala
@@ -34,16 +34,18 @@ case class DbActorNew(
     }
     createChannel(channel, _object)
 
-    val channelData: ChannelData = readChannelData(channel)
-    storage.write(
-      (channel.toString, channelData.addMessage(message.message_id).toJsonString),
-      (s"$channel${Channel.SPECIAL_SEPARATOR}${message.message_id}", message.toJsonString)
-    )
+    this.synchronized{
+      val channelData: ChannelData = readChannelData(channel)
+      storage.write(
+        (channel.toString, channelData.addMessage(message.message_id).toJsonString),
+        (s"$channel${Channel.DATA_SEPARATOR}${message.message_id}", message.toJsonString)
+      )
+    }
   }
 
   @throws [DbActorNAckException]
   private def read(channel: Channel, messageId: Hash): Option[Message] = {
-    Try(storage.read(s"$channel${Channel.SPECIAL_SEPARATOR}$messageId")) match {
+    Try(storage.read(s"$channel${Channel.DATA_SEPARATOR}$messageId")) match {
       case Success(Some(json)) => Some(Message.buildFromJson(json))
       case Success(None) => None
       case Failure(ex) => throw DbActorNAckException(ErrorCodes.SERVER_ERROR.id, ex.getMessage)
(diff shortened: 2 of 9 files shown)


be2-scala/src/main/scala/ch/epfl/pop/storage/DbActorNew.scala
```diff
@@ -37,21 +37,85 @@ case class DbActorNew(
     val channelData: ChannelData = readChannelData(channel)
     storage.write(
       (channel.toString, channelData.addMessage(message.message_id).toJsonString),
-      (s"$channel:${message.message_id}", message.toJsonString)
+      (s"$channel${Channel.SPECIAL_SEPARATOR}${message.message_id}", message.toJsonString)
```
pierluca, line 40: I hadn't realized but there's the same read-modify-write concurrency issue here.

be2-scala/src/test/scala/ch/epfl/pop/storage/DbActorNewSuite.scala
```diff
@@ -274,4 +274,159 @@ class DbActorNewSuite extends TestKit(ActorSystem("DbActorNewSuiteActorSystem"))
...
+    val answer = Await.result(ask, duration)
+
+    answer shouldBe a[DbActorNew.DbActorCatchupAck]
+
+    val list: List[Message] = answer.asInstanceOf[DbActorNew.DbActorCatchupAck].messages
+
+    list.size should equal(2)
+    list should contain(MESSAGE)
+    list should contain(message2)
+
```
pierluca, line 404: the newlines, in this, the next, and the previous tests do not seem to be consistent :)

be2-scala/src/test/scala/ch/epfl/pop/storage/DbActorNewSuite.scala
```diff
@@ -274,4 +274,159 @@ class DbActorNewSuite extends TestKit(ActorSystem("DbActorNewSuiteActorSystem"))
...
+    val ask = dbActor ? DbActorNew.Catchup(channelName1)
+    val answer = Await.result(ask, duration)
+
+    answer shouldBe a[DbActorNew.DbActorCatchupAck]
+
+    val list: List[Message] = answer.asInstanceOf[DbActorNew.DbActorCatchupAck].messages
+
+    list.size should equal(1)
+    list should contain(MESSAGE)
+
```
pierluca, line 429: sometimes when you need to write a rather complex test, it's convenient for the readers to clearly structure the test as the 3 As (arrange, act, assert), using comments to highlight/summarize the structure. Here the test title is clear enough, but as an example of what could be "bare minimum" comments to ease the reading of the test:
```
    // arrange
    val channelName1: Channel = Channel(CHANNEL_NAME)
    ...
    val dbActor: AskableActorRef = system.actorOf(Props(DbActorNew(mediatorRef, initialStorage)))

    // act
    val ask = dbActor ? DbActorNew.Catchup(channelName1)
    val answer = Await.result(ask, duration)

    // assert
    answer shouldBe a[DbActorNew.DbActorCatchupAck]
```

be2-scala/src/test/scala/ch/epfl/pop/storage/DbActorNewSuite.scala
```diff
@@ -274,4 +274,159 @@ class DbActorNewSuite extends TestKit(ActorSystem("DbActorNewSuiteActorSystem"))
...
+    list.size should equal(2)
+    list should contain(MESSAGE)
+    list should contain(message2)
+
+  }
+
+  test("catchup should not fail on a channel with ChannelData containing missing message_ids (and only return valid messages)"){
+    val channelName1: Channel = Channel(CHANNEL_NAME)
+    val initialStorage: InMemoryStorage = InMemoryStorage()
+    val message2 = MESSAGE.copy(message_id = Hash(Base64Data("RmFrZSBtZXNzYWdlX2lkIDopIE5vIGVhc3RlciBlZ2cgcmlnaHQgdGhlcmUhIC0tIE5pY29sYXMgUmF1bGlu")))
```
pierluca, line 410: it seems we don't need a copy of the message here.
I believe a Hash would be enough for our purposes.

be2-scala/src/test/scala/ch/epfl/pop/storage/DbActorNewSuite.scala
```diff
@@ -274,4 +274,159 @@ class DbActorNewSuite extends TestKit(ActorSystem("DbActorNewSuiteActorSystem"))
...
+    val dbActor: AskableActorRef = system.actorOf(Props(DbActorNew(mediatorRef, initialStorage)))
+
+    val ask = dbActor ? DbActorNew.Catchup(channelName1)
+    val answer = Await.result(ask, duration)
+
+    answer shouldBe a[DbActorNew.DbActorCatchupAck]
+
+    val list: List[Message] = answer.asInstanceOf[DbActorNew.DbActorCatchupAck].messages
+
+    list.size should equal(1)
```
pierluca, line 427: as a sanity check, I would also make sure that the second hash is NOT part of the result.
Think of it as a regression test to ensure that a potential, future, very broken implementation, doesn't garble messages together.

pierluca (changes requested):
Very nice job overalls ! 
I've left a bunch of comments, most of them are minor points.
The only critical issue that I can spot is the potential concurrency issue. I'm available to have a chat about it if you think it'd help.

be2-scala/src/main/scala/ch/epfl/pop/storage/DbActorNew.scala
```diff
@@ -110,6 +174,16 @@ case class DbActorNew(
     )
   }
 
+  @throws [DbActorNAckException]
+  private def generateLaoDataKey(channel: Channel): String = {
+    channel.decodeChannelLaoId match {
+      case Some(data) => s"${Channel.ROOT_CHANNEL_PREFIX}$data${Channel.LAO_DATA_LOCATION}"
+      case None =>
+        log.error(s"Actor $self (db) encountered a problem while decoding sub-channel from '$channel'")
```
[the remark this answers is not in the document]
tuomas-p: Yes indeed

tuomas-p:
Concerning the two last comments, I removed the message copy, only keeping the Hash, but there is now no other good way to verify (in my opinion) that the second hash is not part of the result (as we already have the size = 1).

pierluca (approved):
Very good, just a tiny test split suggestion 😄

be2-scala/src/main/scala/ch/epfl/pop/storage/DbActorNew.scala
```diff
@@ -34,24 +34,88 @@ case class DbActorNew(
...
+              log.error(s"Problem encountered while catching up message with id $head")
+              buildCatchupList(tail, acc)
+          }
+      }
+    }
+
+    val channelData: ChannelData = Try(readChannelData(channel)) match {
+      case Success(data) => data
+      case Failure(ex) => throw DbActorNAckException(ErrorCodes.SERVER_ERROR.id, ex.getMessage)
+    }
```
RaulinN, line 112: You can simplify this (since `ex` is the same as what you are throwing at line 111)
```suggestion
    val channelData: ChannelData = readChannelData(channel)
```

be2-scala/src/main/scala/ch/epfl/pop/storage/DbActorNew.scala
```diff
@@ -34,24 +34,88 @@ case class DbActorNew(
...
+    def buildCatchupList(msgIds: List[Hash], acc: List[Message]): List[Message] = {
+      msgIds match {
+        case Nil => acc
+        case head :: tail =>
+          Try(read(channel, head)).recover(
+            ex => None
+          ) match {
+            case Success(Some(msg)) => buildCatchupList(tail, msg :: acc)
+            case _ => 
+              log.error(s"Problem encountered while catching up message with id $head")
```
RaulinN, line 103: I would make the message just a little more explicit. If this occurs, it means that the next batch of students f* up somewhere ^^'. We need to make them understand that this should never happen (i.e. we stored a hash in the channel without its associated message). Something like "critical error: channel $channel contains a message that failed to be stored in the db". Maybe even a `log.critical` if it exists with our logger extension

be2-scala/src/main/scala/ch/epfl/pop/storage/DbActorNew.scala
```diff
@@ -34,24 +34,88 @@ case class DbActorNew(
...
+    }
+
+    val channelData: ChannelData = Try(readChannelData(channel)) match {
+      case Success(data) => data
+      case Failure(ex) => throw DbActorNAckException(ErrorCodes.SERVER_ERROR.id, ex.getMessage)
+    }
+    Try(buildCatchupList(channelData.messages, Nil)) match {
+      case Success(li) => li
+      case Failure(ex) => throw DbActorNAckException(ErrorCodes.SERVER_ERROR.id, ex.getMessage)
+    }
```
RaulinN, line 116: You can also simplify this code but for another reason. Your `buildCatchupList` is not able to throw anything

```suggestion
    buildCatchupList(channelData.messages, Nil)
```

be2-scala/src/main/scala/ch/epfl/pop/storage/DbActorNew.scala
```diff
@@ -34,24 +34,88 @@ case class DbActorNew(
...
+        case Success(data) => data
+        case Failure(_) => LaoData.emptyLaoData
+      }
+      val laoDataKey: String = generateLaoDataKey(channel)
+
+      Try(storage.write(
+        (laoDataKey, laoData.updateWith(message).toJsonString)
+      )).recover(
+        ex => throw DbActorNAckException(ErrorCodes.SERVER_ERROR.id, ex.getMessage)
+      )
```
RaulinN, line 86: No need to recover since you are throwing the same thing :)

```suggestion
      storage.write(laoDataKey -> laoData.updateWith(message).toJsonString)
```

be2-scala/src/main/scala/ch/epfl/pop/model/objects/Channel.scala
```diff
@@ -11,7 +11,7 @@ final case class Channel(channel: String) {
    */
   def decodeChannelLaoId: Option[Base64Data] = channel match {
     case _ if channel.startsWith(Channel.ROOT_CHANNEL_PREFIX) =>
-      Try(Base64Data(channel.substring(Channel.ROOT_CHANNEL_PREFIX.length).split(Channel.SEPARATOR)(0))) match {
+      Try(Base64Data(channel.substring(Channel.ROOT_CHANNEL_PREFIX.length).split(Channel.CHANNEL_SEPARATOR)(0))) match {
```
RaulinN, line 14: Cannot highlight the right line (2 lines above), but is there a reason why you guys decided last semester to return a Base64Data optional here (instead of Hash optional) since we are returning a lao id?

be2-scala/src/test/scala/ch/epfl/pop/storage/DbActorNewSuite.scala
```diff
@@ -1,15 +1,18 @@
...
 import ch.epfl.pop.model.objects._
-import ch.epfl.pop.pubsub.PubSubMediator
+import ch.epfl.pop.pubsub.{AskPatternConstants, PubSubMediator}
 import org.scalatest.{BeforeAndAfterAll, FunSuiteLike, Matchers}
+import scala.concurrent.Await
+import util.examples.MessageExample
 
 
-class DbActorNewSuite extends TestKit(ActorSystem("DbActorNewSuiteActorSystem")) with FunSuiteLike with ImplicitSender with Matchers with BeforeAndAfterAll {
+class DbActorNewSuite extends TestKit(ActorSystem("DbActorNewSuiteActorSystem")) with FunSuiteLike with ImplicitSender with Matchers with BeforeAndAfterAll with AskPatternConstants {
```
RaulinN, line 15: Nice tests ! 👍

RaulinN (commented):
Ye very nice job :D A few minor points in my opinion :)

One more thing I couldn't fit in the review, what do you think about converting the `emptyLaoData` function in LaoData.scala to a constructor accepting 0 parameter (i.e. apply with 0 parameter)? Would make more sense in my mind

be2-scala/src/main/scala/ch/epfl/pop/storage/DbActorNew.scala
RaulinN: EDIT: turns out, there's no log.critical or log.fatal :(

RaulinN: EDIT 2 : will be very useful if we add the modifyLao function or delete message for some reason

be2-scala/src/main/scala/ch/epfl/pop/storage/DbActorNew.scala
RaulinN: Ohhhh I now understand why you had it like that... It's because I did the same mistake ^^' Since the DiskStorage only fails with dbactornackexceptions, we don't need these recover anymore :D

PR #821, opened 2022-02-17
Reorganize core files

Xelowak:
WIP


pierluca:
Merging broken CI because the CI was too lax before. We haven't "worsened" the situation, just made it more obvious.

PR #825, opened 2022-02-18
Reorganize Lao and Witness as features

Xelowak:
I reorganized Lao and Witness as features.


pierluca (approved):
As discussed, we're going to fix the CI in a second step.

pierluca (approved):

PR #828, opened 2022-02-18
Fixed import cycles & switched to dependency injection

pierluca:
This document has been a life-saver: https://redux.js.org/usage/code-splitting

Architecturally, I've also decoupled further the ingestion from the network messages.
`ExtendedMessage` is now part of the ingestion package and unseen outside of it.
However, it does implement `ProcessableMessage`, a new interface defined next to the `MessageRegistry`.


fe1-web/src/core/keypair/index.ts
```diff
@@ -0,0 +1,7 @@
+import { addReducer } from '../redux/Manage';
+import keyPairReducer from './Reducer';
+
+export * from './Reducer';
```
Xelowak, line 4: Why did you rename the reducer ? It being called simply 'reducer' feels like it's generic.

fe1-web/src/core/network/ingestion/Reducer.ts
```diff
@@ -0,0 +1,133 @@
+/**
```
Xelowak, line 1: ... and it makes both these files have the same name.

fe1-web/src/core/network/validation/Checker.ts
```diff
@@ -27,19 +25,23 @@ export function checkWitnessSignatures(witSig: WitnessSignature[], data: Base64U
...
-  // FIXME check modification id
-  return id; // simply to remove eslint warning
-}
-
-const MIN_WITNESS_FACTOR_N = 3; // numerator
-const MIN_WITNESS_FACTOR_D = 5; // denominator, = three fifths = 60%
-
-export function hasWitnessSignatureQuorum(witSigs: WitnessSignature[], lao: Lao): boolean {
+/**
+ * Verify if the witness signatures constitute a quorum of the
```
Xelowak, line 29: I guess the sentence is not finished.

fe1-web/src/core/network/validation/Checker.ts
```diff
@@ -27,19 +25,23 @@ export function checkWitnessSignatures(witSig: WitnessSignature[], data: Base64U
...
+ *
+ * @param witSigs witness signatures
+ * @param lao a LAO providing a list of witness keys
```
Xelowak, line 32:
```suggestion
 * @param witSigs - The witness signatures
 * @param lao - A LAO providing a list of witness keys
```

fe1-web/src/core/objects/__tests__/WitnessSignature.test.ts
```diff
@@ -1,9 +1,7 @@
 import 'jest-extended';
 
 import testKeyPair from 'test_data/keypair.json';
-import { Hash, KeyPair } from 'core/objects';
-
-import { WitnessSignature, WitnessSignatureState } from '../WitnessSignature';
+import { Hash, KeyPair, WitnessSignature, WitnessSignatureState } from '..';
```
Xelowak, line 4:
```suggestion
import { Hash, KeyPair, WitnessSignature, WitnessSignatureState } from '../';
```

Xelowak (commented):
Nice work, it's so clean now !
I think there are some issues for which GitHub warned us, but since you made a push while I was reviewing, they might be gone. By the way, they now appear 5 times, it might be too much 😅
EDIT: Apparently it was only the case before you pushed, now it's 3 times again.

fe1-web/src/App.tsx
```diff
@@ -8,13 +8,12 @@ import { useReduxDevToolsExtension } from '@react-navigation/devtools';
 import { PersistGate } from 'redux-persist/integration/react';
 import { ToastProvider } from 'react-native-toast-notifications';
 import { Provider } from 'react-redux';
-import { store, persist } from 'core/redux/ReduxSetUp';
+import { store, persist } from 'core/redux';
 
 import AppNavigation from 'core/navigation/AppNavigation';
 import { navigationRef } from 'core/navigation/RootNavigation';
```
Xelowak, line 14: The app is probably broken because of this

PR #829, opened 2022-02-19
Create a basis for a error display system

GabrielFleischer:
This PR aims at putting the foundations of an error system that will provide concise messages to the average user while keeping detailed explanations in the logs.

- [x] Create the GeneraticException class
- [x] Create the ErrorUtils class that provides a simple way to display a localized Toast while logging the error
- [x] Port existing exceptions to the new GenericExcepetion
- [x] Apply the change to the rest of the codebase


pierluca (commented):
Looks good so far ! 👍

pierluca (approved):
Much better. My only concern is that `error_retrieve_own_token` is shared in many different circumstances, and the generic message may not always be very eloquent in those specific use cases.
Anyhow, it's "good enough" for now 😄 👍

PR #837, opened 2022-02-22
fix read, writeLaoData and checkChannelExistence tests

tuomas-p:
Change test structure for "failing" test (checkChannelExistence here) as well as fix other previously failing tests


RaulinN (approved):
Very well done :) Thank you! :D

PR #840, opened 2022-02-22
Removed most Scapegoat errors/warnings/infos

RaulinN:


be2-scala/build.sbt
```diff
@@ -39,6 +39,9 @@ scalacOptions ++= Seq(
     "-Ycache-macro-class-loader:last-modified", // and macro definitions. This can lead to performance improvements.
 )
 
+Scapegoat/ scalacOptions -= "-Xfatal-warnings"
+// Temporarily report scapegoat errors as warnings, to avoid broken builds
+Scapegoat/ scalacOptions += "-P:scapegoat:overrideLevels:all=Warning"
```
pierluca, line 44: didn't we say that we were aiming to turn even warnings to errors? 🤔

tuomas-p (approved):

be2-scala/build.sbt
RaulinN: Yes...... I added a complementary to the commit ^^
> Some errors require more time/thinking to resolve. To avoid broken
builds, we will keep Scapegoat's advices as warnings

In other words, we need to find a "clean" way to handle the last Option.get remaining in the code (harder to remove/modify than the ones we cleaned :/ )

RaulinN: PS: the last remaining errors are basically all the same error. Gimme a few days and we will have found a solution. Deal @pierluca ? 😈

PR #844, opened 2022-02-24
Create a dedicated type for the channel in the protocol

GabrielFleischer:
This PR creates a new type for the channel

This makes the code a lot more readable and easier to write.

There are changes almost everywhere, so I will test it manually before merging.

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Channel.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Channel.java
@@ -52,9 +52,9 @@ public class Channel {
   }
 
   private Channel(Channel base, String subElement) {
-    List<String> elements = new ArrayList<>(base.elements);
-    elements.add(subElement);
-    this.elements = Collections.unmodifiableList(elements);
+    List<String> newElems = new ArrayList<>(base.elements);
+    newElems.add(subElement);
+    this.elements = Collections.unmodifiableList(newElems);
 
     if (!CHANNEL_VALIDATOR.test(getAsString()))
       throw new IllegalArgumentException(getAsString() + " is not a valid channel");
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Channel.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Channel.java
@@ -1,0 +1,132 @@
+package com.github.dedis.popstellar.model.objects;
+
+import androidx.annotation.NonNull;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import java.util.function.Predicate;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+/**
+ * This class represent a channel used to subscribe to a certain part of the messages. More detailed
+ * information are given in the protocol specification of the application.
+ *
+ * <p>It is implemented in a functional manner to avoid any unintended modification
+ */
+public class Channel {
+
+  private static final String ROOT_CHANNEL = "/root";
+  private static final String DELIMITER = "/";
+  // Predicate matching a String with the protocol regex of a channel
+  private static final Predicate<String> CHANNEL_VALIDATOR =
+      Pattern.compile("^/root(/[^/]+)*$").asPredicate();
+
+  /** Root channel, every client is subscribed to it. */
+  public static final Channel ROOT = new Channel();
+
+  public static Channel newChannel(String value) {
+    // Avoid creating duplicate instances
+    if (ROOT_CHANNEL.equals(value)) return ROOT;
+
+    // Remove the "/root/" part of the value as it is not relevant and does not need to be stored
+    String relevantPart = value.substring(ROOT_CHANNEL.length() + 1);
+    return new Channel(relevantPart.split(DELIMITER));
+  }
+
+  // ================ Object itself =================
+
+  // This list contains the different parts of the channel except the root as every channel has a
+  // root part
+  // So for the channel /root/lao_id/election_id, the elements will be [lao_id, election_id]
+  @NonNull private final List<String> elements;
+
+  private Channel(String... elements) {
+    this.elements = Arrays.asList(elements);
+
+    if (!CHANNEL_VALIDATOR.test(getAsString()))
+      throw new IllegalArgumentException(getAsString() + " is not a valid channel");
+  }
+
+  private Channel(Channel base, String subElement) {
+    List<String> elements = new ArrayList<>(base.elements);
+    elements.add(subElement);
+    this.elements = Collections.unmodifiableList(elements);
+
+    if (!CHANNEL_VALIDATOR.test(getAsString()))
+      throw new IllegalArgumentException(getAsString() + " is not a valid channel");
+  }
+
+  /**
+   * Return a channel that is this channel with one more element underneath
+   *
+   * @param subElement element to append to this channel
+   * @return a new Channel object
+   */
+  public Channel sub(String subElement) {
+    return new Channel(this, subElement);
+  }
+
+  /** @return true if this channel is of the form '/root/lao_id' */
+  public boolean isLaoChannel() {
+    return elements.size() == 1;
+  }
+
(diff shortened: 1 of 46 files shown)


fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Lao.java
```diff
@@ -48,7 +48,8 @@ public Lao(String id) {
     } else if (id.isEmpty()) {
       throw new IllegalArgumentException(" The id of the Lao is empty");
     }
-    this.channel = "/root/" + id;
+
+    this.channel = Channel.ROOT.sub(id);
```
pierluca, line 52: this is not really more understandable than before...

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Channel.java
```diff
@@ -0,0 +1,132 @@
...
+  private static final String ROOT_CHANNEL = "/root";
+  private static final String DELIMITER = "/";
+  // Predicate matching a String with the protocol regex of a channel
+  private static final Predicate<String> CHANNEL_VALIDATOR =
+      Pattern.compile("^/root(/[^/]+)*$").asPredicate();
+
+  /** Root channel, every client is subscribed to it. */
+  public static final Channel ROOT = new Channel();
+
+  public static Channel newChannel(String value) {
```
pierluca, line 31: How about the following ? 
```suggestion
  public static Channel fromString(String value) {
```

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Channel.java
```diff
@@ -0,0 +1,132 @@
...
+    // Avoid creating duplicate instances
+    if (ROOT_CHANNEL.equals(value)) return ROOT;
```
pierluca, line 33: this looks like an ad-hoc version of the flyweight pattern 😅 
Should we go full flyweight ? I'm not sure if it's needed.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Channel.java
```diff
@@ -0,0 +1,132 @@
...
+
+  // ================ Object itself =================
+
+  // This list contains the different parts of the channel except the root as every channel has a
+  // root part
+  // So for the channel /root/lao_id/election_id, the elements will be [lao_id, election_id]
+  @NonNull private final List<String> elements;
+
+  private Channel(String... elements) {
+    this.elements = Arrays.asList(elements);
```
pierluca, line 48: since we're using a classic URI path format, I would suggest calling them segments (the classical term), rather than elements.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Channel.java
```diff
@@ -0,0 +1,132 @@
...
+      throw new IllegalArgumentException(getAsString() + " is not a valid channel");
+  }
+
+  /**
+   * Return a channel that is this channel with one more element underneath
+   *
+   * @param subElement element to append to this channel
+   * @return a new Channel object
+   */
+  public Channel sub(String subElement) {
```
pierluca, line 69: I know anything is going to be more verbose, but this is not very explicit. 
How about the following ? or do you have a better idea ? I don't, but I'm not very happy my idea either.
```suggestion
  public Channel getSubChannel(String subElement) {
```

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Channel.java
```diff
@@ -0,0 +1,132 @@
...
+
+  private Channel(Channel base, String subElement) {
+    List<String> newElems = new ArrayList<>(base.elements);
+    newElems.add(subElement);
+    this.elements = Collections.unmodifiableList(newElems);
+
+    if (!CHANNEL_VALIDATOR.test(getAsString()))
+      throw new IllegalArgumentException(getAsString() + " is not a valid channel");
+  }
+
```
pierluca, line 62: it might be more readable to have a `public Channel getLaoChannel(String laoId)` than using `Channel.root.getSubChannel`, albeit the former could just be an alias for the latter.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Channel.java
```diff
@@ -0,0 +1,132 @@
...
+  /** @return true if this channel is of the form '/root/lao_id' */
+  public boolean isLaoChannel() {
+    return elements.size() == 1;
+  }
+
+  /** @return true if this channel could be an election channel */
+  public boolean isElectionChannel() {
+    return elements.size() == 2;
```
pierluca, line 80: It could be a number of things though. Not sure the logic here is very robust.
I'd at least put a big warning sign 😅

pierluca (changes requested):
Thank you for going through all of this, it's in pretty good shape.
I would just suggest we improve the Channel API a bit, for added clarity.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Lao.java
GabrielFleischer: It's true that it is not that much more readable, but this can still be improved.

I mostly focused on constancy and robustness when designing the creation API because for me, the way channels were created was already decent except for the fact that is used "magic constants".

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Channel.java
GabrielFleischer: Yes, better.
It also avoids the call `Channel.fromChannel()`

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Channel.java
GabrielFleischer: I might have overdone things a little.

I did it only for the Root channel because I already had a defined constant, and it is easy to isolate.
But for clarity, it is probably better to remove that part.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Channel.java
GabrielFleischer: Yes, good idea

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Channel.java
GabrielFleischer: I'd go with `subChannel()`.

I went for a brief function name because I wanted this call to be easy to chain :
Social Media channel :
`lao.getChannel().sub("social").sub(token)`

subChannel is more meaningful and should not be too long.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Channel.java
GabrielFleischer: Yes, good idea

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Channel.java
GabrielFleischer: I did think at naming this `canBeElectionChannel` 
But this is the kind of artifact of the past that we must keep, at least for now.

This is how it was done before, and I don't see and robust way to do the check with the current codebase.

pierluca: I'd just document it as a javadoc warning or remark that this is brittle and needs improvement. Maybe deprecation warning?

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Lao.java
pierluca: That's how I read it. I agree it was already much better, just not much more readable (yet) 👍

pierluca (approved):

pierluca (approved):

PR #845, opened 2022-02-24
Fix roll call reopening issue and scan attendees feature

### LINKED ISSUE #596: Action for reopen roll call is set to open
When clicking the "RE-OPEN ROLL CALL" button nothing happens.

### Expected behavior
It should be possible to reopen a roll call.

### How to reproduce
1. Create a LAO and roll call
2. Open and close the roll call
3. Click the button "RE-OPEN ROLL CALL"

### Version & Environment
This bug was reproduced on:
- Commit: [`fe1ca33`](https://github.com/dedis/student_21_pop/commit/fe1ca3360fcc3d64059d27634adba9b5023e91ed)

##### Front-ends:
- Fe1-Web 

##### Back-ends:
- Be1-Go

##### Environment:
 - Browser Chrome 95

### Workaround
There is no workaround known.

### Possible root cause
The fe1 sends a `dataOpenRollCall` message with the action as `open` even when reopening a roll call. The action should be `reopen` as specified in [dataOpenRollCall.json](https://github.com/dedis/student_21_pop/blob/master/protocol/query/method/message/data/dataOpenRollCall.json).

0xdieman:
Fixes #596 by sending a correct "reopen" message.
Scanning the attendees after reopening or after changing tabs now works.

### DIFF
--- a/fe1-web/src/features/rollCall/components/EventRollCall.tsx
+++ b/fe1-web/src/features/rollCall/components/EventRollCall.tsx
@@ -4,14 +4,14 @@ import PropTypes from 'prop-types';
 import { useSelector } from 'react-redux';
 import { useNavigation } from '@react-navigation/native';
 
-import { Timestamp } from 'core/objects';
+import { Timestamp, Hash, EventTags } from 'core/objects';
 import { QRCode, WideButtonView } from 'core/components';
 import { makeEventGetter } from 'features/events/reducer';
 import { makeCurrentLao } from 'features/lao/reducer';
 import * as Wallet from 'features/wallet/objects';
 import STRINGS from 'resources/strings';
 
-import { requestOpenRollCall } from '../network';
+import { requestOpenRollCall, requestReopenRollCall } from '../network';
 import { RollCall, RollCallStatus } from '../objects';
 
 /**
@@ -57,7 +57,7 @@ const EventRollCall = (props: IPropTypes) => {
         );
         return;
       }
-      requestOpenRollCall(event.idAlias).catch((e) =>
+      requestReopenRollCall(event.idAlias).catch((e) =>
         console.debug('Unable to send Roll call re-open request', e),
       );
     } else {
@@ -66,12 +66,26 @@ const EventRollCall = (props: IPropTypes) => {
         .then(() => {
           navigation.navigate(STRINGS.roll_call_open, {
             rollCallID: event.id.toString(),
-            time: time.toString(),
+            updateID: Hash.fromStringArray(EventTags.ROLL_CALL, lao.id.toString(), event.id.toString(), time.toString())
           });
         })
         .catch((e) => console.debug('Unable to send Roll call open request', e));
     }
   };
+
+  //Scanning attendees should be available only when the Roll Call is in state opened or reopened => idAlias is defined
+  const onScanAttendees = () => {
+    if (!event.idAlias) {
+      console.debug(
+        'Unable to scan attendees, the event does not have an idAlias',
+      );
+      return;
+    }
+    navigation.navigate(STRINGS.roll_call_open, {
+      rollCallID: event.id.toString(),
+      updateID: event.idAlias.toString()
+    });
+  }
 
   const getRollCallDisplay = (status: RollCallStatus) => {
     switch (status) {
@@ -97,7 +111,7 @@ const EventRollCall = (props: IPropTypes) => {
             {isOrganizer && (
               <WideButtonView
                 title="Scan Attendees"
-                onPress={() => console.error('not implemented yet')}
+                onPress={() => onScanAttendees()}
               />
             )}
           </>
@@ -118,8 +132,18 @@ const EventRollCall = (props: IPropTypes) => {
       case RollCallStatus.REOPENED:
         return (
           <>
-            <Text>Re-Opened</Text>
-            <QRCode visibility value={popToken} />
+            {!isOrganizer && (
+              <>
+                <Text>Let the organizer scan your Pop token</Text>
+                <QRCode visibility value={popToken} />
+              </>
+            )}
+            {isOrganizer && (
+              <WideButtonView
+                title="Scan Attendees"
+                onPress={() => onScanAttendees()}
+              />
+            )}
           </>
         );
       default:
--- a/fe1-web/src/features/rollCall/network/RollCallHandler.ts
+++ b/fe1-web/src/features/rollCall/network/RollCallHandler.ts
@@ -7,7 +7,7 @@ import { getEventFromId } from 'features/events/network/EventHandlerUtils';
 import * as Wallet from 'features/wallet/objects';
 import { makeCurrentLao, setLaoLastRollCall } from 'features/lao/reducer';
 
-import { CloseRollCall, CreateRollCall, OpenRollCall } from './messages';
+import { CloseRollCall, CreateRollCall, OpenRollCall, ReopenRollCall } from './messages';
 import { RollCall, RollCallStatus } from '../objects';
 
 /**
@@ -167,14 +167,43 @@ export function handleRollCallCloseMessage(msg: ProcessableMessage): boolean {
 }
 
 /**
- * TODO: Handles a reopen roll call message.
+ * Handles a RollCallReopen message by reopening the corresponding roll call.
  *
- * @param msg
+ * @param msg - The extended message for reopening a roll call
  */
 export function handleRollCallReopenMessage(msg: ProcessableMessage) {
-  console.warn(
-    'A RollCall reopen message was received but its processing logic is not yet implemented:',
-    msg,
-  );
-  return false;
+  if (
+    msg.messageData.object !== ObjectType.ROLL_CALL ||
+    msg.messageData.action !== ActionType.REOPEN
+  ) {
+    console.warn('handleRollCallReopenMessage was called to process an unsupported message', msg);
+    return false;
+  }
+
+  const makeErr = (err: string) => `roll_call/reopen was not processed: ${err}`;
+
+  const storeState = getStore().getState();
+  const lao = getCurrentLao(storeState);
+  if (!lao) {
+    console.warn(makeErr('no LAO is currently active'));
+    return false;
+  }
+
+  const rcMsgData = msg.messageData as ReopenRollCall;
+  const oldRC = getEventFromId(storeState, rcMsgData.opens) as RollCall;
+  if (!oldRC) {
+    console.warn(makeErr("no known roll call matching the 'opens' field"));
+    return false;
+  }
+  // TODO: delete those logs
+  console.log("Update id from reopen = "+rcMsgData.update_id);
+  const rc = new RollCall({
+    ...oldRC,
+    idAlias: rcMsgData.update_id,
+    openedAt: rcMsgData.opened_at,
+    status: RollCallStatus.REOPENED,
+  });
+
+  dispatch(updateEvent(lao.id, rc.toState()));
+  return true;
 }
--- a/fe1-web/src/features/rollCall/screens/RollCallOpened.tsx
+++ b/fe1-web/src/features/rollCall/screens/RollCallOpened.tsx
@@ -37,7 +37,7 @@ const RollCallOpened = () => {
   // FIXME: navigation and route should user proper type
   const navigation = useNavigation<any>();
   const route = useRoute<any>();
-  const { rollCallID, time } = route.params;
+  const { rollCallID, updateID } = route.params;
   const [attendees, updateAttendees] = useState(new Set<string>());
   const [inputModalIsVisible, setInputModalIsVisible] = useState(false);
   const toast = useToast();
@@ -45,7 +45,7 @@ const RollCallOpened = () => {
   const lao = useSelector(laoSelect);
 
   if (!lao) {
-    throw new Error('Impossible to open a Roll Call without being connected to an LAO');
+    throw new Error('Impossible to open a Roll Call without being connected to a LAO');
   }
 
   // This will run only when the state changes
@@ -104,9 +104,8 @@ const RollCallOpened = () => {
   };
 
   const onCloseRollCall = () => {
-    const updateId = Hash.fromStringArray(EventTags.ROLL_CALL, lao.id.toString(), rollCallID, time);
     const attendeesList = Array.from(attendees).map((key: string) => new PublicKey(key));
-    return requestCloseRollCall(updateId, attendeesList)
+    return requestCloseRollCall(updateID, attendeesList)
       .then(() => {
         navigation.navigate(STRINGS.organizer_navigation_tab_home);
       })


fe1-web/src/features/rollCall/components/EventRollCall.tsx
```diff
@@ -57,7 +57,7 @@ const EventRollCall = (props: IPropTypes) => {
         );
         return;
       }
-      requestOpenRollCall(event.idAlias).catch((e) =>
+      requestReopenRollCall(event.idAlias).catch((e) =>
         console.debug('Unable to send Roll call re-open request', e),
       );
```
Tyratox, line 62: Is there a specific reason why in the case of reopening a roll call you have to press an additional button to get to the QR code scanning screen ("Scan Attendees" with onClick=`onScanAttendees`) rather than directly opening this screen as it is the case when opening the roll call for the first time (Line 67)?

Tyratox (approved):
Other than the question I added the code looks good! 👍 :)
(I assume you already noticed that the CI complains about the test coverage for the new code)

fe1-web/src/features/rollCall/components/EventRollCall.tsx
0xdieman: No specific reason for now. Thinking about it, it would make more sense with the current code to make it the same as the opening behavior.
Yes I noticed the bad coverage thing, might convert the pull request to draft while I take care of those...

Tyratox (dismissed):

fe1-web/src/features/rollCall/components/EventRollCall.tsx
```diff
@@ -57,22 +57,56 @@ const EventRollCall = (props: IPropTypes) => {
...
+            return;
+          }
+          navigation.navigate(STRINGS.roll_call_open, {
+            rollCallID: event.id.toString(),
+            updateID: Hash.fromStringArray(
+              EventTags.ROLL_CALL,
+              lao.id.toString(),
+              event.idAlias.toString(),
+              time.toString(),
+            ),
```
pierluca, line 76: By computing the `updateID` here, you've effectively shifted responsibility for building a correctly formatted network message from the network-related code to a GUI component. I would strongly suggest to keep a clean separation between the two.

fe1-web/src/features/rollCall/components/EventRollCall.tsx
```diff
@@ -57,22 +57,56 @@ const EventRollCall = (props: IPropTypes) => {
...
-      );
+      const time = Timestamp.EpochNow();
+      requestReopenRollCall(event.idAlias, time)
+        .then(() => {
+          if (!event.idAlias) {
+            console.debug(
+              'Unable to send roll call re-open request, the event does not have an idAlias',
+            );
+            return;
+          }
```
pierluca, line 68: why stop here, when you've already sent the network request ? it's either too early or too late.

fe1-web/src/features/rollCall/network/RollCallHandler.ts
```diff
@@ -167,14 +167,42 @@ export function handleRollCallCloseMessage(msg: ProcessableMessage): boolean {
...
+  }
+
+  const makeErr = (err: string) => `roll_call/reopen was not processed: ${err}`;
+
+  const storeState = getStore().getState();
+  const lao = getCurrentLao(storeState);
+  if (!lao) {
+    console.warn(makeErr('no LAO is currently active'));
+    return false;
+  }
```
pierluca, line 190: conveniently, due to the latest refactorings, you don't need this anymore.
You can now retrieve the `lao.id` from `msg.laoId`.
The code wasn't updated everywhere with this change for lack of time.

Feel free to change that elsewhere in this file too 😄

fe1-web/src/features/rollCall/network/RollCallHandler.ts
```diff
@@ -167,14 +167,42 @@ export function handleRollCallCloseMessage(msg: ProcessableMessage): boolean {
...
+
+  const rcMsgData = msg.messageData as ReopenRollCall;
+  const oldRC = getEventFromId(storeState, rcMsgData.opens) as RollCall;
+  if (!oldRC) {
```
pierluca, line 194: Would it make sense to (defensively) check if `oldRc.status` is `CLOSED`, since we're re-opening ?

fe1-web/src/features/rollCall/screens/RollCallOpened.tsx
```diff
@@ -37,15 +37,15 @@ const RollCallOpened = () => {
   // FIXME: navigation and route should user proper type
   const navigation = useNavigation<any>();
   const route = useRoute<any>();
-  const { rollCallID, time } = route.params;
+  const { rollCallID, updateID } = route.params;
```
pierluca, line 40: This then means that 2 components need to deal with network-specific stuff.

fe1-web/src/features/rollCall/screens/RollCallOpened.tsx
```diff
@@ -104,9 +104,8 @@ const RollCallOpened = () => {
   };
 
   const onCloseRollCall = () => {
-    const updateId = Hash.fromStringArray(EventTags.ROLL_CALL, lao.id.toString(), rollCallID, time);
     const attendeesList = Array.from(attendees).map((key: string) => new PublicKey(key));
-    return requestCloseRollCall(updateId, attendeesList)
```
pierluca, line 109: I think it was a great idea to remove the `updateId` from here, but I would suggest moving it into `requestCloseRollCall`, by passing it the `lao.id`and `rollCallID`, instead of just an externally-calculated updateId.

This also ensures that whoever calls `requestCloseRollCall` will always compute the hash in the same way (correctly, hopefully 😅 )

fe1-web/src/features/rollCall/components/EventRollCall.tsx
```diff
@@ -57,22 +57,56 @@ const EventRollCall = (props: IPropTypes) => {
...
+            updateID: Hash.fromStringArray(
+              EventTags.ROLL_CALL,
+              lao.id.toString(),
+              event.idAlias.toString(),
+              time.toString(),
+            ),
+          });
+        })
+        .catch((e) => console.debug('Unable to send Roll call re-open request', e));
     } else {
```
pierluca, line 80: it seems that both branches of this if/else are doing very similar things. Try to see if they can be factored !

fe1-web/src/features/rollCall/components/EventRollCall.tsx
```diff
@@ -57,22 +57,56 @@ const EventRollCall = (props: IPropTypes) => {
...
         .catch((e) => console.debug('Unable to send Roll call open request', e));
     }
   };
 
+  // Scanning attendees should be available only when the Roll Call is in state opened or reopened => idAlias is defined
+  const onScanAttendees = () => {
+    if (!event.idAlias) {
+      console.debug('Unable to scan attendees, the event does not have an idAlias');
+      return;
+    }
```
pierluca, line 103: are you trying to assert a case that should always, absolutely be true ? 
If so, you might want to make this error as loud as possible, e.g. `console.error` or even better throw an exception !

pierluca (changes requested):
Good job on your first PR ! 

I'm being very annoying here, forgive me, but I promise it'll get better over the semester 😅

fe1-web/src/features/rollCall/components/EventRollCall.tsx
0xdieman: Turned it into a console.error

fe1-web/src/features/rollCall/network/RollCallHandler.ts
0xdieman: done !

fe1-web/src/features/rollCall/components/__tests__/EventRollCall.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+import { fireEvent, render } from '@testing-library/react-native';
+import STRINGS from 'resources/strings';
+import { Hash, Timestamp } from 'core/objects';
+import { useNavigation } from '@react-navigation/native';
+
+import * as reactRedux from 'react-redux';
+import { mockLao } from '__tests__/utils/TestUtils';
+import { requestOpenRollCall, requestReopenRollCall } from 'features/rollCall/network';
+import EventRollCall from '../EventRollCall';
+import { EventTypeRollCall, RollCall, RollCallStatus } from '../../objects';
```
Xelowak, line 11: You should write imports in the following order (ideally separated with an empty line) :
1. External libraries
2. Everything outside the current feature: core, resources (and other features if needed, but try to avoid this)
3. Everything inside the feature (in this case contained in `features/rollCall`) that can be imported relatively using `../`

fe1-web/src/features/rollCall/components/__tests__/EventRollCall.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+const ID = new Hash('rollCallId');
+const NAME = 'myRollCall';
+const LOCATION = 'location';
+const TIMESTAMP_START = new Timestamp(1620255600);
+const TIMESTAMP_END = new Timestamp(1620357600);
+const ATTENDEES = ['attendee1', 'attendee2'];
+const rollCallState: any = (mockStatus: RollCallStatus) => {
```
Xelowak, line 19:
```suggestion
const createStateWithStatus: any = (mockStatus: RollCallStatus) => {
```

fe1-web/src/features/rollCall/components/__tests__/EventRollCall.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+beforeEach(() => {
+  jest.resetAllMocks();
+  mockSelector = jest.spyOn(reactRedux, 'useSelector').mockReturnValueOnce(mockLao);
+  (useNavigation as jest.Mock).mockImplementation(() => {
+    return {
+      navigate: mockNavigate,
+    };
+  });
+  (requestOpenRollCall as jest.Mock).mockImplementation(() => Promise.resolve());
+  (requestReopenRollCall as jest.Mock).mockImplementation(() => Promise.resolve());
```
Xelowak, line 64: Since you clear all mock at line 56, you should be able to declare all of this outside of the beforeEach. These mocks will be used in each test, so it's clearer if they are defined before everything.

fe1-web/src/features/rollCall/components/__tests__/EventRollCall.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+let mockSelector: jest.SpyInstance<
+  unknown,
+  [
+    selector: (state: unknown) => unknown,
+    equalityFn?: ((left: unknown, right: unknown) => boolean) | undefined,
+  ]
+>;
+
+beforeEach(() => {
+  jest.resetAllMocks();
```
Xelowak, line 56:
```suggestion
  jest.clearAllMocks();
```
Resetting them will completely delete the mock implementation you wrote, so if you declare mocks before, you just need to clear them here.

fe1-web/src/features/rollCall/components/__tests__/EventRollCall.test.tsx
Xelowak, line 64: I don't think you need to declare these. We already have a file [there](https://github.com/dedis/popstellar/blob/master/fe1-web/src/features/rollCall/network/__mocks__/RollCallMessageApi.ts) where you can add these mock requests. Since you've already written `jest.mock('features/rollCall/network')`, the mock should be correctly imported.

fe1-web/src/features/rollCall/components/__tests__/EventRollCall.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+  mockSelector = jest.spyOn(reactRedux, 'useSelector').mockReturnValueOnce(mockLao);
+  (useNavigation as jest.Mock).mockImplementation(() => {
+    return {
+      navigate: mockNavigate,
+    };
+  });
```
Xelowak, line 62: This mock already exists [there](https://github.com/dedis/popstellar/blob/master/fe1-web/src/__mocks__/useNavigationMock.tsx).

fe1-web/src/features/rollCall/components/__tests__/EventRollCall.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+  };
+};
+
+const mockRollCallClosed = RollCall.fromState(rollCallState(RollCallStatus.CLOSED));
+const mockRollCallCreated = RollCall.fromState(rollCallState(RollCallStatus.CREATED));
+const mockRollCallOpened = RollCall.fromState(rollCallState(RollCallStatus.OPENED));
+
+const mockRenderRollCall = (rollCall: RollCall, isOrganizer: boolean) => {
+  return render(<EventRollCall event={rollCall} isOrganizer={isOrganizer} />);
+};
```
Xelowak, line 41: I'd honestly delete this. I don't think it is clearer and the actual render function isn't that long either.

fe1-web/src/features/rollCall/components/__tests__/EventRollCall.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+    const obj = mockRenderRollCall(mockRollCallCreated, false);
+    expect(obj.toJSON()).toMatchSnapshot();
+  });
+  it('should call requestOpenRollCall when the open button is clicked', () => {
+    const usedMockRollCall = mockRollCallCreated;
+    mockSelector.mockReturnValueOnce(usedMockRollCall);
+    const obj = mockRenderRollCall(usedMockRollCall, true);
+    const openRollCallButton = obj.getByText(STRINGS.roll_call_open);
+    fireEvent.press(openRollCallButton);
+    expect(requestOpenRollCall).toHaveBeenCalled();
```
Xelowak, line 79: It's better if you ensure that it has been called the correct number of times with `toHaveBeenCalledTimes`. The same applies to the 2 other tests in this file.

fe1-web/src/features/rollCall/network/__tests__/RollCallHandler.test.ts
```diff
@@ -0,0 +1,120 @@
...
+  ...rollCallStateClosed,
+  status: RollCallStatus.REOPENED,
+};
+
+const mockRollCallClosed = RollCall.fromState(rollCallStateClosed);
+const mockRollCallCreated = RollCall.fromState(rollCallStateCreated);
+const mockRollCallOpened = RollCall.fromState(rollCallStateOpened);
+const mockRollCallReopened = RollCall.fromState(rollCallStateReopened);
+
+const mockMsg = (type: ActionType, status: RollCallStatus, rollCallState: any) => {
```
Xelowak, line 60:
```suggestion
const createMockMsg = (type: ActionType, status: RollCallStatus, rollCallState: any) => {
```

fe1-web/src/features/rollCall/network/__tests__/RollCallHandler.test.ts
```diff
@@ -0,0 +1,120 @@
...
+      action: type,
+      ...rollCallState,
+      proposed_start: rollCallState.proposedStart,
+      proposed_end: rollCallState.proposedEnd,
+      opened_at: rollCallState?.openedAt,
+      update_id: rollCallState?.idAlias,
+      closed_at: rollCallState?.closedAt,
+    },
+  };
+};
```
Xelowak, line 81: I'd put all of this in a `region`. I allows us to collapse it to make it cleaner and to access the tests without having to scroll a lot. Basically, you just have to write `// region regionName` at the beginning of the data initialization, end an `// endregion` at the end. I'll let you name it however you want.

fe1-web/src/features/rollCall/network/__tests__/RollCallHandler.test.ts
```diff
@@ -0,0 +1,120 @@
...
+
+beforeEach(() => {
+  jest.resetAllMocks();
```
Xelowak, line 84: Again, are you sure resetting is the thing you want ?

fe1-web/src/features/rollCall/network/__tests__/RollCallHandler.test.ts
```diff
@@ -0,0 +1,120 @@
...
+    return {
+      getState: jest.fn(),
+    };
+  });
+});
+
+describe('RollCallHandler', () => {
+  it('should create a correct RollCall object from msgData in handleRollCallCreateMessage', async () => {
+    const usedMockMsg = mockMsg(ActionType.CREATE, RollCallStatus.CREATED, rollCallStateCreated);
+    const mockCall = addEvent as any;
```
Xelowak, line 95: I'm not sure about what you're trying to do here, but it seems weird to me. Do you want to mock the addEvent function ?
The same applies to the next tests.

fe1-web/src/features/rollCall/components/EventRollCall.tsx
```diff
@@ -57,20 +56,25 @@ const EventRollCall = (props: IPropTypes) => {
         );
         return;
       }
-      requestOpenRollCall(event.idAlias).catch((e) =>
+      requestReopenRollCall(event.idAlias).catch((e) =>
```
Xelowak, line 59: Wouldn't it be clearer to split this function into 2 ? Having a function called `onOpenRollCall` that opens and reopens roll call based on a boolean value seems quite odd to me.

Xelowak (changes requested):
That's a nice first PR ! I really apreciate the fact that you dove into jest tests immediately. I've made quite a bunch of comments, so feel free to reach to me if something is unclear.

I'm not really sure about your way of handling mocks. If you want to mock a certain function from a file (and only one) and you won't expect anything from it, you should use the "mock factory" as I call it. It might be easier for you to use and you won't have to do as much casting `as jest.Mock`. For example, if you want to mock `updateEvent` from `features/events/reducer`, you can write :
`jest.mock('features/events/reducer', () => ({
    updateEvent: jest.fn(() => {}),
}));`.
By the way, I'd suggest you separate these `it` with an empty line, they need to breathe !

fe1-web/src/features/rollCall/components/EventRollCall.tsx
```diff
@@ -57,53 +56,24 @@ const EventRollCall = (props: IPropTypes) => {
...
-              EventTags.ROLL_CALL,
-              lao.id.toString(),
-              event.idAlias.toString(),
-              time.toString(),
-            ),
-          });
-        })
-        .catch((e) => console.debug('Unable to send Roll call re-open request', e));
+      requestReopenRollCall(event.idAlias).catch((e) =>
+        console.debug('Unable to send Roll call re-open request', e),
```
pierluca, line 60: we also have toast messages (just search for "toast" in the codebase) to indicate to the user in a clearly visible way that something failed.
It's better than expecting the user to have the developer tools open 😄

fe1-web/src/features/rollCall/components/__tests__/EventRollCall.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+jest.mock('features/rollCall/network');
+const mockNavigate = jest.fn();
+
+let mockSelector: jest.SpyInstance<
+  unknown,
+  [
+    selector: (state: unknown) => unknown,
+    equalityFn?: ((left: unknown, right: unknown) => boolean) | undefined,
+  ]
+>;
```
pierluca, line 53: this looks like black magic to me 😅 
I guess from `mockSelector` what it does, but I'd try to see if there's a cleaner way (e.g. could `spyOn` make your life easier?) or documenting it in a comment.

pierluca (commented):
It's nice to see the progress on the code 😄 
I've just added a couple of points to @Xelowak 's remarks.
Keep up the good work !

fe1-web/src/features/rollCall/components/__tests__/EventRollCall.test.tsx
Xelowak: That's true, I think there might be a better way to handle this.

fe1-web/src/features/rollCall/components/__tests__/EventRollCall.test.tsx
0xdieman: Okay I changed that, the requests works fine when defined before but it seems like the mockSelector does not work as expected when placed before, seems like the mockReturnValueOnce is removed when cleaning, not sure

fe1-web/src/features/rollCall/network/__tests__/RollCallHandler.test.ts
0xdieman: I did but I had a type error when directly using addEvent without the cast to any, now it seems like it's working without it

fe1-web/src/features/rollCall/components/__tests__/EventRollCall.test.tsx
0xdieman: I need to declare them because `EventRollCall` calls a catch on them which does not work when undefined, I used [this](https://github.com/dedis/popstellar/blob/master/fe1-web/src/features/rollCall/screens/__tests__/RollCallOpened.test.tsx#L56) as a template because I wasn't sure about the cleanest way of doing it. Also what would be the reason to use the [external mock request file](https://github.com/dedis/popstellar/blob/master/fe1-web/src/features/rollCall/network/__mocks__/RollCallMessageApi.ts) if the goal is to create a mock function returning a promise, which is a one liner ?

fe1-web/src/features/rollCall/components/EventRollCall.tsx
0xdieman: Yep, seems odd to me too !

fe1-web/src/features/rollCall/components/EventRollCall.tsx
0xdieman: Right, toast seems a better idea !

fe1-web/src/features/rollCall/components/__tests__/EventRollCall.test.tsx
0xdieman: The goal was to have a matching type for the `spyOn` function because my IDE seemed perplex and angry without it, but I changed my IDE and also changed the type to `any`, which does not make it clearer but maybe less confusing ?...

fe1-web/src/features/rollCall/components/__tests__/EventRollCall.test.tsx
Xelowak: Ok, yes it would be quite logical.

Xelowak: Sure, it works fine that way. The purpose if doing it with an external file is that you're not going to specify the function you want to mock in each file where you're going to use it. You just have to write `jest.mock('filePath')` and import the function that you want to use directly. It might be unclear, but we can talk about it in the TypeScript meeting if needed.

Xelowak (commented):
Nicely done ! Thanks a lot for the changes.
I've made some additional comments, but those are minor things.

pierluca (commented):
I agree with @Xelowak 's last (minor) remarks. Very good job overalls 👍

pierluca (commented):
I believe we're close to a mergeable state, but there are some CI checks failing.
Let me know when those are addressed and I'm happy to merge !

fe1-web/src/features/rollCall/components/EventRollCall.tsx
```diff
@@ -2,22 +2,18 @@ import { useNavigation } from '@react-navigation/native';
...
+import { useSelector } from 'react-redux';
 
 import { QRCode, WideButtonView } from 'core/components';
-import { Timestamp } from 'core/objects';
 import { makeEventGetter } from 'features/events/reducer';
 import { selectCurrentLao } from 'features/lao/reducer';
 import * as Wallet from 'features/wallet/objects';
 import STRINGS from 'resources/strings';
 
+import { FOUR_SECONDS } from '../../../resources/const';
```
pierluca, line 14:
```suggestion
import { FOUR_SECONDS } from 'resources/const';
```

pierluca (approved):
Very minor suggestion, but LGTM ! 👍

fe1-web/src/features/rollCall/components/__tests__/__snapshots__/EventRollCall.test.tsx.snap
```diff
@@ -0,0 +1,15 @@
...
+exports[`EventRollCall should correctly render 1`] = `
+Array [
+  <Text>
+    Roll Call
+  </Text>,
+  <Text>
+    Not Open yet
+  </Text>,
+  <Text>
+    Be sure to have set up your Wallet
```
Tyratox, line 12: I think in the future it would be nice having a test for rendering the component multiple times. Once with the wallet not being setup, once with the wallet being setup and the roll call not being open yet, etc. :)

fe1-web/src/features/rollCall/screens/RollCallOpened.tsx
```diff
@@ -71,18 +71,13 @@ const RollCallOpened = () => {
...
-        type: 'danger',
-        placement: 'top',
-        duration: FOUR_SECONDS,
-      },
-    );
+    toast.show(err, {
+      type: 'danger',
+      placement: 'top',
+      duration: FOUR_SECONDS,
+    });
```
Tyratox, line 80: Why is this changed back to the old way of handling errors? Did I miss something?

Xelowak (dismissed):

fe1-web/src/features/rollCall/screens/RollCallOpened.tsx
0xdieman: Nope, that must be a bad conflict fixing thing I did

Tyratox (dismissed):

fe1-web/src/features/rollCall/components/__tests__/__snapshots__/EventRollCall.test.tsx.snap
0xdieman: I'll think about that

0xdieman:
My dream is that one day this becomes mergeable :')

fe1-web/src/features/rollCall/screens/RollCallOpened.tsx
```diff
@@ -71,13 +71,18 @@ const RollCallOpened = () => {
...
     console.error(err);
-    toast.show(err, {
-      type: 'danger',
-      placement: 'top',
-      duration: FOUR_SECONDS,
-    });
+    // The "err" object might be an exception, take the message property if it exists
+    toast.show(
+      err?.message ||
+        (typeof err === 'string' ? err : 'Unkown error, please check the console and report it!'),
```
pierluca, line 79: are you sure you need all of this logic ?
I would tend to believe that `err.toString()` is probably enough to achieve this ?

0xdieman: Now I also tend to believe that !

fe1-web/src/features/rollCall/components/EventRollCall.tsx
```diff
@@ -66,27 +69,26 @@ const EventRollCall = (props: IPropTypes) => {
...
   const onReopenRollCall = () => {
-    if (!event.idAlias) {
+    if (eventHasBeenOpened) {
+      requestReopenRollCall(event.idAlias).catch((e) => {
+        makeToastErr('Unable to send Roll call re-open request');
+        console.debug('Unable to send Roll call re-open request', e);
+      });
+    } else {
       makeToastErr('Unable to send roll call re-open request, the event does not have an idAlias');
       console.debug('Unable to send roll call re-open request, the event does not have an idAlias');
```
pierluca, line 79: if the strings are repeated identically, you can also consider putting them in a variable

pierluca (approved):

PR #846, opened 2022-02-24
[Needs work] Adding a greeting message when connecting to the server

pierluca:
This feature was being added by a student (@celinecamacho) last semester.
It defines a message that the back-end sends to the front-end when a front-end connects to it.
This allows a server to advertise its address and public key, in turn allowing the front-ends to know the key of the server they're talking with. This could enable security measures like *certificate pinning* on the front-end's part.

We believe it would be worth integrating this feature across all front-ends / back-ends.

The code, as it stands, probably needs some (re)work as it wasn't tested or merged with the repository, and the repository has kept moving since.


protocol/query/method/greeting.json
```diff
@@ -0,0 +1,45 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://raw.githubusercontent.com/dedis/student_21_pop/master/protocol/query/method/greeting.json",
```
pierluca, line 3:
```suggestion
  "$id": "https://raw.githubusercontent.com/dedis/popstellar/master/protocol/query/method/greeting.json",
```

protocol/query/method/greeting.json
```diff
@@ -0,0 +1,45 @@
...
+          "type": "string",
+          "contentEncoding": "base64",
+          "$comment": "Note: the string is encoded in Base64"
+        },
+
+        "address": {
+          "description": "Address of the server",
+          "type": "string"
+        }
+      },
```
pierluca, line 34: it might be worth to add a list of `peers` that the server is connected to, i.e. other servers it is communicating with.

Tyratox:
Work was continued in #917

Issue #703, opened 2021-12-26
[BUG] Fe1-Web election end is not implemented correctly


rjaccard:
### Description (Actual behavior)
We cannot close an election early as there is no close election button.

### Expected behavior
The expected behaviour can be discussed. It can be very convenient to have the possibility to close an election early but the organizer can misuse this possibility to close an election when his/her option is winning.

##### Front-ends:
- [x] Fe1-Web (please include browser's names & version)
- [ ] Fe2-Android (please specify if phone or emulation, and Android version)
- [ ] Not applicable

##### Back-ends:
- [x] Be1-Go
- [ ] Be2-Scala
- [ ] Not applicable

### Workaround
Wait the end of the election.

### Impact
There is no big impact, it is just not very convenient.

<img width="1410" alt="Screen Shot 2021-12-26 at 4 11 18 pm" src="https://user-images.githubusercontent.com/61380023/147421429-8655e770-e3d5-4415-b2f2-45afc6b69d23.png">

arnauds5:
Closing an election earlier/later could be done in the future with a consensus between organizer/witnesses to avoid an abuse of organizer where he could choose when to close the election to win.

pierluca:
Currently the implementation is very wrong:
-> after the election end time, the fe1-web updates its _internal state_ , based on the time, without being triggered by a broadcast message, contrary to the global application architecture.
-> when an election is open, there should instead be a button to terminate the election and send the right message [election#end](https://github.com/dedis/popstellar/blob/master/docs/messageData.md#ending-an-election-electionend)

Tyratox:
Currently when sending an [election#end](https://github.com/dedis/popstellar/blob/master/docs/messageData.md#ending-an-election-electionend) message before the end time has passed, the go server replies with

```json
{
  "jsonrpc": "2.0",
  "id": 3,
  "error": {
    "code": -4,
    "description": "failed to handle method: failed to publish: failed to handle a publish message: failed to process \"end\" action: failed to end election: invalid election#end message: election end created at is 1645777760, should be greater or equal to defined end time 1645781233"
  }
}
```

and logs 

```
[...]

PR #848, opened 2022-02-25
Store election result messages

Tyratox:
When working on https://github.com/dedis/popstellar/issues/703 I noticed that election result messages are not sent to a new client connecting after sending a catchup message. The reason is that the election#result messages are not stored when broadcasting them. This pull request fixes this.

Compare
https://github.com/dedis/popstellar/blob/ab162f1cb2122f10dafbac003979abbfee4ac862/be1-go/channel/election/mod.go#L343-L344

and

https://github.com/dedis/popstellar/blob/ab162f1cb2122f10dafbac003979abbfee4ac862/be1-go/channel/election/mod.go#L384-L385

to

https://github.com/dedis/popstellar/blob/ab162f1cb2122f10dafbac003979abbfee4ac862/be1-go/channel/election/mod.go#L430-L435.

### DIFF
--- a/be1-go/channel/election/mod.go
+++ b/be1-go/channel/election/mod.go
@@ -428,7 +428,6 @@ func (c *Channel) broadcastElectionResult() error {
 		MessageID:         messagedata.Hash(newData64, signature),
 		WitnessSignatures: []message.WitnessSignature{},
 	}
-	
 	c.inbox.StoreMessage(electionResultMsg)
 	c.broadcastToAllClients(electionResultMsg)
 
--- a/be1-go/channel/election/mod.go
+++ b/be1-go/channel/election/mod.go
@@ -428,7 +428,8 @@ func (c *Channel) broadcastElectionResult() error {
 		MessageID:         messagedata.Hash(newData64, signature),
 		WitnessSignatures: []message.WitnessSignature{},
 	}
-
+	
+	c.inbox.StoreMessage(electionResultMsg)
 	c.broadcastToAllClients(electionResultMsg)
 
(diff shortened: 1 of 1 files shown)


Warhedgehog (changes requested):
Per the slack message, without further clarifications, I do not think it is a desirable change since the documentation says that the inbox storage is for incoming message

Tyratox:
> Noémien: IIRC we decided to store everything in the inbox, and filter out unnecessary messages when sending a catchup. In this case implementing a persistence layer would only require to persist the inbox, and replay the inbox content to recover the state of an organiser.

It seems then that this comment should be updated as well? Should I include that in this PR or will you do this on a different branch & PR?

Tyratox:
Good points! I do not have a full overview over the project but it seems like a good solution as other servers, for example the witnesses, probably should get the results as well. I will ask in the evoting channel for confirmation and update this reply as soon as I get an answer.

In any case it should just be made sure that this also happens if there is no other server.

Tyratox:
> [Louis-Henri](https://app.slack.com/team/UMRHRP1K5)  [vor 18 Stunden](https://dedis.slack.com/archives/C03442XUBCK/p1646239534435329?thread_ts=1646234776.541329&cid=C03442XUBCK)
> Election Result I believe should be signed by the witnesses to say that we witness and agree to the Election result (bearbeitet) 
> 
> Before it is sent to the clients. Otherwise clients are unsure whether to accept the result or not.
> 
> This would require witness signing between the organizers and witnesses which relies on server-2-server communication.

Issue #843, opened 2022-02-24
[E-Voting] [Protocol]: The cast vote specification currently is ambigous and the implementations differ


Tyratox:
# What is the problem

In the protocol specification https://github.com/dedis/popstellar/blob/master/protocol/query/method/message/data/dataCastVote.json, for a vote the comment states

> Hash : HashLen('Vote', election_id, question_id, (vote_index(es)|write_in)), concatenate vote indexes - must use delimiter

To make this hash unique three things are missing in case there are multiple indices:
- An absolute order
- A concrete specification of what delimiter should be used
- Make sure there are no duplicate entries

# Current implementations

Currently fe1 does not use any delimiter at all (and does not support write-in votes):
https://github.com/dedis/popstellar/blob/9b44393cb36da07751c2bd42fdf9614e5e8da087/fe1-web/src/features/evoting/components/EventElection.tsx#L63-L85

This means that the votes `[1,2]` and `[12]` will end up having the same hash!

In contrast f2 currently uses `List.toString()` which in some cases might be fine but can have different underlying implementations as `List` is just an interface and the `toString()` method can be overridden. Moreover it might be possible that additional strings such as brackets are pre- and appended.

https://github.com/dedis/popstellar/blob/e803d11e2bbd7c91c973c951da37d5e269bace60/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Election.java#L249-L260

# What to do

- [x] Decide on an absolute order of indices. EDIT: Fixed to ascending order on integers, i.e. `<`. ~Suggestions: Since the entries are supposed to be unique (you can only select an option once), the natural order `<` on integers can be used.~
- [x] Decide on a delimiter. EDIT: Fixed to `,` ~Suggestion: `,`~
- [x] Update specification and documentation
- [ ] Adapt fe1 implementation to correctly generate the hash based on the chosen order and delimiter
- [ ] Adapt fe2 implementation to correctly generate the hash based on the chosen order and delimiter
- [ ] Adapt be1 implementation to correctly verify the hash based on the chosen order and delimiter
- [ ] Adapt b2 implementation to correctly verify the hash based on the chosen order and delimiter

# Discussion
Are there options you would prefer over the suggested ones or do you agree? We should have something everyone agrees to since this changes the protocol specification.

Issue #651, opened 2021-12-11
Adding an OpenElection message to the protocol


arnauds5:
Currently the start of an election is only based on the start time from the setup.
Organizer/witnesses can aggree to start it using a consensus, but they need to inform others that it was accepted.
One way of doing that is by adding an OpenElection message and by allowing vote only after it was opened (and before it was ended).

An election will have the following phases:
Setup → **Open** → Cast vote(s) → End → Result

An OpenElection data will be something like this (and will be publish in the election channel) : 
```
{
    "$schema": "http://json-schema.org/draft-07/schema#",
    "$id": "https://raw.githubusercontent.com/dedis/student_21_pop/master/protocol/query/method/message/data/dataOpenElection.json",
    "description": "Match an OpenElection query",
    "type": "object",
    "properties": {
        "object": {
            "const": "election"
        },
        "action": {
            "const": "open"
        },
        "lao": {
            "type": "string",
            "contentEncoding": "base64",
            "$comment": "ID of the LAO"
        },
        "election": {
            "type": "string",
            "contentEncoding": "base64",
            "$comment": "ID of the election"
        },
        "opened_at": {
            "description": "[Timestamp] start time",
            "type": "integer",
            "minimum": 0
        }
    },
    "additionalProperties": false,
    "required": [
        "object",
        "action",
        "lao",
        "election",
        "opened_at"
    ]
}
```

The modifications will need to be implemented in both backend/frontend.

This is up to discussion, what do you think ?

rjaccard:
This is a good idea! Very coherent with the current protocol and this fits the consensus very well. It might be a bit late in the semester to implement this change on the 4 systems but worths investigating the next semester I believe.

Tesa27:
All the sub-systems now support open election and the protocol specification changed to reflect this, since it all the work has been merged to master we can close this issue

PR #850, opened 2022-02-28
Changes to the election process

Tyratox:
This PR changes the election process. Based on the work of @arnauds5 in #657, the protocol is updated to include an election#open message with the following schema:

```json
{
    "$id": "https://raw.githubusercontent.com/dedis/popstellar/master/protocol/query/method/message/data/dataOpenElection.json",
    "description": "Match an OpenElection query",
    "type": "object",
    "properties": {
        "object": {
            "const": "election"
        },
        "action": {
            "const": "open"
        },
        "lao": {
            "type": "string",
            "contentEncoding": "base64",
            "$comment": "ID of the LAO"
        },
        "election": {
            "type": "string",
            "contentEncoding": "base64",
            "$comment": "ID of the election"
        },
        "opened_at": {
            "description": "[Timestamp] start time",
            "type": "integer",
            "minimum": 0
        }
    },
    "required": [
        "object",
        "action",
        "lao",
        "election",
        "opened_at"
    ]
}
```

With this change the state machine for elections is now ```Setup → Open → Cast vote(s) → End → Result``` rather than ```Setup → Cast vote(s) → End → Result```. This allows the frontend UI updates to be completely passive, i.e. they do not have to update the local states by themselves but can instead wait for a message telling them to do so.

Hence after this PR is merged, https://github.com/dedis/popstellar/issues/651 can be closed.

The possibility of closing an election before the `end_time` was discussed [here](https://github.com/dedis/popstellar/issues/703#issuecomment-1049606670) and should also be implemented by the backends in the future.

---

In addition to this new election#open message, there are some other (smaller) changes in this PR:
- Show the "Terminate Election / Tally Votes" button while the election is in the `RUNNING` state. It will send an [election#end](https://github.com/dedis/popstellar/blob/master/docs/messageData.md#ending-an-election-electionend) message and thus this button should work as soon as the used backend implemented the ability to end elections early.
- Fix the display of the election results (There was no issue for this but with the provided code it did not work for me)
- Set the status of an election to `FINISHED` if an [election#end](https://github.com/dedis/popstellar/blob/master/docs/messageData.md#ending-an-election-electionend) message is received.
- Set the status of an election to `RUNNING` if an election#open message is received.
- Refactoring of `EventElection.tsx`, also in preperation of #843
  - Removal of proactive state updates at `start_time` and `end_time` in `EventElection` as described above, also see [here](https://github.com/dedis/popstellar/issues/703#issuecomment-1049606670).
- Add tests for `CastVote.tsx` analogous to `SetupElection.tsx`
- New typescript mapped types called `RemoveMethods` and `MessageDataProperties` that allows more strict type-checking for some constructors such as `SetupElection.ts, OpenElection.ts`, etc. The basic idea is to replace `Partial<T>` with `RemoveMethods<T>` which will make the input of non-optional fields non-optional (`Partial<T>` allows to pass undefined fields).
- Removing some `any` types

I am really sorry for this big PR. It is clear to me that I should have split this into multiple smaller ones, at first I did not know in which direction I will work since I only just started exploring the project in detail. I will make sure that in the future they are smaller!


pierluca:
Careful: I wouldn't assume that the backends know the current time (or that it is correct, for that matter), this is not trivial, as it would require implementing a scheduler and dealing with timezones and geolocating the LAO, in cases where the backend's server is not configured to be in the same timezone as the LAO.

On the other hand, we could keep this simple by imagining that an organizer (or any of the witnesses) can decide to open the election, come the right time.

Tyratox:
Right now when the backend receives an election end message, an error is thrown if the end_time of the election is in the future. (That is why we cannot close elections early for now) Thus I thought that is an assumption we are working with but you are right that this is not trivial. I am not sure if the election end and open time are the only places where the backend currently makes this assumption.

---

That is also a possibility. This would make the `start_time` and `end_time` fields of an election redundant, right?

pierluca:
Indeed, we're using the backend time to check that the timestamp is correct as a "rough" solution, but aligning the solution with the philosophy of trusting the organizers + witnesses to behave (or report on each other's misbehaving), anything that they can agree on would work (-> consensus).

The idea of having "start_time" and "end_time", even if the actual start and end time are managed by the organizer + witnesses, is that they should be communicated to all the humans involved. Witnesses could also decide NOT to authorize / consent to closing an election ahead of the end_time. Think of it as a "social" contract that the organizers and witnesses are "socially" enforcing, and the protocol is just there to support that.

Tyratox:
Ah I see, that makes sense! Then I will for now (until consensus is fully implemented) add a button to the not yet started elections that sends the newly added election#open message.

fe1-web/src/features/evoting/components/EventElection.tsx
```diff
@@ -37,55 +32,21 @@ const styles = StyleSheet.create({
   } as TextStyle,
 });
 
-const EventElection = (props: IPropTypes) => {
+const EventElection: FunctionComponent<IPropTypes> = (props) => {
```
pierluca, line 35: This was recently brought to my attention
https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/function_components/
https://github.com/facebook/create-react-app/pull/8177

Also, in general, the codebase is not relying on the explicit `FunctionComponent` typing.

fe1-web/src/features/evoting/components/EventElection.tsx
```diff
@@ -37,55 +32,21 @@ const styles = StyleSheet.create({
...
-        election.id.toString(),
-        election.questions[idx].id,
-        concatenateIndexes(item),
-      ).valueOf(),
-      question: election.questions[idx].id,
-      vote: item,
-    }));
-    return votes;
-  };
+  const [selectedBallots, setSelectedBallots] = useState<{ [questionIndex: number]: Set<number> }>(
```
pierluca, line 43: I would strongly suggest creating a type for the state. It would help with documenting its expected values.

fe1-web/src/features/evoting/network/ElectionHandler.ts
```diff
@@ -35,7 +37,7 @@ export function handleElectionSetupMessage(msg: ProcessableMessage): boolean {
   const storeState = getStore().getState();
   const lao = getCurrentLao(storeState);
   if (!lao) {
-    console.warn(makeErr('no LAO is currently active'));
+    console.warn(makeErr(STRINGS.no_active_lao));
     return false;
   }
```
pierluca, line 43: Github doesn't allow me to comment on the lines below (48, 60) but it seems that we're only looking for the `lao.id` and we probably shouldn't need the full `Lao` object.

fe1-web/src/features/evoting/network/ElectionHandler.ts
```diff
@@ -64,6 +66,40 @@ export function handleElectionSetupMessage(msg: ProcessableMessage): boolean {
...
+  const ElectionOpenMsg = msg.messageData as OpenElection;
+  const election = getEventFromId(storeState, ElectionOpenMsg.election) as Election;
+  if (!election) {
+    console.warn(makeErr('No active election to end'));
+    return false;
+  }
+
+  // Change election status here such that it will change the election display in the event list
+  election.electionStatus = ElectionStatus.RUNNING;
+  dispatch(updateEvent(lao.id, election.toState()));
```
pierluca, line 99:
```suggestion
  dispatch(updateEvent(msg.laoId, election.toState()));
```

fe1-web/src/features/evoting/network/ElectionHandler.ts
```diff
@@ -64,6 +66,40 @@ export function handleElectionSetupMessage(msg: ProcessableMessage): boolean {
...
+    console.warn('handleElectionOpenMessage was called to process an unsupported message', msg);
+    return false;
+  }
+  const makeErr = (err: string) => `election/open was not processed: ${err}`;
+  const storeState = getStore().getState();
+  const lao = getCurrentLao(storeState);
+  if (!lao) {
+    console.warn(makeErr(STRINGS.no_active_lao));
+    return false;
+  }
```
pierluca, line 89: we may not need this 😄

fe1-web/src/features/evoting/network/ElectionMessageApi.ts
```diff
@@ -41,46 +41,62 @@ export function requestCreateElection(
...
  *
- * @param election_id - The id of the ongoing election
- * @param votes - The votes to be added
+ * @param election - The election for which the vote should be casted
+ * @param votes - The votes to be added, a map from question index to the set of selected answer indices
  */
-export function castVote(election_id: Hash, votes: Vote[]): Promise<void> {
+export function castVote(
+  election: Election,
+  votes: { [questionIndex: number]: Set<number> },
```
pierluca, line 56: an explicit type might improve readability

fe1-web/src/features/evoting/network/ElectionMessageApi.ts
```diff
@@ -41,46 +41,62 @@ export function requestCreateElection(
...
+      // sort it by index
+      .sort((a, b) => a.index - b.index)
+      // and add an id to all votes as well as the matching question id
+      .map<Vote>(({ index, selectionOptions }) => ({
+        id: CastVote.computeVoteId(election, index, selectionOptions).valueOf(),
+        question: election.questions[index].id,
+        // sort votes in ascending order
+        vote: [...selectionOptions].sort((a, b) => a - b),
+      })),
   });
```
pierluca, line 79: This code looks very smart, but in general you want your code to look boring.
The more boring and obvious it is, the less effort it requires to maintain 😁 

Because the data processing logic is nested inside the object creation, it's both very hard to read and nearly impossible to step through in a debugger. I would strongly suggest to:
1. split this in multiple separate statements.
2. consider hiding the data manipulation in a dedicated function, taking in your `votes` variable and returning your `Vote[]` result.

fe1-web/src/features/evoting/network/messages/CastVote.ts
```diff
@@ -90,4 +91,35 @@ export class CastVote implements MessageData {
...
+    selectionOptions: Set<number>,
+  ): Hash {
+    return Hash.fromStringArray(
+      EventTags.VOTE,
+      election.id.toString(),
+      election.questions[questionIndex].id,
+      // Important: A standardized order is required, otherwise the hash cannot be verified
+      // Even more important: A standardized delimiter has to be used to disambiguate [1,0] from [10]
+      // See https://github.com/dedis/popstellar/issues/843 for details
+      // TODO: Update after discussion in #843 is finished
```
pierluca, line 115: TODO - just a reminder 😄

fe1-web/src/features/evoting/network/messages/EndElection.ts
```diff
@@ -17,11 +19,7 @@ export class EndElection implements MessageData {
 
   public readonly registered_votes: Hash;
 
-  constructor(msg: Partial<EndElection>) {
-    if (!msg.election) {
-      throw new ProtocolError("Undefined 'id' parameter encountered during 'EndElection'");
-    }
```
pierluca, line 23: 😿 ?

fe1-web/src/features/evoting/network/messages/__tests__/SetupElection.test.ts
```diff
@@ -24,6 +25,10 @@ const TIMESTAMP_BEFORE = new Timestamp(1609445600);
 let electionId: Hash;
 let mockQuestionObject1: Question;
 let mockQuestionObject2: Question;
+
+// As discussed on slack, in these tests we should assume that the input to the messages is
```
pierluca, line 29: I'm sure the Slack discussion will be quickly forgotten or unknown to most readers 😄 
```suggestion
// In these tests, we should assume that the input to the messages is
```

fe1-web/src/features/evoting/objects/Election.ts
```diff
@@ -127,8 +128,12 @@ export class Election {
     this.end = obj.end;
     this.questions = obj.questions;
     this.questionResult = obj.questionResult;
-    // Sets the election status automatically
-    this.electionStatus = Election.getElectionStatus(obj.start, obj.end);
+    // Sets the election status automatically if none is passed
+    if (obj.electionStatus === undefined) {
+      this.electionStatus = Election.getElectionStatus(obj.start, obj.end);
```
pierluca, line 133: Would it make sense to default to `NOT_RUNNING` as a more robust option ? 
We'd expect messages to trigger the change anyway, under the new functioning.

protocol/examples/messageData/election_open.json
```diff
@@ -0,0 +1,7 @@
+{
+    "object": "election",
+    "action": "open",
+    "lao": "fzJSZjKf-2cbXH7kds9H8NORuuFIRLkevJlN7qQemjo=",
+    "election": "zG1olgFZwA0m3mLyUqeOqrG0MbjtfqShkyZ6hlyx1tg=",
```
pierluca, line 5: is this based on the same data as the other example jsons ? 😄

fe1-web/src/features/evoting/screens/CreateElection.tsx
```diff
@@ -97,7 +97,6 @@ const CreateElection = ({ route }: any) => {
   const buttonsVisibility: boolean = electionName !== '' && !getQuestionObjects().some(isInvalid);
 
   const createElection = () => {
-    console.log(getQuestionObjects());
```
pierluca, line 100: 👍

pierluca (changes requested):
Are you always going to make me work this hard? 🤣 
Jokes aside, pretty good job ! There are a few minor points here and there, but it looks pretty good already.

Disclaimer: I have not reviewed the tests, but @Xelowak will kindly do that tomorrow.

fe1-web/src/__tests__/utils/TestUtils.ts
```diff
@@ -4,11 +4,19 @@ import { Lao, LaoState } from 'features/lao/objects';
...
+});
+
+export const mockPublicKey2 = testKeyPair.publicKey2;
+export const mockPrivateKey2 = testKeyPair.privateKey2;
+
+export const mockKeyPair2 = KeyPair.fromState({
+  publicKey: mockPublicKey2,
+  privateKey: mockPrivateKey2,
 });
+
```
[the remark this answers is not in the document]
Tyratox: Ah I did not realize this, thank you :)

docs/messageData.md
```diff
@@ -1211,6 +1213,71 @@ The election may allow write-in or have ballot options.
 
 ```
 
+## Opening an Election (election#open)
+🧭 **RPC Message** > **RPC payload** (*Query*) > **Query payload** (*Publish*) >
+**Mid Level** > **High level** (*election#open*)
+
+The election can be opened by publishing an election/open message on the election channel. This message indicates when votes begin to be valid. Until consensus or other changes are implemented, this message is automatically sent by the backends at the start_time of an election and allows the frontend to trigger state changes.
```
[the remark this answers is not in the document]
Tyratox: Whoops! Yes I have written this before the discussion on GitHub 😅

docs/messageData.md
```diff
@@ -1211,6 +1213,71 @@ The election may allow write-in or have ballot options.
...
+        "lao": {
+            "type": "string",
+            "contentEncoding": "base64",
+            "$comment": "ID of the LAO"
+        },
+        "election": {
+            "type": "string",
+            "contentEncoding": "base64",
+            "$comment": "ID of the election"
+        },
```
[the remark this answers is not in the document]
Tyratox: I literally copy-pasted this part of @arnauds5's work in #657 as the protocol part seemed approved already :)

fe1-web/src/core/redux/__mocks__/GlobalStore.ts
```diff
@@ -1,13 +1,30 @@
...
 
-const store: Store = createStore(reducers, applyMiddleware(thunkMiddleware));
+const reducers = wrapWithClearStorageReducer(
+  combineReducers({
+    ...laoReducer,
+    ...walletReducer,
+    ...eventsReducer,
+  }),
+);
+
```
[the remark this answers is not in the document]
Tyratox: Oh I did not realize this sets up the store as well, thank you for the hint! I just saw this pattern used in [WalletStore.test.ts](https://github.com/dedis/popstellar/blob/master/fe1-web/src/features/wallet/store/__tests__/WalletStore.test.ts) but then it should probably be changed there instead!

fe1-web/src/features/evoting/network/messages/__tests__/SetupElection.test.ts
Tyratox: Good point 👍

fe1-web/src/core/types/index.ts
```diff
@@ -0,0 +1,19 @@
+// Copied from https://stackoverflow.com/a/49062616/2897827
+/**
+ * Given a type T, this returns a type with all keys set
+ * keyof T returns a union of the keys of T, i.e. if T = {a: number, b: string, c: () => 0}, then keyof T = 'a' | 'b' | 'c'
+ * T['a' | 'b' | 'c'] returns a union of types for the given keys, i.e. using the previous example T['a' | 'b'] = number | string | () => 0
+ * Combining this we have the following:
+ * { [P in keyof T]: T[P] extends Function ? never : P } = {'a': 'a', 'b': 'b', 'c': never} with the previous example
+ * Then selecting of this the union of keys gives 'a' | 'b' | never = 'a' | 'b'
```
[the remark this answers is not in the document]
Tyratox: 👍

fe1-web/src/features/evoting/components/EventElection.tsx
Tyratox: Interesting link! :) The only time I encountered issues when not using it is when you pass react components to other components but since this is not something we need to do here I am happy to remove it.

fe1-web/src/features/evoting/network/ElectionHandler.ts
Tyratox: Seems like it yes and maybe not even that depending on whether we keep the laoId in the election messages. Also depends what we should see as the source of truth 🤔 For the election#open handler I now return false if the message laoId and the active laoId do not match. This should probably be added to the other handlers as well but I will do that in a seperate PR.

fe1-web/src/features/evoting/network/ElectionHandler.ts
Tyratox: 👍 Should do the same for `handleCastVoteMessage`, `handleElectionEndMessage`, `handleElectionResultMessage` in a future PR and also add tests for them!

fe1-web/src/features/evoting/network/ElectionHandler.ts
Tyratox: Yes I thought the same but I was not sure whether this again is a case of defensive programming as this code is found in all election handlers 😅

fe1-web/src/features/evoting/network/ElectionMessageApi.ts
Tyratox: 👍

fe1-web/src/features/evoting/network/ElectionHandler.ts
pierluca: All `ProcessableMessage`s  anyway contain a `.laoId` parameter which is derived from the channel on which they're posted. This is relatively new in the codebase, hence we may need neither the `lao` parameter in the election messages nor the check of the current active LAO. In principle, we could be connected to multiple LAOs (passively) but only one being currently active 😉

fe1-web/src/features/evoting/network/ElectionHandler.ts
pierluca: Before the `ProcessableMessage.laoId` field existed, this was necessary to associate an incoming message with a Lao. It's basically a remnant of past code organization.

fe1-web/src/features/evoting/network/ElectionMessageApi.ts
Tyratox: 👍

fe1-web/src/features/evoting/network/messages/CastVote.ts
Tyratox: Yep - But this will probably have to wait at least until the meeting today

fe1-web/src/features/evoting/network/messages/EndElection.ts
Tyratox: This check was redundant, it is checked once more here:

https://github.com/dedis/popstellar/blob/work-fe1-nico/fe1-web/src/features/evoting/network/messages/EndElection.ts#L41-L43

Moreover the error message was not accurate and sonarcloud complained about uncovered lines as the code was unreachable due to the double check 😹

fe1-web/src/features/evoting/objects/Election.ts
Tyratox: Yes, definitely! This was also one of the things I wanted to change after the discussion here on GitHub. Before this was the only way the election status was determined and there was some really interesting code making it all work.

I actually removed this default state and required it to be always passed explicitly now as I feel like this is an even clearer option.

protocol/examples/messageData/election_open.json
Tyratox: Not sure tbh as this is copied from #657 😅

fe1-web/src/features/evoting/network/messages/EndElection.ts
pierluca: 😸

fe1-web/src/features/evoting/objects/__tests__/Election.test.ts
```diff
@@ -53,14 +55,15 @@ const initializeData = () => {
 
   electionState = {
     id: 'electionId',
-    eventType: EventTypeElection,
+    eventType: LaoEventType.ELECTION,
```
pierluca, line 58: there was a reason for not using `LaoEventType`, but it's clearly not obvious from the current code-base.

One of the goal of the current separation into features is that each feature should be as independent as possible.

Ideally, we would like `meeting`, `rollCall` and `election` to each have their own reducer, + an `events` reducer which is type-agnostic and just allows other features to associate their own keyword (e.g. `EventTypeElection`) to a given way of displaying the event. 
This would allow us to reach the state where features do not import each other *at all*. 
Remind me to talk about this in one of our next meetings.

pierluca (commented):
Much better, here are a few more remarks :)

Xelowak (changes requested):
Nice work ! I'm impressed by how familiar you seem to be with the project already.

Regarding tests, I have 2 general comments to make : 
1. It could be nice to add an empty line after each `it` in tests for clarity. You have done it somewhere but not everywhere.
2. I see there are a lot of `@ts-ignore`. I'd suggest you try as much as possible to not use them, because there are here for a reason and you can use another notation in tests if needed. By the way, the issues that you are ignoring (especially in `SetupElection.test`) don't show anything in my IDE.

To stay consistent with the `RollCallStatus` names (and the fact that the message is called `OpenElection`), I would rename `ElectionStatus.RUNNING` to `ElectionStatus.OPENED`.

fe1-web/src/features/evoting/objects/__tests__/Election.test.ts
```diff
@@ -13,12 +14,13 @@ import {
   RegisteredVote,
   Vote,
 } from '../Election';
+import { ElectionState } from '..';
```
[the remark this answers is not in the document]
pierluca: Yup, I can't readily find the link to the bug description though 😞 ah well, you'll have to trust us 😆

fe1-web/src/features/evoting/network/messages/__tests__/CastVote.test.ts
```diff
@@ -0,0 +1,238 @@
...
+      votes: [mockVoteObject1, mockVoteObject2],
+    };
+    const createFromJson = () => CastVote.fromJson(obj);
+    expect(createFromJson).toThrow(ProtocolError);
+  });
+
+  describe('constructor', () => {
+    it('should throw an error if election is undefined', () => {
+      const createWrongObj = () =>
+        // @ts-ignore. Here was pass data to the constructor that is missing a field even though ts would require it
```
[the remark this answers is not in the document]
pierluca: I guess @Tyratox wants to try and break the message parsing.
I don't like it either, but I don't see how we can avoid the ts-ignore in this scenario 🤷

fe1-web/src/features/evoting/components/__tests__/EventElection.test.tsx
```diff
@@ -0,0 +1,236 @@
...
+  describe('Not started election', () => {
+    it('renders correctly for an organizer', () => {
+      const component = render(
+        <EventElection election={notStartedElection} isOrganizer />,
+      ).toJSON();
+      expect(component).toMatchSnapshot();
+    });
+    it('renders correctly for an attendee', () => {
+      const component = render(
+        <EventElection election={notStartedElection} isOrganizer={false} />,
```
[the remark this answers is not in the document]
pierluca: Wouldn't the property then be defined as `undefined` (which evaluates to falsy) ?

fe1-web/src/features/evoting/network/messages/__tests__/CastVote.test.ts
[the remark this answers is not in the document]
Xelowak: We already did it in other tests and the IDE (and ESLint) didn't seem to complain, at least on my end.

fe1-web/src/features/evoting/components/__tests__/EventElection.test.tsx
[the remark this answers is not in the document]
Xelowak: It is actually a rule from the Airbnb style convention, as you can find [there](https://airbnb.io/javascript/react/). Normally, the IDE would complain about it by the way.

image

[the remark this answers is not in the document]
Xelowak: It might be the case, since `EventElection` doesn't have default property value for `isOrganizer`. It was how we handled false booleans in props until now though, so it is probably done this way in the application (I actually have never seen a boolean being set to false in React props). I guess it's not an important issue anyway.

fe1-web/src/features/evoting/components/__tests__/EventElection.test.tsx
```diff
@@ -0,0 +1,236 @@
...
+  lao: mockLaoIdHash,
+  id: mockElectionId,
+  name: 'An election',
+  version: STRINGS.election_version_identifier,
+  createdAt: TIMESTAMP,
+  start: TIMESTAMP,
+  end: CLOSE_TIMESTAMP,
+  questions: [question],
+  // @ts-ignore: Sonarcloud wants high test coverage, this is to hit the default case
+  electionStatus: 'undefined',
```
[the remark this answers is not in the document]
Tyratox: True, I can use `'undefined' as ElectionStatus` but `undefined` will not work as the Election constructor requires it to be truthy :)

fe1-web/src/features/evoting/network/__tests__/ElectionHandler.test.ts
```diff
@@ -0,0 +1,201 @@
...
+  laoId: mockLaoIdHash,
+  data: Base64UrlData.encode(''),
+  sender: mockKeyPair.publicKey,
+  signature: Base64UrlData.encode('') as Signature,
+  channel: '',
+  message_id: Hash.fromString(''),
+  witness_signatures: [],
+};
+
+const initializeData = () => {
```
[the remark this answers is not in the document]
Tyratox: True 👍 . Actually I saw that clearing only happened once before all tests and it should have been done before each test, hence I now moved it to `beforeEach`.

fe1-web/src/features/evoting/network/__tests__/ElectionHandler.test.ts
```diff
@@ -0,0 +1,201 @@
...
+beforeAll(() => {
+  configureTestFeatures();
+  initializeData();
+});
+
+describe('ElectionHandler', () => {
+  describe('election#open', () => {
+    it('should return false if the object is not "election"', () => {
+      const mockFn = jest.fn();
+      console.warn = mockFn;
```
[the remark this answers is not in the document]
Tyratox: How would I do that? The reason for doing it that way was to be able to check how many times it was called during that test and if I just did it in a `beforeEach` I will get typescript errors since it does not know we overwrote `console.warn` with a mock function. (i.e. `mockFn.mock.calls[0][0]` will not be typechecked correctly)

fe1-web/src/features/evoting/components/__tests__/EventElection.test.tsx
```diff
@@ -0,0 +1,236 @@
...
+  registeredVotes: [registeredVote],
+  questionResult: [questionResult],
+});
```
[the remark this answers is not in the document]
Tyratox: 👍

fe1-web/src/features/evoting/network/messages/__tests__/CastVote.test.ts
```diff
@@ -0,0 +1,238 @@
...
+
+  CastVoteJson = `{
+    "object": "${ObjectType.ELECTION}",
+    "action": "${ActionType.CAST_VOTE}",
+    "lao": "${mockLaoIdHash}",
+    "election": "${electionId}",
+    "created_at": ${TIMESTAMP},
+    "votes": ${JSON.stringify(mockVotes)}
+  }`;
+};
```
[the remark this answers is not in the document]
Tyratox: Cool, I prefer that too! 👍 :)

fe1-web/src/features/evoting/network/messages/__tests__/CastVote.test.ts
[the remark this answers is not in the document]
Tyratox: Yes but I changed the typing of the constructors to now allow undefined inputs anymore. The only place this can still happen is in the `fromJson` function where we work with an `any` type. Since I still wanted to test these cases a ts-ignore was necessary, otherwise VSCode reports the following error:

```
Argument of type '{ lao: Hash; created_at: Timestamp; votes: Vote[]; }' is not assignable to parameter of type 'MessageDataProperties<CastVote>'.
  Property 'election' is missing in type '{ lao: Hash; created_at: Timestamp; votes: Vote[]; }' but required in type 'MessageDataProperties<CastVote>'.ts(2345)
[CastVote.ts(17, 19): ]()'election' is declared here.
```

fe1-web/src/features/evoting/network/messages/__tests__/EndElection.test.ts
```diff
@@ -0,0 +1,232 @@
...
+      });
+
+      expect(msg.object).toEqual(ObjectType.ELECTION);
+      expect(msg.action).toEqual(ActionType.END);
+    });
+  });
+
+  describe('computeRegisteredVotesHash', () => {
+    // It seems a bit silly to test this function apart from it not returning an error as
+    // we would simply write the same code here again?
```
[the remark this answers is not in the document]
Tyratox: 👍

fe1-web/src/features/evoting/components/__tests__/EventElection.test.tsx
```diff
@@ -0,0 +1,232 @@
...
+  createdAt: TIMESTAMP,
+  start: TIMESTAMP,
+  end: CLOSE_TIMESTAMP,
+  questions: [question],
+  electionStatus: 'undefined' as ElectionStatus,
+  registeredVotes: [registeredVote],
+  questionResult: [questionResult],
+});
+
+// #endregion
```
Xelowak, line 139:
```suggestion
// endregion
```
Same issue

fe1-web/src/features/evoting/network/messages/__tests__/CastVote.test.ts
```diff
@@ -0,0 +1,234 @@
...
+import { EventTags, Hash, Timestamp, ProtocolError } from 'core/objects';
+import { ActionType, ObjectType } from 'core/network/jsonrpc/messages';
+import { OpenedLaoStore } from 'features/lao/store';
+
+import { MessageDataProperties } from 'core/types';
+import STRINGS from 'resources/strings';
+import { Election, ElectionStatus, Question, Vote } from '../../../objects';
+import { CastVote } from '../CastVote';
+
+// #region test data initialization
```
Xelowak, line 20:
```suggestion
// region test data initialization
```

fe1-web/src/features/evoting/network/messages/__tests__/CastVote.test.ts
```diff
@@ -0,0 +1,234 @@
...
+const CastVoteJson: string = `{
+  "object": "${ObjectType.ELECTION}",
+  "action": "${ActionType.CAST_VOTE}",
+  "lao": "${mockLaoIdHash}",
+  "election": "${electionId}",
+  "created_at": ${TIMESTAMP},
+  "votes": ${JSON.stringify(mockVotes)}
+}`;
+
+// #endregion
```
Xelowak, line 121:
```suggestion
// endregion
```

fe1-web/src/features/evoting/network/messages/__tests__/EndElection.test.ts
```diff
@@ -0,0 +1,232 @@
...
+import { Hash, Timestamp, ProtocolError, EventTags } from 'core/objects';
+import { ActionType, ObjectType } from 'core/network/jsonrpc/messages';
+import { OpenedLaoStore } from 'features/lao/store';
+
+import { MessageDataProperties } from 'core/types';
+import { Election, ElectionStatus, Question } from 'features/evoting/objects';
+import STRINGS from 'resources/strings';
+import { EndElection } from '../EndElection';
+
+// #region test data initialization
```
Xelowak, line 20:
```suggestion
// region test data initialization
```

fe1-web/src/features/evoting/network/messages/__tests__/EndElection.test.ts
```diff
@@ -0,0 +1,232 @@
...
+const endElectionJson: string = `{
+  "object": "${ObjectType.ELECTION}",
+  "action": "${ActionType.END}",
+  "election": "${electionId}",
+  "lao": "${mockLaoIdHash}",
+  "created_at": ${TIMESTAMP},
+  "registered_votes": "${mockElectionResultHash}"
+}`;
+
+// #endregion
```
Xelowak, line 111:
```suggestion
// endregion
```

fe1-web/src/features/evoting/network/messages/__tests__/SetupElection.test.ts
```diff
@@ -13,90 +13,99 @@ import { ActionType, ObjectType } from 'core/network/jsonrpc/messages';
 import { OpenedLaoStore } from 'features/lao/store';
 import STRINGS from 'resources/strings';
 
+import { MessageDataProperties } from 'core/types';
 import { Question } from '../../../objects';
 import { SetupElection } from '../SetupElection';
 
+// #region test data initialization
```
Xelowak, line 20:
```suggestion
// region test data initialization
```

fe1-web/src/features/evoting/network/messages/__tests__/SetupElection.test.ts
```diff
@@ -13,90 +13,99 @@ import { ActionType, ObjectType } from 'core/network/jsonrpc/messages';
...
+  "lao": "${mockLaoIdHash}",
+  "name": "${mockLaoName}",
+  "version": "${VERSION}",
+  "created_at": ${TIMESTAMP},
+  "start_time": ${TIMESTAMP},
+  "end_time": ${CLOSE_TIMESTAMP},
+  "questions": ${JSON.stringify(mockQuestions)}
+}`;
+
+// #endregion
```
Xelowak, line 97:
```suggestion
// endregion
```

Xelowak (commented):
Good job ! There's just a tiny issue about regions that cannot be collapsed if written with a '#'.

fe1-web/src/features/evoting/components/__tests__/EventElection.test.tsx
```diff
@@ -0,0 +1,232 @@
...
+import {
+  Election,
+  ElectionStatus,
+  Question,
+  QuestionResult,
+  RegisteredVote,
+} from 'features/evoting/objects';
+import EventElection from '../EventElection';
+
+// #region test data initialization
```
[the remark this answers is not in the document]
Tyratox: Huh, interesting. In VSCode it works like this and according to [this](https://www.jetbrains.com/help/rider/Coding_Assistance__Surrounding_with_Region.html#managing-regions-in-the-editor) it should also work in webstorm? Does it work for you without the `#`? Because then it at least does not work in VSCode (which would be fine for me, I don't mind scrolling a bit haha)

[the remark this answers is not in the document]
Tyratox: Found this: https://github.com/madskristensen/WebEssentials2015/issues/103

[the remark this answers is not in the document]
Tyratox: Oh no! Apparently `#region` works in webstorm but `# region` does not. And `eslint` wants spaces before a comment 😂

https://youtrack.jetbrains.com/issue/WEB-13948:

> //region
> // region
> //#region
> all work in 2019.2, but
> // #region
> does not.
> 
> This sucks because only the ones with # work with VS Code, and most eslint rulesets require a space after //, so there's no way to make everyone happy.

So I guess we can go with the very funny

```
region // #region NameOfRegion

// endregion // #endregion
```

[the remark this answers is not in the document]
Xelowak: Hahaha okay I definitely didn't think it would be that much of a mess.
Does this work with VSCode using this method at least ?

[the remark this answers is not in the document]
Tyratox: The `// #region` worked in VSCode but since Webstorm is the recommended IDE we should probably go with `// region x` and I will look for a plugin or something :)

pierluca (approved):

fe1-web/src/features/evoting/network/messages/__tests__/SetupElection.test.ts
```diff
@@ -120,14 +129,31 @@ describe('SetupElection', () => {
...
     const obj = {
       object: ObjectType.ELECTION,
+      action: ActionType.NOTIFY_ADD,
+      id: electionId.toString(),
+      lao: mockLaoIdHash.toString(),
+      name: mockLaoName,
+      version: VERSION,
+      created_at: parseInt(TIMESTAMP.toString(), 10),
+      start_time: parseInt(TIMESTAMP.toString(), 10),
+      end_time: parseInt(CLOSE_TIMESTAMP.toString(), 10),
```
Xelowak, line 139: These can be simply replaced by 
```suggestion
      created_at: TIMESTAMP.valueOf(),
      start_time: TIMESTAMP.valueOf(),
      end_time: CLOSE_TIMESTAMP.valueOf(),
```
But it's not a big issue.

Xelowak (approved):
LGTM ! The only thing I have to say is that it's easier to use `Timestamp.valueOf()` instead of using `parseInt`, but it is not a big issue at all.

0xdieman (dismissed):

fe1-web/src/features/evoting/network/messages/__tests__/SetupElection.test.ts
Tyratox: Thank you for the hint! I just pushed another commit to this PR regarding this 😊

Xelowak (approved):

PR #851, opened 2022-03-01
DbActor: fill decodedData

lourkeur:
Whenever a message is retrieved from the db through the `DbActor` (e.g. by sending a `Read` or `Catchup` signal), the `decodedData` field is not filled.

### DIFF
--- a/be2-scala/src/main/scala/ch/epfl/pop/json/MessageDataProtocol.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/json/MessageDataProtocol.scala
@@ -9,7 +9,7 @@ import ch.epfl.pop.model.network.method.message.data.meeting._
 import ch.epfl.pop.model.network.method.message.data.rollCall._
 import ch.epfl.pop.model.network.method.message.data.socialMedia._
 import ch.epfl.pop.model.network.method.message.data.witness._
-import ch.epfl.pop.model.network.method.message.data.{ActionType, ObjectType}
+import ch.epfl.pop.model.network.method.message.data.{MessageData, ActionType, ObjectType}
 import ch.epfl.pop.model.objects._
 import ch.epfl.pop.pubsub.MessageRegistry
 import spray.json._
@@ -18,6 +18,8 @@ import scala.collection.immutable.ListMap
 import scala.util.Try
 
 object MessageDataProtocol extends DefaultJsonProtocol {
+  final private val PARAM_OBJECT = "object"
+  final private val PARAM_ACTION = "action"
 
   // ----------------------------------- ENUM FORMATTERS ----------------------------------- //
   implicit object objectTypeFormat extends RootJsonFormat[ObjectType] {
@@ -45,7 +47,7 @@ object MessageDataProtocol extends DefaultJsonProtocol {
   // Succeeds if both 'object' and 'action' are present and are strings
   def parseHeader(data: String): Try[(ObjectType, ActionType)] =
     Try {
-      data.parseJson.asJsObject.getFields("object", "action") match {
+      data.parseJson.asJsObject.getFields(PARAM_OBJECT, PARAM_ACTION) match {
         case Seq(objectString @ JsString(_), actionString @ JsString(_)) =>
           (objectString.convertTo[ObjectType], actionString.convertTo[ActionType])
         case _ => throw new IllegalArgumentException("parseHeader: header fields not found")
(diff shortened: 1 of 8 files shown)


be2-scala/src/main/scala/ch/epfl/pop/storage/DbActor.scala
```diff
@@ -47,7 +48,11 @@ final case class DbActor(
   @throws [DbActorNAckException]
   private def read(channel: Channel, messageId: Hash): Option[Message] = {
     Try(storage.read(s"$channel${Channel.DATA_SEPARATOR}$messageId")) match {
-      case Success(Some(json)) => Some(Message.buildFromJson(json))
+      case Success(Some(json)) =>
+        val msg = Message.buildFromJson(json)
+        val channelData = readChannelData(channel)
+        val builder = registry.getBuilder(channelData.channelType, ActionType.CREATE).get
```
lourkeur, line 54: I'm not sure how to properly summon the correct action type...

lourkeur:
If we want a proper `(JsonString, MessageRegistry) => MessageData` procedure to be possible, the `MessageData` json schema has to include these fields...

lourkeur:
As far as the current state of the PR is concerned: the plan would be to replace all the `jsonFormatN` calls in `MessageDataProtocol` like I have already done 3 times. It just requires more time to sit down and do. (or maybe we can meta-program our way out of that)

Ajkunas (dismissed):
Tested!

be2-scala/src/main/scala/ch/epfl/pop/json/MessageDataProtocol.scala
```diff
@@ -217,25 +244,25 @@ object MessageDataProtocol extends DefaultJsonProtocol {
     }
   }
 
-  implicit val closeRollCallFormat: JsonFormat[CloseRollCall] = jsonFormat[Hash, Hash, Timestamp, List[PublicKey], CloseRollCall](CloseRollCall.apply, "update_id", "closes", "closed_at", "attendees")
-  implicit val createRollCallFormat: JsonFormat[CreateRollCall] = jsonFormat[Hash, String, Timestamp, Timestamp, Timestamp, String, Option[String], CreateRollCall](CreateRollCall.apply, "id", "name", "creation", "proposed_start", "proposed_end", "location", "roll_call_description")
-  implicit val openRollCallFormat: JsonFormat[OpenRollCall] = jsonFormat[Hash, Hash, Timestamp, OpenRollCall](OpenRollCall.apply, "update_id", "opens", "opened_at")
-  implicit val reopenRollCallFormat: JsonFormat[ReopenRollCall] = jsonFormat[Hash, Hash, Timestamp, ReopenRollCall](ReopenRollCall.apply, "update_id", "opens", "opened_at")
+  implicit val closeRollCallFormat: JsonFormat[CloseRollCall] = annotateHeader(jsonFormat[Hash, Hash, Timestamp, List[PublicKey], CloseRollCall](CloseRollCall.apply, "update_id", "closes", "closed_at", "attendees"))
```
tuomas-p, line 247: the new formats look good to me

be2-scala/src/main/scala/ch/epfl/pop/json/MessageDataProtocol.scala
```diff
@@ -37,8 +41,31 @@ object MessageDataProtocol extends DefaultJsonProtocol {
...
 
+  // ------------------------------- METADATA UTILITY -------------------------------------- //
+  // retrieve information from the header if it is correct.
+  //
+  // Succeeds if both 'object' and 'action' are present and are strings
+  def parseHeader(data: String): Try[(ObjectType, ActionType)] =
+    Try {
+      data.parseJson.asJsObject.getFields(PARAM_OBJECT, PARAM_ACTION) match {
+        case Seq(objectString @ JsString(_), actionString @ JsString(_)) =>
+          (objectString.convertTo[ObjectType], actionString.convertTo[ActionType])
```
tuomas-p, line 52: I don't exactly know whether this is really a relevant point, as it's wrapped in the Try, but I'd still add a case _ => just to be sure

tuomas-p (changes requested):
Apart from the small change with the match case (which I'm not completely sure about), looks good to me, very clean overall!

be2-scala/src/main/scala/ch/epfl/pop/json/MessageDataProtocol.scala
lourkeur: That is absolutely right, fixed it up!

be2-scala/src/main/scala/ch/epfl/pop/json/MessageDataProtocol.scala
```diff
@@ -37,8 +41,32 @@ object MessageDataProtocol extends DefaultJsonProtocol {
...
+        case _ => throw new IllegalArgumentException("parseHeader: header fields not found")
```
tuomas-p, line 53: 👍

tuomas-p (approved):

PR #852, opened 2022-03-01
Fix the issue where android in not able to launch an LAO

GabrielFleischer:
The issue had multiple factors, and they got fixed individually.

A test was also added to assert that multiple simultaneous requests does not conflict with one-another.


jbsv (approved):
LGTM (Looks Good To Me)
👍

PR #854, opened 2022-03-02
Fix Roll Call Organizer Camera Error

0xdieman:
Fixes an error when declining the access to the camera or when no camera is found => an error popup will appear instead of the whole application crashing.

### DIFF
--- a/fe1-web/src/features/rollCall/screens/RollCallOpened.tsx
+++ b/fe1-web/src/features/rollCall/screens/RollCallOpened.tsx
@@ -71,9 +71,9 @@ const RollCallOpened = () => {
     addOwnToken().catch((e) => console.error(e));
   }, [lao, rollCallID, toast]);
 
-  const handleError = (err: string) => {
+  const handleError = (err: any) => {
     console.error(err);
-    toast.show(err, {
+    toast.show(err?.message || err, {
       type: 'danger',
       placement: 'top',
       duration: FOUR_SECONDS,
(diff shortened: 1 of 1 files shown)


fe1-web/src/features/rollCall/screens/RollCallOpened.tsx
```diff
@@ -71,9 +71,10 @@ const RollCallOpened = () => {
     addOwnToken().catch((e) => console.error(e));
   }, [lao, rollCallID, toast]);
 
-  const handleError = (err: string) => {
+  const handleError = (err: any) => {
     console.error(err);
-    toast.show(err, {
+    // The "err" object might be an exception, take the message property if it exists
+    toast.show(err?.message || err, {
```
Tyratox, line 77: This might be a little picky but since we are already changing this line we can also make the fix slightly more robust by doing something like

```ts
err?.message || (typeof err === "string" ? err : "Unkown error, please check the console and report it!")
```

As with an any type in theory it could also be `err = {}` for which the first condition is falsy and thus `{}` is passed to toast.show. If I remember correctly you said yesterday it only accepts strings, this means that we could still end up in a broken state in that case?

Tyratox (changes requested):

fe1-web/src/features/rollCall/screens/RollCallOpened.tsx
0xdieman: Yep correct, I'll change that !

Xelowak (approved):
LGTM, thanks for fixing this

PR #855, opened 2022-03-02
Bump AGP version and fix Hilt entry point declarations

### LINKED ISSUE #836: [FE2] The current AGP (7.1.1) has compatibility issues with Hilt on Local Tests
This is a known issue from the side of Dagger:
https://github.com/google/dagger/issues/1956

But it seems to be a bit more complex from the side of the AGP team.
https://issuetracker.google.com/issues/213534628
The current fix works, but is not pretty.

GabrielFleischer:
Bump the Android Gradle Plugin version to 7.1.2.

Fixes #836.


pierluca (approved):
I haven't tested this, but I fully support the change if it works 😅

fe2-android/build.gradle
```diff
@@ -7,7 +7,7 @@ buildscript {
 }
 
 plugins {
-    id 'com.android.application' version '7.1.1' apply false
+    id 'com.android.application' version '7.1.2' apply false
```
jbsv, line 10: any idea what apply 'false' means ? why not 'true' ?

jbsv (approved):

jbsv:
thanks for this clean up 👍

PR #856, opened 2022-03-03
Correctly update poptoken in the qr code

Tyratox:
This change fixes the bug we encountered during the global meeting where the data encoded in the QR code is invalid (i.e. an empty string).

In a future PR we should also show a nice message when the wallet was not set up first as this results in the same issue and no error is shown in the UI.

The other change (`console.error`) should help as previously a generic error message was shown and the actual error was discarded.


Xelowak (approved):
LGTM ! Thanks for fixing this issue

PR #857, opened 2022-03-03
Update file for Windows compatibility

MaximeZmt:
This Pull Request contains the following modifications:

README.md:
After having lost a lot of time looking at how to compile the project on windows. 
We finally found a GCC 64 bits compiler that works well.
I added the URL into the documentation to avoid other Windows users losing time.

make.bat:
I've updated the build script to automatically rename pop into pop.exe. Like that, Windows will automatically be able to run it from the terminal.

### DIFF
--- a/be1-go/README.md
+++ b/be1-go/README.md
@@ -84,6 +84,16 @@ sudo apt-get install build-essential # Ubuntu/Debian
 You may build the `pop` CLI to interact with the server by executing `make
 build` on Linux/OSX or `make.bat build` in Windows.
 
+> Special Note for Windows:
+>
+> To use the command `.\make.bat build`  you should first install GCC compiler 64 bits.
+>
+> After having looked on the internet a while we recommend [TDM-GCC](https://jmeubank.github.io/tdm-gcc/), using the tdm64-gcc-%VERSION%.exe that contains MinGW-w64.
+>
+> Then verify that the path of `gcc` exists (by typing gcc in the terminal)
+>
+> You will be able to compile using `.\make.bat build` which will create `pop.exe` that you will be able to execute from the terminal. 
+
 You can see the CLI usage with the `-h` flag.
 
 ```bash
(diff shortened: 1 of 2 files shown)


be1-go/README.md
```diff
@@ -84,6 +84,16 @@ sudo apt-get install build-essential # Ubuntu/Debian
 You may build the `pop` CLI to interact with the server by executing `make
 build` on Linux/OSX or `make.bat build` in Windows.
 
+> Special Note for Windows:
```
nkcr, line 87: Would be better to have a subsection for each environment:

```
#### Unix

...

#### Windows

....
```

nkcr (commented):
Thanks! 👍

Warhedgehog:
Should we add that the user should make sure the environment variable are correctly set ?

MaximeZmt:
> Should we add that the user should make sure the environment variable are correctly set ?

I don't know. It is just that it is not done necessarily by default on windows (to set the path)! So It might be a case of failure. Then there is room for debate.

Warhedgehog:
You're probably right, making sure it is correctly set is part of installing it

Warhedgehog (approved):
LGTM 👍

MaximeZmt:
> You're probably right, making sure it is correctly set is part of installing it

Yet I've removed it following your comment. Should I add it back 😅 ? 

Otherwise, I've refactored a small part of the execution paragraph.

Warhedgehog:
> > You're probably right, making sure it is correctly set is part of installing it
> 
> Yet I've removed it following your comment. Should I add it back 😅 ?
> 
> Otherwise, I've refactored a small part of the execution paragraph.

Either way is fine with me, I'll leave you the judge of it. 
What just came to mind is that for the suggestion of the compiler we should add the date at which we consider it the best option. In the coming months or years, it might change quite a bit.

MaximeZmt:
> > > You're probably right, making sure it is correctly set is part of installing it
> > 
> > 
> > Yet I've removed it following your comment. Should I add it back 😅 ?
> > Otherwise, I've refactored a small part of the execution paragraph.
> 
> Either way is fine with me, I'll leave you the judge of it. What just came to mind is that for the suggestion of the compiler we should add the date at which we consider it the best option. In the coming months or years, it might change quite a bit.

I've added the date :)

Warhedgehog (approved):
Good job !

PR #860, opened 2022-03-04
Create the JSON schema of a Digital Cash transaction

0xdieman:
Draft schema of a Digital Cash transaction posting.
Each object representation in an array is given by the field `"items"`.
Not sure about fields data types, so some objects are still of `"type"`: `"bytes"` with a length attribute.

### DIFF
--- a/protocol/query/method/message/data/dataCashTransaction.json
+++ b/protocol/query/method/message/data/dataCashTransaction.json
@@ -1,7 +1,7 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
   "$id": "https://raw.githubusercontent.com/dedis/popstellar/master/protocol/query/method/message/data/dataPostTransaction.json",
-  "description": "Post a digital cash transaction",
+  "title": "Post a digital cash transaction",
   "type": "object",
   "properties": {
     "object": {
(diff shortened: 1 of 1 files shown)


protocol/query/method/message/data/dataCashTransaction.json
```diff
@@ -0,0 +1,100 @@
...
+                "Type": {
+                  "type": "string",
+                  "description": "Type of script"
+                },
+                "PubkeyHash": {
+                  "type": "string",
+                  "contentEncoding": "base64",
+                  "description": "Hash of the recipient’s public key"
+                },
+                "Sig": {
```
lourkeur, line 83: the signature should be in the `TxIn`s and be signed by the recipients of the inputs, not the current recipient.

lourkeur (changes requested):
found a couple small things...

protocol/query/method/message/data/dataCashTransaction.json
0xdieman: Riight I swapped the script part of the TxIn with the TxOut one

lourkeur (commented):
Re: endianness
Bitcoin seems to use little-endian so if we put a comment saying all integer fields are little-endian I'm satisfied.

Re: fe1 interface work
It seems to be independent from the JSON schema work.

Re: JSON schema
I would like to see a sample post transaction message and a unit test that validates it against the schema. I can write one myself if requested.

0xdieman:
Forgot to checkout to my other branch, I reverted the commits and now the only change is concerning the json schema.

PR #861, opened 2022-03-04
Adding a default case in the switch-case in HomeActivity.java

Tesa27:
Currently there are only two possibilities for an event (Scanning and requesting for permission to use the camera) but we might want to add some more later, for now there is no default behavior which is perhaps wanted but should be addressed in a cleaner way.

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeViewModel.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeViewModel.java
@@ -47,8 +47,11 @@ public class HomeViewModel extends AndroidViewModel
     implements CameraPermissionViewModel, QRCodeScanningViewModel {
 
   public static final String TAG = HomeViewModel.class.getSimpleName();
-  public static final String SCAN = "SCAN";
-  public static final String REQUEST_CAMERA_PERMISSION = "REQUEST_CAMERA_PERMISSION";
+//  public static final String SCAN = "SCAN";
+//  public static final String REQUEST_CAMERA_PERMISSION = "REQUEST_CAMERA_PERMISSION";
+
+  public static enum HOMEVIEWACTION {SCAN, REQUEST_CAMERA_PERMISSION}
+
   private static final ScanningAction scanningAction = ScanningAction.ADD_LAO_PARTICIPANT;
 
   /*
@@ -58,7 +61,7 @@ public class HomeViewModel extends AndroidViewModel
   private final MutableLiveData<SingleEvent<Boolean>> mOpenHomeEvent = new MutableLiveData<>();
   private final MutableLiveData<SingleEvent<Boolean>> mOpenConnectingEvent =
       new MutableLiveData<>();
-  private final MutableLiveData<SingleEvent<String>> mOpenConnectEvent = new MutableLiveData<>();
+  private final MutableLiveData<SingleEvent<HOMEVIEWACTION>> mOpenConnectEvent = new MutableLiveData<>();
   private final MutableLiveData<SingleEvent<Boolean>> mOpenLaunchEvent = new MutableLiveData<>();
   private final MutableLiveData<SingleEvent<Boolean>> mLaunchNewLaoEvent = new MutableLiveData<>();
   private final MutableLiveData<SingleEvent<Boolean>> mCancelNewLaoEvent = new MutableLiveData<>();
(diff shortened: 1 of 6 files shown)


ahominal (dismissed):

fe2-android/app/src/test/java/com/github/dedis/popstellar/ui/detail/LaoDetailActivityTest.java
```diff
@@ -146,6 +149,13 @@ public void witnessButtonShowsWitnessTab() {
     fragmentContainer().check(matches(withChild(withId(witnessFragmentId()))));
   }
 
+  @Test
+  public void OpenQRSCanningSetsValueCorrectly(){
+    HomeViewModel.HOMEVIEWACTION scan = HomeViewModel.HOMEVIEWACTION.SCAN;
+    HomeViewModel.HOMEVIEWACTION cameraPermission = HomeViewModel.HOMEVIEWACTION.REQUEST_CAMERA_PERMISSION;
+    //TODO: create an instance of HomeViewModel and check if event is well set
+  }
+
```
GabrielFleischer, line 158: I'd avoid having empty tests except if you plan on filling them in a very short time.
But even then, I think it is best to just push this test once it is complete.

It will avoid issues where dead code is left in the codebase and forgotten.

fe2-android/app/src/test/java/com/github/dedis/popstellar/ui/home/CameraPermissionFragmentTest.java
```diff
@@ -70,6 +70,11 @@ public void denyPermissionDoesNothing() {
     assertFalse("A response was received when none was expected", receiver.received());
   }
 
+  @Test
+  public void openAddWitnessSetsWitnessOnLAO(){
+
+  }
+
```
GabrielFleischer, line 77: Same issue as above. I don't like empty functions pushed on the main branch.

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeViewModel.java
```diff
@@ -47,8 +47,9 @@
     implements CameraPermissionViewModel, QRCodeScanningViewModel {
 
   public static final String TAG = HomeViewModel.class.getSimpleName();
-  public static final String SCAN = "SCAN";
-  public static final String REQUEST_CAMERA_PERMISSION = "REQUEST_CAMERA_PERMISSION";
+
+  public enum HOMEVIEWACTION {SCAN, REQUEST_CAMERA_PERMISSION}
```
GabrielFleischer, line 51: `HOMEVIEWACTION` does not respect the usual naming convention.

A better name would be
```suggestion
  public enum HomeViewAction {SCAN, REQUEST_CAMERA_PERMISSION}
```

GabrielFleischer (changes requested):
There are some tests that should not be pushed on main in my opinion and the name
`HOMEVIEWACTION` doesn't fit the usual naming convention.

(You could probably also use static imports to avoid redundant `HomeViewAction.[...]`)

Otherwise, this is a great step forward, nicely done ! 👍

fe2-android/app/src/test/java/com/github/dedis/popstellar/ui/detail/LaoDetailActivityTest.java
Tesa27: Yeah, I agree forgot that I started writing the test should definitely finish it, but since I don't have enough coverage in the new code there is no fear of merging something unfinished in this case. But yeah should be avoided in the upcoming PRs, thanks

fe2-android/app/src/test/java/com/github/dedis/popstellar/ui/home/CameraPermissionFragmentTest.java
Tesa27: Yup agreed

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeViewModel.java
```diff
@@ -47,8 +47,9 @@
     implements CameraPermissionViewModel, QRCodeScanningViewModel {
 
   public static final String TAG = HomeViewModel.class.getSimpleName();
-  public static final String SCAN = "SCAN";
-  public static final String REQUEST_CAMERA_PERMISSION = "REQUEST_CAMERA_PERMISSION";
+
+  public enum HomeViewAction {SCAN, REQUEST_CAMERA_PERMISSION}
```
GabrielFleischer, line 51: Actually, I just realized that this Action is also used in `LaoDetailViewModel`.

So, does it make sense to keep it inside `HomeViewModel` ?
Just asking, this is really nitpicking.

GabrielFleischer (approved):
Looks good to me, except a very small detail that you can address if you think it is important.

PR #865, opened 2022-03-07
Remove the direct dependency of the evoting feature on other features

Tyratox:
As discussed in the last meeting we want to reduce the inter-feature dependencies. As a first try I did this for the evoting feature and depending on whether this goes into a good direction a similar pattern can later be used for the other features.

Since no other feature depends on the evoting feature, not functionality had to be exported which made it easier.

While at it I also added some tests for the uncovered parts in the evoting feature.


fe1-web/src/features/index.ts
```diff
@@ -17,10 +22,49 @@ export function configureFeatures() {
...
+   */
+  const getCurrentLaoId = () => {
+    const currentLaoId = getCurrentLaoIdUnbound(getStore().getState());
+
+    if (!currentLaoId) {
+      throw new Error('Error encountered while accessing storage : no currently opened LAO');
+    }
+
+    return currentLaoId;
+  };
```
pierluca, line 53: ideally, these should very much be exported from the `lao` feature, without needing to declare them here. 
Maybe as part of the `laoConfig` below.

fe1-web/src/features/evoting/objects/Configuration.ts
```diff
@@ -0,0 +1,96 @@
...
+  | 'addEvent'
+  | 'updateEvent'
+  | 'onConfirmEventCreation'
+>;
+
+/**
+ * The interface the evoting feature exposes
+ */
+export interface EvotingInterface extends FeatureInterface {
+  context: EvotingReactContext;
```
pierluca, line 95: Why do we need a `EvotingReactContext` and can't just use an `EvotingConfiguration` here ?

fe1-web/src/features/evoting/network/messages/__tests__/OpenElection.test.ts
```diff
@@ -1,51 +1,37 @@
...
-  configureTestFeatures,
-} from '__tests__/utils';
+import { mockLaoIdHash, configureTestFeatures } from '__tests__/utils';
 
 import { Hash, Timestamp, ProtocolError } from 'core/objects';
 import { ActionType, ObjectType } from 'core/network/jsonrpc/messages';
-import { OpenedLaoStore } from 'features/lao/store';
 
 import { MessageDataProperties } from 'core/types';
+import { mockElectionId } from 'features/evoting/objects/__tests__/utils';
```
pierluca, line 9: nice clean-up, both here and in the other test files

fe1-web/src/features/evoting/network/index.ts
```diff
@@ -8,31 +8,51 @@ import {
...
+ * @param addEvent - A function creating a redux action to add a new event to the store of the currently active lao
+ * @param updateEvent - A function returning a redux action for update an event in the currently active lao store
  * @param registry - The MessageRegistry where we want to add the mappings
  */
-export function configureNetwork(registry: MessageRegistry) {
+export const configureNetwork = (
+  getCurrentLao: EvotingConfiguration['getCurrentLao'],
+  getEventFromId: EvotingConfiguration['getEventFromId'],
+  addEvent: EvotingConfiguration['addEvent'],
+  updateEvent: EvotingConfiguration['updateEvent'],
```
pierluca, line 25: do we need this here, or can we just pass a `config: EvotingConfiguration` and then use `config.registry`, `config.addEvent`, etc?

fe1-web/src/features/evoting/network/__tests__/ElectionMessageApi.test.ts
```diff
@@ -0,0 +1,172 @@
...
+        mockElectionNotStarted.start,
+        mockElectionNotStarted.end,
+        mockElectionNotStarted.questions,
+        mockElectionNotStarted.createdAt,
+      );
+
+      expect(channelFromIds).toHaveBeenCalledWith(mockLaoIdHash);
+      expect(channelFromIds).toHaveBeenCalledTimes(1);
+
+      expect((publish as jest.Mock).mock.calls[0][0]).toEqual(mockChannelId);
```
pierluca, line 60: you could define a `const publishMock = (publish as jest.Mock).mock` and use that everywhere here

fe1-web/src/features/evoting/network/__tests__/ElectionMessageApi.test.ts
```diff
@@ -0,0 +1,172 @@
...
+
+      expect((publish as jest.Mock).mock.calls[0][1]).toBeInstanceOf(SetupElection);
```
pierluca, line 62: what is `calls[0][0]` and `calls[0][1]` here?
I'd use a variable to alias the name to something more meaningful :)

fe1-web/src/features/evoting/network/__tests__/ElectionHandler.test.ts
```diff
@@ -104,31 +229,151 @@ describe('ElectionHandler', () => {
...
+            object: ObjectType.CHIRP,
+            action: ActionType.CAST_VOTE,
           },
         }),
       ).toBeFalse();
 
       expect(warn).toHaveBeenCalledTimes(1);
       // check if the printed warning message contains substring
-      expect(warn.mock.calls[0][0]).toMatch(/LAO/i);
+      expect(warn.mock.calls[0][0]).toMatch(/unsupported message/i);
```
pierluca, line 320: This kind of logic makes me think that the API is wrong.
We should probably throw exceptions instead of returning true/false.
This, however, would be a rather bigger change to do
Definitely outside the scope of this PR.

fe1-web/src/features/evoting/network/__tests__/ElectionHandler.test.ts
```diff
@@ -70,13 +90,115 @@ beforeEach(() => {
...
 afterEach(() => {
   warn.mockClear();
+  (channelFromIds as jest.Mock).mockClear();
+  (subscribeToChannel as jest.Mock).mockClear();
+});
+
+afterAll(() => {
+  (channelFromIds as jest.Mock).mockReset();
+  (subscribeToChannel as jest.Mock).mockReset();
+  warn.mockReset();
```
pierluca, line 100: is there a reason not to reset all mocks at once?

fe1-web/src/features/evoting/hooks/EvotingHooks.ts
```diff
@@ -0,0 +1,38 @@
...
+    }
+    return featureContext[EVOTING_FEATURE_IDENTIFIER] as EvotingReactContext;
+  };
+  /**
+   * Gets the current lao id
+   * @returns The current lao id
+   */
+  export const useCurrentLaoId = () => {
+    return useEvotingContext().getCurrentLaoId();
+  };
```
pierluca, line 21: This might not behave the way you want, we probably need to discuss this in-person.
There are subtleties to the use of `make`Selectors, useSelectors and memoization of those.
The same applies to the other hooks here.

fe1-web/src/features/events/reducer/EventsReducer.ts
```diff
@@ -106,15 +106,14 @@ const eventsSlice = createSlice({
...
-        return { payload: { laoId: laoId.valueOf(), event: event, alias: alias?.valueOf() } };
+      prepare(laoId: Hash | string, event: LaoEventState): any {
+        return { payload: { laoId: laoId.valueOf(), event: event } };
       },
       reducer(
         state,
         action: PayloadAction<{
           laoId: string;
           event: LaoEventState;
-          alias?: string;
```
pierluca, line 117: are you sure about this? 🤔

fe1-web/src/features/evoting/objects/Feature.ts
```diff
@@ -0,0 +1,29 @@
...
+    readonly start: number;
+
+    readonly end?: number;
+  }
+
+  export interface Event {
+    readonly id: Hash;
+
+    toState(): EventState;
+  }
```
pierluca, line 28: we may not need this re-declaration (later on...), for now, I agree that it's a minor evil 😄

fe1-web/src/features/evoting/objects/Configuration.ts
```diff
@@ -0,0 +1,96 @@
+import { PayloadAction } from '@reduxjs/toolkit';
+import { MessageRegistry } from 'core/network/jsonrpc/messages';
+import { Hash, Timestamp } from 'core/objects';
+import FeatureInterface from 'core/objects/FeatureInterface';
+import { EvotingFeature } from './Feature';
+
+export interface EvotingConfiguration {
```
pierluca, line 7: I'm wondering if this should be in the `objects` folder. Effectively, this is just a bunch of "stuff" - mostly functions - that comes from other features

fe1-web/src/features/evoting/objects/index.ts
```diff
@@ -1 +1,3 @@
 export * from './Election';
+export * from './Configuration';
+export * from './Feature';
```
pierluca, line 3: Not too comfortable about these here. I don't have a better idea for now, though.

Tyratox: I agree that a different place would be nicer. What about features/evoting/ directly since this defines the evoting feature interface? Or a subfolder called interface?

fe1-web/src/features/evoting/interface/Feature.ts
```diff
@@ -0,0 +1,29 @@
+import { Hash, PublicKey } from 'core/objects';
+
+export namespace EvotingFeature {
```
Xelowak, line 3: I'm not sure I understand the purpose of this file. If we remove dependencies the way you did in other features that have events (like `rollCall`), will we need to have this exact same file in each of them ?

fe1-web/src/features/evoting/network/ElectionMessageApi.ts
```diff
@@ -42,45 +41,68 @@ export function requestCreateElection(
...
   // publish on the general LAO channel
-  return publish(channelFromIds(currentLao.id), message);
+  return publish(channelFromIds(laoId), message);
+}
+
+/**
+ * Sends a query to open an election.
+ *
+ * @param laoId - The id of the lao in which the given election should be terminated
+ * @param election - The election that should be terminated
```
Xelowak, line 51: I guess there is a copy/paste mistake here.

fe1-web/src/features/evoting/network/__tests__/ElectionHandler.test.ts
```diff
@@ -69,14 +91,118 @@ beforeEach(() => {
...
+            created_at: mockElectionNotStarted.createdAt,
+            start_time: mockElectionNotStarted.start,
+            end_time: mockElectionNotStarted.end,
+            questions: mockElectionNotStarted.questions,
+          }),
+        }),
+      ).toBeTrue();
+
+      // no warning should have been printed
+      expect(warn).toHaveBeenCalledTimes(0);
```
Xelowak, line 173: Is it really relevant ? If at some point someone wants to write a warning in this function for whatever reason, I don't think it should make tests fail.

fe1-web/src/features/evoting/network/__tests__/ElectionMessageApi.test.ts
```diff
@@ -0,0 +1,151 @@
...
+    channelFromIds: jest.fn().mockImplementation(() => mockChannelId),
+  };
+});
+
+jest.mock('core/network', () => {
+  return {
+    ...jest.requireActual('core/network'),
+    publish: jest.fn().mockImplementation(() => Promise.resolve()),
+  };
+});
```
Xelowak, line 33: I believe this mock already exists [here](https://github.com/dedis/popstellar/blob/master/fe1-web/src/core/network/__mocks__/JsonRpcApi.ts).

fe1-web/src/features/evoting/objects/__tests__/utils.ts
```diff
@@ -0,0 +1,157 @@
...
+import { CastVote, ElectionResult, EndElection } from 'features/evoting/network/messages';
+import STRINGS from 'resources/strings';
+import { mockLaoId, mockLaoIdHash } from '__tests__/utils';
+import { Election, ElectionStatus, Question, RegisteredVote, Vote } from '../Election';
+
+const TIMESTAMP = new Timestamp(1609455600); // 1st january 2021
+const VERSION = STRINGS.election_version_identifier;
+const CLOSE_TIMESTAMP = new Timestamp(1609542000); // 2nd january 2021
+
+export const mockElectionName = 'An election';
```
Xelowak, line 11: This file is a great addition, thanks ! But I'll probably put it higher in the folders (closer to `evoting`). The problem being that all tests of the feature will use these constants, and here they're only in `objects`. I can see the logic behind it, but it might be better to put it somewhere else, I'll let you figure it out.

fe1-web/src/features/evoting/objects/__tests__/utils.ts
```diff
@@ -0,0 +1,157 @@
...
+  id: mockVoteId2.toString(),
+  question: mockQuestionId2.valueOf(),
+  vote: [1],
+};
+
+export const mockVotes = [mockVote1];
+
+export const mockElectionResultHash = Hash.fromStringArray(mockVoteId1.valueOf());
+
+// id: string; result: { ballot_option: string; count: number }[]
```
Xelowak, line 140:
```suggestion
// ElectionResult format : { id: string; result: [{ ballot_option: string; count: number }] }
```
But is this really useful ?

fe1-web/src/features/lao/hooks.ts
```diff
@@ -1,29 +1,60 @@
...
-  return useSelector(laosMap);
+  return useSelector(selectLaosMap);
+};
+
+/**
+ * Returns the current lao and throws an error if there is none
+ * @returns The current lao
+ */
+
+export const useCurrentLao = () => {
```
Xelowak, line 38:
```suggestion
/**
 * Returns the current lao and throws an error if there is none
 * @returns The current lao
 */
export const useCurrentLao = () => {
```

Xelowak (commented):
Really nice job ! I've seen some problems in imports order, but I know you have another PR about it so it's ok.
Thanks a lot for all your work about tests, that's really appreciated. However, I'm not sure it's relevant to make "something works fine" test fail on `console.warn`. From a maintenance point of view, I don't think someone adding a warning should worry about causing tests to fail. It's not meant to be an error throw, but more like an information.
I've made some other comments, but those are minor things.

fe1-web/src/features/evoting/interface/Configuration.ts
```diff
@@ -0,0 +1,108 @@
...
+   * Returns the currently active lao id. Should be used outside react components
+   * @returns The current lao id
+   */
+  getCurrentLaoId: () => Hash;
+
+  /**
+   * Returns the currently active lao id. Should be used inside react components
+   * @returns The current lao id
+   */
+  useCurrentLaoId: () => Hash;
```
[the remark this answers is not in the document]
Tyratox: Yes, one of them is a hook (useXX) and one of them is not (getXX). The hook is supposed to be used in react components and the getters outside. We could also use the getXX functions in react components but there were some uncertainties whether this works as intended.

It was related to the selectors (see @pierluca's comment [here](https://github.com/dedis/popstellar/pull/865#discussion_r822617325)), i.e. useSelector() directly uses the redux state and apparently automatically updates the component when needed but the getXX functions will not. I am slightly unsure about the implementation of useSelector() but this is what we discussed 😅

fe1-web/src/features/evoting/interface/Feature.ts
Tyratox: Yes.. I could for now not think of a better way since the evoting feature needs some way to determine a subset type of what it receives?

fe1-web/src/features/evoting/network/__tests__/ElectionHandler.test.ts
```diff
@@ -102,33 +234,162 @@ describe('ElectionHandler', () => {
...
+          messageData: new OpenElection({
+            lao: mockLaoIdHash,
+            election: mockElectionId,
+            opened_at: TIMESTAMP,
+          }),
+        }),
+      ).toBeTrue();
+
+      // no warning should have been printed
+      expect(warn).toHaveBeenCalledTimes(0);
```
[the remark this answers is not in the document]
Tyratox: 👍

fe1-web/src/features/evoting/network/__tests__/ElectionMessageApi.test.ts
Tyratox: Oh, thank you! :)

fe1-web/src/features/evoting/network/__tests__/ElectionMessageApi.test.ts
```diff
@@ -0,0 +1,151 @@
...
+
+afterEach(() => {
+  jest.clearAllMocks();
+});
+afterAll(() => {
+  jest.restoreAllMocks();
```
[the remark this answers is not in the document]
Tyratox: 👍

fe1-web/src/features/lao/hooks.ts
Tyratox: 👍

fe1-web/src/features/evoting/interface/Feature.ts
Xelowak: I guess it's ok for now, even if it's not ideal.

Xelowak (approved):
LGTM ! Thanks for the changes

PR #869, opened 2022-03-08
Create the JSON schema of a Digital Cash transaction 

0xdieman:
Schema of a Digital Cash transaction posting.
Each object representation in an array is given by the field "items".

### DIFF
--- a/protocol/query/method/message/data/dataCashTransaction.json
+++ b/protocol/query/method/message/data/dataCashTransaction.json
@@ -52,7 +52,8 @@
                     "contentEncoding": "base64",
                     "description": "Signature on all txins and txouts using the recipient's private key"
                   }
-                }
+                },
+                "required": ["Type", "Pubkey", "Sig"]
               },
               "Sequence": {
                 "type": "integer",
--- a/protocol/query/method/message/data/dataCashTransaction.json
+++ b/protocol/query/method/message/data/dataCashTransaction.json
@@ -1,0 +1,100 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://raw.githubusercontent.com/dedis/popstellar/master/protocol/query/method/message/data/dataPostTransaction.json",
+  "title": "Post a digital cash transaction",
+  "type": "object",
+  "properties": {
+    "object": {
+      "const": "transaction"
+    },
+    "action": {
+      "const": "post"
+    },
+    "transaction": {
+      "Version": {
+        "type": "integer",
+        "description": "The version of the transaction inputs"
+      },
+      "TxIn": {
+        "type": "array",
+        "description": "[Array[Objects]] array of output transactions to use as inputs",
+        "items": {
+          "type": "object",
+          "description": "Object representing a transaction output to use as an input for this transaction",
+          "properties": {
+            "TxOutHash": {
+              "type": "string",
+              "contentEncoding": "base64",
+              "description": "Previous (to-be-used) transaction hash"
+            },
+            "TxOutIndex": {
+              "type": "bytes",
+              "length": "4",
+              "description": "index of the previous to-be-used transaction"
+            },
+            "Script": {
+              "type": "object",
+              "description": "The script describing the unlock mechanism",
+              "properties": {
+                "Type": {
+                  "type": "string",
+                  "description": "Type of script"
+                },
+                "Pubkey": {
+                  "type": "string",
+                  "contentEncoding": "base64",
+                  "description": "The recipient’s public key"
+                },
+                "Sig": {
+                  "type": "string",
+                  "contentEncoding": "base64",
+                  "description": "Signature on all txins and txouts using the recipient's private key"
+                }
+              }
+            },
+            "Sequence": {
+              "type": "bytes",
+              "length": "4",
+              "$comment": "Set as 0xFFFFFFFF"
+            }
+          }
+        }
+      },
+      "TxOut": {
+        "type": "array",
+        "description": "[Array[Objects]] array of outputs from this transactions",
+        "items": {
+          "type": "object",
+          "description": "Object representing an output of this transaction",
+          "properties": {
+            "Value": {
+              "type": "bytes",
+              "length": "8",
+              "description": "the value of the output transaction, expressed in miniLAOs"
+            },
+            "Script": {
+              "type": "object",
+              "description": "The script describing the TxOut unlock mechanism",
+              "properties": {
+                "Type": {
+                  "type": "string",
+                  "description": "Type of script"
+                },
+                "PubkeyHash": {
+                  "type": "string",
+                  "contentEncoding": "base64",
+                  "description": "Hash of the recipient’s public key"
+                }
+              }
+            }
+          }
+        }
+      },
+      "LockTime": {
+        "type": "bytes",
+        "length": "4"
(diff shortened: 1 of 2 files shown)


protocol/query/method/message/data/dataCashTransaction.json
```diff
@@ -0,0 +1,100 @@
...
+        "type": "integer",
+        "description": "The version of the transaction inputs"
+      },
+      "TxIn": {
+        "type": "array",
+        "description": "[Array[Objects]] array of output transactions to use as inputs",
+        "items": {
+          "type": "object",
+          "description": "Object representing a transaction output to use as an input for this transaction",
+          "properties": {
```
lourkeur, line 24: This node should specify `required` properties. Right now `[{}]` is a valid TxIn

protocol/query/method/message/data/dataCashTransaction.json
```diff
@@ -0,0 +1,100 @@
...
+          }
+        }
+      },
+      "TxOut": {
+        "type": "array",
+        "description": "[Array[Objects]] array of outputs from this transactions",
+        "items": {
+          "type": "object",
+          "description": "Object representing an output of this transaction",
+          "properties": {
```
lourkeur, line 69: Same here

protocol/query/method/message/data/dataCashTransaction.json
```diff
@@ -0,0 +1,100 @@
...
+  "title": "Post a digital cash transaction",
+  "type": "object",
+  "properties": {
+    "object": {
+      "const": "transaction"
+    },
+    "action": {
+      "const": "post"
+    },
+    "transaction": {
```
lourkeur, line 13: `"transaction": {}` should not be valid. I think this should contain `{ "object": { "properties": ... }}` instead of `"Version"` directly.

lourkeur (changes requested):
I think we should link into the parent schema:  48c981a08deac4f1212716fcaa48bdc3d0cf5d18

I have played with an example transaction, the schema is too permissive because there are things about it that should not validate: 3d592923d4457f3a9c1bdd633a40c536279e3884

zoemarin:
And perhaps also change the label Script like 35 and 75,And give some other name like Lock and Unllock ? Zoé MarinVon: Louis Bettens ***@***.***>Gesendet: Sonntag, 13. März 2022, 11:54An: dedis/popstellar ***@***.***>Cc: zoemarin ***@***.***>; Review requested ***@***.***>Betreff: Re: [dedis/popstellar] Create the JSON schema of a Digital Cash transaction  (PR #869)
@lourkeur requested changes on this pull request.

I think we should link into the parent schema:  48c981a
I have played with an example transaction, the schema is too permissive because there are things about it that should not validate: 3d59292

In protocol/query/method/message/data/dataCashTransaction.json:
> +    },
+    "action": {
+      "const": "post"
+    },
+    "transaction": {
+      "Version": {
+        "type": "integer",
+        "description": "The version of the transaction inputs"
+      },
+      "TxIn": {
+        "type": "array",
+        "description": "[Array[Objects]] array of output transactions to use as inputs",
+        "items": {
+          "type": "object",
+          "description": "Object representing a transaction output to use as an input for this transaction",
+          "properties": {

This node should specify required properties. Right now [{}] is a valid TxIn



In protocol/query/method/message/data/dataCashTransaction.json:
> +            },
+            "Sequence": {
+              "type": "bytes",
+              "length": "4",
+              "$comment": "Set as 0xFFFFFFFF"
+            }
+          }
+        }
+      },
+      "TxOut": {
+        "type": "array",
+        "description": "[Array[Objects]] array of outputs from this transactions",
+        "items": {
+          "type": "object",
+          "description": "Object representing an output of this transaction",
+          "properties": {

Same here



In protocol/query/method/message/data/dataCashTransaction.json:
> @@ -0,0 +1,100 @@
+{
+  "$id": "https://raw.githubusercontent.com/dedis/popstellar/master/protocol/query/method/message/data/dataPostTransaction.json",
+  "title": "Post a digital cash transaction",
+  "type": "object",
+  "properties": {
+    "object": {
+      "const": "transaction"
+    },
+    "action": {
+      "const": "post"
+    },
+    "transaction": {

"transaction": {} should not be valid. I think this should contain { "object": { "properties": ... }} instead of "Version" directly.

protocol/query/method/message/data/dataCashTransaction.json
```diff
@@ -0,0 +1,108 @@
...
+                  "Sig": {
+                    "type": "string",
+                    "contentEncoding": "base64",
+                    "description": "Signature on all txins and txouts using the recipient's private key"
+                  }
+                }
+              },
+              "Sequence": {
+                "type": "bytes",
+                "length": "4",
```
lourkeur, line 59:
```suggestion
                "type": "integer",
```

I think `bytes` isn't a type:
`$.transaction.TxIn[0].Sequence: object found, unknown expected`

protocol/query/method/message/data/dataCashTransaction.json
```diff
@@ -0,0 +1,108 @@
...
+              "TxOutHash": {
+                "type": "string",
+                "contentEncoding": "base64",
+                "description": "Previous (to-be-used) transaction hash"
+              },
+              "TxOutIndex": {
+                "type": "integer",
+                "description": "index of the previous to-be-used transaction"
+              },
+              "Script": {
```
lourkeur, line 37: The `{}` script is allowed. there should be required fields.

lourkeur (changes requested):

lourkeur (approved):
LGTM.

As far as what Zoé said about script naming conventions I am ambivalent.

protocol/query/method/message/data/dataCashTransaction.json
```diff
@@ -56,7 +56,8 @@
                 "required": ["Type", "Pubkey", "Sig"]
               },
               "Sequence": {
-                "type": "integer",
+                "type": "string",
+                "contentEncoding": "base64",
```
lourkeur, line 60: why base64

lourkeur (changes requested):
Honestly let's just remove this field if it's just going to be a meaningless constant

Pros:
- it looks like bitcoin.

Cons:
- it holds 0 useful information.
- It's more work to implement and maintain for no benefit.
- code is read more often than it is written. If you have code that serves no purpose it's very confusing, so you should remove it.
- We can add it later. Unlike bitcoin transactions this isn't a byte-offset format so we don't need to plan ahead.
- "Simplicity 👈" —Pierluca, 2022

haoqianzhang (dismissed):
Looks good to me.

protocol/query/method/message/data/dataCashTransaction.json
haoqianzhang: Feel free to omit the sequence.

MaximeZmt (approved):
👍

lourkeur (approved):

haoqianzhang (approved):
Looks good.

MaximeZmt (changes requested):
I think we should update that file too in order to reference it in the protocol  https://github.com/dedis/popstellar/blob/master/protocol/query/method/message/data/data.json 😅

MaximeZmt (approved):

haoqianzhang (approved):

PR #871, opened 2022-03-09
[fe1-web] Normalize import order in all files and add eslint rule as well as an autofix

Tyratox:


fe1-web/src/core/network/validation/schemas/dataSchemas.ts
```diff
@@ -1,22 +1,22 @@
+import dataAddChirp from 'protocol/query/method/message/data/dataAddChirp.json';
```
pierluca, line 1: Should we disable the rule in this file ? 
It's a lot harder to find something or identify a missing import now 😅

Tyratox: That seems to be a good idea yeah... 😅

Xelowak (approved):
Thanks for the addition, it will be much easier to manage imports now !
Will imports get fixed automatically by running eslint ?

Tyratox:
Yes, they will! 😁

Xelowak (approved):

Xelowak (approved):

PR #873, opened 2022-03-10
Update protocol specification for election#castVote to be more precise

Tyratox:
See discussion in issue #843 for more details.

### DIFF
--- a/docs/messageData.md
+++ b/docs/messageData.md
@@ -1287,6 +1287,10 @@ A member of the LAO can cast a vote by publishing an election/cast_vote message 
 election’s channel. Each member may cast multiple votes, only the last one will be counted.
 If write-in is allowed for the election then the vote has to have a write-in.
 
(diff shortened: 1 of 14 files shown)


pierluca (approved):
It's not super unambiguous yet, but it's better, hence my approval.

I would strongly encourage you to futher update messageData.md , in particular the "Casting a vote (election#cast_vote)" section, to precisely and unambiguously specifiy how this should be done.

Tyratox:
I will do that! :)

Tyratox:
I updated it to this which also includes an example :)

> For the generated vote ids, it has to be made sure that the hash is unique and consistent across all subsystems. The hash is computed based on the list of ballot option indices. To disambiguated between `[1, 0]` and `[10]`, the usage of the comma delimiter `,` in between the indices is mandatory and to make the hash unique, the list of indicies should be sorted in ascending order (from small numbers/indicies to bigger ones, from left to right). Last but not least, it must be made sure that this list does not contain any duplicates, i.e. represents a set of indices. (Each ballot option can only be selected once).
> 
> For example if the user select the ballot options with indices 5 and 2, then the hash function should be applied only to the string `2,5`, not to `2,2,5`, `5,2`, `25`, `[2,5]` or `2, 5` (note the additional space).

https://github.com/dedis/popstellar/blob/work-fe1-nico-update-castVote-spec/docs/messageData.md#casting-a-vote-electioncast_vote

PR #874, opened 2022-03-10
Center qr code scanner and rename container style

Tyratox:


0xdieman (approved):
The re-centering seems ok ! :')

PR #875, opened 2022-03-12
Changing the used checkSelfPermissionAccess function to the static one 

Tesa27:
As indicated in this [code smell](https://sonarcloud.io/project/issues?resolved=false&severities=CRITICAL&types=CODE_SMELL&id=dedis_popstellar_fe2&open=AXhvC-zxJfsYgmhMRsCn) this is a critical code smell to be addressed the method used should be the static method defined in the mother class ContextCompat.java rather than the inherited function by ActivityCompat.java this is to avoid confusion and illusion that too such method exist in both classes instead of just one defined in the mother class ContextCompat.java.


ahominal (approved):

GabrielFleischer (approved):
LGTM

PR #876, opened 2022-03-13
changed the return code in tests from -4 to -6 and tests on backend g…

Tesa27:
The return error code seems to be giving -6 instead of the -4 when an invalid lao creation request is issued.


lhmerino:
@Tesa27, what's the status on this pull request. Seems relatively simple change, do you need approval?

lhmerino:
@Tesa27, what's the status on this pull request. Seems relatively simple change, do you need approval?

pierluca:
I believe this PR should be closed, as we discussed that the problem is about the back-ends not adhering to the specification.

Tesa27:
> I believe this PR should be closed, as we discussed that the problem is about the back-ends not adhering to the specification.

Yes I agree

Issue #881, opened 2022-03-17
[BUG] Trying to display toast on non-ui thread makes the app crash


Warhedgehog:
### Description (Actual behavior)
When trying to launch an LAO without a backend running, the app crashes a few seconds after pressing the launch button

### Expected behavior
When trying to launch an Lao with no server running one would expect the app of displaying an error and continue functioning normally

### How to reproduce
1. Go to 'Launch'
2. Click on the EditText with hint  'Organization name'
3. Fill text with anything
4. Click on 'Launch' below
6. See error : app crash

### Version & Environment
This bug was reproduced on: master
- Commit: 4c032f8531b75dbaf1f9358ca02abec24c3a1b8f

##### Front-ends:
- [ ] Fe1-Web (please include browser's names & version)
- [x] Fe2-Android (emulator, API 30)
- [ ] Not applicable

##### Back-ends:
- [ ] Be1-Go
- [ ] Be2-Scala
- [x] Not applicable

##### Environment (as applicable):
 - Device: [Pixel 5 and Pixel 4a emulators]
 - OS: [Windows 10]
 - 
### Workaround
Not getting an error message displayed as a callback. In the case it was discovered, having a running server avoids making the app crash on pressing the launching button for Lao creation

### Impact
The app crashes when we try to inform the user via toast using a callback  function. The app can be restarted without consequences.

### Possible root cause
The problem is that the displaying of toasts must be made on the UI thread. By waiting for a 5 seconds connection delay, and using a callback to display the toast in case of error, the callback code is executed by another thread. This causes a crash at least  each time the `logAndShow()` function of the `ErrorUtils` class is used.

Warhedgehog:
Fixed by #882

PR #882, opened 2022-03-17
Fix app crash when trying to launch an LAO without a server running

Warhedgehog:
Previous implementation tried to toasts displayed by callback function on the thread the callback was executed on. This cause crashes. Toasts can only be showed by the UI thread. This fixes the bug by forcing the toast only to be executed by the UI thread.

This solves #881

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/error/ErrorUtils.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/error/ErrorUtils.java
@@ -41,12 +41,8 @@ public class ErrorUtils {
 
     //This makes it so that the toast is run on the UI thread
     //Otherwise it would crash
-    new Handler(context.getMainLooper()).post(new Runnable() {
-      @Override
-      public void run() {
-        Toast.makeText(context, exceptionMsg, Toast.LENGTH_LONG).show();
-      }
-    });
+    new Handler(context.getMainLooper()).post(
+            () -> Toast.makeText(context, exceptionMsg, Toast.LENGTH_LONG).show());
   }
 
   private static String getLocalizedMessage(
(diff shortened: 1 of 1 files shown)


Tesa27 (approved):
Definitely a problem good catch!

jbsv (commented):
As I I don't see any restriction in the Android [documentation](https://developer.android.com/reference/android/widget/Toast#makeText(android.content.Context,%20int,%20int)) I wonder if the "context" passed to the function is always correct ?

Warhedgehog:
From https://stackoverflow.com/questions/5018545/getapplication-vs-getapplicationcontext:
"Although in current Android Activity and Service implementations, getApplication() and getApplicationContext() return the same object, there is no guarantee that this will always be the case (for example, in a specific vendor implementation)."

This could indicate that it is better to replace `getApplication()` with `getApplication().getApplicationContext()` 


For avoiding having a null context : 
From https://developer.android.com/reference/android/content/Context#getApplicationContext():
"Return the context of the single, global Application object of the current process. This generally should only be used if you need a Context whose lifecycle is separate from the current context, that is tied to the lifetime of the process rather than the current component."

From my understanding it should be fine. I've tested with what used to crash but go into settings before the toast is displayed and it behaves as expected.

What do you think ?

jbsv:
Hmmm, so could we just use 'getApplication().getApplicationContext()' when context is null to avoid crashing ? or do we get another error which is an unexpected behaviour (like not displaying the toast) ? In which case I would stick with the proposed fix, indeed.
Alternatively, you might want to give the [snackbar](https://developer.android.com/training/snackbar/) a quick try.

GabrielFleischer:
To me the current implementation seems completely fine.
From my expirience, the context is only used to know where in view hierachy the Toast should be displayed. And that's way there is no real restriction on the input context. There is no reason to as long as the context is part of the application, it will work fine.

And as @Warhedgehog pointed out, the `context` should never be `null`. And in the rare case it is, we probably have bigger problems to consider.

The probable only safer way to implement this is to create a new Service centered around the creation of Toast. It would have access to the Application context at startup. That way we are always sure that the context is valid.

I approve this PR as I don't see any reason not to.

GabrielFleischer (approved):
LGTM

PR #883, opened 2022-03-17
LAO creation by someone that is not an organizer should fail

Tesa27:
Added a Karate test where the backend should send an error message if the organizer is not the one trying to create a lao, in the current state the backend accepts a lao creation  this can be detected by the new test failing.


tests/karate/src/test/java/be/createLAO/create.feature
```diff
@@ -75,3 +75,13 @@ Feature: Create a pop LAO
...
       Then match err contains deep {jsonrpc: '2.0', id: 1, error: {code: -4, description: '#string'}}
+
+  Scenario: Create Lao with different public key from the organizer should fail with error response
+    Given string invalidIdLao = read('classpath:data/lao/bad_lao_create_sender_not_organizer.json')
+    And   def socket = karate.webSocket(wsURL,handle)
+    When  eval socket.send(invalidIdLao)
+    *  karate.log('Sent: '+ karate.pretty(invalidIdLao))
+    And  json err = socket.listen(timeout)
+    *  karate.log('Received: '+ karate.pretty(err) )
+    Then match err contains deep {jsonrpc: '2.0', id: 1, error: {code: -4, description: '#string'}}
```
pierluca, line 86: Do we really expect a `-4` code here? 

Just thinking: these tests might be more explicit if we define the error constants somewhere, instead of relying on their numeric representation.

Tesa27: I agree constant are a better solution. I put a -4 here to be consistent with the remaining tests until we figure out what is the issue with different returning errors is

pierluca:
@GabrielFleischer please follow up on this, as I'm off.

pierluca:
@Tesa27 what should happen to this PR ?

Tesa27:
> @Tesa27 what should happen to this PR ?

Sorry I wanted to adapt the test after the self contained tests were done, I pushed the changes now

Tesa27:
I am unsure why the fe1 tests keep failing here but I opened another PR where I just omitted the json file since we wanted to have a self contained test in this PR  #1032  and it seems to work well

Tesa27:
The enhancement is done in the PR #1032 where everything is up to date and the test of sender not being the organizer for lao create is there as well, every comment and suggestion from this PR has been taken into account in other PRs or that one, so this one can be closed.

PR #885, opened 2022-03-21
Roll call Karate tests

Tesa27:
Tests if the backend sends the correct response for the following scenarios:

1. Create a roll call :
   -  if it returns a result on a valid roll call request
   -  if it returns an error if a field in roll call create request is erroneous
   -  if it returns an error if the roll call was not sent on a previously created lao 

2. Open roll call: 
   - if it returns a result message if a lao has been created and a roll call as well and a valid open roll call request was sent
   - if it returns an error if a field in roll call open is invalid
   -  if it return an error if the corresponding roll call for which an open request was sent was never created
  
3. Close roll call:
   - if it return a result message if lao, create roll call and open roll call where sent before and a valid roll call close afterwards
   - if it returns an error if a field in roll call close is invalid
   - if it returns an error if the corresponding roll call for which the roll call request was sent was never opened


tests/karate/src/test/java/be/createRollCall/closeRollCall.feature
```diff
@@ -0,0 +1,33 @@
+@env=go,scala
+Feature: Close a Roll Call
+  Background:
```
Warhedgehog, line 3: I realize this comment was copy pasted from the preexisting codebase. Nevertheless it seems to me the first line does not make sense and misses the end. Also has the double ## a meaning over the single # one ?

I think it would be good to fix them all.

Warhedgehog (commented):
Well good job 👍
Looks good to me except for the one comment.
As it is not very readable for now, we might want to add soon constants for error codes and serialize and deserialize on the fly.

tests/karate/src/test/java/be/createRollCall/closeRollCall.feature
Tesa27: Done

tests/karate/src/test/java/be/utils/simpleScenarios.feature
```diff
@@ -0,0 +1,65 @@
+@ignore @report=false
+  Feature: create lao
+    @name=valid_lao
+    Scenario: Creates valid lao
+
+      * string laoCreateReq  = read('classpath:data/lao/valid_lao_create.json')
+      * eval frontend.send(laoCreateReq)
+      * frontend_buffer.takeTimeout(timeout)
+
+      * def subscribe =
```
Warhedgehog, line 10: Is there a reason why this string is defined here rather than read as usual from a json file ?

tests/karate/src/test/java/be/createRollCall/closeRollCall.feature
```diff
@@ -0,0 +1,29 @@
...
+        # This is feature will be called to test Roll Call close
+        # The following calls makes this feature, mockFrontEnd.feature and server.feature share the same scope
+    * call read('classpath:be/utils/server.feature')
+    * call read('classpath:be/mockFrontEnd.feature')
+
+  Scenario: Close a valid roll should succeed
```
Warhedgehog, line 9: String are usually in camelCase and here in snake_case. It would probably be good if we have a convention.
I see json are always in snake_case, is it intentional ?

There's also a typo in every rol_call_close at the start of every roll call close scenario

Warhedgehog (changes requested):
Very good job overall. Apologies for not addressing everything with my first review.
A few minor details to change. 
On top of that, the file organization of the data/rollCall folder is weird. Is there a reason that roll call create jsons are in the top level file and others in sub directories ? It would seem that a create folder would help.

tests/karate/pom.xml
```diff
@@ -21,6 +21,12 @@
       <version>${karate.version}</version>
       <scope>test</scope>
     </dependency>
+      <dependency>
+          <groupId>junit</groupId>
+          <artifactId>junit</artifactId>
+          <version>4.12</version>
+          <scope>test</scope>
+      </dependency>
```
GabrielFleischer, line 29: This looks strange to me.
The project wasn't working without it ?

tests/karate/src/test/java/be/createLAO/create.feature
```diff
@@ -2,14 +2,40 @@
...
+  Scenario: Create Lao with invalid id hash should fail with an error response
+      Given string invalidIdLao = read('classpath:data/lao/bad_lao_create_id_invalid_hash.json')
+      And   def socket = karate.webSocket(wsURL,handle)
+      When  eval socket.send(invalidIdLao)
+      *  karate.log('Sent: '+ karate.pretty(invalidIdLao))
+      And  json err = socket.listen(timeout)
+      *  karate.log('Received: '+ karate.pretty(err) )
+      Then match err contains deep {jsonrpc: '2.0', id: 1, error: {code: -6, description: '#string'}}
+
+  @name=valid_lao
```
GabrielFleischer, line 38: Is this annotation used ?

I thought you were using simpleScenarios to isolate usual procedures.

GabrielFleischer (commented):
This looks promising.

I would add documentation to explain what the tests are doing.
Tests are a bit too magical to me right now.

tests/karate/src/test/java/be/utils/simpleScenarios.feature
Tesa27: subscribe and catchup are used only here, and since this is a scripting file, I think it is okay to leave it as it is since these too messages will be used only in conjunction with the underlying valid lao.

Warhedgehog (dismissed):
LGTM 👍

tests/karate/src/test/java/be/createLAO/create.feature
```diff
@@ -2,14 +2,39 @@
...
+
+  Scenario: Create Lao with invalid id hash should fail with an error response
+      Given string invalidIdLao = read('classpath:data/lao/bad_lao_create_id_invalid_hash.json')
+      And   def socket = karate.webSocket(wsURL,handle)
+      When  eval socket.send(invalidIdLao)
+      *  karate.log('Sent: '+ karate.pretty(invalidIdLao))
+      And  json err = socket.listen(timeout)
+      *  karate.log('Received: '+ karate.pretty(err) )
+      Then match err contains deep {jsonrpc: '2.0', id: 1, error: {code: -6, description: '#string'}}
+
```
GabrielFleischer, line 37: Is this really part of this PR ?
And I'm pretty sure I have seen these changes in #876

tests/karate/src/test/java/be/createLAO/create.feature
```diff
@@ -2,14 +2,39 @@
...
+        # This is feature will be called  to test an LAO creation
         # This call makes this feature and server.feature share the same scope
         # Meaning they share def variables, configurations ...
         # Especially JS functions defined in server.feature can be directly used here thanks to Karate shared scopes
-        # * call wait <timeout>
-        # * karate.set(varName, newValue)
     * call read('classpath:be/utils/server.feature')
 
+  Scenario: Create Lao request with empty lao name should fail with an error response
+    Given string  emptyNameReq = read('classpath:data/lao/bad_lao_create_empty_name.json')
```
GabrielFleischer, line 12: I think it would be good to use only one implementation of a web socket throughout the whole codebase.

It might create confusion in the future.

This doesn't need to be done in this PR in my opinion, but it should be noted.
What do you think ?

GabrielFleischer (changes requested):
LGTM, except for the part of #876 that found its way into this PR.

tests/karate/src/test/java/be/createLAO/create.feature
Tesa27: No it is not you are right, I changed this now.

tests/karate/src/test/java/be/createLAO/create.feature
Tesa27: Yes, this is a great consistency idea, I will address this in my next PR.

GabrielFleischer (approved):
LGTM 👍

PR #888, opened 2022-03-21
Enhancement to the wallet home screen

0xdieman:
- Added a button to directly use the generated seed on wallet setup
- Added a dropdown selector to select among the available roll call Poptokens
- Added a mock state to be able to use some mock Poptokens
- As only one Poptoken is selected at a time, the QR code is shown by default
- Fixed the wallet returning to setup page when connecting to/creating an LAO


fe1-web/src/features/wallet/components/RollCallTokensDropDown.tsx
```diff
@@ -0,0 +1,50 @@
...
+const RollCallTokensDropDown = (props: IPropTypes) => {
+  const { onTokenChange } = props;
+  const { rollCallTokens } = props;
+
+  const options = useMemo(() => {
+    return rollCallTokens.map((rc) => {
+      const value = rc?.token.publicKey.valueOf() || '';
+      return <Picker.Item key={value} label={value} />;
+    });
+  }, [rollCallTokens]);
```
Tyratox, line 27: Maybe that is just me but as long as there are no performance problems with this code (are there any?), I find it confusing to have some UI components memoized instead of just having this code inline on line 38?

fe1-web/src/features/wallet/navigation/WalletNavigation.tsx
```diff
@@ -18,15 +13,22 @@ const Stack = createStackNavigator();
  * Allows to navigate between the wallet screens.
  */
 export default function WalletNavigation() {
+  const home = () => {
+    return <Stack.Screen name={STRINGS.navigation_wallet_home_tab} component={WalletHome} />;
+  };
+  const setup = () => {
+    return <Stack.Screen name={STRINGS.navigation_wallet_setup_tab} component={WalletSetup} />;
+  };
```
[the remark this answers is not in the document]
Tyratox: Or even better:

```
const home = {name: STRINGS.navigation_wallet_home_tab, Component: WalletHome};
const setup = {name: STRINGS.navigation_wallet_setup_tab, Component: WalletSetup};
```

fe1-web/src/features/wallet/objects/RollCallToken.ts
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    if (obj.token === undefined) {
+      throw new Error("Undefined 'id' when creating 'RollCallToken'");
+    }
+    if (obj.laoId === undefined) {
+      throw new Error("Undefined 'name' when creating 'RollCallToken'");
+    }
+    if (obj.rollCallId === undefined) {
+      throw new Error("Undefined 'creation' when creating 'RollCallToken'");
```
Tyratox, line 30: :)

fe1-web/src/features/wallet/objects/RollCallToken.ts
```diff
@@ -0,0 +1,49 @@
...
+  public static fromState(rct: RollCallTokenState): RollCallToken {
+    return new RollCallToken({
+      token: PopToken.fromState(rct.token),
+      laoId: new Hash(rct.laoId),
+      rollCallId: new Hash(rct.rollCallId),
+    });
+  }
+
+  public toState(): RollCallTokenState {
+    return JSON.parse(JSON.stringify(this));
```
Tyratox, line 47: I guess that works and is also done in various other places but I feel like it is not a very clean way to remove methods etc. (TBH I am even surprised JSON.stringify properly works and does not output something like `toState: [function]`)

@0xdieman @pierluca @Xelowak What do you think?

fe1-web/src/features/wallet/objects/__mocks__/mockWallet.ts
```diff
@@ -0,0 +1,43 @@
...
+  });
+};
+const mockRCID0: string = '0mock';
+const mockRCID1: string = '1mock';
+const hashMock0 = new Hash(mockRCID0);
+const hashMock1 = new Hash(mockRCID1);
+/*
+ * Generates a mock state with some mock popTokens
+ */
+export function useMockWalletState() {
```
Tyratox, line 25: The `use` prefix is usually used for react hooks, this does not seem to be one or am I missing something?

fe1-web/src/features/wallet/objects/Wallet.ts
```diff
@@ -43,3 +44,44 @@ export async function recoverWalletPoPTokens(): Promise<Record<string, Record<st
...
   return tokens;
 }
+/**
+ * Recovers all PoP tokens associated with this wallet.
+ * @return Promise<RollCallToken[]>
+ * @remarks
+ * This is implemented by checking through all known Roll Calls of all known LAOs,
+ * generating tokens for them and checking if the token is a verified attendee.
+ */
+export async function recoverWalletRollCallTokens(): Promise<RollCallToken[]> {
```
Tyratox, line 54: You could also pass the roll calls as input to this function, would make testing a bit easier :)

fe1-web/src/features/wallet/screens/WalletHome.tsx
```diff
@@ -1,51 +1,111 @@
...
+      .then((rct) => {
+        setTokens(rct);
+        setSelectedToken(rct[0]);
+      })
+      .catch((e) => {
+        console.debug(e);
+      });
+  }, [rollCalls, isDebug]);
+
+  const setDebug = (isOn: boolean) => {
```
Tyratox, line 58: Suggestion: Rename to `toggleDebugMode` and directly access the `isDebug` variable without accepting it as an argument?

I was a bit confused by the similar names at first 😅

fe1-web/src/features/wallet/components/RollCallTokensDropDown.tsx
0xdieman: True, I thought about the lot of tokens case but even then it might not be an issue

fe1-web/src/features/wallet/navigation/WalletNavigation.tsx
```diff
@@ -18,15 +13,22 @@ const Stack = createStackNavigator();
...
+  const hasSeed = WalletStore.hasSeed();
```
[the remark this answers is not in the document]
0xdieman: Yep, but if I'm correct it solves the issue I wanted to solve, that is the synced wallet screen disappearing for the wallet setup screen when joining/creating an lao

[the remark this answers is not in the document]
0xdieman: At least for now

fe1-web/src/features/wallet/navigation/WalletNavigation.tsx
```diff
@@ -18,15 +13,22 @@ const Stack = createStackNavigator();
...
     <Stack.Navigator
       screenOptions={{
         headerShown: false,
       }}>
-      <Stack.Screen name={STRINGS.navigation_home_tab_wallet} component={WalletHome} />
-      <Stack.Screen name={STRINGS.navigation_show_seed_wallet} component={WalletShowSeed} />
-      <Stack.Screen name={STRINGS.navigation_insert_seed_tab_wallet} component={WalletSetSeed} />
-      <Stack.Screen name={STRINGS.navigation_synced_wallet} component={WalletSyncedSeed} />
+      {hasSeed ? home() : setup()}
+      {hasSeed ? setup() : home()}
```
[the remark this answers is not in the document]
0xdieman: Agreed !

fe1-web/src/features/wallet/objects/RollCallToken.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  constructor(obj: Partial<RollCallToken>) {
+    if (obj === undefined || obj === null) {
+      throw new Error(
+        'Error encountered while creating a RollCallToken object: undefined/null parameters',
+      );
+    }
+
+    if (obj.token === undefined) {
+      throw new Error("Undefined 'id' when creating 'RollCallToken'");
```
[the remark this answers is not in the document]
0xdieman: Woops

fe1-web/src/__tests__/utils/TestUtils.ts
```diff
@@ -39,6 +43,28 @@ export const mockLaoState: LaoState = {
...
+  name: mockName,
+  location: mockLocation,
+  creation: mockTimestampStart.valueOf(),
+  proposedStart: mockTimestampStart.valueOf(),
+  proposedEnd: mockTimestampEnd.valueOf(),
+  status: RollCallStatus.CLOSED,
+  attendees: mockAttendees,
+};
+
+export const mockRollCall = RollCall.fromState(mockRollCallState);
```
Xelowak, line 66: Do you really need all of these to be exported ? Try to export only what you will need elsewhere. If you realize that it is used in only one test file, you'd better define it in it directly. 
In addition, I would rename these variables to make clear that they are all linked to roll call tests (for example, `mockId` and `mockName` are clearly related to roll calls if we see what is in there, but their name doesn't reflect this).

Xelowak (approved):
Thanks for the changes ! Nevertheless, it seems that you don't have enough coverage to merge for now. I think you should be able to achieve the 50%.

Tyratox:
`architecture-data-flow.png` seems to have changed, was that intended?

fe1-web/src/features/wallet/objects/RollCallToken.ts
```diff
@@ -0,0 +1,64 @@
+import { Hash, KeyPairState, PopToken } from 'core/objects';
+
+export interface RollCallTokenState {
+  token: KeyPairState;
+  laoId: string;
+  rollCallId: string;
+  rollCallName: string;
+}
```
Tyratox, line 8: I don't think I currently see the purpose of this class but I am probably missing something 😅 I think previously we did not store the token because we could rely on the deterministic generation based on the seed, why does that not work anymore?

Tyratox: Oh wait, you do not actually store it but return it in a nicer format from `recoverWalletRollCallTokens`, right? Then we probably don't need `RollCallTokenState`, `fromState` and `toState`?

fe1-web/src/features/wallet/objects/Wallet.ts
```diff
@@ -1,45 +1,32 @@
...
+      if (rc.containsToken(popToken)) {
+        return new RollCallToken({
+          token: popToken,
+          laoId: currentLao.id,
+          rollCallId: rc.id,
+          rollCallName: rc.name,
+        });
+      }
+      return undefined;
+    });
```
Tyratox, line 29: Since we return a promise anyway, it might help readability to use an async function instead?

fe1-web/src/features/wallet/screens/WalletHome.tsx
```diff
@@ -1,51 +1,121 @@
...
 
+const rollCallSelector = makeEventByTypeSelector<RollCall>(LaoEventType.ROLL_CALL);
+
 /**
- * Wallet home screen
+ * Wallet UI once the wallet is synced
  */
-const WalletHome = () => {
-  // FIXME: Navigation should use a defined type here (instead of any)
-  const navigation = useNavigation<any>();
```
Tyratox, line 25: Is there a reason you prefer the prop over the hook? Because I usually prefer the hook as it (at least in the future) allows you to have the proper typing without  modifying the IPropTypes returned by `InferProps` 😅 But I'd be interested in your thoughts :)

fe1-web/src/features/wallet/screens/WalletHome.tsx
```diff
@@ -1,51 +1,121 @@
...
+      createDummyWalletState().then(() => setIsDebug(true));
+    }
+  };
+  const tokenInfos = () => {
+    if (selectedToken) {
+      const rollCallName = `Roll Call name: ${selectedToken.rollCallName.valueOf()}`;
+      return (
+        <View style={containerStyles.centeredXY}>
+          <TextBlock size={18} text={rollCallName} />
+          <QRCode value={selectedToken.token.publicKey.valueOf()} visibility />
```
Tyratox, line 78: When testing this file make sure to mock the QRCode, the jest snapshot will be huge otherwise 😂

tests/karate/README.md
```diff
@@ -4,8 +4,6 @@ This folder implements tests for the back-end and front-end.
 
 ## First Setup
 
-### All
-Make sure that you have [maven](https://maven.apache.org/download.cgi) installed
 ### Backend
 
 To test a backend, you don't need more setup that what is needed to build that backend. Use the resources provided by
```
Tyratox, line 9: What happened here? Did you actually change that or is that a merging issue?

tests/karate/src/test/java/fe/FrontEndTest.java
```diff
@@ -6,6 +6,6 @@
 
   @Karate.Test
   Karate testCreateLAO() {
-    return Karate.run("classpath:fe/LAO/create_lao.feature");
+    return Karate.run("classpath:fe/create_lao.feature");
```
Tyratox, line 9: same question, merge issue?

tests/karate/src/test/java/fe/create_lao.feature
```diff
@@ -25,4 +25,4 @@ Feature: Create LAO
 
     Then match create_lao contains deep { method: 'publish', params: { channel: '/root' }}
     Then match subscribe contains deep { method: 'subscribe' }
-    Then match catchup contains deep { method: 'catchup' }
\ No newline at end of file
+    Then match catchup contains deep { method: 'catchup' }
```
Tyratox: once more

tests/karate/src/test/java/fe/utils/android.feature
```diff
@@ -20,19 +20,12 @@ Feature: android page object
...
-    * def tab_wallet_confirm_selector = '#com.github.dedis.popstellar:id/button_confirm_seed'
-
-    # Lao Event List
-    * def add_event_selector = '#com.github.dedis.popstellar:id/add_future_event_button'
-
   @name=basic_setup
-  Scenario: Setup connection to the backend and complete wallet initialization
-    Given driver driverOptions
+    Scenario: Setup connection to the backend and complete on the home page
+      Given driver driverOptions
```
Tyratox, line 25: and another one

tests/karate/src/test/java/fe/utils/web.feature
```diff
@@ -23,7 +22,7 @@ Feature: web test
     Given driver driverOptions
 
     # Create and import mock backend
-    And call read('classpath:fe/net/mockBackend.feature')
+    And call read('classpath:fe/net/mockbackend.feature')
```
Tyratox, line 25: aaand another one 😂

tests/karate/src/test/java/karate-config.js
```diff
@@ -35,13 +35,13 @@ function fn() {
...
-        "app" : "../../fe2-android/app/build/outputs/apk/debug/app-debug.apk",
+        "app" : "../../fe2-android/app/build/outputs/apk/prod/debug/app-prod-debug.apk",
         "newCommandTimeout" : 300,
         "platformVersion" : "9.0",
         "platformName" : "Android",
         "connectHardwareKeyboard" : true,
         "deviceName" : "emulator-5554",
-        "avd" : "Pixel_4_API_30",
+        "avd" : "Pixel_2_API_29",
         "automationName" : "UiAutomator2"
```
Tyratox, line 45: And I think that is the last one haha

Tyratox (changes requested):
I think the most important thing to check is the changed architecture data flow image and the karate tests that to me appear to be changed by accident. Otherwise you can look at my comments (feel free to ignore the code style related ones) and also then see what Kilian thinks :)

fe1-web/src/features/wallet/objects/DummyWallet.ts
```diff
@@ -0,0 +1,115 @@
...
+import { dispatch } from 'core/redux';
+import { addEvent, removeEvent } from 'features/events/reducer';
+import { connectToLao, disconnectFromLao, removeLao } from 'features/lao/reducer';
+import { EventTypeRollCall, RollCall, RollCallStatus } from 'features/rollCall/objects';
+
+import { Lao, LaoState } from '../../lao/objects';
+import { generateToken } from './Token';
+
+// region Dummy values definition
+// MOCK LAO
```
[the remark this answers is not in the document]
0xdieman: That's to define a comment region to be able to hide it in the IDE

fe1-web/src/features/wallet/objects/RollCallToken.ts
0xdieman: Right, as I wasn't sure exactly how I was going to use those objects I defined a state anyways, but for now at least it is not of any use.

fe1-web/src/features/wallet/components/RollCallTokensDropDown.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const onChange = (pk: string) => {
+    const found = rollCallTokens.find((rct) => rct?.token.publicKey.valueOf() === pk);
+    if (found) {
+      onTokenChange(found);
+    }
+  };
+
+  return (
+    <Picker
+      onValueChange={(val: any) => onChange(val)}
```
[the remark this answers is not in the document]
0xdieman: Right, I changed the implementation to make use of the itemValue that now corresponds to the index of the selected token in the rollCallTokens array, which makes it clearer !

fe1-web/src/features/wallet/components/RollCallTokensDropDown.tsx
```diff
@@ -0,0 +1,53 @@
...
+  );
+};
+
+const propTypes = {
+  onTokenChange: PropTypes.func.isRequired,
+  rollCallTokens: PropTypes.arrayOf(PropTypes.instanceOf(RollCallToken)).isRequired,
+  selectedToken: PropTypes.instanceOf(RollCallToken).isRequired,
+};
+RollCallTokensDropDown.propTypes = propTypes;
+type IPropTypes = PropTypes.InferProps<typeof propTypes>;
```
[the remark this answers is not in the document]
0xdieman: That would be nice

fe1-web/src/features/wallet/objects/Wallet.ts
0xdieman: In the callback function ?

fe1-web/src/__tests__/components/MockNavigator.tsx
```diff
@@ -7,7 +7,7 @@ const MockNavigator = ({
   component,
   params = {},
 }: {
-  component: React.ComponentType<unknown>;
+  component: React.ComponentType<any>;
```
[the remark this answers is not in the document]
0xdieman: That was related to the `proptypes` navigation, but migrated to the hook !

fe1-web/src/features/wallet/components/RollCallTokensDropDown.tsx
```diff
@@ -0,0 +1,53 @@
...
+    margin: 'auto',
+  },
+});
+const RollCallTokensDropDown = (props: IPropTypes) => {
+  const { onTokenChange } = props;
+  const { rollCallTokens } = props;
+  const { selectedToken } = props;
+
+  const options = rollCallTokens.map((rc) => {
+    const value = rc?.token.publicKey.valueOf() || '';
```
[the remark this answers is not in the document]
0xdieman: The rollCallTokens should not be empty when passed in the component, so I fixed this by making it a `Proptypes` requirement.

fe1-web/src/features/wallet/screens/WalletHome.tsx
```diff
@@ -1,49 +1,114 @@
...
-import { WalletStore } from '../store';
+import { RollCallTokensDropDown } from '../components';
+import * as Wallet from '../objects';
+import { createDummyWalletState, clearDummyWalletState } from '../objects/DummyWallet';
+import { RollCallToken } from '../objects/RollCallToken';
 
 const styles = StyleSheet.create({
+  homeContainer: {
+    ...containerStyles.centeredXY,
+    padding: '30px',
```
Xelowak, line 22:
```suggestion
    padding: 30,
```
I think both work, but we're not used to this notation in the project.

fe1-web/src/features/wallet/screens/WalletHome.tsx
```diff
@@ -1,49 +1,114 @@
...
-    if (WalletStore.hasSeed()) {
-      navigation.navigate(STRINGS.navigation_synced_wallet);
+  useEffect(() => {
+    if (lao && rollCalls[lao.id.valueOf()]) {
+      Wallet.recoverWalletRollCallTokens(rollCalls, lao)
+        .then((rct) => {
+          if (rct.length > 0) {
+            setTokens(rct);
+            setSelectedTokenIndex(0);
+          }
```
Xelowak, line 54: What is supposed to happen if `rct.length = 0` ?

fe1-web/src/features/rollCall/screens/RollCallOpened.tsx
```diff
@@ -105,7 +105,7 @@ const RollCallOpened = () => {
 
   const onCloseRollCall = () => {
     const attendeesList = Array.from(attendees).map((key: string) => new PublicKey(key));
-    return requestCloseRollCall(rollCallID, attendeesList)
+    return requestCloseRollCall(rollCallIDAlias, attendeesList)
```
Xelowak, line 108: Are you sure this works as intended ? I'm concerned because I'm pretty sure you need to close the roll call using the base `rollCallID` and not its alias.

Xelowak (commented):
Nice job, and thanks for all the changes.

As a general remark, I suggest you take more care into separating blocks with an empty line. No one really likes code that is written in one entire block without spaces.

I think you're using the wrong library for your tests. You should use jest functions like `.toThrow()` or `.toEqual()`, so be sure to import `jest-extended` for unit tests.

Since this PR contains changes from [this one](https://github.com/dedis/popstellar/pull/951), you can take this branch as the base branch of this PR so that we don't see changes twice in 2 different PRs.

fe1-web/src/features/wallet/components/RollCallTokensDropDown.tsx
```diff
@@ -0,0 +1,43 @@
...
+    fontSize: 20,
+    textAlign: 'center',
+    margin: 'auto',
+  },
+});
+const RollCallTokensDropDown = (props: IPropTypes) => {
+  const { onIndexChange, rollCallTokens, selectedTokenIndex } = props;
+
+  const options = rollCallTokens.map((rc, index) => {
+    return <Picker.Item value={index} label={rc.token.publicKey.valueOf()} />;
```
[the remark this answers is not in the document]
0xdieman: Yes it's a good idea, I think I'll do that in a next PR, along with showing the names of the roll calls instead of the poptoken public keys

fe1-web/src/features/rollCall/screens/RollCallOpened.tsx
0xdieman: The message API actually expects us to use the idAlias because it uses it to create a new updateId, which depends on the last idAlias

fe1-web/src/features/wallet/screens/WalletHome.tsx
0xdieman: No valid roll call tokens where retrieved so nothing...

0xdieman:
> Nice job, and thanks for all the changes.
> 
> As a general remark, I suggest you take more care into separating blocks with an empty line. No one really likes code that is written in one entire block without spaces.
> 
> I think you're using the wrong library for your tests. You should use jest functions like `.toThrow()` or `.toEqual()`, so be sure to import `jest-extended` for unit tests.
> 
> Since this PR contains changes from [this one](https://github.com/dedis/popstellar/pull/951), you can take this branch as the base branch of this PR so that we don't see changes twice in 2 different PRs.

This PR contains changes from the other one but the entire files including their names changed so I would prefer not to take it as a base to avoid creating some weird merging conflicts.

fe1-web/src/features/wallet/objects/__tests__/RollCallToken.test.ts
```diff
@@ -0,0 +1,23 @@
+import { mockKeyPair, mockLao, mockRC } from '__tests__/utils';
+import { PopToken } from 'core/objects';
+
+import { RollCallToken } from '../RollCallToken';
+
+const mockToken = PopToken.fromState(mockKeyPair.toState());
+
+describe('Roll call token object', () => {
+  it('throws when token is undefined', () => {
+    expect(() => new RollCallToken({ token: undefined })).toThrow();
```
Xelowak, line 10:
```suggestion
    expect(() => new RollCallToken({ token: undefined })).toThrow(Error);
```
The same goes for the other ones. It allows us to check if the correct type of error is thrown.

fe1-web/src/features/wallet/objects/__tests__/Wallet.test.ts
```diff
@@ -0,0 +1,51 @@
+import { expect } from 'chai';
```
Xelowak, line 1: This is not the `expect` we use. As I said in my first comment, you should use jest.

fe1-web/src/features/wallet/objects/__tests__/Wallet.test.ts
```diff
@@ -0,0 +1,51 @@
...
+    const expected = new RollCallToken({
+      token: popToken,
+      laoId: mockLao.id,
+      rollCallId: mockRC.id,
+      rollCallName: mockRC.name,
+    });
+
+    const state = { [mockLao.id.valueOf()]: { [mockRC.id.valueOf()]: mockRC } };
+    const rct = await recoverWalletRollCallTokens(state, mockLao);
+    expect(rct).to.eql([expected]);
```
Xelowak, line 42: ... and this could be replaced
```suggestion
    expect(rct).toEqual([expected]);
```

fe1-web/src/features/wallet/screens/__tests__/WalletCreateSeed.test.tsx
```diff
@@ -0,0 +1,23 @@
+import { render } from '@testing-library/react-native';
+import React from 'react';
+
+import MockNavigator from '__tests__/components/MockNavigator';
+
+import { WalletCreateSeed } from '../index';
+
+const mockSeed = 'one two three';
+
+jest.mock('../../objects/Seed', () => {
```
Xelowak, line 10: This path is still relative.

Xelowak (commented):
Thanks for the changes. I wrote some small remarks but this should be merged soon!

fe1-web/src/features/wallet/objects/__tests__/Wallet.test.ts
0xdieman: Right, I forgot about this one :(

fe1-web/src/features/wallet/objects/DummyWallet.ts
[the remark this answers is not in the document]
Tyratox: I meant the Mock lao comment :)

fe1-web/src/features/wallet/objects/__tests__/RollCallToken.test.ts
```diff
@@ -0,0 +1,23 @@
...
+    expect(() => new RollCallToken({ token: mockToken })).toThrow(Error);
+  });
+  it('throws when rollCallId is undefined', () => {
+    expect(() => new RollCallToken({ token: mockToken, laoId: mockLao.id })).toThrow(Error);
+  });
+  it('throws when rollCallName is undefined', () => {
+    expect(
+      () => new RollCallToken({ token: mockToken, laoId: mockLao.id, rollCallId: mockRC.id }),
+    ).toThrow(Error);
+  });
```
Tyratox, line 22: Nice! Maybe just add one more test where you call the constructor with all required arguments and check `expect().not.toThrow(Error)` and optionally whether the fields of the return object are set to the expected values?

fe1-web/src/features/wallet/screens/WalletHome.tsx
```diff
@@ -1,49 +1,117 @@
...
+ * Wallet UI once the wallet is synced
  */
 const WalletHome = () => {
-  // FIXME: Navigation should use a defined type here (instead of any)
+  const [tokens, setTokens] = useState<RollCallToken[]>();
+  const [selectedTokenIndex, setSelectedTokenIndex] = useState(-1);
+  const [isDebug, setIsDebug] = useState(false);
+  const rollCalls = useSelector(rollCallSelector);
+  const lao = useSelector(selectCurrentLao);
   const navigation = useNavigation<any>();
```
Tyratox, line 45: It is probably a good idea to leave the `// FIXME: Navigation should use a defined type here (instead of any)` comment so that it is not forgotten when somebody takes care of defining the navigation types? :)

fe1-web/src/features/wallet/screens/WalletHome.tsx
```diff
@@ -1,49 +1,117 @@
...
+      <View style={styles.tokenSelectContainer}>
+        {tokens && (
+          <RollCallTokensDropDown
+            rollCallTokens={tokens}
+            onIndexChange={setSelectedTokenIndex}
+            selectedTokenIndex={selectedTokenIndex}
+          />
+        )}
+      </View>
+      {tokens && tokenInfos()}
```
Tyratox, line 99: I would either inline the function or make it a component, having a non-component returning jsx is a bit counterintuitive to me? 😅 But I guess this is a design choice

Tyratox (approved):
`architecture-data-flow.png` is still changed :)

Otherwise just some minor suggestions but otherwise lgtm 😁

fe1-web/src/features/wallet/screens/WalletHome.tsx
0xdieman: Why would that not be intuitive ?

fe1-web/src/features/wallet/objects/__tests__/RollCallToken.test.ts
0xdieman: Yes !

Tyratox (approved):

fe1-web/src/features/wallet/screens/WalletHome.tsx
Tyratox: A component is nothing else than a function returning jsx so I usually expect non-component functions (non-captialized) not to return jsx :)

fe1-web/src/features/wallet/objects/Wallet.ts
pierluca: I've reopend the conversation as it was marked resolved but I don't see a decision here 😅

fe1-web/src/features/wallet/objects/Wallet.ts
```diff
@@ -1,45 +1,32 @@
...
-        }),
-    );
-
-    ops = ops.concat(newOps);
-  }
-
-  await Promise.all(ops);
-
-  return tokens;
+export async function recoverWalletRollCallTokens(
```
pierluca, line 11: I have a strong concern here: it appears we've significantly modified the behaviour of the wallet and reduced the scope of its functionality.
Effectively, the wallet was (previously) able to display all the tokens from all the roll calls in all the LAOs.
Now, it's impossible to see the tokens from a LAO we're not connected to, despite this being in the [Digital Wallet specifications](https://docs.google.com/document/d/1bIEg31ckdsCnOT97uRgaDtkWHFbK2IJ4Zt5QyTNJxt8/edit?disco=AAAAYqLrbD4)

fe1-web/src/features/wallet/screens/WalletHome.tsx
```diff
@@ -1,49 +1,119 @@
...
+    } else {
+      createDummyWalletState().then(() => setIsDebug(true));
+    }
+  };
+  const tokenInfos = () => {
+    if (selectedTokenIndex !== -1 && tokens) {
+      const rollCallName = `Roll Call name: ${tokens[selectedTokenIndex].rollCallName.valueOf()}`;
+      return (
+        <View style={containerStyles.centeredXY}>
+          <TextBlock size={18} text={rollCallName} />
```
pierluca, line 81: It looks as though a regular `<Text>` component would work here.

fe1-web/src/features/wallet/screens/WalletHome.tsx
```diff
@@ -1,49 +1,119 @@
...
+  }, [rollCalls, isDebug, lao]);
+
+  const toggleDebugMode = () => {
+    if (isDebug) {
+      clearDummyWalletState();
+      setIsDebug(false);
+    } else {
+      createDummyWalletState().then(() => setIsDebug(true));
+    }
+  };
```
pierluca, line 75: In the past, I hit (several times) a bug which prevented reloading the application after enabling the dummy wallet. Has this been fixed ? 

Also, do we really need the dummy wallet in the production code of the application ? Or was it just a debugging convenience ?

fe1-web/src/features/wallet/screens/WalletHome.tsx
```diff
@@ -1,49 +1,119 @@
...
+      Wallet.recoverWalletRollCallTokens(rollCalls, lao)
+        .then((rct) => {
+          if (rct.length > 0) {
+            setTokens(rct);
+            setSelectedTokenIndex(0);
+          }
+        })
+        .catch((e) => {
+          console.debug(e);
+        });
```
pierluca, line 60: In this kind of cases, where you have a clear "normal" flow and a short alternate or exception flow (e.g., showing tokens OR dealing with the lao or roll call not existing), I'd suggest the following structure to limit indentation and make your code clearer:

```suggestion
  useEffect(() => {
    if (!lao || !rollCalls[lao.id.valueOf()]) {
      // Clear tokens screen state
      setSelectedTokenIndex(-1);
      setTokens(undefined);
      return;
    }
    
    Wallet.recoverWalletRollCallTokens(rollCalls, lao)
      .then((rct) => {
        if (rct.length > 0) {
          setTokens(rct);
          setSelectedTokenIndex(0);
        }
      })
      .catch((e) => {
        console.debug(e);
      });   
```

By returning early and limiting indentation and "jumps" in code flow, it makes following the logic much easier

fe1-web/src/features/wallet/screens/__tests__/WalletHome.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+beforeEach(() => {
+  jest.clearAllMocks();
+});
+describe('Wallet home', () => {
+  it('renders correctly with an empty wallet', () => {
+    const component = render(<MockNavigator component={WalletHome} />).toJSON();
+    expect(component).toMatchSnapshot();
+  });
+  it('renders correctly with a non empty wallet', async () => {
```
pierluca, line 27: for readability you might want to add empty lines between the different `it` clauses

pierluca (commented):
Disclaimer: I'm running a fever and I've only had a quick high-level look at this PR.
Not all comments may be on target and I may have missed a number of things.

fe1-web/src/core/components/TextBlock.tsx
```diff
@@ -37,13 +39,15 @@ const propTypes = {
   bold: PropTypes.bool,
   visibility: PropTypes.bool,
   color: PropTypes.string,
+  size: PropTypes.number,
```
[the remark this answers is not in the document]
Tyratox: Hmm yes this is not optimal indeed.. Yes I agree, I think just using the `Text` component with the correct styles is a better solution!
Also with this component it is easy to create text components in a dozen different sizes / colors whereas usually in UIs you try to limit the number of different font sizes / colors etc (which each can be its own style in a StyleSheet).

fe1-web/src/features/wallet/objects/Wallet.ts
0xdieman: It appeared to me that we did not need to see the tokens from every LAO, but only the ones in which we are connected because every action we could take with this token would be related to this particular LAO(or even this particular RollCall) ?

fe1-web/src/features/wallet/screens/WalletHome.tsx
0xdieman: It should be fixed, but yes it is just for debugging convenience

fe1-web/src/features/wallet/screens/WalletHome.tsx
0xdieman: True ! 👍

fe1-web/src/core/components/TextBlock.tsx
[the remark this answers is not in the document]
Xelowak: I agree that the `TextBlock` component is not practical to use at all. I struggled a little bit with it last semester and that's why I used a lot of `Text`s in the end. However, I think that having a generic text for the application is a good idea for consistency (like font and color), but it can be done with styles.

fe1-web/src/features/wallet/screens/__tests__/WalletHome.test.tsx
Xelowak: I agree!

Xelowak (commented):
Thanks for applying the changes I asked for! I let you deal with Pierluca's remarks.

fe1-web/src/features/wallet/objects/Wallet.ts
pierluca: It depends how you look at the Wallet. In principle, the wallet could be used offline / while disconnected from a LAO, though in relation with it. I believe that's one of the aims of the digital cash project.

If you think about it that way, it makes sense that you'd like to see (at least) all the current PoP tokens (from all LAOs) and possibly also historical (now invalid) PoP tokens.

Don't get me wrong, you're absolutely right that it might make sense to be able to restrict the view to a given LAO, and maybe a way to do that in the UI would be nice, I just wouldn't drop the ability to see everything else

fe1-web/src/core/components/TextBlock.tsx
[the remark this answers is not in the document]
pierluca: It seems there's a consensus, let's try to slowly phase out the `TextBlock` then 😄

fe1-web/src/features/wallet/objects/Wallet.ts
0xdieman: Yes, that is true, I think it makes more sense for now to keep this branch as is and I'll work around a way to be able to see every PopToken on another branch, for example by having an LAO selector, showing the LAO we're currently connected to and every other LAO and being able to select the one we are interested in.
Tell me if it makes sense to you

fe1-web/src/features/wallet/objects/DummyWallet.ts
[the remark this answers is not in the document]
0xdieman: oh right

fe1-web/src/features/wallet/objects/Wallet.ts
pierluca: It makes sense, given the already-wide scope of this PR.
Do try to address that issue in the next PR though, lest we lose functionality along the way :)

pierluca (dismissed):

Xelowak (approved):
Re-approving

PR #895, opened 2022-03-23
Fix fe-1 bugs

Tyratox:
This PR fixes many of the bugs discovered yesterday. For one of the fixes (storing the correct server address for a lao for attendees), the server address had to be attached to the requests, responses and messages as otherwise it is no longer clear from which server a message came.


0xdieman (dismissed):

fe1-web/src/core/network/jsonrpc/__tests__/ExtendedJsonRpcRequest.test.ts
```diff
@@ -0,0 +1,29 @@
...
+      { receivedFrom: mockAddress },
+      mockJsonRequest,
+    );
+    expect(extendedRequest.request).toEqual(new JsonRpcRequest(mockJsonRequest));
+    expect(extendedRequest.receivedFrom).toEqual(mockAddress);
+  });
+
+  it('cannot create a new instance with missing receivedFrom parameter', () => {
+    const fn = () => new ExtendedJsonRpcRequest({}, mockJsonRequest);
+    expect(fn).toThrow();
```
Xelowak, line 27:
```suggestion
    expect(fn).toThrow(ProtocolError);
```
It might be better to ensure it is the right error.

fe1-web/src/core/network/jsonrpc/__tests__/ExtendedJsonRpcResponse.test.ts
```diff
@@ -0,0 +1,23 @@
...
+      { receivedFrom: mockAddress },
+      mockJsonResponse,
+    );
+    expect(extendedResponse.response).toEqual(new JsonRpcResponse(mockJsonResponse));
+    expect(extendedResponse.receivedFrom).toEqual(mockAddress);
+  });
+
+  it('cannot create a new instance with missing receivedFrom parameter', () => {
+    const fn = () => new ExtendedJsonRpcResponse({}, mockJsonResponse);
+    expect(fn).toThrow();
```
Xelowak, line 21:
```suggestion
    expect(fn).toThrow(ProtocolError);
```

Xelowak (commented):
Nice job ! I just left a couple of comments about test data that could be generic (like `'some address'`) and ignored issues/warnings. I don't know if you can remove them all, but it could be nice to search for another solution if possible.

fe1-web/src/core/network/__tests__/JsonRpcApi.test.ts
```diff
@@ -0,0 +1,175 @@
+/* eslint-disable @typescript-eslint/dot-notation */
```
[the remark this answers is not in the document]
Tyratox: Yes and no. The problem is that private class properties in typescript can only be accessed using the bracket notation, see https://github.com/microsoft/TypeScript/issues/19335.

Another way would be to cast to any but I find the bracket notation a nicer solution, what do you think?

fe1-web/src/core/network/__tests__/JsonRpcApi.test.ts
```diff
@@ -0,0 +1,175 @@
...
+// endregion
+
+beforeAll(() => {
+  // set up mock registries
+  configureJsonRpcApi(mockMessageRegistry, mockKeyPairRegistry);
+  configureMessages(mockMessageRegistry);
+  // replace sending strategy of network manager with a jest mock function
+
+  // @ts-ignore
+  networkManager['sendingStrategy'] = mockSendingStrategy;
```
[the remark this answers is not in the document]
Tyratox: Similar to the issue with private fields, typescript will complain when overriding a `readonly` field, hence the `ts-ignore`. Do you see a better solution?

fe1-web/src/core/network/__tests__/JsonRpcApi.test.ts
```diff
@@ -0,0 +1,175 @@
...
+      throw new Error('The generator should contain at least one message');
+    }
+
+    const expected = ExtendedMessage.fromMessage(mockResponseMessage, mockChannel, mockAddress);
+
+    // the receivedAt value can differ
+    // @ts-ignore
+    delete value.receivedAt;
+    // @ts-ignore
+    delete expected['receivedAt'];
```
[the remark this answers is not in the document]
Tyratox: Again the `received` field is `readonly` so I simply to not know of another way to make this test work 😅
Casting to any would work but it is basically the same?

fe1-web/src/core/network/ingestion/__tests__/IngestionHandler.test.ts
```diff
@@ -0,0 +1,73 @@
...
+
+// these can only be instantiated after mocking the message registry
+let mockMessage: Message;
+let mockJsonRequest: Partial<JsonRpcRequest>;
+let extendedRequest: ExtendedJsonRpcRequest;
+
+const mockMessageRegistry = {
+  getSignatureType: jest.fn().mockImplementation(() => 'some signature'),
+  buildMessageData: jest.fn().mockImplementation((input) => JSON.stringify(input)),
+} as unknown as MessageRegistry;
```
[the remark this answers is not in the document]
Tyratox: 👍

fe1-web/src/core/network/ingestion/__tests__/IngestionHandler.test.ts
```diff
@@ -0,0 +1,73 @@
...
+
+describe('handleExtendedRpcRequests', () => {
+  it('dispatches the correct redux action', () => {
+    handleExtendedRpcRequests(extendedRequest);
+    const obj = addMessages(
+      ExtendedMessage.fromMessage(mockMessage, mockChannel, mockAddress).toState(),
+    );
+
+    // the receivedAt value can differ. calls[i][j] => jth argument of ith call
+    // @ts-ignore
```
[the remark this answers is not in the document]
Tyratox: I try, would you prefer a cast to any or do you know of a better solution to compare to objects where one property is allowed to be different? (That is not testing all other properties individually 😂)

I guess I could implement a `toBeJsonEqualExcept` that disregards certain properties?

fe1-web/src/core/network/jsonrpc/__tests__/ExtendedJsonRpcRequest.test.ts
```diff
@@ -0,0 +1,29 @@
...
+import { JsonRpcMethod } from '../JsonRpcMethods';
+import { JsonRpcRequest } from '../JsonRpcRequest';
+
+const mockAddress = 'some address';
+const mockChannel = 'some channel';
+const mockJsonRequest: Partial<JsonRpcRequest> = {
+  jsonrpc: 'some data',
+  method: JsonRpcMethod.BROADCAST,
+  params: { channel: mockChannel },
+};
```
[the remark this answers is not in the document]
Tyratox: 👍

fe1-web/src/core/network/jsonrpc/__tests__/ExtendedJsonRpcRequest.test.ts
Tyratox: 👍

fe1-web/src/core/network/__tests__/JsonRpcApi.test.ts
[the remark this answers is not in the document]
pierluca: to fix this issue, what you really want to do is the following:
1. replace the `getNetworkManager` implementation in this test with a mock
2. have the mock`getNetworkManager` return a `Networkmanager` with your desired strategy

By doing so, there's no need to change the private field, which is an implementation detail and really shouldn't be manipulated in the tests anyways :)

fe1-web/src/core/network/ingestion/__tests__/IngestionHandler.test.ts
```diff
@@ -0,0 +1,67 @@
...
+    delete obj.payload[0].receivedAt;
+    delete (dispatch as jest.Mock).mock.calls[0][0].payload[0].receivedAt;
+
+    expect(dispatch).toHaveBeenCalledWith(obj);
```
pierluca, line 65: While I could forgive the `@ts-ignore`, the intention behind your code is not very understandable, and that's a much bigger problem.

It looks like this is what you want to do:
https://stackoverflow.com/a/69291149/9533693
... and I think you'll agree this is much easier to understand as well 😄

fe1-web/src/core/keypair/index.ts
```diff
@@ -11,7 +11,14 @@ export * from './KeyPairStore';
...
+export function initializeKeyPair() {
   // initialize the keypair
-  KeyPairStore.get();
+  const keyPair = KeyPairStore.get();
+  console.log(`Using the public key: ${keyPair.publicKey.toString()}`);
 }
+
+// only initialize keypair after the store has been rehydrated, otherwise we get a new key
+// on every page load..
+addRehydrationCallback(initializeKeyPair);
```
pierluca, line 24: nice catch !

pierluca (changes requested):
I haven't reviewed the full thing, but I see a lot of good things.

I've just tried pointing you towards a few better solution with respect to the issues that @Xelowak had identified, as those are patterns you probably want to become aware of, unless you're already familiar with them and I missed something 😄

Xelowak (commented):
I think @pierluca brought a solution to the `@ts-ignore` problem regarding readonly fields.

fe1-web/src/core/network/ingestion/__tests__/IngestionHandler.test.ts
Tyratox: That seems to be working, thank you very much!! I tried something similar before and it didn't work but apparently I just messed up somewhere :)

fe1-web/src/core/network/__tests__/JsonRpcApi.test.ts
```diff
@@ -0,0 +1,175 @@
...
+        channel: mockChannel,
+      }),
+      id: AUTO_ASSIGN_ID,
+    });
+
+    expect(mockSendingStrategy).toHaveBeenCalledWith(request, expect.anything());
+    expect(mockSendingStrategy).toHaveBeenCalledTimes(1);
+  });
+
+  it('returns the a valid message generator', async () => {
```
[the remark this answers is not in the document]
Tyratox: 👍

fe1-web/src/core/network/__tests__/JsonRpcApi.test.ts
```diff
@@ -36,34 +35,36 @@ import {
...
   .fn()
   .mockImplementation(() =>
     Promise.resolve([new ExtendedJsonRpcResponse({ receivedFrom: mockAddress }, mockResponse)]),
   );
 
+jest.mock('core/network/NetworkManager', () => {
+  const actual = jest.requireActual('/core/network/NetworkManager');
+  return {
+    ...actual,
+    getNetworkManager: () => new actual.TEST_ONLY_EXPORTS.NetworkManager(mockSendingStrategy),
```
pierluca, line 58: okay, that's one (effective) way to go about it that I didn't expect, I thought you'd just be mocking the one function 😅 
What's `TEST_ONLY_EXPORTS` ? I've never seen this 😅

fe1-web/src/core/network/__tests__/JsonRpcApi.test.ts
```diff
@@ -151,13 +152,13 @@ describe('catchup', () => {
...
-    delete expected.receivedAt;
-
-    expect(value).toBeJsonEqual(expected);
+    // the receivedAt value is allowed differ
+    // cannot use expect.any(Number) here because of toBeJsonEqual and cannot use toBeEqual
+    // because once it is an instance of Base64Url and once of Hash
+    expect({ ...value, receivedAt: 0 }).toBeJsonEqual({
+      ...expected,
+      receivedAt: 0,
+    });
```
pierluca, line 161:
```suggestion
    // the receivedAt value is allowed differ
    expect({ ...value, receivedAt: 0 })
      .toBeJsonEqual({...expected, receivedAt: 0 });
```

I don't think the comments about expect.any / toBeJsonEqual / etc. will make much sense to someone who's not actively fighting with jest on these few lines of code 😅 but the comment itself is very useful to a reviewer.
In general, I try to put these comments as a self-review in the PR itself, so that the reviewers can benefit from the context, but it's also not confusing a subsequent developer reading through this code.

Tyratox: That makes sense!

pierluca (dismissed):

Xelowak (approved):

PR #900, opened 2022-03-27
Modified CreateRollCall to create channel in db with tests

rombirli:


be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
```diff
@@ -40,8 +43,42 @@ class RollCallHandler(dbRef: => AskableActorRef) extends MessageHandler {
...
+        // we are using the rollcall id instead of the message_id at rollcall creation
+        val rollCallChannel: Channel = Channel(s"${Channel.ROOT_CHANNEL_PREFIX}${data.id}")
+        val combined = for {
+          // check whether the rollcall already exists in db
+          _ <- {
+            (dbActor ? DbActor.ChannelExists(rollCallChannel)).transformWith {
+              case Success(_) => Future {
+                throw DbActorNAckException(ErrorCodes.INVALID_ACTION.id, "rollcall already exists in db")
+              }
+              case _ => Future {
```
Ajkunas, line 58: Quick question: in case it enters in this case, the function will continue by then creating the channel, right? 
Otherwise, all seems good to me :)

rombirli: Yes when the RollCall channel doesn't already exist, it enters this case : 'case _ => Future' and the for continues with 'dbActor ? DbActor.CreateChannel(rollCallChannel, ObjectType.ROLL_CALL)' and 'dbAskWritePropagate(rpcRequest)'. Otherwise the whole for return an error

Ajkunas: Thank you for the answer!

Ajkunas (approved):

rombirli:
I will reopen it with commits from my github account

PR #901, opened 2022-03-28
Create json messages from data

Tesa27:
As base64 is not really human readable, having a json message that contains a base64 encoded data field does not really "speak", it is hard to distinguish 2 messages only by looking at 2 base64 encodings. To cope with that having a cleartext json data file is a much cleaner approach, we can clearly see what the data represents. So instead of storing raw json messages, we store json data messages and construct utility functions that will use the human readable json data file to construct the json message that will be sent to the backend (that contains the base64 encoding as data field as before but now the conversion is done implicitly by the program).


tests/karate/src/test/java/be/utils/JsonConverter.java
```diff
@@ -0,0 +1,85 @@
...
+  public String convertJson(Json json){
+
+    String stringJson = json.toString();
+    byte[] jsonBytes = stringJson.getBytes(StandardCharsets.UTF_8);
+    Base64.Encoder encoder = Base64.getEncoder();
+    String base64 = encoder.encodeToString(jsonBytes);
+    return base64;
+  }
+
+
```
Warhedgehog, line 31: I think some documentation would help understand which part of the protocol this generates

tests/karate/src/test/java/be/utils/JsonConverter.java
```diff
@@ -19,22 +19,27 @@ public Json fromMapToJson(Map<String,String> map){
   private String signature = "ONylxgHA9cbsB_lwdfbn3iyzRd4aTpJhBMnvEKhmJF_niE_pUHdmjxDXjEwFyvo5WiH1NZXWyXG27SYEpkasCA==";
   private String messageIdForced = "";
 
+  /*
+    Takes a json representation the data field of a message and produces it's base64 counterpart
```
Warhedgehog, line 23: This comment does not feel grammatically correct

Warhedgehog (dismissed):
LGTM 👍

tests/karate/src/test/java/be/utils/server.feature
```diff
@@ -82,6 +82,15 @@ Feature: This feature starts a server and stops it after every scenario.
...
     * karate.log('Executing tests')
 
+    * def convertData =
+        """
+          function(){
+            var JsonConverter = Java.type('be.utils.JsonConverter')
+            return new JsonConverter()
+          }
+        """
+    * def converter = call convertData
```
pierluca, line 92: since all the tests are doing `convert` + `send` (with `"publish"` method) , wouldn't it make sense to define an API here that allows you to just `publish(message, id)` ?

tests/karate/src/test/java/be/utils/JsonConverter.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.Base64;
+
+public class JsonConverter {
+
+  public Json fromMapToJson(Map<String,String> map){
+    return Json.of(map);
+  }
+
+  private String senderPk = "J9fBzJV70Jk5c-i3277Uq4CmeL4t53WDfUghaK0HpeM=";
+  private String signature = "ONylxgHA9cbsB_lwdfbn3iyzRd4aTpJhBMnvEKhmJF_niE_pUHdmjxDXjEwFyvo5WiH1NZXWyXG27SYEpkasCA==";
```
pierluca, line 19: this can't be a constant & correct. We're not really building the messages correctly here...

tests/karate/src/test/java/data/lao/valid_lao_create.json
```diff
@@ -7,7 +7,7 @@
       "data": "eyJvYmplY3QiOiJsYW8iLCJhY3Rpb24iOiJjcmVhdGUiLCJuYW1lIjoiTEFPIiwiY3JlYXRpb24iOjE2MzMwMzU3MjEsIm9yZ2FuaXplciI6Iko5ZkJ6SlY3MEprNWMtaTMyNzdVcTRDbWVMNHQ1M1dEZlVnaGFLMEhwZU09Iiwid2l0bmVzc2VzIjpbXSwiaWQiOiJwX0VZYkh5TXY2c29wSTVRaEVYQmY0ME1PX2VOb3E3Vl9MeWdCZDRjOVJBPSJ9",
       "sender": "J9fBzJV70Jk5c-i3277Uq4CmeL4t53WDfUghaK0HpeM=",
       "signature": "ONylxgHA9cbsB_lwdfbn3iyzRd4aTpJhBMnvEKhmJF_niE_pUHdmjxDXjEwFyvo5WiH1NZXWyXG27SYEpkasCA==",
-      "message_id": "2mAAevx61TZJi4groVGqqkeLEQq0e-qM6PGmTWuShyY=",
+      "message_id": "e15ae2a533d22237d5288c659dfb3f88c52acd6a573a98d67ad2247a13825504",
```
pierluca, line 10: are you changing the specification to match the tests? 🤔

pierluca (changes requested):
I haven't reviewed this completely, but this seems very broken.
Is the work finished or should this be still a draft ?

tests/karate/src/test/java/be/utils/JsonConverter.java
```diff
@@ -0,0 +1,92 @@
...
+    Base64.Encoder encoder = Base64.getEncoder();
+    String base64 = encoder.encodeToString(jsonBytes);
+    return base64;
+  }
+
+  /*
+    Produces a valid Json representation of a message given the message data, the id of the message
+    and the channel where the message is supposed to be sent
+   */
+  public Json messageFromData(String stringData, String method, int id, String channel){
```
GabrielFleischer, line 37: I am not very convinced by this function name.

From the implementation, you can only create publish messages and broadcast messages.

And in the current implementation, you only produces publish messages.

Why not droping the method argument and going with `createPublish`

The call would look like
```
jsonConverter.createPubish(data, id, channel)
```
instead of
```
jsonConverter.messageFromData(data, "publish", id, channel)
```

I personally find it easier to understand. What do you think ?

tests/karate/src/test/java/be/utils/JsonConverterTest.java
```diff
@@ -0,0 +1,77 @@
...
+    validStringMessage.put("jsonrpc","2.0");
+    Json toCompare = Json.of(validStringMessage);
+    assert toCompare.toString().equals(jsonValidLaoMessage.toString());
+
+  }
+
+
+
+
+
```
GabrielFleischer, line 76: There are empty lines here

GabrielFleischer (changes requested):
It seems that between my review start and end, Pierluca raised everything I found and more.

Hope it is not too redundant.

This is definitely a step forward. The PR just needs a bit polishing 😉

tests/karate/src/test/java/be/utils/JsonConverter.java
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  private String senderPk = "J9fBzJV70Jk5c-i3277Uq4CmeL4t53WDfUghaK0HpeM=";
+  private String senderSk = "0leCDBokllJXKXT72psnqF5UYFVRxnc1BNDShY05KHQn18HMlXvQmTlz6LfbvtSrgKZ4vi3ndYN9SCForQel4w==";
+
+  private String signature = "ONylxgHA9cbsB_lwdfbn3iyzRd4aTpJhBMnvEKhmJF_niE_pUHdmjxDXjEwFyvo5WiH1NZXWyXG27SYEpkasCA==";
+  private String messageIdForced = "";
+  /*
+    Produces the base64 variant of the json file passed as argument
+   */
```
GabrielFleischer, line 24: This does not match the usual javadoc style
```suggestion
  
  /**
    * Produces the base64 variant of the json file passed as argument
    */
```

tests/karate/src/test/java/be/utils/JsonConverter.java
```diff
@@ -0,0 +1,143 @@
...
+  public String convertJson(Json json){
```
GabrielFleischer, line 25: I am not convinced by the function name.

The Javadoc explains it (even if it is a Json object more than a file, but this is nitpicking) but the row function is not clear.

```suggestion
  public String convertJsonToBase64(Json json){
```

tests/karate/src/test/java/be/utils/JsonConverter.java
```diff
@@ -0,0 +1,143 @@
...
+    return paramsPart;
+  }
+
+  /*
+      Constructs a valid signature on given data
+   */
+  public String constructSignature(String messageDataBase64){
+    try {
+      // Hex representation of the private key
+      String privateKeyHex = "1498b5467a63dffa2dc9d9e069caf075d16fc33fdd4c3b01bfadae6433767d93";
```
GabrielFleischer, line 84: This is bad practice.

This should be set as constant.

But what I don't understand is how `senderPk` can be mutable while the private key is fixed.
Isn't that an issue ?

tests/karate/src/test/java/be/utils/JsonConverter.java
```diff
@@ -0,0 +1,143 @@
...
+      byte[] privateKeyBytes = new byte[privateKeyHex.length() / 2];
+
+      for (int i = 0; i < privateKeyBytes.length; i++) {
+        int index = i * 2;
+
+        int val = Integer.parseInt(privateKeyHex.substring(index, index + 2), 16);
+        privateKeyBytes[i] = (byte)val;
+      }
+      PublicKeySign publicKeySign = new Ed25519Sign(privateKeyBytes);
+      byte[] singie = publicKeySign.sign(messageDataBase64.getBytes(StandardCharsets.UTF_8));
```
GabrielFleischer, line 94: You should not sign the UTF-8 encoding of the string. But the base64 encoding.

Remember, in this project, we chose to represent bytes as base64 strings.
But it could also have been hex strings, and it could have worked the same way.

tests/karate/src/test/java/be/utils/JsonConverterTest.java
```diff
@@ -0,0 +1,78 @@
...
+    Json jsonValidLaoMessage= jsonConverter.publishМessageFromData(laoDataJsonString,1,"/root");
+    Map<String,Object> validStringMessage = new LinkedHashMap<>();
+    validStringMessage.put("method","publish");
+    validStringMessage.put("id",1);
+    Map<String,Object> paramsMap = new LinkedHashMap<>();
+    paramsMap.put("channel","/root");
+    Map<String,Object> messageMap  = new LinkedHashMap<>();
+    messageMap.put("data","eyJvYmplY3QiOiJsYW8iLCJhY3Rpb24iOiJjcmVhdGUiLCJuYW1lIjoiTEFPIiwiY3JlYXRpb24iOjE2MzMwMzU3MjEsIm9yZ2FuaXplciI6Iko5ZkJ6SlY3MEprNWMtaTMyNzdVcTRDbWVMNHQ1M1dEZlVnaGFLMEhwZU09Iiwid2l0bmVzc2VzIjpbXSwiaWQiOiJwX0VZYkh5TXY2c29wSTVRaEVYQmY0ME1PX2VOb3E3Vl9MeWdCZDRjOVJBPSJ9");
+    messageMap.put("sender","J9fBzJV70Jk5c-i3277Uq4CmeL4t53WDfUghaK0HpeM=");
+    messageMap.put("signature","PLGEtiVT2KFFIxJGJLdrNF6dK5wkrRkGeOS0ALdYdbWm1jJE4srWDNMg3vB0EaDWK-07WqSDWNBYVajF66MGCw==");
```
GabrielFleischer, line 61: How did you construct the signature ?

GabrielFleischer (changes requested):
I see a lot of questionable code style choices.

It could be an idea to, like in the other subprojects, choose a code style for the project and stick to it.

I also found some suspect things in the signature generation.

tests/karate/src/test/java/be/utils/JsonConverter.java
```diff
@@ -0,0 +1,163 @@
...
+  public Map<String,Object> constructMessageField(String stringData, boolean forcedMessageId){
+    Map<String,Object> messagePart  = new LinkedHashMap<>();
+    Json messageData = Json.of(stringData);
+    String messageDataBase64 = convertJsonToBase64(messageData);
+    messagePart.put("data", messageDataBase64);
+    messagePart.put("sender", senderPk);
+
+    String signature = constructSignature(stringData);
+    messagePart.put("signature", signature);
+    String messageId = hash(messageDataBase64.getBytes(),signature.getBytes());
```
GabrielFleischer, line 64: So, `messageDataBase64` is the base64 encoding of stringData.getBytes().

What you are doing here is retrieving the UTF-8 encoding of `messageDataBase64` which is not what we want to hash.

We want to hash the encoded bytes (stringData.getBytes())

tests/karate/src/test/java/be/utils/JsonConverter.java
```diff
@@ -0,0 +1,163 @@
...
+      byte[] privateKeyBytes = new byte[privateKeyHex.length() / 2];
+
+      for (int i = 0; i < privateKeyBytes.length; i++) {
+        int index = i * 2;
+
+        int val = Integer.parseInt(privateKeyHex.substring(index, index + 2), 16);
+        privateKeyBytes[i] = (byte)val;
+      }
+      PublicKeySign publicKeySign = new Ed25519Sign(privateKeyBytes);
+      byte[] singie = publicKeySign.sign(messageDataBase64.getBytes());
```
GabrielFleischer, line 104: This was correct actually. Because messageDataBase64 is not not actually in base64 in your implementation.

It is simply the stringData which makes this line correct.

(Although defining the charset as UTF-8 was good so you should put it back 😅)

But you definitely need to rename this field as the current name is definitely missleading.

GabrielFleischer (changes requested):
I made a quick review regarding the point I raised yesterday.

tests/karate/src/test/java/be/utils/JsonConverter.java
Tesa27: I am not sure that this is what we want or at least there is a missmatch.  I compared the messageId that that is present /protocol/examples/answer/general_message.json  with the message id  constructed with encoding of `messageDataBase64` and they match

tests/karate/src/test/java/be/utils/JsonConverter.java
Tesa27: I agree on this point, and applied the changes

GabrielFleischer (changes requested):
There are still quite a lot of formatting inconcistencies.

If you're using intellij, you can right-click a file `Refomat Code`.

It will at least be consistent in the whole file.

tests/karate/src/test/java/be/utils/JsonConverter.java
```diff
@@ -0,0 +1,165 @@
...
+public class JsonConverter {
+
+  public Json fromMapToJson(Map<String,String> map){
+    return Json.of(map);
+  }
+
+  private String senderPk = "J9fBzJV70Jk5c-i3277Uq4CmeL4t53WDfUghaK0HpeM=";
+  private String senderSkObject = "0leCDBokllJXKXT72psnqF5UYFVRxnc1BNDShY05KHQn18HMlXvQmTlz6LfbvtSrgKZ4vi3ndYN9SCForQel4w==";
+  private String senderSk = "0leCDBokllJXKXT72psnqF5UYFVRxnc1BNDShY05KHQ=";
+  private String privateKeyHex = "d257820c1a249652572974fbda9b27a85e54605551c6773504d0d2858d392874";
```
GabrielFleischer, line 21: These variables are never used

tests/karate/src/test/java/be/utils/JsonConverter.java
```diff
@@ -0,0 +1,165 @@
...
+
+  /**
+   * Produces the base64 variant of the json file passed as argument
+   */
+  public String convertJsonToBase64(Json json){
+    String stringJson = json.toString();
+    byte[] jsonBytes = stringJson.getBytes();
+    Base64.Encoder encoder = Base64.getEncoder();
+    String base64 = encoder.encodeToString(jsonBytes);
+    return base64;
```
GabrielFleischer, line 34: You can collapse this in one line

tests/karate/src/test/java/be/utils/JsonConverter.java
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  /**
+   * Produces a valid Json representation of a message given the message data, the id of the message
+   * and the channel where the message is supposed to be sent
+   */
+  public Json publishМessageFromData(String stringData, int id, String channel){
+    Json messageData = Json.of(stringData);
+    String messageDataBase64 = convertJsonToBase64(messageData);
```
GabrielFleischer, line 43: This variable is never used

tests/karate/src/test/java/be/utils/JsonConverter.java
```diff
@@ -0,0 +1,165 @@
...
+        int index = i * 2;
+
+        int val = Integer.parseInt(privateKeyHex.substring(index, index + 2), 16);
+        privateKeyBytes[i] = (byte)val;
+      }
+      PublicKeySign publicKeySign = new Ed25519Sign(privateKeyBytes);
+      byte[] singie = publicKeySign.sign(messageData.getBytes(StandardCharsets.UTF_8));
+      String signature = Base64.getUrlEncoder().encodeToString(singie);
+      return signature;
+
```
GabrielFleischer, line 109: This can be collapsed in one line

GabrielFleischer (changes requested):
There are still some style inconsistencies, especially in the generic types definitions :
`Map<A,B>` instead of `Map<A, B>`.

If you're using intellij, you can right-click a file and `Refomat Code`.

The best would be to specify the project style instead of using the IDE's.
But at least it will be consistent with the whole file.

GabrielFleischer (approved):
LGTM

tests/karate/src/test/java/be/utils/JsonConverterTest.java
```diff
@@ -0,0 +1,95 @@
...
+    laoData.put("action","create");
+    laoData.put("name","LAO");
+    laoData.put("creation",1633035721);
+    laoData.put("organizer","J9fBzJV70Jk5c-i3277Uq4CmeL4t53WDfUghaK0HpeM=");
+    String witness[] = new String[0];
+    laoData.put("witnesses",witness);
+    laoData.put("id","p_EYbHyMv6sopI5QhEXBf40MO_eNoq7V_LygBd4c9RA=");
+    Json laoDataJson = Json.of(laoData);
+    return laoDataJson;
+  }
```
GabrielFleischer, line 23: This file should also be reformatted.

tests/karate/src/test/java/be/utils/JsonConverter.java
```diff
@@ -0,0 +1,160 @@
...
+        String dataLength = Integer.toString(data.length);
+        digest.update(dataLength.getBytes());
+        digest.update(data);
+      }
+      return Base64.getUrlEncoder().encodeToString(digest.digest());
+    } catch (Exception e) {
+      e.printStackTrace();
+      System.out.println("Exception happened in hash construction");
+    }
+    return "Hash is not constructed correctly";
```
GabrielFleischer, line 157: I would prefer to throw an error rather than returning an invalid value.

You can even throw the error and remove the try block.

tests/karate/src/test/java/be/utils/JsonConverter.java
```diff
@@ -0,0 +1,160 @@
...
+      PublicKeySign publicKeySign = new Ed25519Sign(privateKeyBytes);
+      byte[] singie = publicKeySign.sign(messageData.getBytes(StandardCharsets.UTF_8));
+      return Base64.getUrlEncoder().encodeToString(singie);
+
+    } catch (Exception e) {
+      System.out.println("CANNOT SIGN");
+      e.printStackTrace();
+    }
+    // This should never happen
+    return signatureForced;
```
GabrielFleischer, line 110: I would prefer to throw an error rather than returning an invalid value.

You can throw the error and remove the try block, or create a new error that is thrown here.

tests/karate/src/test/java/be/utils/JsonConverter.java
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  public Map<String, Object> constructParamsField(String channel, String messageDataBase64) {
+    Map<String, Object> paramsPart = new LinkedHashMap<>();
+    paramsPart.put("channel", channel);
+
+    Map<String, Object> messagePart = constructMessageField(messageDataBase64, false);
+    paramsPart.put("message", messagePart);
+
+    return paramsPart;
```
GabrielFleischer, line 84: Instead of using maps that are transformed in Json, you can create Json Objects with `Json.object()` and use `Json.set(path, value)`

tests/karate/src/test/java/be/utils/JsonConverter.java
```diff
@@ -0,0 +1,160 @@
...
+    messagePart.put("sender", senderPk);
+
+    String signature = constructSignature(stringData);
+    messagePart.put("signature", signature);
+    String messageId = hash(messageDataBase64.getBytes(), signature.getBytes());
+
+    if (forcedMessageId && !messageIdForced.isEmpty()) {
+      messageId = this.messageIdForced;
+      messageIdForced = "";
+    }
```
GabrielFleischer, line 67: I am not a fan of this method, as we talked about in the last meeting, I think this should be removed completely.

tests/karate/src/test/java/be/utils/JsonConverter.java
```diff
@@ -0,0 +1,160 @@
...
+  /**
+   * Produces a valid Json representation of a message given the message data, the id of the message
+   * and the channel where the message is supposed to be sent
+   */
+  public Json publishМessageFromData(String stringData, int id, String channel) {
+    Json messageData = Json.of(stringData);
+    Map<String, Object> messageJson = new LinkedHashMap<>();
+
+    messageJson.put("method", "publish");
+    messageJson.put("id", id);
```
GabrielFleischer, line 43: Instead of using maps that are transformed in Json, you can create Json Objects with `Json.object()` and use `Json.set(path, value)`

tests/karate/src/test/java/be/utils/JsonConverter.java
```diff
@@ -0,0 +1,160 @@
...
+      byte[] privateKeyBytes = new byte[privateKeyHex.length() / 2];
+
+      for (int i = 0; i < privateKeyBytes.length; i++) {
+        int index = i * 2;
+
+        int val = Integer.parseInt(privateKeyHex.substring(index, index + 2), 16);
+        privateKeyBytes[i] = (byte) val;
+      }
+      PublicKeySign publicKeySign = new Ed25519Sign(privateKeyBytes);
+      byte[] singie = publicKeySign.sign(messageData.getBytes(StandardCharsets.UTF_8));
```
GabrielFleischer, line 102: I think there is a typo in `singie`

tests/karate/src/test/java/be/utils/JsonConverter.java
```diff
@@ -0,0 +1,160 @@
...
+    try {
+      // Hex representation of the private key
+      byte[] privateKeyBytes = new byte[privateKeyHex.length() / 2];
+
+      for (int i = 0; i < privateKeyBytes.length; i++) {
+        int index = i * 2;
+
+        int val = Integer.parseInt(privateKeyHex.substring(index, index + 2), 16);
+        privateKeyBytes[i] = (byte) val;
+      }
```
GabrielFleischer, line 100: Could this be defined in another function ? I think it would easier to read.

GabrielFleischer (changes requested):
I reread the java code and there were some things I should have raised last time.

I am sorry for the big delay 😅

GabrielFleischer (approved):
I have a small nitpicking, you can address it if you want.

Otherwise, it is a great PR overall.

tests/karate/src/test/java/be/utils/JsonConverter.java
```diff
@@ -14,16 +16,15 @@ public class JsonConverter {
 
   private String senderPk = "J9fBzJV70Jk5c-i3277Uq4CmeL4t53WDfUghaK0HpeM=";
   private String privateKeyHex = "d257820c1a249652572974fbda9b27a85e54605551c6773504d0d2858d392874";
-  private String signatureForced = "ONylxgHA9cbsB_lwdfbn3iyzRd4aTpJhBMnvEKhmJF_niE_pUHdmjxDXjEwFyvo5WiH1NZXWyXG27SYEpkasCA==";
+  private String signatureForced =
```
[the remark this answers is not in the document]
Tesa27: The idea was to have some default values for these variables and change them with setters if necessary

PR #902, opened 2022-03-28
Be1 public key check on Lao Creation

Warhedgehog:
Right now anyone can create an LAO on an organizer's server. This PR prevents by responding with an error when someone other than the organizer, defined by the publick key given in the command line on start, tries to start an LAO.


MaximeZmt (dismissed):
👍

Robingoumaz (dismissed):
Very good

PR #913, opened 2022-03-31
Fix app crash on receiving a GenericError 

### LINKED ISSUE #912: App crashes when some errors are received
For some errors on Android the app crashes instead of treating it. It is probably when the app tries to handle GenericErrors.

### Expected behavior
The app is expected to handle normal errors and be al

### How to reproduce
1. Launch the android app
2. Initialize the wallet
3. Launch the Go server with a publick key different than the one from the app but of valid format nevertheless
4. Launch an LAO from the app

### Version & Environment
This bug was reproduced on:
- Commit: master [688f814](https://github.com/dedis/popstellar/commit/688f814c153dcbe4e6fcf6993140b28a4d4f4889)

##### Front-ends:
- [ ] Not applicable

##### Back-ends:
- [ ] Be1-Go
- [ ] Be2-Scala
- [x] Not applicable

 - Device: [e.g. Pixel 4 emulator on Windows 10]
 - OS: [Windows 10]

### Workaround
Use functionnalities in a way that do not cause GenericError on the Android app

### Impact
It is quite nasty, the app would probably crash on a wide variety of occasion.

### Possible root cause
There is a formatting error in a utility functions for error logging and displaying

Warhedgehog:
This fixes #912

### DIFF
--- a/tests/karate/src/test/java/fe/utils/android.feature
+++ b/tests/karate/src/test/java/fe/utils/android.feature
@@ -8,12 +8,6 @@ Feature: android page object
     * capabilities.app = karate.toAbsolutePath('file:' + capabilities.app)
     * def driverOptions = { webDriverSession: { desiredCapabilities : "#(capabilities)" } }
 
-    # Create and import mock backend
-    * call read('classpath:fe/net/mockbackend.feature')
-    * def backendURL = 'ws://10.0.2.2:' + backend.getPort()
-    # Import message filters
-    * call read('classpath:common/net/filters.feature')
-
     # ================= Page Object Start ====================
 
     # Tab buttons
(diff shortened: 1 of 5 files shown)


tests/karate/src/test/java/fe/utils/android.feature
```diff
@@ -20,9 +20,12 @@ Feature: android page object
     * def tab_launch_lao_name_selector = '#com.github.dedis.popstellar:id/entry_box_launch'
     * def tab_launch_create_lao_selector = '#com.github.dedis.popstellar:id/button_launch'
 
+    # Lao Detail
+    * def
+
```
Tesa27, line 25: I think these 3 lines should be removed

Tesa27 (changes requested):
Looks good overall, just a small remark.

Warhedgehog:
I failed to undo the changes in the karate tests so I made a new branch and PR rather than lose more time.

PR #915, opened 2022-04-02
Checking the validity of the message id

Tesa27:
When a message is received by the backend the message id was not checked, which created a potential integrity violation. A user could send a message containing the wrong message id or a message that has the message id of a valid message and backend will accept them both if the remaining fields are correctly set.


be1-go/hub/standard_hub/message_handling.go
```diff
@@ -269,6 +269,17 @@ func (h *Hub) handlePublish(socket socket.Socket, byteMessage []byte) (int, erro
...
 
+	signature := publish.Params.Message.Signature
+	messageId := publish.Params.Message.MessageID
+	data := publish.Params.Message.Data
+
+	expectedMessageId := messagedata.Hash(data, signature)
+
+	if expectedMessageId != messageId {
+
+		return publish.ID, xerrors.Errorf("message_id is wrong expected is %q but was %v", expectedMessageId, messageId)
```
Warhedgehog, line 280: I may be missing something but would it not be massively vulnerable to inform the sender of the hash that we expected ? Could he not resend the same message with the appropriate message_id thus impersonating someone else ?

Warhedgehog (changes requested):
I may misunderstand the message_id but there seems to be a big problem with the current proposed changes.

Nevertheless the check is crucial, so nice change.  I think you should also do it for handleBroadcast below

be1-go/hub/standard_hub/message_handling.go
Tesa27: I do agree, the attacker could just send a random message id to get information about the true message id and would just resend the same message with the correct message id.

Warhedgehog (commented):
It would be good to do a unit test for the handleBroadcast too

Warhedgehog (dismissed):
LGTM

be1-go/hub/standard_hub/mod_test.go
```diff
@@ -121,6 +125,91 @@ func Test_Create_LAO_Bad_Key(t *testing.T) {
...
+
+	hub.handleMessageFromClient(&socket.IncomingMessage{
+		Socket:  sock,
+		Message: publishBuf,
+	})
+
+	require.EqualError(
+		t, sock.err,
+		fmt.Sprintf(
+			"failed to handle method: message_id %vis wrong", badMessageID))
```
Robingoumaz, line 210: there is a missing space

MaximeZmt (approved):
👍

Robingoumaz (changes requested):
Great work, just a little typo

be1-go/hub/standard_hub/message_handling.go
```diff
@@ -269,6 +269,16 @@ func (h *Hub) handlePublish(socket socket.Socket, byteMessage []byte) (int, erro
 		return -1, xerrors.Errorf("failed to unmarshal publish message: %v", err)
 	}
 
+	signature := publish.Params.Message.Signature
+	messageId := publish.Params.Message.MessageID
+	data := publish.Params.Message.Data
+
+	expectedMessageId := messagedata.Hash(data, signature)
```
lhmerino, line 276: From a security perspective, there should be a delimiter between data and signature. https://crypto.stackexchange.com/questions/55162/best-way-to-hash-two-values-into-one

be1-go/hub/standard_hub/message_handling.go
```diff
@@ -269,6 +269,16 @@ func (h *Hub) handlePublish(socket socket.Socket, byteMessage []byte) (int, erro
...
+
+	if expectedMessageId != messageId {
+		return publish.ID, xerrors.Errorf("message_id  is wrong")
```
lhmerino, line 279: Extra space.

be1-go/hub/standard_hub/message_handling.go
lhmerino: Although this change would likely require the cooperation of all subsystems but generally its essential to add a delimiter when concatenating strings that is to be hashed.

lhmerino (changes requested):
LGTM; left some comments.

be1-go/hub/standard_hub/message_handling.go
Tesa27: I see I see, for the moment the delimiter is not included in any subsystem if I am not mistaken, so this should be a global change in the system and I guess in the specifications as well

pierluca: Actually, we don't need a delimiter because we prefix each data element with its length.

Robingoumaz (dismissed):

be1-go/hub/standard_hub/message_handling.go
```diff
@@ -123,6 +123,7 @@ func (h *Hub) handleRootChannelBroadcastMesssage(sock socket.Socket, broadcast m
 	if err != nil {
 		h.log.Err(err).Msg("invalid lao#create message")
 		sock.SendError(&id, err)
+		return err
```
nkcr, line 126: nice catch! 🙀

be1-go/hub/standard_hub/message_handling.go
```diff
@@ -269,6 +270,15 @@ func (h *Hub) handlePublish(socket socket.Socket, byteMessage []byte) (int, erro
 		return -1, xerrors.Errorf("failed to unmarshal publish message: %v", err)
 	}
 
+	signature := publish.Params.Message.Signature
+	messageId := publish.Params.Message.MessageID
+	data := publish.Params.Message.Data
+
+	expectedMessageId := messagedata.Hash(data, signature)
+	if expectedMessageId != messageId {
+		return publish.ID, xerrors.Errorf("message_id is wrong")
```
nkcr, line 279:
```suggestion
		return publish.ID, xerrors.Errorf("message_id is wrong: expected %q found %q", expectedMessageId, messageId)
```

nkcr: same bellow

be1-go/hub/standard_hub/mod_test.go
```diff
@@ -677,6 +779,90 @@ func Test_Create_LAO_Broadcast(t *testing.T) {
...
+	keypair := generateKeyPair(t)
+
+	fakeChannelFac := &fakeChannelFac{
+		c: &fakeChannel{},
+	}
+
+	hub, err := NewHub(keypair.public, nolog, fakeChannelFac.newChannel, hub.OrganizerHubType)
+	require.NoError(t, err)
+
+	now := time.Now().Unix()
```
nkcr, line 793: tests should be reproducible. Can't we use a fixed value ?

be1-go/hub/standard_hub/mod_test.go
```diff
@@ -677,6 +779,90 @@ func Test_Create_LAO_Broadcast(t *testing.T) {
...
+	require.NoError(t, err)
+
+	sock := &fakeSocket{}
+
+	hub.handleMessageFromServer(&socket.IncomingMessage{
+		Socket:  sock,
+		Message: publishBuf,
+	})
+
+	require.EqualError(
```
nkcr, line 860: Mmmm we avoid that type of formatting. 
In the case of a test it is fine to have
```go
require.EqualError(t, sock.err, fmt.Sprintf("failed to handle method: message_id %v is wrong", fakeMessageID))
```

nkcr: If that wasn't a test, we would format it as
```go
	require.EqualError(t, sock.err, fmt.Sprintf("failed to handle method: "+
		"message_id %v is wrong", fakeMessageID))
```

nkcr (commented):
Good job, have a couple of minor comments 👍

nkcr (approved):

PR #917, opened 2022-04-02
Add greeting message specification

Tyratox:
Here we can further discuss whether everyone is happy with the specification of the greeting message.

As this message is required for all further steps in the encrypted voting, it should be a priority to implement this message.

## Goals

As discussed on Tuesday, the primary purpose of the message is to enable the frontends to verify messages that originate from backends such as the message that will return a new encryption key for an election.

A secondary goal (with the `peers` property) is to make client to multiple server communication more user-friendly by not requiring them to connect to all servers manually.

## How to progress

Every subsystem should first confirm that they are happy with the specification. After that (or even in parallel), the different subsystems can start pushing changes to this branch where they add support for the greeting message.

The priority should be on the public key. Support for the peers property can also be added with a later commit and can be empty if the time does not suffice for both.

If a subsystem has added their implementation to the `work-fe1-evoting-greeting-message` branch, edit this message and tick of the respective checkbox. As soon as one frontend and one backend implemented support it, it should be tested and merged.

Since the specification of the message is largely based on #846, @Robingoumaz you might be able to make use some of the code already written there.

- [x] be-1
- [x] be-2 Support in #1002  PR
- [x] fe-1. Support for the not finalized specification has been implemented in #922 and #957.
- [x] fe-2. Support for the not finalized specification has been implemented in #952.


ahominal (approved):
I think It's fine for me. So the server doesn't need any kind of acknowledgment after sending the greeting right?

Tyratox:
No or at least I do not see why it would need one 😅

protocol/examples/query/greeting/wrong_greeting_invalid_address.json
```diff
@@ -0,0 +1,17 @@
+{
+    "jsonrpc": "2.0",
+    "method": "greeting",
+    "params": {
+        "channel": "a channel",
```
Xelowak, line 5: Isn't the channel invalid there too ? I don't know what is intended, but since the file is called `invalid_address`, I prefer to mention it.

protocol/examples/query/publish/publish.json
```diff
@@ -1,6 +1,6 @@
 {
     "jsonrpc": "2.0",
-    "id": 4,
+    "id": 9,
```
Xelowak, line 3: Why is this change needed ?

Tyratox: Oh whoops, not sure how this happened!

protocol/examples/query/greeting/wrong_greeting_invalid_address.json
Tyratox: Oh right, thank you!

docs/protocol.md
```diff
@@ -186,12 +187,125 @@ and its arguments (`params`).
...
 * Subscribe
 * Unsubscribe
 * Catchup
 * Broadcast
 * Publish
 
+### Greeting new clients
+🧭 **RPC Message** > **Query** > **Greeting**
+
+This is a message that is sent to clients after they have connected to a server.
```
pierluca, line 200: This is incoherent with the JSON-Schema specification.
When I connect to a server, I don't know (yet) which lao_id I'm connecting to.
Yet the `params` object contains a `channel`. 
We can either send this message only after the client `subscribe`s to a valid LAO, _or_ exclude the `channel` from the `params`.

pierluca: Conversely, if it is indeed related to a `channel`, why put it at the JSON-RPC level and not encapsulated in a usual `Message` object ? (I'm not sure if it'd be a good idea, just asking a possibly-dumb question)

Tyratox: I guess it also depends on whether we want a greeting message per LAO or per organizer server. Do we want to support the case where a given server hosts multiple LAOs and the set of peers/witnesses is different for each lao?
I would argue yes, and thus I think it actually might make more sense to have this message being sent on the lao channel?

Regarding why it is  at the JSON-RPC level, I do not know haha. I just assumed that there was a good reason for doing so when PR #846 was created 😂

Tyratox: @Ajkunas @romain46 @ahominal @Robingoumaz
What do you think? Per LAO (including the `channel` parameter to prevent replays, at least I think that this is its purpose)? And should we pack it into a message instead of using a different `method`?

lhmerino: Doesn't the QR code that the organizer supplies contain the IP address of the server + LAO ID? If so, would make sense to send LAO ID along as a type of "Server Name Indication" like in SSL.

Sounds to me that maybe it should be encapsulated in the message under the `subscribe`  method _only_ when subscribing to the overall LAO ID.

Tyratox: Yes I think that makes sense! So we just send the greeting on the lao channel after subscribing and replace the channel field with LAO ID?

Tyratox: What I am not sure about is the subscribe method since this message originates from the server 🤔 A `broadcast` of a `publish` might be more fitting?

Tyratox: Something like this?

https://github.com/dedis/popstellar/blob/efef6116a71dab7854b76305003bdd5603230d4f/protocol/query/method/message/data/dataGreetLao.json#L1-L49

lhmerino: Looks great!

Tyratox: This is from a discussion during the frontend meeting in the week before the easter break:

For now there is no way we can determine what backend is the organizer's since we could also connect through witnesses servers and the greeting message does not necessarily have to come from the organizer. Determining who the organizer is, is important for the secret ballot since (for now) only the organizer should be able to send election keys and the frontends need to verify the signature and check whether it was indeed the organizer who signed the message.

To do this we can bind the organizer's backend public key to the organizer's frontend public key (which is part of the lao#create message). Since the same might be required for the witnesses in the future, just treating the lao#create message as a one-sided claim and then requiring a witness signature might not be ideal.

We could for example add another field to the greeting message called `frontend` where the backend claims that it is the backend of some frontend public key. The corresponding frontend could then either just sign this message using the witnessing feature or send a separate message to everyone in order to finish the bidirectional claim. When going with the witnessing approach, a given lao#greet message would only be valid with the signature of the frontend. (And since the corresponding frontend is usually the first to connect to its backend, this signature will usually already be there)
With this approach, backends can be bound to the corresponding frontends and it is possible to determine the organizers backend. What are your thoughts on this?

Tyratox: Also I just realized that we do not need the `sender` field, this value can be retrieved from the layer above :)

pierluca (commented):
Good to approve, in principle, but fe2 CI fails.

ahominal:
> Good to approve, in principle, but fe2 CI fails.

Thanks for noticing, now it works. forgot to add a slash

Robingoumaz (approved):

jbsv (approved):
Looks Good To Me

PR #919, opened 2022-04-03
Fixed the writing of the public keys of attendees in LAO

Ajkunas:
The previous implementation of the handler of CloseRollCall did not write the public keys of the attendees in the LAO data. Hence, when we tried to retrieve the public keys of the attendees of the Roll Call, the correct public keys were not registered.

### DIFF
--- a/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
@@ -87,7 +87,7 @@ class RollCallHandler(dbRef: => AskableActorRef) extends MessageHandler {
               case Some(_) =>
                 val combined = for {
                   _ <- dbActor ? DbActor.ReadLaoData(rpcRequest.getParamsChannel)
-                  _ <- dbActor ? DbActor.Write(rpcRequest.getParamsChannel, message)
+                  _ <- dbActor ? DbActor.WriteLaoData(rpcRequest.getParamsChannel, message)
                 } yield ()
 
                 Await.ready(combined, duration).value match {


lourkeur:
So this adds a `writeLaoData` operation and removes the plain `write` operation. Is the latter superfluous? How? It's not clear to me.

Ajkunas:
> So this adds a `writeLaoData` operation and removes the plain `write` operation. Is the latter superfluous? How? It's not clear to me.

As I understood it, the writeLaoData function updates the laoData of the LAO. The Write function only writes the message in the database, not updating the LAO. 
I saw the bug while trying to implement the validators of the Election. Since we needed to access the PublicKeys of the attendees, we needed to have access to the attendees registered in the LAO (this is done after calling the CloseRollCallHandler). But, I saw that the PublicKeys were not updated in the LAO. Hence, the test of the attendees in the validators was not correct.

lourkeur:
> > So this adds a `writeLaoData` operation and removes the plain `write` operation. Is the latter superfluous? How? It's not clear to me.
> 
> As I understood it, the writeLaoData function updates the laoData of the LAO. The Write function only writes the message in the database, not updating the LAO.

That seems to be true, but does `writeLaoData` also write the message in the database? And if it doesn't, does it matter if it does?

Ajkunas:
> > > So this adds a `writeLaoData` operation and removes the plain `write` operation. Is the latter superfluous? How? It's not clear to me.
> > 
> > 
> > As I understood it, the writeLaoData function updates the laoData of the LAO. The Write function only writes the message in the database, not updating the LAO.
> 
> That seems to be true, but does `writeLaoData` also write the message in the database? And if it doesn't, does it matter if it does?

Yes it also writes it in the database

lourkeur (approved):
LGTM my concerns have been addressed

lhmerino (approved):

PR #922, opened 2022-04-04
Add support for the greeting message to fe-1

Tyratox:
This PR adds support for the newly specified greeting message. See #917.

The main purpose for now is to store the public key that is associated with a backend. A secondary goal was to bootstrap client-to-multiple server communication using the `peers` property.


fe1-web/src/core/network/ingestion/Handler.ts
```diff
@@ -61,6 +65,25 @@ export function handleExtendedRpcRequests(req: ExtendedJsonRpcRequest) {
...
+      addServer(
+        new Server({ address: greetingParams.address, publicKey: greetingParams.sender }).toState(),
+      ),
+    );
+
+    // connect to all received peer addresses
+    // IMPORTANT: The network manager deduplicates connections to the same address (string)
+    // and the received peer addresses are supposed to be the canonical ones.
+    // Hence it just has to be made sure that the first connection also is to the canonical
+    // address, otherwiese a client will connect to the same server twice (e.g. using its IP and then
```
pierluca, line 81:
```suggestion
    // address, otherwise a client will connect to the same server twice (e.g. using its IP and then
```

fe1-web/src/core/network/ingestion/Handler.ts
```diff
@@ -61,6 +65,25 @@ export function handleExtendedRpcRequests(req: ExtendedJsonRpcRequest) {
...
+    dispatch(
+      addServer(
+        new Server({ address: greetingParams.address, publicKey: greetingParams.sender }).toState(),
+      ),
+    );
+
+    // connect to all received peer addresses
+    // IMPORTANT: The network manager deduplicates connections to the same address (string)
+    // and the received peer addresses are supposed to be the canonical ones.
+    // Hence it just has to be made sure that the first connection also is to the canonical
```
pierluca, line 80:
```suggestion
    // Hence we just have to make sure that the first connection is also to the canonical
```
(tiny bit simpler / clearer)

fe1-web/src/core/objects/Server.ts
```diff
@@ -0,0 +1,63 @@
...
+  // NOTE: There is no need to store peers: ServerAddress[] here.
+  // As soon as a greeting message arrives, we connect to all peers. The server addresses
+  // will be added to the lao state as soon as a lao creation message is received
+  // over each connection
+
+  /**
+   * Constructs a new server instance
+   * @param server The properties of the new server instance
+   */
+  constructor(server: RemoveMethods<Server>) {
```
pierluca, line 32: Outside the scope of this PR:
I can't help but feel that `RemoveMethods` is the wrong name when reading this.
It is a type, but the name implies a verb (describing how it's implemented).
If we focus on its usage, how about something like `AttributesOnly<>` or `DataFields<>` ?

fe1-web/src/core/redux/ServerReducer.ts
```diff
@@ -0,0 +1,105 @@
...
+
+export default {
+  [SERVER_REDUCER_PATH]: serverSlice.reducer,
+};
+
+export const getServerState = (state: any): ServerReducerState => state[SERVER_REDUCER_PATH];
+
+/**
+ * A function to directly retrieve the public key from the redux store.
+ * NOTE: This function does not memoize the result. If you need this, use makeServerSelector instead
```
pierluca, line 75: should this be a `@remarks` field ?

fe1-web/src/core/redux/ServerReducer.ts
```diff
@@ -0,0 +1,105 @@
...
+/**
+ * Creates a server selector for a given address. Can for example be used in useSelector()
+ * @param address The server address
+ * @returns A selector for the server object for the given address or undefined if there is none
+ */
+export const makeServerSelector = (address: ServerAddress) =>
+  createSelector(
+    // First input: map of addresses to servers
+    (state) => getServerState(state).byAddress,
+    // Selector: returns an array of LaoStates -- should it return an array of Lao objects?
```
pierluca, line 102: comment seems out of date

fe1-web/src/core/objects/Server.ts
```diff
@@ -0,0 +1,63 @@
...
+export interface ServerState {
+  address: string;
+  publicKey: string;
+}
+
+export class Server {
+  /**
+   * The canonical address of the server
+   */
+  address: ServerAddress;
```
pierluca, line 16: should this be a `URL` object ?

fe1-web/src/core/objects/Server.ts
```diff
@@ -0,0 +1,63 @@
...
+import { PublicKey } from './PublicKey';
+
+export type ServerAddress = string;
+
+export interface ServerState {
+  address: string;
+  publicKey: string;
+}
+
+export class Server {
```
pierluca, line 12: 1. The name is a problem. This class doesn't implement a server. I'd go with something clearer, such as `ServerInfo`, `ServerReference`, etc. 
2. Should this class be in `core/objects` ? Where do we really need this ?

fe1-web/src/core/network/ingestion/Handler.ts
Tyratox: 👍

fe1-web/src/core/network/ingestion/Handler.ts
Tyratox: 👍

fe1-web/src/core/objects/Server.ts
Tyratox: Yeah I agree!

Tyratox: I like `AttributesOnly` or `OmitMethods` (analogous to `Omit`) would work too

fe1-web/src/core/redux/ServerReducer.ts
Tyratox: Seems like it, thank you! :) Apart from `@param` and `@returns?  I do not know the tsdoc tags very well 😅

fe1-web/src/core/redux/ServerReducer.ts
Tyratox: Uff, yes 😅 Thank you 👍

fe1-web/src/core/objects/Server.ts
Tyratox: I guess that could work..? 😅 I will have to check the URL type since I am not sure what it does and whether it is supported on mobile devices

fe1-web/src/core/objects/Server.ts
Tyratox: 1. I agree 👍
2. This really boils down to what we want to have in the core and what not. Basically the core could also just be a feature that does not accept a config but only exports its interface. My thought was that it might be used by a other features as well and since it (for now) is not a message but a json rpc method that stores instances I put it in the core. If we had it as a feature and as a json rpc method at the same time, we would need to extend the interface to intercept other types of json rpc calls?

fe1-web/src/core/network/jsonrpc/__tests__/Greeting.test.ts
```diff
@@ -0,0 +1,104 @@
...
+
+describe('Greeting', () => {
+  describe('constructor', () => {
+    it('can correctly construct a greeting instance', () => {
+      const peers = [otherAddress];
+
+      const g = new Greeting({
+        address: mockAddress,
+        channel: mockChannel,
+        peers,
```
Xelowak, line 16: Is it intended that it's not `peers: peers` ?

fe1-web/src/core/network/jsonrpc/__tests__/Greeting.test.ts
```diff
@@ -0,0 +1,104 @@
...
+  });
+
+  describe('fromJson', () => {
+    it('can correctly construct a greeting instance', () => {
+      const peers = [otherAddress];
+
+      const g = Greeting.fromJson({
+        address: mockAddress,
+        channel: mockChannel,
+        peers,
```
Xelowak, line 94: Same comment.

fe1-web/src/core/redux/ServerReducer.ts
```diff
@@ -0,0 +1,105 @@
...
+const serverSlice = createSlice({
+  name: SERVER_REDUCER_PATH,
+  initialState,
+  reducers: {
+    addServer: (state: Draft<ServerReducerState>, action: PayloadAction<ServerState>) => {
+      const server = action.payload;
+
+      state.byAddress[server.address] = server;
+      state.allAddresses.push(server.address);
+    },
```
Xelowak, line 35: If you add another server with the same address, but other properties (by mistake), it will get overwritten, right ? Is that the behaviour you want ?

Xelowak (commented):
Nice work !
I agree with what @pierluca said, and I added other comments on minor issues.

fe1-web/src/core/network/jsonrpc/__tests__/Greeting.test.ts
Tyratox: That is a shorthand notation for property names :) ("Object Property Value Shorthand") (It allows you to write `x` instead of `x:x` and is for example described here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#new_notations_in_ecmascript_2015)

Xelowak: I didn't know about it, thanks !

Xelowak (approved):

PR #924, opened 2022-04-04
Implement passive witnessing functionality

Tyratox:
For the following messages, the frontend will automatically send witness messages for the following object / action combinations:

- election#cast_vote
- chirp#add
- chirp#delete
- reaction#add

These are exactly the messages that are signed using pop tokens and whose content is determined by other people and hence should not require a manual verification.

https://github.com/dedis/popstellar/blob/dabd3f94ff5329781d5e564b4d73f3a608b76adb/fe1-web/src/core/network/jsonrpc/messages/MessageRegistry.ts#L55-L90

https://github.com/dedis/popstellar/blob/dabd3f94ff5329781d5e564b4d73f3a608b76adb/fe1-web/src/features/witness/network/messages/WitnessRegistry.ts#L35-L69

There is a feature toggle in place that prevents the actual network messages from being sent.


lhmerino:
Perfect, sounds good.

lhmerino (dismissed):
Looks great!

fe1-web/src/core/network/jsonrpc/messages/__tests__/MessageRegistry.test.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/dot-notation */
```
pierluca, line 1: why do you need to disable this ? are you trying to bypass the `private` field protection ? 

If so, a typical pattern is to create a `TestableMessageRegistry` which inherits from `MessageRegistry` and would somehow expose the ~~private~~ protected field. It makes the intention a tiny bit clearer.

fe1-web/src/core/network/jsonrpc/messages/MessageRegistry.ts
```diff
@@ -108,7 +113,17 @@ export class MessageRegistry {
...
+
+    const success = messageEntry.handle!(msg);
+
+    if (success) {
+      // execute all after processing handlers
+      for (const handler of this.afterProcessingHandlers) {
+        handler(msg);
+      }
+    }
+
```
pierluca, line 125: this feels very ad-hoc.

It's true that we could image a pre-processing and post-processing of `ProcessableMessage`, but I could imagine that we may have scenarios where a given pre-processing or post-processing applies to one message but not the other.

Taking a step back, do we really need something like this ? I'd opt for YAGNI 😅

fe1-web/src/features/witness/network/messages/WitnessRegistry.ts
```diff
@@ -0,0 +1,78 @@
+import { ActionType, MessageData, ObjectType } from 'core/network/jsonrpc/messages';
```
pierluca, line 1: why duplicate this and not add a field in `MessageRegistry` ?

fe1-web/src/features/witness/network/index.ts
```diff
@@ -1,18 +1,63 @@
...
     ActionType.WITNESS,
-    handleWitnessMessage,
+    handleWitnessMessage(config.getCurrentLao),
     WitnessMessage.fromJson,
   );
-}
+
+  config.messageRegistry.addAfterProcessingHandler(
+    afterMessageProcessingHandler(config.enabled /* config.isLaoWitness */),
+  );
```
pierluca, line 62: here we're hooking up to message registry / message processing directly.
It feels a bit ad-hoc, even though the solution was developed to be generic.

what if instead we hooked up to the store and watched messages being marked as processed (e.g. the `processMessages` event in the `MessageReducer`) ? (as we discussed today for the election key problem)

I see more code / more logic as a clear disadvantage of what I'm saying, but it'd also be purely centralized in the `witness` feature and a feature couldn't, on its own, break the message registry logic.

as a completely different approach, we could also envision the `processMessages` action to be wrapped in a thunk, so that pre-processors and post-processors can be added at that level (in a completely transparent way to the message registry).

Any thoughts ? The current implementation is fine anyhow, I'm just wondering what would look better😄

fe1-web/src/core/network/jsonrpc/messages/MessageRegistry.ts
Tyratox: I will re-implement it using redux 👍

fe1-web/src/core/network/jsonrpc/messages/__tests__/MessageRegistry.test.ts
Tyratox: Given that `afterProcessingHandlers` are removed anyway, this problem is solved as well :)
But yes, I agree. But it is not straight forward to do so, since I actually want to test the `MessageRegistry` and not some `TestableMessageRegistry`. Extending the class should work though 🤔

fe1-web/src/features/witness/network/messages/WitnessRegistry.ts
Tyratox: My thought was that it should be inside the feature and not part of the core?

fe1-web/src/features/witness/network/index.ts
Tyratox: Yes I think that is much nicer! I'll try to implement this using a store watcher now :)

fe1-web/src/core/keypair/__mocks__/KeyPairStore.ts
```diff
@@ -0,0 +1,19 @@
...
+import { mockKeyPair } from '__tests__/utils';
+
+import { KeyPair } from '../../objects';
+
+let keyPair: KeyPair = mockKeyPair;
+
+export namespace KeyPairStore {
+  export const store = jest.fn<void, [KeyPair]>().mockImplementation((kp: KeyPair) => {
+    keyPair = kp;
+  });
```
Xelowak, line 12: Can't you write it like this? I feel like it's clearer and less verbose. Not sure about it though. 
```suggestion
  export const store = jest.fn<void, [KeyPair]>((kp: KeyPair) => {
    keyPair = kp;
  });
```
The same would apply to the function below and to every function where you called `.mockImplementation()`.

fe1-web/src/features/lao/functions/lao.ts
```diff
@@ -23,6 +23,12 @@ export const getCurrentLao = () => {
  */
 export const getCurrentLaoId = () => selectCurrentLaoId(getStore().getState());
 
+/**
+ * Returns whether the user is witness of the current lao
+ * @returns Whether the user is witness of the current lao
```
Xelowak, line 28: This feels a little bit repetitive.

fe1-web/src/features/lao/interface/Configuration.ts
```diff
@@ -65,11 +65,17 @@ export interface LaoConfigurationInterface extends FeatureInterface {
     useLaoList: () => Lao[];
 
     /**
-     * Checks whether the current user is the organizer of the current lao
-     * @returns Whether the current user is the organizer of the current lao
+     * Checks whether the current user is an organizer of the current lao
+     * @returns Whether the current user is an organizer of the current lao
```
Xelowak, line 69: Same comment.

fe1-web/src/features/witness/network/WitnessStoreWatcher.ts
```diff
@@ -0,0 +1,102 @@
...
+export const afterMessageProcessingHandler =
+  (
+    enabled: WitnessConfiguration['enabled'],
+    /* isLaoWitness: WitnessConfiguration['isLaoWitness'] */
+  ) =>
+  (msg: ExtendedMessage) => {
+    const entry = getWitnessRegistryEntry(msg.messageData);
+
+    if (entry) {
+      // we have a wintessing entry for this message type
```
Xelowak, line 25: Just a typo.
```suggestion
      // we have a witnessing entry for this message type
```

fe1-web/src/features/witness/network/messages/WitnessRegistry.ts
Xelowak: I agree that it's part of the feature, but on the other side it could easily fit in the `MessageRegistry`, as all features depend on this too. Maybe you can talk about it with @0xdieman.

fe1-web/src/features/witness/network/messages/WitnessRegistry.ts
```diff
@@ -0,0 +1,78 @@
...
+  NOTIFY_ADD,
+  DELETE,
+  NOTIFY_DELETE,
+} = ActionType;
+
+export enum WitnessingType {
+  'NO_WITNESSING',
+  'PASSIVE',
+  'ACTIVE',
+}
```
Xelowak, line 27: Making a string-valued enum might help when debugging.

fe1-web/src/features/witness/network/messages/WitnessRegistry.ts
```diff
@@ -0,0 +1,78 @@
...
+  [k(CHIRP, ADD), { type: WitnessingType.PASSIVE }],
+  [k(CHIRP, NOTIFY_ADD), { type: WitnessingType.NO_WITNESSING }],
+  [k(CHIRP, DELETE), { type: WitnessingType.PASSIVE }],
+  [k(CHIRP, NOTIFY_DELETE), { type: WitnessingType.NO_WITNESSING }],
+
+  // Reactions
+  [k(REACTION, ADD), { type: WitnessingType.PASSIVE }],
+]);
+
+export const getWitnessRegistryEntry = (data: MessageData): WitnessEntry => {
```
Xelowak, line 71: Is it really relevant to get the whole `WitnessEntry`? I think it would be better to get each part individually (so the `WitnessingType` only for now), so that you don't have to use the object once retrieved.

fe1-web/src/features/witness/network/messages/__tests__/WitnessRegistry.test.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import { getWitnessRegistryEntry } from '../WitnessRegistry';
+
+describe('WitnessRegistry', () => {
+  it('returns entries for valid message data input', () => {
+    getWitnessRegistryEntry({ object: ObjectType.ELECTION, action: ActionType.SETUP });
+    getWitnessRegistryEntry({ object: ObjectType.ELECTION, action: ActionType.OPEN });
+    getWitnessRegistryEntry({ object: ObjectType.ELECTION, action: ActionType.CAST_VOTE });
+    getWitnessRegistryEntry({ object: ObjectType.ELECTION, action: ActionType.END });
+  });
```
Xelowak, line 14: You're not actually testing anything there, right? For it to become relevant, you should add a check to the entries, but then it means that updating the registry would imply updating the test, which is not practical.

Xelowak (commented):
Nice addition !
Thanks for all the documentation you added to your code, but I think that some of it is really too repetitive. When a function only returns a boolean, I think one sentence is enough to describe it, otherwise it becomes a little bit cumbersome.

fe1-web/src/features/witness/network/messages/WitnessRegistry.ts
Tyratox: 👍 I will do that! :)

fe1-web/src/core/keypair/__mocks__/KeyPairStore.ts
Tyratox: Oh, indeed this is much nicer! I did not know about this, thank you! 😁

fe1-web/src/features/lao/functions/lao.ts
Tyratox: 👍

fe1-web/src/features/lao/interface/Configuration.ts
Tyratox: 👍

fe1-web/src/features/witness/network/messages/__tests__/WitnessRegistry.test.ts
Tyratox: Well yes & no, the purpose of this test was to check whether it throws an error or not but I'll make this more explicit!

fe1-web/src/features/witness/network/index.ts
```diff
@@ -1,18 +1,28 @@
...
     WitnessMessage.fromJson,
   );
-}
+
+  // listen for new processable messages
+  const store = getStore();
+  store.subscribe(makeWitnessStoreWatcher(store, afterMessageProcessingHandler(config.enabled)));
+  /* config.messageRegistry.addAfterProcessingHandler(
+    afterMessageProcessingHandler(config.enabled, config.isLaoWitness),
+  ); */
```
0xdieman, line 27: Did you leave this on purpose ?
Also, I'm not sure I understand the difference between those two, is it that if the handler fails, the entire message processing fails, when the subscribing way only "watches" (does not influence the processing logic even if it fails)?

0xdieman (approved):

fe1-web/src/features/witness/network/index.ts
Tyratox: Oh whoops no, thank you!

Both approaches would have not make the message processing fail, the watcher approach is just a bit more "clean" since it directly reacts to state changes (when messages are added) and not to network events :)

For the other approach I had to add an additional `addAfterProcessingHandler` which is no longer needed with the store watcher :)

Xelowak (commented):
Thanks a lot for all the changes! Since I don't know if you made a decision about the WitnessRegistry, I'll let you merge this branch when ready.

Xelowak (approved):

PR #925, opened 2022-04-04
Implement a basic / ugly version for active witnessing

Tyratox:
This PR adds a notification screen to the lao navigation and a reducer where notifications can be stored. Moreover it implements a first type of notification that can be stored there: messages that have to be witnessed.

The implementation looks as follows: notifications have a `type` property that in the case of witness notifications is set to `message-to-witness`. The witness feature registers the WitnessNotification.tsx component with the notification feature and the SingleNotificationScreen.tsx then passes the notification to the component that is registered for the type of notification, i.e. in case of `message-to-witness` to WitnessNotification.tsx.

This screen then shows JSON.stringified version of the message that was received and two buttons: One that just deletes the notification and one that first sends a notification message to the server and then deletes the notification.

The placement of the notification screen is temporary and will be moved as soon as I have time to implement the UI changes discussed in the UX meeting last week. For now it seems as if it is an easy way to test this feature. Also the notification indicator is not very pretty right now.

Also at the moment only `lao#create` is set to active witnessing in

https://github.com/dedis/popstellar/blob/43bc7448690bae7cb51f13f9acab62f76f9ededf/fe1-web/src/features/witness/network/messages/WitnessRegistry.ts#L35-L69

There is a feature toggle in place that prevents the actual network messages from being sent.


Tyratox:
I just realized that the notification reducer should be separate per lao id!

fe1-web/src/features/notification/screens/NotificationScreen.tsx
```diff
@@ -0,0 +1,153 @@
...
+        notificationIds: notifications.map((n) => n.id),
+      }),
+    );
+  };
+
+  return (
+    <ScreenWrapper>
+      <FlatList
+        data={notificationData}
+        keyExtractor={(item) => ('id' in item ? item.id.toString() : item.title)}
```
Xelowak, line 118: Be careful with that. If 2 items happen to have the same id, the application will crash immediately.

fe1-web/src/features/notification/screens/NotificationScreen.tsx
```diff
@@ -0,0 +1,153 @@
...
+                  <Text>{item.title}</Text>
+                </View>
+              </TouchableOpacity>
+            );
+          }
+          // separator
+          return <Text style={Typography.important as TextStyle}>{item.title}</Text>;
+        }}
+      />
+      <Button title="Clear notifications" onPress={onClearNotifications} />
```
Xelowak, line 148: All strings that are displayed in the UI shall be defined in `strings` to allow easier translation.

fe1-web/src/features/notification/screens/SingleNotificationScreen.tsx
```diff
@@ -0,0 +1,55 @@
...
+type NavigationProps = StackScreenProps<NotificationStackParamList, 'Notification'>;
+
+const SingleNotificationScreen = () => {
+  const route = useRoute<NavigationProps['route']>();
+  const navigation = useNavigation<NavigationProps['navigation']>();
+  const { notificationId } = route.params;
+
+  const store = useStore();
+  const laoId = NotificationHooks.useCurrentLaoId();
+  const notification = getNotification(laoId.valueOf(), notificationId, store.getState());
```
Xelowak, line 22: I find it weird to have to pass the redux storage to this function. As I commented earlier, it might be better to treat it as a selector instead.

fe1-web/src/features/witness/components/WitnessNotification.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+const propTypes = {
+  notification: PropTypes.shape({
+    // WitnessFeature.MessageToWitnessNotification
+    id: PropTypes.number.isRequired,
+    timestamp: PropTypes.number.isRequired,
+    title: PropTypes.string.isRequired,
+    type: PropTypes.string.isRequired,
+    messageId: PropTypes.string.isRequired,
+  }).isRequired,
```
Xelowak, line 74: Couldn't it be treated as an object instance? I mean, if an object contains all those fields, you can just write `Prototypes.instanceOf(Object)`.
EDIT: It seems to match the `Notification` interface.

fe1-web/src/features/witness/components/__tests__/WitnessNotification.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+  WitnessFeature,
+  WITNESS_FEATURE_IDENTIFIER,
+} from 'features/witness/interface';
+import { addMessageToWitness, witnessReducer } from 'features/witness/reducer';
+
+import WitnessNotification from '../WitnessNotification';
+
+const mockMessageId = 'some message id';
+
+// otherwise .formData won't work
```
Xelowak, line 35: Small typo.
```suggestion
// otherwise .fromData won't work
```

fe1-web/src/features/witness/interface/Configuration.ts
```diff
@@ -21,27 +25,68 @@ export interface WitnessConfiguration {
...
+  discardNotifications: (args: { laoId: string; notificationIds: number[] }) => AnyAction;
 }
 
 /**
  * The type of the context that is provided to react witness components
  */
-export type WintessReactContext = Pick<WitnessConfiguration, 'useCurrentLao'>;
+export type WintessReactContext = Pick<
+  WitnessConfiguration,
+  | 'enabled'
```
Xelowak, line 64:
```suggestion
  'enabled'
```
Or does it work the way you wrote it?

fe1-web/src/features/witness/reducer/__tests__/WitnessReducer.test.ts
```diff
@@ -0,0 +1,203 @@
...
+      const newState = witnessReduce(
+        { allIds: [], byId: {} },
+        addMessageToWitness(extendedMessage),
+      );
+
+      expect(newState.allIds).toEqual([message.message_id.valueOf()]);
+      expect(newState.byId).toHaveProperty(message.message_id.valueOf(), extendedMessage);
+    });
+
+    it("doesn't add a messages a second time to the store", () => {
```
Xelowak, line 54:
```suggestion
    it("doesn't add a message a second time to the store", () => {
```

Xelowak (commented):
That's a big PR, good job! I made a bunch of comments for you, but it's pretty nice already.

Even if you're not satisfied by your current UI, it could be nice to add some screenshots to your PR description. This way, we can see how it looks from a user point of view without having to run your code locally.

I saw that you added 2 new typography types that are centered, and it seems that the 2 old ones (`base` and `important`) have been replaced in a lot of places. Are these still relevant or could they be deleted?

Thanks for all the tests, it's really appreciated!

Tyratox:
@Xelowak Regarding the styles: The `base` and `important` styles were centered which were confusing and thus I renamed them to `baseCentered` and `importantCentered`.

For the notification screen I now wanted the text to be left-aligned and thus there is a "new" `base` style with the text being left-aligned. As soon as I have time I would like to work on the UI a bit and make this a bit more uniform :)

fe1-web/src/features/notification/reducer/NotificationReducer.ts
```diff
@@ -0,0 +1,201 @@
...
+        };
+      }
+
+      const notification: NotificationState = {
+        ...action.payload,
+        id: laoState.nextId,
+        hasBeenRead: false,
+      };
+
+      laoState.nextId += 1;
```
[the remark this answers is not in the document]
pierluca: You do highlight a very important problem, which I believe doesn't affect us because... JavaScript !

JavaScript is not strictly concurrent, it's runtime model is based on an event loop.
Cf. https://developer.mozilla.org/en-US/docs/Web/JavaScript/EventLoop
Furthermore, reducers are basically called synchronously. 
Effectively, this means that there's only ever going to be one `addNotification` being executed.

fe1-web/src/features/notification/screens/NotificationScreen.tsx
```diff
@@ -0,0 +1,149 @@
...
+const NotificationScreen = () => {
+  const navigation = useNavigation<NavigationProps['navigation']>();
+
+  const laoId = NotificationHooks.useCurrentLaoId();
+  const selectAllNotifications = useMemo(
+    () => makeAllNotificationsSelector(laoId.valueOf()),
+    [laoId],
+  );
+  const notifications = useSelector(selectAllNotifications);
+  const notificationData = useMemo(() => {
```
pierluca, line 60: if possible, I'd try to extract a function out of this (and out of the component), as it seems very much independent of everything except for the `notifications` input.
It'd also clarify the goal of the code, as having `notificationData = useMemo(() => {....` doesn't tell me anything about what you're trying to achieve here.

fe1-web/src/features/witness/components/WitnessNotification.tsx
```diff
@@ -0,0 +1,106 @@
...
+export default WitnessNotification;
+
+export const WitnessNotificationType = {
+  /**
+   * Checks whether the given notification is a witness notification
+   * @param notification The notification whose type should be checked
+   * @returns True if the notification is a witness notification, false otherwise
+   */
+  isOfType: (notification: WitnessFeature.Notification) =>
+    'type' in notification && notification.type === MESSAGE_TO_WITNESS_NOTIFICATION_TYPE,
```
pierluca, line 90: I believe you might be able to define this lambda as a TypeScript [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) for extra clarity as to its behavior and intended use, which will then allow you to take any `notification` as parameter.

fe1-web/src/core/components/ScreenWrapper.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { StyleSheet, ViewStyle } from 'react-native';
+import { ScrollView } from 'react-native-gesture-handler';
+
+const styles = StyleSheet.create({
+  view: {
+    padding: 16,
+  } as ViewStyle,
+});
+
+const ScreenWrapper = ({ children }: IPropTypes) => (
```
pierluca, line 12: some documentation would be nice here (and elsewhere) 😉

fe1-web/src/features/lao/navigation/LaoNavigation.tsx
```diff
@@ -37,47 +35,20 @@ const styles = StyleSheet.create({
...
-  const publicKeyRaw = getKeyPairState(store.getState()).keyPair?.publicKey;
-  const publicKey = publicKeyRaw ? new PublicKey(publicKeyRaw) : undefined;
-
-  const isOrganizer = !!(lao && publicKey && publicKey.equals(lao.organizer));
-  const isWitness = !!(lao && publicKey && lao.witnesses.some((w) => publicKey.equals(w)));
-
-  const tabName: string = getLaoTabName(isOrganizer, isWitness);
-  const laoName: string = lao ? lao.name : STRINGS.unused;
+  const isOrganizer = useSelector(selectIsLaoOrganizer);
+  const isWitness = useSelector(selectIsLaoWitness);
```
pierluca, line 42: so much nicer 😍

fe1-web/src/features/notification/navigation/NotificationNavigation.tsx
```diff
@@ -0,0 +1,33 @@
...
+  return (
+    <NotificationStackNavigator.Navigator
+      initialRouteName={STRINGS.notification_navigation_tab_notifications}>
+      <NotificationStackNavigator.Screen
+        name={STRINGS.notification_navigation_tab_notifications}
+        component={NotificationScreen}
+        options={{
+          title: STRINGS.notification_navigation_tab_notifications_title,
+          // show not back button for this screen
+          headerLeft: () => null,
```
pierluca, line 22: I'm not sure I understand this.

fe1-web/src/features/witness/components/WitnessNotification.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  // if the notification state somehow gets out of sync, remove the corresponding notification
+  useEffect(() => {
+    if (!message) {
+      dispatch(
+        discardNotifications({ laoId: laoId.valueOf(), notificationIds: [notification.id] }),
+      );
+      navigateToNotificationScreen();
+    }
+  }, [laoId, navigateToNotificationScreen, discardNotifications, notification.id, message]);
```
pierluca, line 54: logically, I'd expect this to appear before the `onWitness`/ `onDecline` definitions.
Also, is this just a safeguard ? When is the "somehow gets out of sync" expected ?
If it's just added for robustness, I'd log debugging information to make sure that we know and understand what happens (and we aren't just redirected to some other page).

fe1-web/src/features/notification/reducer/NotificationReducer.ts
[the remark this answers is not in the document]
Tyratox: Ah and the event loop ["runs to completion"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/EventLoop#run-to-completion). Thank you for the pointer it is good to have a reference for this! 😄

fe1-web/src/__tests__/utils/TestUtils.ts
```diff
@@ -72,3 +73,14 @@ export const mockJsonRequest: Partial<JsonRpcRequest> = {
...
+export function wrapWithClearMockStorageReducer(reducers: Reducer): Reducer {
+  return (state: any, action: AnyAction) => {
+    let newState = state;
+
+    if (action.type === 'CLEAR_STORAGE') {
+      newState = undefined;
+    }
+    return reducers(newState, action);
+  };
+}
```
Xelowak, line 86: To be honest, once I saw this I realized that I might have been wrong when asking you to declare the store before the tests (right after the `describe`). I think it was cleaner and more readable to create a new one for each test (it was in `LaoProperties.test.ts`). Wrapping it as you did is fine, but it seems more complicated to me.

fe1-web/src/features/notification/reducer/NotificationReducer.ts
```diff
@@ -0,0 +1,201 @@
...
+  /* the id of the lao this notification is associated with */
+  laoId: string;
+  /* whether the notification has been read. is automatically assigned */
+  hasBeenRead: boolean;
+  /* the time associated with the notification */
+  timestamp: number;
+  /* the title that is shown in the notification */
+  title: string;
+  /* this field can be used to differentiate various types of notifications */
+  type: string;
```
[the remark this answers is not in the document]
Xelowak: Maybe it could be cast to a string-valued enum in each feature if needed.

fe1-web/src/features/notification/reducer/NotificationReducer.ts
```diff
@@ -0,0 +1,201 @@
...
+}
+
+export const NOTIFICATION_REDUCER_PATH = 'notifications';
+
+export interface NotificationReducerState {
+  byLaoId: {
+    [laoId: string]: { byId: Record<number, NotificationState>; allIds: number[]; nextId: number };
+  };
+}
```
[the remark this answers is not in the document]
Xelowak: Yes, it seems like a good idea, but the problem might be that you need to copy data from one to the other when a notification is read. Another method could be to maintain a counter of unread notifications, if you know that you won't need to retrieve notifications by their type (read/unread).

fe1-web/src/features/notification/screens/NotificationScreen.tsx
Xelowak: Yeah that's what I was trying to point out, thanks @pierluca . The id is fine, but the title could be the same if the id happens to be undefined.

fe1-web/src/features/witness/interface/Configuration.ts
Xelowak: Okay, I find it weird but why not 😅

Xelowak (commented):
Thanks for the changes!
I added some minor comments, and I agree with what Pierluca pointed out.

fe1-web/src/__tests__/utils/TestUtils.ts
Tyratox: Yes I agree that it is simpler to create a new one for each test :) Should I leave `wrapWithClearMockStorageReducer` in the codebase or remove it again?

fe1-web/src/features/witness/reducer/WitnessReducer.ts
```diff
@@ -0,0 +1,109 @@
...
+import { createSelector, createSlice, Draft, PayloadAction } from '@reduxjs/toolkit';
+
+import { ExtendedMessage, ExtendedMessageState } from 'core/network/ingestion/ExtendedMessage';
+
+export const WITNESS_REDUCER_PATH = 'witness';
+
+export interface MessagesToWitnessReducerState {
+  byId: Record<string, ExtendedMessageState>;
+  allIds: string[];
+}
```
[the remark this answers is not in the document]
Tyratox: Okay :) Just wasn't sure how independent the feature can be but I guess it should be fine to rely on everything in the core

fe1-web/src/features/notification/reducer/NotificationReducer.ts
```diff
@@ -0,0 +1,217 @@
...
+          // this message was never stored?
+          return;
+        }
+
+        delete state.byLaoId[laoId].byId[notificationId];
+        // remove the id from allIds
+        state.byLaoId[laoId].allIds.splice(
+          state.byLaoId[laoId].allIds.findIndex((id) => id === notificationId),
+          1,
+        );
```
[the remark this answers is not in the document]
Tyratox: Thank you for the link :)

fe1-web/src/features/notification/navigation/NotificationNavigation.tsx
Tyratox: I want to prevent this arrow from showing.
Screenshot 2022-05-03 at 08 35 49

I now updated the comment to
```
// show no back button for this screen in the top navigation bar
// the back button should only be used from notification detail/single views
// to get back to this screen but not to go from this screen "back" to anywhere
```

Is this clearer?

fe1-web/src/features/notification/reducer/NotificationReducer.ts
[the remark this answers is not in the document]
Tyratox: True, that should work! String-valued enums are a subset of string :)

fe1-web/src/features/notification/reducer/NotificationReducer.ts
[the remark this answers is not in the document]
Tyratox: Actually I need to have them separated, right now I just check their field to separate them but if the store already has them in this format, that would no longer be needed :)

fe1-web/src/features/notification/reducer/NotificationReducer.ts
```diff
@@ -0,0 +1,201 @@
...
+
+/**
+ * Retrives a single notification state by id
+ * NOTE: This function does not memoize since there is no computation taking place
+ * @param laoId The id of the lao
+ * @param notificationId The id of the notification to retrieve
+ * @param state The redux state
+ * @returns A single notificatio n state
+ */
+export const getNotification = (laoId: string, notificationId: number, state: unknown) => {
```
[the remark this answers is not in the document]
Tyratox: Okay that makes much more sense! I was very confused about this statement in the documentation, especially the word `incorrect` which made it sound as if something is going to break for some reason 😂
From your explanation I would have expected `inefficient` / `does not need memoization` or something but oh well :)

fe1-web/src/features/notification/screens/NotificationScreen.tsx
Tyratox: I hope it is clearer now? 😅

fe1-web/src/features/notification/screens/NotificationScreen.tsx
Tyratox: I added comments and a check at the end of the `useMemo` that checks for uniqueness and throws an error otherwise. Do you think this suffices or did you have anything else in mind? I think this is easier than adding a complex mechanism for generating unique identifiers for the headings?

fe1-web/src/features/notification/screens/NotificationScreen.tsx
```diff
@@ -0,0 +1,149 @@
...
+    );
+  };
+
+  return (
+    <ScreenWrapper>
+      <FlatList
+        data={notificationData}
+        keyExtractor={(item) => ('id' in item ? item.id.toString() : item.title)}
+        renderItem={({ item }) => {
+          if ('id' in item) {
```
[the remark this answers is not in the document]
Tyratox: Ah yes of course but I just used this check to differentiate between headings (that never have an `id`) and notifications (that always have one) :) Since they are not classes, I could not use `instanceof`

fe1-web/src/features/notification/screens/NotificationScreen.tsx
```diff
@@ -0,0 +1,149 @@
...
+                  }>
+                  <Text>{item.title}</Text>
+                </View>
+              </TouchableOpacity>
+            );
+          }
+          // separator
+          return <Text style={Typography.important as TextStyle}>{item.title}</Text>;
+        }}
+      />
```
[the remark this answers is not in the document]
Tyratox: Sounds like an good idea :)

fe1-web/src/features/notification/screens/SingleNotificationScreen.tsx
Tyratox: That is true and I fully agree that `useSelector` is more appropriate but since I thought we cannot use `createSelector` if no computation takes place, this was the only option left 😅

fe1-web/src/features/witness/components/WitnessNotification.tsx
Tyratox: Ah yes for sure, that is horrible! I am just not sure whether it is a good idea to add a class just for the sake of being able to validate it here using PropTypes 🤔 But maybe it is actually worth it, what do you think? Maybe there will be some other use for this class at a later point in time 😂

fe1-web/src/features/witness/network/WitnessStoreWatcher.ts
```diff
@@ -16,21 +24,60 @@ import { requestWitnessMessage } from './WitnessMessageApi';
...
+
+    const storedMessage = getMessageToWitness(msg.message_id.valueOf(), getStore().getState());
+    if (storedMessage) {
+      // this message is already stored in the witness reducer
+      // and hence does not have to be stored a second time
+      return;
+    }
+
     const type = getWitnessRegistryEntryType(msg.messageData);
+    const lao = getCurrentLao();
```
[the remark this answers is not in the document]
Tyratox: Oh yes of course, not sure why I did this

fe1-web/src/features/witness/components/WitnessNotification.tsx
Tyratox: Makes sense!

Yes it is just as a safeguard :) I will add debugging information!

fe1-web/src/features/witness/network/WitnessStoreWatcher.ts
```diff
@@ -16,21 +24,60 @@ import { requestWitnessMessage } from './WitnessMessageApi';
...
 
     if (type) {
-      // we have a witnessing entry for this message type
+      // we have a wintessing entry for this message type
```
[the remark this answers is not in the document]
Tyratox: 🙈🙈 According to the search it was the last one 😂🥳

fe1-web/src/features/witness/network/WitnessStoreWatcher.ts
```diff
@@ -16,21 +24,60 @@ import { requestWitnessMessage } from './WitnessMessageApi';
...
+
       switch (type) {
         case WitnessingType.PASSIVE:
-          if (enabled) {
-            requestWitnessMessage(msg.channel, msg.message_id);
+          if (!enabled) {
+            return;
           }
+
+          requestWitnessMessage(msg.channel, msg.message_id);
```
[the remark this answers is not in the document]
Tyratox: Hmmm interesting yes! I mean probably we would then add the ability to send messages to different LAOs (as a parameter to `requestWitnessMessage`) but I agree that this check would be good for the moment :)

And yes, I should definitely add a catch for `requestWitnessMessage`!

fe1-web/src/features/witness/components/WitnessNotification.tsx
Tyratox: Oh I did not know about this, thank you! 😁

fe1-web/src/resources/strings.ts
```diff
@@ -1,239 +1,260 @@
-const STRINGS = {
+/* eslint-disable @typescript-eslint/naming-convention */
```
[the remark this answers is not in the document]
Tyratox: Yes.. I will google for that :)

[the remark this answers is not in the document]
Tyratox: I only found a way to completely exclude the file.. Should we change all variable names or should I set them to wontfix on sonarcloud?

fe1-web/src/__tests__/utils/TestUtils.ts
Xelowak: As you wish, but I guess it won't be used anywhere else for now, so you can delete it.

fe1-web/src/features/witness/components/WitnessNotification.tsx
```diff
@@ -0,0 +1,120 @@
...
+import { WitnessHooks } from '../hooks';
+import { WitnessFeature } from '../interface';
+import { requestWitnessMessage } from '../network/WitnessMessageApi';
+import { removeMessageToWitness } from '../reducer';
+
+const WitnessNotification = ({ notification, navigateToNotificationScreen }: IPropTypes) => {
+  const messageSelector = useMemo(
+    () => makeMessageSelector(notification.messageId),
+    [notification.messageId],
+  );
```
Xelowak, line 19: Same remark as others.

fe1-web/src/features/witness/components/WitnessNotification.tsx
Xelowak: Another way of handling this could be to add this `PropTypes.shape` in [this file](https://github.com/dedis/popstellar/blob/master/fe1-web/src/resources/Props.ts). The problem is that it would again need to be updated when the notification implementation changes (as Pierluca said), but this time we're sure that it will be updated only once.
Perhaps a class could be worth it to ensure this validation, yes.

fe1-web/src/features/witness/components/WitnessNotification.tsx
```diff
@@ -0,0 +1,120 @@
...
+    }
+  };
+
+  return (
+    <View>
+      <Text>{JSON.stringify(notification)}</Text>
+      <Text>{JSON.stringify(message)}</Text>
+      <Button title="Witness Message" onPress={onWitness} />
+      <Button title="Decline Message" onPress={onDecline} />
+    </View>
```
Xelowak, line 73: Is it temporary? I'm not sure that a user which sees a JSON would understand it directly.

fe1-web/src/features/witness/network/WitnessStoreWatcher.ts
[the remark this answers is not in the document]
Xelowak: Is it resolved? It seems the lao selector is still in here.

Xelowak (commented):
Nice job! Thanks a lot for the changes. I made a couple of comments, but those are minor and I believe we'll soon be able to merge this big PR!

fe1-web/src/features/notification/screens/NotificationScreen.tsx
```diff
@@ -0,0 +1,186 @@
...
+const NotificationScreen = () => {
+  const laoId = NotificationHooks.useCurrentLaoId();
+  const selectUnreadNotifications = useMemo(
+    () => makeUnreadNotificationsSelector(laoId.valueOf()),
+    [laoId],
+  );
+  const selectReadNotifications = useMemo(
+    () => makeReadNotificationsSelector(laoId.valueOf()),
+    [laoId],
+  );
```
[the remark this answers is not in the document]
Tyratox: True but `makeReadNotificationsSelector` is not a selector itself, it is a function that returns a selector 😅 And if I understood correctly, the same selector reference should be used for memoization to work, i.e.

`useSelector(makeReadNotificationsSelector(laoId.valueOf()))` would not work because the selector is a new one every time.

https://redux.js.org/usage/deriving-data-selectors#creating-unique-selector-instances

fe1-web/src/features/notification/screens/SingleNotificationScreen.tsx
```diff
@@ -0,0 +1,62 @@
...
+          notification={notification}
+          navigateToNotificationScreen={() => navigation.goBack()}
+        />
+      ) : (
+        <Text>
+          <Text>
+            No notification type component has been provided to render the following notification:
+          </Text>
+          <Text>{JSON.stringify(notification)}</Text>
+        </Text>
```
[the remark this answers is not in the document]
Tyratox: That is true! But just crashing the application is not a good solution either, right? So maybe just show the notification type and tell the user to report this as a bug?

fe1-web/src/features/witness/components/WitnessNotification.tsx
Tyratox: Yes this is temporary. (One of the reason for say `ugly` version) 😅
I am not sure yet what information should be displayed (since it also depends on the message) and what the best way of doing so is :) I thought I'd be better to improve this in a separate PR?

fe1-web/src/features/witness/network/WitnessStoreWatcher.ts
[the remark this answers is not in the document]
Tyratox: It is still here but there is a check right below whether the lao corresponding to the message matches the currently connected to lao. For messages that are witnessed passively, we currently would not have a way of saying "send this message on some connection to the lao with this id".

fe1-web/src/features/witness/components/WitnessNotification.tsx
Tyratox: At the moment I am not sure whether a class is a good idea 🤔 I mean it would work but it I think it would require passing the constructor of a class:

The `WitnessNotification.tsx` component is passed to the notification feature and it is used of a given notification is of type `WitnessNotification`. We could define a class for `WitnessNotification` that extends some `BasicNotification` class defined in the `notification` feature so that the witness feature knows the exact type and the notification feature still has some idea of the datatypes it is operating on.

But then in order to extend `BasicNotification` in the `witness` feature we would have to pass this via dependency injection to the witness feature. This should work but I am not 100% how intuitive it is. Do you see a better solution or do you think this is intuitive?

fe1-web/src/features/notification/hooks/__tests__/NotificationHooks.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+  (store: Store) =>
+  ({ children }: { children: React.ReactChildren }) =>
+    (
+      <Provider store={store}>
+        <FeatureContext.Provider value={contextValue}>{children}</FeatureContext.Provider>
+      </Provider>
+    );
+
+describe('NotificationHooks', () => {
+  describe('NotificationHooks.useNotificationTypes', () => {
```
Xelowak, line 38:
```suggestion
  describe('useNotificationTypes', () => {
```
It will already be printed by the describe on the previous line.

Xelowak (dismissed):
LGTM! I left tiny comments but you can safely ignore them and merge if you want to.

fe1-web/src/features/notification/screens/SingleNotificationScreen.tsx
```diff
@@ -0,0 +1,61 @@
...
+        <Component
+          notification={notification}
+          navigateToNotificationScreen={() => navigation.goBack()}
+        />
+      ) : (
+        <Text>
+          <Text>
+            Unkown notification type &apos{notification?.type}&apos. Please report this as a bug.
+          </Text>
+        </Text>
```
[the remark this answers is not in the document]
Tyratox: Hahaha oh no! How could did this happen 😂

Tyratox:
@Xelowak I updated the PR once more with the mentioned changes :) But unfortunately it now requires re-approval 😅

Xelowak (approved):
Re-approving!

PR #926, opened 2022-04-04
Adapt Karate front-ends tests

Warhedgehog:
Adapts the karate front-ends tests to be runnable again. Slightly improve the documentation for the front-end part.


tests/karate/src/test/java/fe/utils/web.feature
```diff
@@ -1,6 +1,7 @@
 Feature: web test
 
   Background: App Preset
+    #* configure driver = { type: 'chrome', executable: 'C:/Program Files/Google/Chrome/Application/chrome.exe'}
```
Tesa27, line 4: I think this should not be present here, if it will be used in an upcoming PR better place it there

tests/karate/src/test/java/fe/utils/android.feature
```diff
@@ -34,10 +41,16 @@ Feature: android page object
...
-      And click('#com.github.dedis.popstellar:id/button_apply')
-      And match backend.waitForConnection(5000) == true
+    And input('#com.github.dedis.popstellar:id/entry_box_server_url', backendURL)
+    And click('#com.github.dedis.popstellar:id/button_apply')
+    And match backend.waitForConnection(5000) == true
+
+    And click(tab_wallet_selector)
+    And click(tab_wallet_new_wallet_selector)
+    And click(tab_wallet_confirm_selector)
+    And dialog(true)
```
Tesa27, line 55: Having a lot of Ands like this is not very informative IMO, better use some other conjunctions like Then and When, to make it more readable. This has no impact on the functionality I just think that it is more clear for newcomers who try to understand the code. If we only want to perform scripting without trying to satisfy the reader I think going with * instead of And is a better option.

Tesa27 (changes requested):
Good job overall, I just have some small comments.

lhmerino (dismissed):
Looks great!

tests/karate/src/test/java/fe/utils/android.feature
Warhedgehog: I agree I've updated it and the existing code

tests/karate/src/test/java/fe/utils/web.feature
Warhedgehog: This is on purpose. Karate considers that the default installation file for Chrome is `/Program Files(x86)/` but another common default file is `/Program Files/` from what I've read Google went a bit back and forth on this and furthermore it depends if the installation is for all users or not. 
So the idea of the commented line is to be able to quickly change for people that need it.

Tesa27: Okay I agree than

Tesa27 (approved):
LGTM

lhmerino (approved):

PR #930, opened 2022-04-07
Removed election end time check

Robingoumaz:
Just a small PR to remove election end time check to end an election before the set end time according to [this comment](https://github.com/dedis/popstellar/pull/850#issuecomment-1054593279)


Warhedgehog (approved):
I imagine that this change is approved by the election team and consistent across all system.

nkcr (approved):

PR #931, opened 2022-04-07
[FE2] Create a Button on the main activity

zoemarin:
This PR Just adds a button on the main interface 
It correspond to a new option Transaction

Signed-off-by: zoemarin <zoe@marin.info>


Tesa27 (approved):

jbsv (changes requested):
Please double check the toast text. Having the same text for 2 different buttons is suspicious and might  bring confusion.

fe2-android/app/src/main/res/layout/tab_home.xml
```diff
@@ -17,7 +17,7 @@
     android:layout_width="match_parent"
     android:layout_height="match_parent"
     android:orientation="vertical"
-    app:layout_constraintGuide_percent="@dimen/guideline_vertical_tab_home" />
+    app:layout_constraintGuide_percent="0.16" />
```
GabrielFleischer, line 20: It is a good practice to use placeholders for this kind of buttons.

It allows that magic number to be reused and the dimen file can be easily swapped depending on the device model.
(We do not do such a thing, but it might be a thing to consider in the future)

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeActivity.java
```diff
@@ -262,6 +264,16 @@ public void setupSocialMediaButton() {
     socialMediaButton.setOnClickListener(v -> mViewModel.openSocialMedia());
   }
 
+  public void setupDigitalCashButton() {
+    Button digitalCashButton = findViewById(R.id.tab_digital_cash);
+    digitalCashButton.setOnClickListener(
+        new View.OnClickListener() {
+          public void onClick(View v) {
+            Toast.makeText(HomeActivity.this, "Button for Transaction", Toast.LENGTH_LONG).show();
```
GabrielFleischer, line 272: In Android, we try to use string.xml to store any text that will be displayed to the user.

This makes localization (translation) very easy.

GabrielFleischer (changes requested):
This looks good, I just raised some small bad practices in android.

I am also scared that this PR will be broken with @Warhedgehog 's changes.
You might want to ask him help for the merge.

PR #932, opened 2022-04-07
Work fe2 marin digital cash UI button lao

zoemarin:
This PR Just adds a button on the lao interface
It correspond to a new option Transaction which is use in a lao

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/LaoDetailActivity.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/LaoDetailActivity.java
@@ -233,10 +233,7 @@ public class LaoDetailActivity extends AppCompatActivity {
     Button digitalCashButton = findViewById(R.id.tab_digital_cash);
     digitalCashButton.setOnClickListener(
         new View.OnClickListener() {
-          public void onClick(View v) {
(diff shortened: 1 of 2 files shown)


Tesa27 (approved):

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/LaoDetailActivity.java
```diff
@@ -226,6 +229,17 @@ public void setupSocialMediaButton() {
...
   }
 
+  public void setupDigitalCashButton() {
+    Button digitalCashButton = findViewById(R.id.tab_digital_cash);
+    digitalCashButton.setOnClickListener(
+        new View.OnClickListener() {
+          public void onClick(View v) {
+            Toast.makeText(LaoDetailActivity.this, "Button for Transaction", Toast.LENGTH_LONG)
+                .show();
+          }
```
GabrielFleischer, line 239: You can simplify this with a lambda :

```suggestion
         v -> Toast.makeText(LaoDetailActivity.this, "Button for Transaction", Toast.LENGTH_LONG).show()
```

And as stated in the last PR, you should use `string.xml` values when text is shown to the user.

GabrielFleischer (changes requested):
Looks almost good to me.

I have a small nitpick on style

PR #933, opened 2022-04-07
Work fe2 marin digital add some events 

zoemarin:
Add some events in both view model 
Connect them in the activity

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/LaoDetailActivity.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/LaoDetailActivity.java
@@ -89,6 +89,8 @@ public class LaoDetailActivity extends AppCompatActivity {
     setupIdentityFragment();
     // Subscribe to " open social media " event
     setupSocialMediaActivity();
+    // Subscribe to " open digital cash " event
+    setupDigitalCashMainActivity();
     // Subscribe to " open witness message" event
     setupWitnessMessageFragment();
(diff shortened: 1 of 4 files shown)


fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/LaoDetailActivity.java
```diff
@@ -291,6 +294,10 @@ private void setupSocialMediaActivity() {
             });
   }
 
+  private void setupDigitalCashMainActivity() {
+    // TODO do the things here when activity is created
+  }
+
```
Tesa27, line 300: If the content of this function will be addressed in another PR I would remove it here since it does nothing. I think having empty functions is a bad practice

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeActivity.java
```diff
@@ -330,6 +345,10 @@ private void setupSocialMediaActivity() {
     }
   }
 
+  private void setupDigitalCashActivity() {
+    // TODO : EMPTY
+  }
```
Tesa27, line 350: Same as above

Tesa27 (changes requested):
As a general trend I think having empty functions or empty tests is not a a good thing. Leaving an empty function and filling in the functionality later seems a bit countraintuitive, better have the function  declaration and it's content in the same PR.

PR #936, opened 2022-04-08
Work fe2 marin digital cash gson object

zoemarin:
Create a Clean Json Representation of the Transaction 
(very similar to the draft)

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/digitalcash/PostTransaction.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/digitalcash/PostTransaction.java
@@ -1,0 +1,50 @@
+package com.github.dedis.popstellar.model.network.method.message.data.digitalcash;
+
+import com.github.dedis.popstellar.model.network.method.message.data.Action;
+import com.github.dedis.popstellar.model.network.method.message.data.Data;
+import com.github.dedis.popstellar.model.network.method.message.data.Objects;
+import com.google.gson.annotations.SerializedName;
+
+public class PostTransaction extends Data {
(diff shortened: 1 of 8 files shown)


fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/digitalcash/Transaction.java
```diff
@@ -0,0 +1,88 @@
...
+        + '\''
+        + ", TxIns="
+        + Arrays.toString(TxIns.toArray())
+        + '\''
+        + ", TxOuts="
+        + Arrays.toString(TxOuts.toArray())
+        + '\''
+        + ", timestamp="
+        + timestamp
+        + '\''
```
zoemarin, line 85: Are you agreeing on the '\'' at each time ? 
I wasn't sure (because sometimes the auto build don't put them)
I put them on my own

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/digitalcash/PostTransaction.java
```diff
@@ -0,0 +1,56 @@
...
+
+  @SerializedName(value = "transaction")
+  private final Transaction transaction; // the transaction object // String
+
+  /**
+   * @param transaction the transaction object
+   */
+  public PostTransaction(Transaction transaction) {
+    this.transaction = transaction;
+    this.transaction_id = Base64.getEncoder().encodeToString(transaction.toString().getBytes());
```
zoemarin, line 22: I have a question here 
How do we encode a transaction in base64 ?  
And then how do we hash it in sha256 and get a String at the end ?

PR #943, opened 2022-04-08
Fix roll call issue

Tyratox:
This reverts a changed that was most likely introduced by accident.

Compare to an earlier version

https://github.com/dedis/popstellar/blob/6736b0b65a3908a03c0d56e2582a67d459ab5b95/fe1-web/src/features/rollCall/components/EventRollCall.tsx#L57-L79

Without this fix, the pop tokens are not valid since they are not generated using the rollcalls id after creation but rather the new one after opening it.


Xelowak (approved):
You're right, the id alias is only used to keep up what's happening in the roll call, but it should not be used to generate tokens.

PR #945, opened 2022-04-11
Fix home navigation bug

Warhedgehog:


GabrielFleischer (approved):
I see, it was that kind of issue.

Great fix !

PR #949, opened 2022-04-13
Fix roll call closing bug

0xdieman:
The correct roll call ID was used to generate the organizer's token, but the one used for closing the roll call was not the updated idAlias anymore.
Fixed it by passing the two IDs in the scanning screen.


Tyratox (approved):
LGTM, thank you!

pierluca (approved):
LGTM !

PR #951, opened 2022-04-13
Fix wallet navigation

0xdieman:
This PR fixes the wallet navigation issues when navigating through the tabs, once connected the setup screen should not appear anymore.
Also migrates to a hook usage for navigation rather than `Proptypes`.


0xdieman:
It seemed to be a routing issue, calling reset lets no choice to the navigator but to show the correct screen.
Also I discovered the `initialRouteName` thing for the navigator, which fixes the initial route only on the first load of the navigator, so when connecting to/creating an LAO it loads the right screen if already connected.

fe1-web/src/features/wallet/screens/WalletSetSeed.tsx
```diff
@@ -28,7 +28,10 @@ const WalletSetSeed = ({ navigation }: IPropTypes) => {
   const initWallet = async () => {
     try {
       await Wallet.importMnemonic(seed);
-      navigation.navigate(STRINGS.navigation_synced_wallet);
+      navigation.reset({
```
pierluca, line 31: I would recommend you rely on `useNavigation()` instead of passing it through prop-types, it's just so much simpler 😉

0xdieman: Ook !

pierluca (dismissed):

fe1-web/src/features/wallet/screens/WalletSetSeed.tsx
```diff
@@ -21,14 +20,18 @@ const styles = StyleSheet.create({
...
+const WalletSetSeed = () => {
   /* used to set the mnemonic seed inserted by the user */
   const [seed, setSeed] = useState('');
+  const navigation = useNavigation<any>();
 
   const initWallet = async () => {
     try {
       await Wallet.importMnemonic(seed);
-      navigation.navigate(STRINGS.navigation_synced_wallet);
+      navigation.reset({
```
pierluca, line 31: I'm okay with this, but I find it a bit odd that we need to use `reset` though. 
What's wrong with `navigate` ? I'm not sure I understand.

0xdieman: Yes,it is odd, the thing is I still do not understand exactly why the bug was happening, navigate should move the selected screen on top of the stack, and the wallet screen stack should not change even when the "parent" navigator(the one used for changing tabs) changes. But it did for whatever reason. 
I figured that maybe resetting the routes history would help, but that still does not explain the bug.

fe1-web/src/features/wallet/screens/WalletError.tsx
```diff
@@ -16,22 +15,18 @@ const styles = StyleSheet.create({
...
-      title={STRINGS.back_to_wallet_home}
-      onPress={() => navigation.navigate(STRINGS.navigation_home_tab_wallet)}
-    />
-  </View>
-);
-
-const propTypes = {
-  navigation: PROPS_TYPE.navigation.isRequired,
+const WalletError = () => {
+  const navigation = useNavigation<any>();
```
Xelowak, line 19: You should maybe add a `FIXME` comment here, as it's not intended to use `any` and it should be fixed in the future.

fe1-web/src/features/wallet/screens/WalletSetSeed.tsx
```diff
@@ -21,14 +20,18 @@ const styles = StyleSheet.create({
 /**
  * Wallet screen to set an already existing mnemonic
  */
-const WalletSetSeed = ({ navigation }: IPropTypes) => {
+const WalletSetSeed = () => {
   /* used to set the mnemonic seed inserted by the user */
   const [seed, setSeed] = useState('');
+  const navigation = useNavigation<any>();
```
Xelowak, line 26: Same comment.

fe1-web/src/features/wallet/screens/__tests__/WalletError.test.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { render } from '@testing-library/react-native';
+import React from 'react';
+
+import MockNavigator from '__tests__/components/MockNavigator';
+
+import { WalletError } from '../index';
+
+const mockSeed = 'one two three';
+
+jest.mock('../../objects/Seed', () => {
```
Xelowak, line 11: It might be better to use an absolute path for consistency with other tests. 
By the way, I think it's nicer if blocks are separated by an empty line.

Xelowak (commented):
Thanks for the bug fix !
I'm not sure about the usage of `useNagivation<any>`. I know we've done it elsewhere in the project, but if you know exactly the type of the navigation you're using it might be better to create a type for it, especially if it is only accross the same feature. If it seems too complicated or would take too much time, just keep it as it is and add `FIXME`'s.

fe1-web/src/features/wallet/screens/WalletError.tsx
0xdieman: I will add the navigation type on another PR I think, this PR is meant to be a small fix for the navigation system.

Xelowak (approved):

Issue #948, opened 2022-04-12
[BUG] Attendee cannot connect to LAO on Fe 2


Warhedgehog:
### Description (Actual behavior)
When scanning the QR code of the organizer, the user is notified that a timeout occured.

### Expected behavior
The user expects that when they scan the organizer's QR code they will be able to join the LAO.

### How to reproduce
1. Scan the QR code of an organizer that has created an LAO 
2. See the toast indicating that a timeout occured

### Version & Environment
This bug was reproduced on:
- Commit: [df27a83](https://github.com/dedis/popstellar/commit/df27a8336f5a06dc62f66412b3f9ce66e98384ac)

##### Front-ends:
- [ ] Fe1-Web (please include browser's names & version)
- [x] Fe2-Android (please specify if phone or emulation, and Android version)
- [ ] Not applicable

##### Back-ends:
- [ ] Be1-Go
- [ ] Be2-Scala
- [x] Not applicable

##### Environment (as applicable):
 - Device: [e.g. Diverse android phones, Samsung A50, S8]
 - OS: [e.g. Android 11, Android 9]

### Workaround
There is no workaround

### Impact
The app is unusable for an attendee

PR #955, opened 2022-04-17
Fix Lao Connection Bug

Warhedgehog:
This solves #948 

This is the implementation of a workaround where the change of websocket connection and the subscription/catch up process are done in separate activities. Therefore this adds a separate connection activity rather than a `HomeActivity` fragment.

The code smell is a false positive since the tests that makes no assertion, tests the opening of the `HomeActivity` upon pressing cancel.

### DIFF
--- a/fe2-android/app/src/main/AndroidManifest.xml
+++ b/fe2-android/app/src/main/AndroidManifest.xml
@@ -17,9 +17,7 @@
     android:roundIcon="@mipmap/ic_launcher_round"
     android:supportsRtl="true"
     android:theme="@style/AppTheme">
-    <activity
-      android:name=".ui.home.connecting.ConnectingActivity"
-      android:exported="false" />
+    <activity android:name=".ui.home.connecting.ConnectingActivity" />
     <activity
       android:name=".ui.home.HomeActivity"
       android:exported="true">
--- a/fe2-android/app/src/main/AndroidManifest.xml
+++ b/fe2-android/app/src/main/AndroidManifest.xml
@@ -18,7 +18,7 @@
     android:supportsRtl="true"
     android:theme="@style/AppTheme">
     <activity
-      android:name=".ui.home.ConnectingActivity"
+      android:name=".ui.home.connecting.ConnectingActivity"
       android:exported="false" />
     <activity
       android:name=".ui.home.HomeActivity"
--- a/fe2-android/app/src/main/AndroidManifest.xml
+++ b/fe2-android/app/src/main/AndroidManifest.xml
@@ -8,17 +8,20 @@
   <uses-permission android:name="android.permission.INTERNET" />
 
   <application
+    android:name=".PoPApplication"
     android:allowBackup="true"
     android:fullBackupContent="@xml/backup_rules"
     android:icon="@mipmap/ic_launcher"
     android:label="@string/app_name"
-    android:name="com.github.dedis.popstellar.PoPApplication"
     android:networkSecurityConfig="@xml/network_security_config"
     android:roundIcon="@mipmap/ic_launcher_round"
     android:supportsRtl="true"
     android:theme="@style/AppTheme">
     <activity
-      android:name="com.github.dedis.popstellar.ui.home.HomeActivity"
+      android:name=".ui.home.ConnectingActivity"
+      android:exported="false" />
+    <activity
+      android:name=".ui.home.HomeActivity"
       android:exported="true">
       <intent-filter>
         <action android:name="android.intent.action.MAIN" />
@@ -27,7 +30,7 @@
       </intent-filter>
     </activity>
     <activity
-      android:name="com.github.dedis.popstellar.ui.detail.LaoDetailActivity"
+      android:name=".ui.detail.LaoDetailActivity"
       android:windowSoftInputMode="adjustPan" />
     <activity android:name=".ui.settings.SettingsActivity" />
     <activity android:name=".ui.socialmedia.SocialMediaActivity" />
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/remote/LAONetworkManager.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/remote/LAONetworkManager.java
@@ -104,6 +104,7 @@ public class LAONetworkManager implements MessageSender {
 
   @Override
   public Completable catchup(Channel channel) {
+    Log.d(TAG, "sending a catchup to the channel " + channel);
     Catchup catchup = new Catchup(channel, requestCounter.incrementAndGet());
 
     return request(catchup)
@@ -184,6 +185,7 @@ public class LAONetworkManager implements MessageSender {
             .map(Answer.class::cast)
             // This specific request has an id, only let the related Answer pass
             .filter(answer -> answer.getId() == query.getRequestId())
+            .doOnNext(answer -> Log.d(TAG, "request id: " + answer.getId()))
             // Transform from an Observable to a Single
             // This Means that we expect a result before the source is disposed and an error will be
             // produced if no value is received.
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/remote/LAONetworkManager.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/remote/LAONetworkManager.java
@@ -111,7 +111,6 @@ public class LAONetworkManager implements MessageSender {
         .map(ResultMessages.class::cast)
         .map(ResultMessages::getMessages)
         .doOnSuccess(messages -> Log.d(TAG, "Catchup on " + channel + " retrieved : " + messages))
-        .doOnError(msg -> Log.d(TAG, "Error in catchup is " + msg))
         .doOnSuccess(messages -> handleMessages(messages, channel))
         .ignoreElement();
   }
@@ -152,8 +151,8 @@ public class LAONetworkManager implements MessageSender {
   }
 
   @Override
-  public Connection getConnection() {
-    return connection;
+  public Observable<WebSocket.Event> getConnectEvents() {
+    return connection.observeConnectionEvents();
   }
 
   private void handleBroadcast(Broadcast broadcast) {
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/remote/LAONetworkManager.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/remote/LAONetworkManager.java
@@ -105,7 +105,6 @@ public class LAONetworkManager implements MessageSender {
   @Override
   public Completable catchup(Channel channel) {
     Catchup catchup = new Catchup(channel, requestCounter.incrementAndGet());
-    Log.d(TAG, "sending a catchup to the channel " + channel + " with id " + requestCounter.get());
 
     return request(catchup)
         .map(ResultMessages.class::cast)
@@ -136,12 +135,9 @@ public class LAONetworkManager implements MessageSender {
         // This is used when reconnecting after a lost connection
         .doOnSuccess(answer -> subscribedChannels.add(channel))
         // Catchup already sent messages after the subscription to the channel is complete
-        // || TODO This should be used instead of the two next uncommented lines as it allows the
-        // || returned Completable to complete only when both subscribe and catchup are complete.
-        // || But right now, the LAO creation need this specific behavior
+        // This allows for the completion of the returned completable only when both subscribe
+        // and catchup are completed
         .flatMapCompletable(answer -> catchup(channel));
-    // .doAfterSuccess(answer -> catchup(channel).subscribe())
-    // .ignoreElement();
   }
 
   @Override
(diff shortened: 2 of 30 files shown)


Tesa27 (approved):

fe2-android/app/src/main/AndroidManifest.xml
```diff
@@ -8,17 +8,20 @@
...
     android:label="@string/app_name"
-    android:name="com.github.dedis.popstellar.PoPApplication"
     android:networkSecurityConfig="@xml/network_security_config"
     android:roundIcon="@mipmap/ic_launcher_round"
     android:supportsRtl="true"
     android:theme="@style/AppTheme">
     <activity
-      android:name="com.github.dedis.popstellar.ui.home.HomeActivity"
+      android:name=".ui.home.connecting.ConnectingActivity"
+      android:exported="false" />
```
GabrielFleischer, line 22: I don't think the exported field is needed here.

fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/remote/LAONetworkManager.java
```diff
@@ -106,10 +106,12 @@ private void processIncomingMessages() {
   public Completable catchup(Channel channel) {
     Log.d(TAG, "sending a catchup to the channel " + channel);
     Catchup catchup = new Catchup(channel, requestCounter.incrementAndGet());
+
     return request(catchup)
         .map(ResultMessages.class::cast)
         .map(ResultMessages::getMessages)
         .doOnSuccess(messages -> Log.d(TAG, "Catchup on " + channel + " retrieved : " + messages))
+        .doOnError(msg -> Log.d(TAG, "Error in catchup is " + msg))
```
GabrielFleischer, line 114: Isn't `doOnError()` supposed to provide a `Throwable` ?
And by design, we chose to keep the error handling to the caller. But this is a questionable choice.

fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/remote/LAONetworkManager.java
```diff
@@ -134,12 +136,9 @@ public Completable subscribe(Channel channel) {
...
         // Catchup already sent messages after the subscription to the channel is complete
-        // || TODO This should be used instead of the two next uncommented lines as it allows the
-        // || returned Completable to complete only when both subscribe and catchup are complete.
-        // || But right now, the LAO creation need this specific behavior.
-        // .flatMapCompletable(answer -> catchup(channel))
-        .doAfterSuccess(answer -> catchup(channel).subscribe())
-        .ignoreElement();
+        // This allows for the completion of the returned completable only when both subscribe
+        // and catchup are completed
+        .flatMapCompletable(answer -> catchup(channel));
```
GabrielFleischer, line 141: The reason we used `doAfterSuccess()` is to catchup after the subscribe is finished. Because in the current app implementation, the Lao needs to be created before message are retrieved through the catchup. (which is a very questionable design choice 😅)

This change does fix a problem, but it creates another one that can be very tricky to debug as it introduce concurrency.

fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/remote/MessageSender.java
```diff
@@ -60,4 +60,6 @@ public interface MessageSender extends Disposable {
    * @return a {@link Completable} the will complete once the unsubscribe is finished
    */
   Completable unsubscribe(Channel channel);
+
+  Connection getConnection();
```
GabrielFleischer, line 64: Why do you need this ? We decided to abstract the connection away when implementing this part of the app. Is this really necessary ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/LaoDetailViewModel.java
```diff
@@ -1181,8 +1181,8 @@ public void subscribeToLao(String laoId) {
                 lao -> {
                   Log.d(TAG, "got an update for lao: " + lao.getName());
                   mCurrentLao.postValue(lao);
+                  assert keyManager != null;
```
GabrielFleischer, line 1184: Isn't `keyManager` injected through Hilt ?
When is this assert needed ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeActivity.java
```diff
@@ -316,8 +317,10 @@ private void setupLaunchFragment() {
     setCurrentFragment(R.id.fragment_launch, LaunchFragment::newInstance);
   }
 
-  private void setupConnectingFragment() {
-    setCurrentFragment(R.id.fragment_connecting, ConnectingFragment::newInstance);
+  private void setupConnectingActivity(String laoId) {
+    Intent intent = new Intent(this, ConnectingActivity.class);
+    intent.putExtra("LAO_ID", laoId);
```
GabrielFleischer, line 322: I generally try to avoid "magic values" as much as possible.
Especially when this is used as a key.

Could you create a constant for this ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/connecting/ConnectingActivity.java
```diff
@@ -0,0 +1,111 @@
...
+
+  private ConnectingViewModel mViewModel;
+
+  @Override
+  protected void onCreate(Bundle savedInstanceState) {
+    super.onCreate(savedInstanceState);
+    setContentView(R.layout.connecting_activity);
+    setConnectingFragment();
+
+    String channelId = (String) (getIntent().getExtras()).get("LAO_ID");
```
GabrielFleischer, line 34: Unnecessary parenthesis

```suggestion
    String channelId = (String) getIntent().getExtras().get("LAO_ID");
```

GabrielFleischer, line 34: Unnecessary parenthesis

```suggestion
    String channelId = (String) getIntent().getExtras().get("LAO_ID");
```

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/connecting/ConnectingActivity.java
```diff
@@ -0,0 +1,111 @@
...
+  public static ConnectingViewModel obtainViewModel(FragmentActivity activity) {
+    return new ViewModelProvider(activity).get(ConnectingViewModel.class);
+  }
+
+  private void openLaoDetailActivity(String laoId) {
+    Log.d(TAG, "Trying to open lao detail for lao with id " + laoId);
+
+    Intent intent = new Intent(this, LaoDetailActivity.class);
+    intent.putExtra("LAO_ID", laoId);
+    intent.putExtra("FRAGMENT_TO_OPEN", "LaoDetail");
```
GabrielFleischer, line 77: Here also, it would be better to use constants.

fe2-android/app/src/test/java/com/github/dedis/popstellar/ui/home/ConnectingFragmentTest.java
```diff
@@ -0,0 +1,85 @@
...
+import dagger.hilt.android.testing.HiltAndroidRule;
+import dagger.hilt.android.testing.HiltAndroidTest;
+
+@HiltAndroidTest
+@RunWith(AndroidJUnit4.class)
+public class ConnectingFragmentTest {
+  private static final String laoName = "Lao Name";
+  private static final PublicKey pk = Base64DataUtils.generatePublicKey();
+  private static final Lao lao = new Lao(laoName, pk, 10223421);
+  private static final String laoId = lao.getId();
```
GabrielFleischer, line 41: In Java, we name constants with UPPERCASE and in snake_case.

fe2-android/app/src/test/java/com/github/dedis/popstellar/ui/home/ConnectingFragmentTest.java
```diff
@@ -0,0 +1,85 @@
...
+  public void dynamicTextTest() {
+    laoConnectingText().check(matches(withText(laoId)));
+  }
+
+  @Test
+  public void cancelButtonLaunchesHomeTest() {
+    Intents.init();
+    cancelButton().perform(click());
+    intended(hasComponent(HomeActivity.class.getName()));
+  }
```
GabrielFleischer, line 84: I have not used much of Intents when testing, but I think you should use something along the line of `Intents.release()` at the end of the test.

fe2-android/app/src/test/java/com/github/dedis/popstellar/ui/home/WalletFragmentTest.java
```diff
@@ -41,7 +40,6 @@
 import dagger.hilt.android.testing.HiltAndroidRule;
 import dagger.hilt.android.testing.HiltAndroidTest;
 
-@LargeTest
```
GabrielFleischer, line 44: Why did you remove this ?

GabrielFleischer (changes requested):
Great work fixing this bug ! 👍 
I have some question here and there and there are some unintuitive things in the codebase that you might break.

So I made a lot of comments, don't hesitate to reply if something seems wrong.

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeActivity.java
Warhedgehog: Sure, I lazily followed was already there. I've updated everything

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/LaoDetailViewModel.java
Warhedgehog: I forgot to remove it sorry

fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/remote/MessageSender.java
Warhedgehog: I add a Listener to avoid for the subscribe and catchup process to begin before the change of connection is finished

fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/remote/LAONetworkManager.java
Warhedgehog: My mistake, I misunderstood the purpose of onError

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/connecting/ConnectingActivity.java
Warhedgehog: It does not like it at all if remove them :) 
Alternatively I could do something cleaner like

 `(String) Objects.requireNonNull(getIntent().getExtras()).get(getString(R.string.lao_id)))`

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/LaoDetailActivity.java
```diff
@@ -60,8 +60,11 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
     setContentView(R.layout.lao_detail_activity);
     mViewModel = obtainViewModel(this);
     mViewModel.subscribeToLao(
-        (String) Objects.requireNonNull(getIntent().getExtras()).get("LAO_ID"));
-    if (getIntent().getExtras().get("FRAGMENT_TO_OPEN").equals("LaoDetail")) {
+        (String) Objects.requireNonNull(getIntent().getExtras()).get(getString(R.string.lao_id)));
```
GabrielFleischer, line 63: The `string.xml` should be used for text that will be displayed to the user, and you should keep using constants defined in the code for things like keys and so.

https://stackoverflow.com/questions/16160047/android-strings-xml-vs-static-constants

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/connecting/ConnectingViewModel.java
```diff
@@ -0,0 +1,136 @@
...
+
+    Log.d(TAG, "connecting to lao " + channelId);
+    disposables.add(
+        networkManager
+            .getMessageSender()
+            .subscribe(lao.getChannel())
+            .subscribe(
+                () -> {
+                  Log.d(TAG, "subscribing to LAO with id " + lao.getId());
+                  Log.d(TAG, "got success result for subscribe to lao");
```
GabrielFleischer, line 68: Are the two logs really needed here ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/data/LaoHandler.java
```diff
@@ -41,7 +42,11 @@ public static void handleCreateLao(HandlerContext context, CreateLao createLao)
     Channel channel = context.getChannel();
 
     Log.d(TAG, "handleCreateLao: channel " + channel + ", msg=" + createLao);
-    Lao lao = laoRepository.getLaoByChannel(channel);
+    Lao lao = new Lao(createLao.getId());
+
+    // Adding the newly created LAO to the repository
+    laoRepository.getLaoById().put(lao.getId(), new LAOState(lao));
+    laoRepository.setAllLaoSubject();
```
GabrielFleischer, line 49: Nice fix !

GabrielFleischer (dismissed):
Almost done to me, I just have a small issue with the way you define constants.

Usually, I like to see string.xml as the localizable strings -> what the user will see
And the rest of the constants as usual static final strings.

But this is a big nitpick so I let you fix it if you want and approve for now.

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/Constants.java
```diff
@@ -0,0 +1,16 @@
+package com.github.dedis.popstellar.utility;
+
+public class Constants {
```
GabrielFleischer, line 3: Pure nitpicking, but why creating a whole new class ?
You could have defined it in the fragment classes.

GabrielFleischer (approved):
LGTM

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/Constants.java
Warhedgehog: Because they are extras passed across activities, I thought it was cleaner that way. Plus I thought it might be useful to add some other constants in the future

PR #958, opened 2022-04-26
Add support for the election#request_key and election#key message proposals

Tyratox:
This PR only adds the underlying message handling for the two messages specified in #918 and nothing more. Extending the election#setup message and updating the UI will be done in a separate PR.


fe1-web/src/features/evoting/reducer/ElectionKeyWatcher.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import { getElectionKeyByElectionId } from './ElectionKeyReducer';
+
+/**
+ * Watches the redux store for new message#witness messages for lao#greet messages since they only
+ * become valid after they are witnessed by the corresponding frontend
+ * @remark Implemented analogous to 'makeMessageStoreWatcher'
+ * @param store The redux store to watch
+ * @param laoGreetSignatureHandler The function to call when a signature is added to a lao#greet message
+ */
```
Tyratox, line 13: This comment is wrong here😅 I fixed in in #968:

https://github.com/dedis/popstellar/blob/work-fe1-nico-secret-ballot-election-ui/fe1-web/src/features/evoting/reducer/ElectionKeyWatcher.ts#L7-L12

fe1-web/src/core/objects/Channel.ts
```diff
@@ -53,6 +53,22 @@ export function getReactionChannel(laoIdHash: Hash): Channel {
...
+ * @param channel - The channel whose first component we want to obtain
+ *
+ * @remarks
+ * Using this function is equivalent to making a lot of assumptions about the channel.
+ * This can be brittle and using a more validation-heavy logic may be preferable.
+ */
+export function getFirstPartOfChannel(channel: Channel): Hash {
+  const channels = channel.split('/');
+  return new Hash(channels[2]);
+}
```
pierluca, line 70: Since this is pretty much guaranteed to return a Lao ID, can we just call it `getLaoIdFromChannel` and do some proper error management ? 😅 
I'd rather make the function below (`getLastPartOfChannel`) go away, replaced by something cleaner, than give it a new sibling😅

fe1-web/src/resources/strings.ts
```diff
@@ -141,7 +141,8 @@ const STRINGS = {
   election_voting_method: 'Voting Method',
   election_method_Plurality: 'Plurality',
   election_method_Approval: 'Approval',
-  election_version_identifier: '1.0.0',
+  election_version_open_ballot: 'open-ballot',
+  election_version_secret_ballot: 'secret-ballot',
```
pierluca, line 145: was there a change in the specification ? It'd make sense, I must've missed it 😅

fe1-web/src/features/evoting/reducer/ElectionKeyReducer.ts
```diff
@@ -0,0 +1,110 @@
...
+  byElectionId: {
+    [electionId: string]: string;
+  };
+}
+
+const initialState: ElectionKeyReducerState = {
+  byElectionId: {},
+};
+
+export const ELECTION_KEY_REDUCER_PATH = 'electionKey';
```
pierluca, line 25: why have a separate reducer from the `election` reducer ?

fe1-web/src/features/evoting/reducer/ElectionKeyWatcher.ts
```diff
@@ -0,0 +1,20 @@
...
+ * @param callback The function to call when a the corresponding election#key message is received
+ */
+export const makeElectionKeyStoreWatcher =
+  (electionId: string, store: Store, callback: (electionKey: PublicKey) => void) => () => {
+    const electionKey = getElectionKeyByElectionId(electionId, store.getState());
+
+    if (electionKey) {
+      callback(electionKey);
+    }
+  };
```
pierluca, line 20: I don't see this being used anywhere.
Do you really need this ? 
Wouldn't it be enough to update the election object in its reducer and use that in whichever components you need it ?

fe1-web/src/core/objects/Channel.ts
Tyratox: When updating the previous PR I just realized that there already is a function doing exactly this! Somehow I completely missed it 😅

fe1-web/src/resources/strings.ts
Tyratox: Yes that is one of the changes of #918. Since we use the version field to differentiate between open and secret ballot elections and given there are no other "versions" at the moment I thought we might as well chose names instead of numbers?

https://github.com/dedis/popstellar/blob/work-fe1-evoting-request-election-key/protocol/query/method/message/data/dataSetupElection.json#L13-L16

fe1-web/src/features/evoting/reducer/ElectionKeyReducer.ts
Tyratox: For now there is no election reducer or am I missing something? At the moment, the evoting feature still uses the events feature to store the events. As soon as this is no longer the case, we can merge these reducers, I agree :)

fe1-web/src/features/evoting/reducer/ElectionKeyWatcher.ts
Tyratox: Ah yes, I forgot to remove this! Before the protocol was a bit more complicated and I thought I would need a way to wait for the election#key message being received but this is no longer the case :)

fe1-web/src/resources/strings.ts
pierluca: there might be multiple versions of the secret ballot, but it's no different than any other aspect of the communication protocol in the application. For now, we've just ignored the problem of protocol versioning and assumed that everyone would be running the same version of the software, which is Good Enough (TM) for now 😄 
If and when we'll need to address that, we'll have to use a more global solution anyways.

fe1-web/src/features/evoting/reducer/ElectionKeyReducer.ts
pierluca: Oh, I thought we'd moved away from having the events feature store everything about the events. My bad.

pierluca (approved):

fe1-web/src/resources/strings.ts
Tyratox: Hmm yes that is a good point as well.. I saw that the evoting document mentions semantic versioning 🤔 So maybe use the version field as before and add another type field?

fe1-web/src/features/evoting/reducer/ElectionKeyReducer.ts
Tyratox: Not yet, unfortunately. But I will try to look into this as soon as I merged all of the current PRs :)

pierluca (approved):

PR #959, opened 2022-04-26
Extend election#setup message definition

Tyratox:
The changes are based on the specification in #918


Tyratox:
Not necessary anymore due to changes in the protocol

PR #961, opened 2022-04-26
Fix roll call reopening attendees bug

0xdieman:
Reopening a roll call did not take in account the already scanned attendees during the last opening, this PR fixes that.


Xelowak (approved):
LGTM, thanks for the bug fix!

PR #968, opened 2022-04-28
UI for secret ballot elections

Tyratox:
This PR adds a basic UI for secret ballot election along with the code required to send the corresponding election#setup message.


fe1-web/src/core/components/DropdownSelector.tsx
```diff
@@ -13,26 +13,35 @@ import containerStyles from '../styles/stylesheets/containerStyles';
...
-  const { selected } = props;
-  const { onChange } = props;
-  const { values } = props;
-  const options: any = [];
-  values.forEach((value) => options.push(<Picker.Item key={value} label={value || ''} />));
+const DropdownSelector = (
+  props: Omit<IPropTypes, 'onChange'> & {
+    // make the type for 'onChange' more concrete than the inferred type
+    onChange: (itemValue: string | null, itemIndex: number) => void;
+  },
```
pierluca, line 20: I'm wondering if casting `onChange` wouldn't be easier to read than this.
At the end of the day, neither options _really_ checks what is being passed beyond what `PropTypes` already does.

fe1-web/src/core/network/JsonRpcApi.ts
```diff
@@ -83,6 +84,28 @@ export async function subscribe(
   // propagate the catch() with the full error message, as it needs to be handled on a higher level
 }
 
+/**
+ * Unsubscribe from a channel
+ *
+ * @param channel - The channel from which we want to unsubscribe
+ * @param connections - An optional list of network connection if the message should only be sent on a subset of connections
+ */
+export async function unsubscribe(
```
pierluca, line 93: thanks for finally implementing this. I thought it'd been implemented a long time ago 😅

fe1-web/src/core/network/JsonRpcApi.ts
```diff
@@ -83,6 +84,28 @@ export async function subscribe(
...
+  const request = new JsonRpcRequest({
+    method: JsonRpcMethod.UNSUBSCRIBE,
+    params: new Unsubscribe({
+      channel,
+    }),
+    id: AUTO_ASSIGN_ID,
+  });
+
+  await getNetworkManager().sendPayload(request, connections);
+  // propagate the catch() with the full error message, as it needs to be handled on a higher level
```
pierluca, line 106: it might be worthwhile to update the comment, given the `async/await` pattern usage.
```suggestion
  // do not catch exceptions at this level, as the full error message needs to be handled at a higher level
```

fe1-web/src/features/evoting/components/EventElection.tsx
```diff
@@ -86,77 +100,81 @@ const EventElection = (props: IPropTypes) => {
...
-  };
+  const canCastVote = election.version !== ElectionVersion.SECRET_BALLOT || electionKey;
 
-  return (
-    <>
-      <TimeDisplay start={election.start.valueOf()} end={election.end.valueOf()} />
-      {getElectionDisplay(election.electionStatus)}
-    </>
-  );
+  switch (election.electionStatus) {
```
pierluca, line 105: this switch is getting kinda big. Should we split it in multiple sub-components ?

fe1-web/src/features/evoting/screens/CreateElection.tsx
```diff
@@ -133,34 +157,77 @@ const CreateElection = ({ route }: any) => {
...
+            <View style={[styles.view, { padding: 5 }]}>
+              <ParagraphBlock text={STRINGS.election_create_start_time} />
+              <DatePicker
+                selected={startTime.toDate()}
+                onChange={(date: Date) =>
+                  onChangeStartTime(date, setStartTime, setEndTime, DEFAULT_ELECTION_DURATION)
+                }
+              />
+            </View>
+            <View style={[styles.view, { padding: 5, zIndex: 'initial' }]}>
```
pierluca, line 192: since you're sort-of changing this file, I'd try to address the lint warnings about inline styles 😉

fe1-web/src/features/evoting/screens/CreateElection.tsx
```diff
@@ -23,95 +24,118 @@ import STRINGS from 'resources/strings';
...
 const DEFAULT_ELECTION_DURATION = 3600;
 
 // for now only plurality voting is supported (2022-03-16, Tyratox)
 const VOTING_METHOD = STRINGS.election_method_Plurality;
 
+// the type used for storing questions in the react state
+// does not yet contain the id of the questions, this is computed
+// only on creation of the election
+// ALSO: for now the write_in feature is disabled (2022-03-16, Tyratox)
+type NewQuestion = Omit<Question, 'id' | 'write_in'>;
```
pierluca, line 38: if the `write_in` is disabled, shouldn't we remove it altogether from `Question` for the time being ? 
It might be more straightforward than having this kind of annotations here and there in the codebase.

fe1-web/src/features/evoting/components/EventElection.tsx
Tyratox: I agree. Subcomponents are an option but will introduce a lot of redundant code since many parameters need to be passed to all of them but I it is also the best option I can think of at the moment

fe1-web/src/features/evoting/screens/CreateElection.tsx
Tyratox: Yes I agree but this would require the protocol to change as well and I thought for the moment it might be easier to leave the field in the protocol and just fix its value to `false`?

0xdieman (approved):
LGTM, as usual

fe1-web/src/core/components/DropdownSelector.tsx
pierluca: That seems an oddly IDE-specific behaviour and I'd take code clarity over IDE happiness 😅
Which IDE are you using ? Are you redefining a new `const` variable taking the cast property ?

fe1-web/src/features/evoting/screens/CreateElection.tsx
pierluca: good point, I hadn't thought about the protocol implications.

fe1-web/src/core/components/DropdownSelector.tsx
```diff
@@ -13,26 +15,38 @@ import containerStyles from '../styles/stylesheets/containerStyles';
...
-  const { onChange } = props;
-  const { values } = props;
-  const options: any = [];
-  values.forEach((value) => options.push(<Picker.Item key={value} label={value || ''} />));
+const DropdownSelector = (
+  props: ExtendType<
+    IPropTypes,
+    {
+      // make the type for 'onChange' more concrete than the inferred type
+      onChange: (itemValue: string | null, itemIndex: number) => void;
```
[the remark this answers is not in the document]
pierluca: I would love that too, but unfortunately, prop-types are a runtime construct in React.
They have no notion of TypeScript types because that only exists at compile-time, in JavaScript.
JavaScript is not strongly typed, so it'd be impossible to extract the right types from the function, even if it were possible to indicate to prop-types its expected parameters and return value.

fe1-web/src/features/evoting/components/ElectionOpened.tsx
```diff
@@ -0,0 +1,102 @@
...
+import { useToast } from 'react-native-toast-notifications';
+
+import { CheckboxList, TimeDisplay, WideButtonView } from 'core/components';
+import { FOUR_SECONDS } from 'resources/const';
+import STRINGS from 'resources/strings';
+
+import { castVote, terminateElection } from '../network/ElectionMessageApi';
+import { Election, SelectedBallots } from '../objects';
+
+const ElectionOpened = ({ election, questions, isOrganizer, canCastVote }: IPropTypes) => {
```
pierluca, line 14: `canCastVote` seems to be only a concern to this specific component.
Couldn't we compute it internally to this component ? 
Ideally, it'd be great to be able to do `election.canCastVote()`, but I understand this still requires a refactoring and is probably beyond the scope of this PR.

fe1-web/src/features/evoting/network/ElectionMessageApi.ts
```diff
@@ -56,62 +56,55 @@ export async function requestCreateElection(
 /**
  * Sends a query to open an election.
  *
- * @param laoId - The id of the lao in which the given election should be opened
  * @param election - The election that should be opened
  */
-export function openElection(laoId: Hash, election: Election): Promise<void> {
+export function openElection(election: Election): Promise<void> {
```
pierluca, line 61: nice changes, so much cleaner !

fe1-web/src/features/evoting/screens/CreateElection.tsx
```diff
@@ -109,7 +114,7 @@ const createElection = (
  */
 
 const CreateElection = ({ route }: any) => {
-  const styles = route.params;
+  const routeStyles = route.params;
```
pierluca, line 117: isn't it a bit weird here that we get the styles from the route params? 🤔 
why is that ?

fe1-web/src/features/evoting/components/EventElection.tsx
```diff
@@ -1,176 +1,52 @@
 import PropTypes from 'prop-types';
-import React, { FunctionComponent, useMemo, useState } from 'react';
-import { SectionList, StyleSheet, Text, TextStyle } from 'react-native';
-import { Badge } from 'react-native-elements';
-import { useToast } from 'react-native-toast-notifications';
+import React, { FunctionComponent, useMemo } from 'react';
```
pierluca, line 2: so much cleaner, this whole file !

pierluca (commented):
So much cleaner, great job ! 
Just a few minor remarks 😉

fe1-web/src/core/network/CommunicationApi.ts
```diff
@@ -1,25 +1,38 @@
...
+ * Subscribes to a channel and optionally catches up to previously sent messages
+ * @param channel The channel to subscribe to
+ * @param connections An optional list of network connection if the message should only be sent on a subset of connections
+ * @param sendChatchup Whether to send a catchup message after subscribing to the channel. Default value is 'true'
+ * @returns A promise to wait on the subscription (and the optional catchup)
+ */
+export async function subscribeToChannel(
+  channel: Channel,
+  connections?: NetworkConnection[],
+  sendChatchup: boolean = true,
```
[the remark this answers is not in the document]
Tyratox: Uh, thank you!

fe1-web/src/features/evoting/components/ElectionOpened.tsx
Tyratox: Haha yes, I noticed that too and changed it in https://github.com/dedis/popstellar/pull/981! :) I will quickly port that change to this PR 👍

fe1-web/src/features/evoting/screens/CreateElection.tsx
Tyratox: It is very weird indeed! I have not yet figured out where these styles come from.. The same code is in CreateMeeting and CreateRollCall and the styles come from here:

https://github.com/dedis/popstellar/blob/master/fe1-web/src/features/events/screens/CreateEvent.tsx#L16-L27

I will look into changing this in a separate PR :)

fe1-web/src/core/components/DropdownSelector.tsx
Tyratox: I mean I don't think any IDE should be able to find out the correct type? The PropTypes just specify a generic function?

Tyratox: What I meant is that the IDE does not know the type when doing

```tsx
<DropdownSelector onChange={...} />
```

Within the function it is fine using a cast

fe1-web/src/features/evoting/components/ElectionOpened.tsx
Tyratox: `election.canCastVote()` would be cool indeed! :) I can to this in the same PR or in a separate one, whatever you prefer :) Maybe it is easiest if I do it in the PR for encrypting the votes?

pierluca (approved):
I could be annoying with `ElectionOpened` and ask why we need the electionKey as a parameter, rather than just retrieving it in this component, much like the `canCastVote`, as it doesn't seem to be used anywhere else 😅 

Overalls though, LGTM !

Tyratox:
Oh true! 😂 I will do that in the encrypted votes PR 😄

PR #969, opened 2022-04-28
Fix bug in witnessing feature

Tyratox:
Previously the witnessing feature tried to call `ExtendedMessage.fromState` without being connected to a LAO when fe-1 was refreshed (since the watcher thinks all stored messages are new messages). Since certain message constructors such as CreateRollCall throw an error if the application is not connected to a LAO, this causes the application to error and fail.

This PR fixes this issue by delaying all actions triggered by the watcher until the application is connected to a LAO.

### DIFF
--- a/fe1-web/src/features/witness/interface/Configuration.ts
+++ b/fe1-web/src/features/witness/interface/Configuration.ts
@@ -24,7 +24,7 @@ export interface WitnessConfiguration {
 
   /**
    * Returns the currently active lao id. Should be used outside react components
-   * @returns The current lao  or undefined if there is none.
+   * @returns The current lao or undefined if there is none.
    */
   getCurrentLaoId: () => Hash | undefined;
 
--- a/fe1-web/src/features/witness/interface/Configuration.ts
+++ b/fe1-web/src/features/witness/interface/Configuration.ts
@@ -1,4 +1,5 @@
 import { MessageRegistry } from 'core/network/jsonrpc/messages';
+import { Hash } from 'core/objects';
 import FeatureInterface from 'core/objects/FeatureInterface';
 
 import { WitnessFeature } from './Feature';
@@ -20,6 +21,12 @@ export interface WitnessConfiguration {
    * @returns The current lao
    */
   getCurrentLao: () => WitnessFeature.Lao;
+
+  /**
+   * Returns the currently active lao id. Should be used outside react components
+   * @returns The current lao  or undefined if there is none.
+   */
+  getCurrentLaoId: () => Hash | undefined;
 
   /**
    * Returns whether the user is witness of the current lao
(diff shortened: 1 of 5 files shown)


fe1-web/src/features/witness/interface/Configuration.ts
```diff
@@ -21,6 +22,12 @@ export interface WitnessConfiguration {
    */
   getCurrentLao: () => WitnessFeature.Lao;
 
+  /**
+   * Returns the currently active lao id. Should be used outside react components
+   * @returns The current lao  or undefined if there is none.
```
pierluca, line 27:
```suggestion
   * @returns The current lao or undefined if there is none.
```

fe1-web/src/features/witness/network/WitnessStoreWatcher.ts
```diff
@@ -61,7 +62,16 @@ export const makeWitnessStoreWatcher = (
   let currentAllIds: string[] = [];
   let currentUnprocessedIds: string[] = [];
   return () => {
+    // we have to be careful with ExtendedMessage.fromState
+    // since some message constructors assume that we are connected to a lao
+    // thus we delay this watcher until we are connected to a lao
+    // (sending witness messages would also be difficult under these circumstances)
+    if (!getCurrentLaoId()) {
+      return;
+    }
```
pierluca, line 71: this seems like a viable fix for the time being, but in the long run I'm moderately concerned by the constructors assuming we're connected to a LAO. What if we're connected to multiple LAOs in the future ? 
I believe the ExtendedMessage now contains the associated LAO, so do we still need to rely on the currently connected LAO, to process messages ?

Tyratox: Yes I agree! The problem specifically are messages like `CreateRollCall` where the lao Id is used to verify the hash:

https://github.com/dedis/popstellar/blob/master/fe1-web/src/features/rollCall/network/messages/CreateRollCall.ts#L77-L90

The same is the case in `CloseRollCall` and `ReopenRollCall`. It might make sense to do these checks in the `RollCallHandler` instead where the messages are received? Then we could take the lao id from the channel and verify the hash before creating the instance.

In general it seems as if a lot of message handlers use the current lao instead of taking the lao id from the channel or the message. This will have to change as well as soon as you can be connected to multiple LAOs at the same time.

pierluca: What you suggest make sense, but we lose the ability to verify that the objects we build ourselves are built correctly.
How about we make the lao id an optional parameter, both to the constructor and `fromJson`, which gets passed from the ExtendedMessage. In the absence of it, we assume it's the current LAO and fetch it.
This should nicely cover both the current usage and future usages, and it's probably intuitive enough, what do you think ?

Tyratox: True! Yes I think your suggestion seems like a good solution to this :)

pierluca (approved):
Approved, with the understanding that a _separate_ PR might provide a better fix 😅

PR #972, opened 2022-04-29
[BE2] compute and validate `transaction_id`

lourkeur:
The current specification is: communication protocol style hash (length-prefixed thing) of string representations of all relevant fields in some fixed order.

- [x] add a bad transaction id testcase


be2-scala/src/main/scala/ch/epfl/pop/model/objects/Transaction.scala
```diff
@@ -30,7 +30,32 @@ final case class Transaction(
   Version: Int,
   TxIn: List[TxIn],
   TxOut: List[TxOut],
-  )
+  LockTime: Int,
+) {
+  lazy val transaction_id = {
+    import scala.util.chaining._
```
tuomas-p, line 36: I would probably still leave the imports at the top of the file for clarity unless there is a very specific reason for this

tuomas-p (approved):
Looks good to me, no major remarks

be2-scala/src/main/scala/ch/epfl/pop/model/objects/Transaction.scala
```diff
@@ -27,10 +27,30 @@ final case class TxIn(
...
-  TxIn: List[TxIn],
-  TxOut: List[TxOut],
-  )
+  version: Int,
+  inputs: List[TxIn],
+  outputs: List[TxOut],
+  lockTime: Int,
+) {
+  lazy val transactionId = {
+    val strings = collection.mutable.ListBuffer.empty[String]
```
RaulinN, line 36: Interesting use of buffer, wouldn't have thought of using them :) ! I don't know what's the typical size of `inputs`, but remember that the list buffer will only be more efficient than a list if the collection is "large"

RaulinN (approved):
Overall, really clean code! Good job! 👍 Not much else to say :)

PR #950, opened 2022-04-13
Be2-Scala CreateRollCall 

rombirli:

PR #977, opened 2022-04-30
Be2-Scala Election : ResultElection, ElectionValidator, Tests & Examples

rombirli:


be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/ElectionValidator.scala
```diff
@@ -113,26 +121,95 @@ sealed class ElectionValidator(dbActorRef: => AskableActorRef) extends MessageDa
...
           Left(rpcMessage)
         }
-
       case _ => Right(validationErrorNoMessage(rpcMessage.id))
     }
   }
 
+  private def getOpenMessage(electionChannel: Channel): Option[OpenElection] = {
+    var result: Option[OpenElection] = None
+    Await.ready(dbActor ? DbActor.ReadChannelData(electionChannel), duration).value match {
```
lourkeur, line 171: Might have been mentioned at the last meeting but this should probably be a `flatMap`.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/ElectionValidator.scala
```diff
@@ -168,23 +245,88 @@ sealed class ElectionValidator(dbActorRef: => AskableActorRef) extends MessageDa
...
           Right(validationError("unexpected lao id"))
-        } else if (!validateOwner(sender, channel, dbActorRef)) {
+        } else if (!validateOwner(sender, channel, dbActor)) {
           Right(validationError(s"invalid sender $sender"))
-        } else if (!validateChannelType(ObjectType.ELECTION, channel, dbActorRef)) {
+        } else if (!validateChannelType(ObjectType.ELECTION, channel, dbActor)) {
           Right(validationError(s"trying to send a EndElection message on a wrong type of channel $channel"))
+          // TODO : check it - is this check ok ?
+          /*} else if (!compareResults(getLastVotes(channel, dbActor).flatMap(_.votes.map(_.id)), data.registered_votes)) {
+            Right(validationError(s"Incorrect verification hash"))*/
```
lourkeur, line 262: needs to be addressed

be2-scala/src/test/scala/ch/epfl/pop/json/MessageDataProtocolSuite.scala
```diff
@@ -77,17 +77,17 @@ class MessageDataProtocolSuite extends FunSuite with Matchers {
...
-    val questionResult = ElectionQuestionResult(Hash(Base64Data("2PLwVvqxMqW5hQJXkFpNCvBI9MZwuN8rf66V1hS-iZU=")), result1 :: result2 :: Nil)
-    val expected = ResultElection(questionResult :: Nil, Nil)
+    val lao = Hash(Base64Data("fzJSZjKf-2cbXH7kds9H8NORuuFIRLkevJlN7qQemjo="))
+    val election = Hash(Base64Data("zG1olgFZwA0m3mLyUqeOqrG0MbjtfqShkyZ6hlyx1tg="))
+    val opened_at = Timestamp(1633099883L)
+    val expected = OpenElection(lao, election, opened_at)
 
-    messageData shouldBe a[ResultElection]
+    messageData shouldBe a[OpenElection]
     messageData should equal(expected)
```
lourkeur, line 91: Can we keep both tests?

lourkeur (changes requested):
What is the relationship between this and #950?

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandler.scala
```diff
@@ -71,15 +74,84 @@ class ElectionHandler(dbRef: => AskableActorRef) extends MessageHandler {
```
Ajkunas, line 74: I tried to test the back end with the front end 1: there is a NullPointerException when we try to fetch the election#setup message from the election channel, because the message is only in the lao channel, so I suggest this change: 
In the handleSetupElection function, to add this at line 54: 
`_ <- dbActor ? DbActor.WriteAndPropagate(electionChannel, message)` 
I tried it and it should work.

be2-scala/src/test/scala/ch/epfl/pop/json/MessageDataProtocolSuite.scala
Ajkunas: Yep we could have kept both tests

lourkeur (changes requested):
To be discussed, maybe at meeting:
- `Await` cascades are probably an probably an anti-pattern
- Pointless diffs, like the ones that change `.getOrElse(` into ` getOrElse ` make reviewing harder so if we could avoid those in the future.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandler.scala
```diff
@@ -71,15 +75,83 @@ class ElectionHandler(dbRef: => AskableActorRef) extends MessageHandler {
...
+      val ballots = question2Ballots(question)
+      val ballot = voteElection.vote match {
+        case Some(List(index)) => ballots.toArray.apply(index)
+        case _ => "<NOT_VOTED>"
+      }
+      println(f"Your ballot is $ballot !")
+      if (results.contains(question)) {
+        val questionresult = results(question)
+        if (questionresult.contains(ballot))
+          results.update(question, questionresult.updated(ballot, questionresult(ballot) + 1))
```
tuomas-p, line 144: I would suggest getOrElse with the default value 0, to avoid an additional if

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandler.scala
```diff
@@ -71,15 +75,83 @@ class ElectionHandler(dbRef: => AskableActorRef) extends MessageHandler {
...
+        else
+          results.update(question, questionresult.updated(ballot, 1))
+      }
+      else
+        results.update(question, Map(ballot -> 1))
+    }
+    println(results)
```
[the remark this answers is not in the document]
tuomas-p: Is it normal that it didn't seem to go away?

tuomas-p (changes requested):
Good work addition-wise, a few remarks concerning the code, as well as some of @lourkeur 's comments that seem to need to be addressed

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
```diff
@@ -40,8 +40,35 @@ class RollCallHandler(dbRef: => AskableActorRef) extends MessageHandler {
...
+        // we are using the rollcall id instead of the message_id at rollcall creation
+        val rollCallChannel: Channel = Channel(s"${Channel.ROOT_CHANNEL_PREFIX}${data.id}")
+        val ask =
+          for (
+            _ <- dbActor ? DbActor.ChannelExists(rollCallChannel) transformWith {
+              case Success(_) => Future {
+                throw DbActorNAckException(ErrorCodes.INVALID_ACTION.id, "rollCall already exists in db")
+              }
+              case _ => Future {
+                ()
```
lourkeur, line 55:
```suggestion
```
For stylistic points: an empty `{}` block is implicitly `Unit`

be2-scala/src/test/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandlerTest.scala
```diff
@@ -123,4 +203,55 @@ class ElectionHandlerTest extends TestKit(ActorSystem("Election-DB-System")) wit
...
+
+  /*test("EndElection should fail if the database fails storing the message") {
+    val mockedDB = mockDbWithNAckEndElection
+    val rc = new ElectionHandler(mockedDB)
+    val request = EndElectionMessages.endElection
+
+    rc.handleEndElection(request) shouldBe an[Right[PipelineError, _]]
+
+    system.stop(mockedDB.actorRef)
+  }*/
```
lourkeur, line 256: Should do something with that commented-out code

be2-scala/src/main/scala/ch/epfl/pop/model/objects/ElectionChannel.scala
```diff
@@ -0,0 +1,64 @@
...
+import ch.epfl.pop.pubsub.AskPatternConstants
+import ch.epfl.pop.storage.DbActor
+import ch.epfl.pop.storage.DbActor.{DbActorReadAck, DbActorReadChannelDataAck}
+
+import scala.collection.mutable
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.{Await, Future}
+
+object ElectionChannel {
+  implicit class ElectionChannel(channel: Channel) extends AskPatternConstants {
```
lourkeur, line 15:
```suggestion
  implicit class ElectionChannelExtensionMethods(channel: Channel) extends AskPatternConstants {
```
Since this is not really a channel, it might be better to give it this kind of naming convention, in line with https://www.scala-lang.org/api/2.13.8/scala/collection/IterableOnceExtensionMethods.html. That should prevent confusion.

be2-scala/src/main/scala/ch/epfl/pop/model/objects/ElectionChannel.scala
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     *
+     * @param dbActor AskableActorRef
+     * @tparam T type of data to extract from the election channel
+     * @return Future of a list of tuple containing the message and the data extracted
+     */
+    def extractMessages[T: Manifest](dbActor: AskableActorRef = DbActor.getInstance): Future[List[(Message, T)]] = {
```
lourkeur, line 23:
```suggestion
    def extractMessages[T <: MessageData: Manifest](dbActor: AskableActorRef = DbActor.getInstance): Future[List[(Message, T)]] = {
```

Optional, but I noticed the typing system inferred `T with MessageData` during the session so might as well.
(need to import `MessageData`)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandler.scala
```diff
@@ -59,27 +62,80 @@ class ElectionHandler(dbRef: => AskableActorRef) extends MessageHandler {
...
+              val questionResult = resultsTable(question)
+              resultsTable.update(question, questionResult.updated(ballot, questionResult(ballot) + 1))
+            case _ =>
+          }
+        } {}
+        Future(Success(()))
+      }
+      }
+      results = resultsTable.map(tuple => {
+        val (qid, ballotToCount) = tuple
```
lourkeur, line 134: In 52b4cbd, I pointed out that there's an elegant way to avoid binding a name there. You can also write the lambda as `{ case (qid, ballotToCount) =>`, that works too

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandler.scala
```diff
@@ -59,27 +62,80 @@ class ElectionHandler(dbRef: => AskableActorRef) extends MessageHandler {
...
+        // build ElectionBallotVotes objects
+        val electionBallotVotes = ballotToCount.map(tuple => ElectionBallotVotes(tuple._1, tuple._2))
```
lourkeur, line 136: same idea here

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/ElectionValidator.scala
```diff
@@ -101,6 +104,18 @@ sealed class ElectionValidator(dbActorRef: => AskableActorRef) extends MessageDa
...
+    votes.forall(
+      voteElection => {
+        val vote = voteElection.vote match {
+          case Some(v :: _) => v
+          case _ => -1
+        }
+        // check if the ballot is available
+        vote < q2ballots(voteElection.question).size
+      }
+    )
```
lourkeur, line 117:
```suggestion
    votes.forall { voteElection =>
        val vote = voteElection.vote match {
          case Some(v :: _) => v
          case _ => -1
        }
        // check if the ballot is available
        vote < q2ballots(voteElection.question).size
      }
```
Usually, this is the more conventional lambda syntax.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
```diff
@@ -40,8 +40,35 @@ class RollCallHandler(dbRef: => AskableActorRef) extends MessageHandler {
...
   def handleCreateRollCall(rpcRequest: JsonRpcRequest): GraphMessage = {
-    val ask: Future[GraphMessage] = dbAskWritePropagate(rpcRequest)
-    Await.result(ask, duration)
+    rpcRequest.getParamsMessage match {
+      case Some(message: Message) =>
+        val data: CreateRollCall = message.decodedData.get.asInstanceOf[CreateRollCall]
+        // we are using the rollcall id instead of the message_id at rollcall creation
+        val rollCallChannel: Channel = Channel(s"${Channel.ROOT_CHANNEL_PREFIX}${data.id}")
+        val ask =
+          for (
```
lourkeur, line 49:
```suggestion
          for {
```
especially for a `for yield`, this saves you the semicolons.

be2-scala/src/main/scala/ch/epfl/pop/model/objects/ElectionChannel.scala
```diff
@@ -0,0 +1,64 @@
...
+      extractMessages[CastVoteElection](dbActor).map(votes => {
+        val attendeeIdToLastVote = mutable.Map[PublicKey, (Message, CastVoteElection)]()
+        for ((message, castvote) <- votes) {
+          if (!attendeeIdToLastVote.contains(message.sender) ||
+            attendeeIdToLastVote(message.sender)._2.created_at < castvote.created_at) {
+            attendeeIdToLastVote.update(message.sender, (message, castvote))
+          }
+        }
+        attendeeIdToLastVote.values.toList
+      })
```
lourkeur, line 61:
```suggestion
      extractMessages[CastVoteElection](dbActor).map { votes =>
        val attendeeIdToLastVote = mutable.Map[PublicKey, (Message, CastVoteElection)]()
        for ((message, castvote) <- votes) {
          if (!attendeeIdToLastVote.contains(message.sender) ||
            attendeeIdToLastVote(message.sender)._2.created_at < castvote.created_at) {
            attendeeIdToLastVote.update(message.sender, (message, castvote))
          }
        }
        attendeeIdToLastVote.values.toList
      }
```
This is the most common syntax for a lambda like that.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandler.scala
```diff
@@ -1,17 +1,20 @@
...
+import ch.epfl.pop.model.network.method.message.data.election._
+import ch.epfl.pop.model.objects._
 import ch.epfl.pop.pubsub.graph.{ErrorCodes, GraphMessage, PipelineError}
 import ch.epfl.pop.storage.DbActor
 
+import scala.collection.mutable
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.{Await, Future}
 import scala.util.{Failure, Success}
+import ch.epfl.pop.model.objects.ElectionChannel.ElectionChannel
```
lourkeur, line 17:
```suggestion
import ch.epfl.pop.model.objects.ElectionChannel._
```
usually you would do that in Scala 2 because the names of implicits are not really relevant. Rather their types are.

lourkeur (changes requested):

RaulinN (commented):
Last few small comments before merge :)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/ElectionHelper.scala
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  /**
+   * Read every castvote in the channel and keep the last per attendee
+   *
+   * @param electionChannel : the election channel
+   * @param dbActor         the AskableActorRef we use (by default the main DbActor, obtained through getInstance)
+   * @return the final vote for each attendee that has voted
+   */
+  def getLastVotes(electionChannel: Channel, dbActor: AskableActorRef = DbActor.getInstance): List[(Message, CastVoteElection)] = {
```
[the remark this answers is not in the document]
Ajkunas: We need the Message part in our function `compareResults` in the class `ElectionValidator`, because we need to sort the votes by message ids.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandler.scala
```diff
@@ -48,6 +51,7 @@ class ElectionHandler(dbRef: => AskableActorRef) extends MessageHandler {
     val combined = for {
       _ <- dbActor ? DbActor.WriteAndPropagate(rpcMessage.getParamsChannel, message)
       _ <- dbActor ? DbActor.CreateChannel(electionChannel, ObjectType.ELECTION)
+      _ <- dbActor ? DbActor.WriteAndPropagate(electionChannel, message)
     } yield ()
```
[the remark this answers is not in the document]
Ajkunas: This was addressed on slack, since both backends intially did not have the setup message written in the channel election, and it was a problem when we wanted to retrieve the setup message. Pierluca affirmed that the setup message should also be included in both the lao and election channel.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandler.scala
```diff
@@ -71,15 +75,83 @@ class ElectionHandler(dbRef: => AskableActorRef) extends MessageHandler {
 
   def handleCastVoteElection(rpcMessage: JsonRpcRequest): GraphMessage = {
     // no need to propagate here, hence the use of dbAskWrite
-    val ask: Future[GraphMessage] = dbAskWrite(rpcMessage)
+    val ask: Future[GraphMessage] = dbAskWritePropagate(rpcMessage)
```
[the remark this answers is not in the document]
Ajkunas: Yes. We need to send it back to the frontend so that they can compute `registered_votes` of the EndElection message. We already discussed that with the evoting team, so yes we need to propagate the message.

lourkeur (approved):
all good on my side

tuomas-p (approved):

PR #979, opened 2022-05-02
lao create self contained tests

Tesa27:
Karate is not a straightforward framework and gets some time to get used to, in order for developers that do not want to understand the functionality of the karate tests we decide to create tests that are somewhat similar to human readable language, so that anyone who has the basic understanding of the PoP protocol can read and even write a karate test more easily. This wasy it is more developer and reader friendly. 

This PR addresses the first step towards this easy reading tests and addresses the basic building blocks for these kind of tests and transforms the already existent LAO create karate tests into tests that respect this paradigm we want to achieve.

### DIFF
--- a/tests/karate/src/test/java/be/LAO/create.feature
+++ b/tests/karate/src/test/java/be/LAO/create.feature
@@ -9,7 +9,6 @@ Feature: Create a pop LAO
     * call read('classpath:be/utils/server.feature')
     * call read('classpath:be/mockFrontEnd.feature')
     * call read('classpath:be/constants.feature')
-    * def id = 1
     * string channel = "/root"
 
   Scenario: Create Lao request with empty lao name should fail with an error response 2
@@ -25,7 +24,7 @@ Feature: Create a pop LAO
           "witnesses": []
         }
       """
-    When frontend.publish(JSON.stringify(badLaoReq), id, channel)
+    When frontend.publish(JSON.stringify(badLaoReq), channel)
     And json answer = frontend.getBackendResponseWithoutBroadcast()
     Then match answer contains INVALID_MESSAGE_FIELD
     And match frontend.receiveNoMoreResponses() == true
@@ -44,7 +43,7 @@ Feature: Create a pop LAO
           "witnesses": []
         }
       """
-    When frontend.publish(JSON.stringify(badLaoReq), id, channel)
+    When frontend.publish(JSON.stringify(badLaoReq), channel)
     And json answer = frontend.getBackendResponseWithoutBroadcast()
     Then match answer contains INVALID_MESSAGE_FIELD
     And match frontend.receiveNoMoreResponses() == true
@@ -61,7 +60,7 @@ Feature: Create a pop LAO
           "witnesses": []
         }
       """
-    When frontend.publish(JSON.stringify(badLaoReq), id, channel)
+    When frontend.publish(JSON.stringify(badLaoReq), channel)
     And json answer = frontend.getBackendResponseWithoutBroadcast()
     Then match answer contains INVALID_MESSAGE_FIELD
     And match frontend.receiveNoMoreResponses() == true
@@ -78,7 +77,7 @@ Feature: Create a pop LAO
           "witnesses": []
         }
       """
-    When frontend.publish(JSON.stringify(laoCreateRequest), id, channel)
+    When frontend.publish(JSON.stringify(laoCreateRequest), channel)
     And json answer = frontend.getBackendResponseWithoutBroadcast()
     Then match answer contains VALID_MESSAGE
     And match frontend.receiveNoMoreResponses() == true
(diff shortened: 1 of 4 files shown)


tests/karate/src/test/java/be/LAO/create.feature
```diff
@@ -7,47 +7,78 @@ Feature: Create a pop LAO
...
+          "name": "",
+          "creation": 1633098234,
+          "organizer": '#(getOrganizer)',
+          "witnesses": []
+        }
+      """
+    When frontend.publish(JSON.stringify(badLaoReq), id, channel)
+    And json answer = frontend.getBackendResponseWithoutBroadcast()
+    Then match answer contains INVALID_MESSAGE_FIELD
+    And match frontend.receiveNoMoreResponses() == true
```
GabrielFleischer, line 31: I wonder if the `== true` is necessary.
As well as the `id` argument.

GabrielFleischer (changes requested):
The  java code could benefit from a bit more documentation.
And I have 2 questions on the way you wrote the tests.

But other than that, it is going in the right direction.

tests/karate/src/test/java/be/LAO/create.feature
Tesa27: It does not seem to work without `==true` though

tests/karate/src/test/java/common/net/MultiMsgWebSocketClient.java
```diff
@@ -37,4 +41,27 @@ public synchronized Object listen(long timeout) {
...
+  }
+
+  public String getBackendResponseWithoutBroadcast(){
+    String result = getBuffer().takeTimeout(5000);
+    return result;
+  }
+
+  public boolean receiveNoMoreResponses(){
+    String result_dummy = getBuffer().takeTimeout(5000);
+    String result = getBuffer().takeTimeout(5000);
```
pierluca, line 64: why do this twice ?

tests/karate/src/test/java/common/net/MultiMsgWebSocketClient.java
```diff
@@ -37,4 +41,27 @@ public synchronized Object listen(long timeout) {
...
+
+  public void publish(String data, int id, String channel){
+    JsonConverter jsonConverter = new JsonConverter();
+    Json request =  jsonConverter.publishМessageFromData(data, id, channel);
+    this.send(request.toString());
+  }
+
+  public String getBackendResponseWithBroadcast(String a){
+    String broadcast = getBuffer().takeTimeout(5000);
+    String result = getBuffer().takeTimeout(5000);
```
pierluca, line 53: how do you know which one is the broadcast and which one is the result ?
I don't think the order was specified in the protocol, and it's probably fine to leave it undefined.

tests/karate/src/test/java/be/constants.feature
```diff
@@ -0,0 +1,36 @@
...
+        function(){
+          return "J9fBzJV70Jk5c-i3277Uq4CmeL4t53WDfUghaK0HpeM="
+        }
+      """
+    * def getLaoIdEmptyName = call createLaoIdEmptyName
+    * def getLaoIdNegativeTime = call createLaoIdNegativeTime
+    * def getLaoValid = call createLaoValid
+    * def getOrganizer = call organizerPk
+    * def INTERNAL_SERVER_ERROR = {error: {code: -6, description: '#string'}}
+    * def INVALID_MESSAGE_FIELD = {error: {code: -4, description: '#string'}}
```
pierluca, line 35: while you're at it, for completeness, you might want to list them all out.

tests/karate/src/test/java/be/constants.feature
```diff
@@ -0,0 +1,36 @@
+@ignore @report=false
+Feature: Constants
+  Scenario: Creates constants that will be used by other features
+    # TODO: make the function depend on all the attributes the lao id depends on
```
pierluca, line 4: I'm thinking it might make sense to have a generic `HashLen` function in Java, and then maybe a `fillValidLaoId` / `fillValidElectionId` / etc. functions in JS

tests/karate/src/test/java/common/net/MultiMsgWebSocketClient.java
```diff
@@ -37,4 +41,27 @@ public synchronized Object listen(long timeout) {
...
   }
+
+  public void publish(String data, int id, String channel){
+    JsonConverter jsonConverter = new JsonConverter();
+    Json request =  jsonConverter.publishМessageFromData(data, id, channel);
+    this.send(request.toString());
+  }
+
+  public String getBackendResponseWithBroadcast(String a){
+    String broadcast = getBuffer().takeTimeout(5000);
```
pierluca, line 52: what do you do with the broadcast? how do you assert that you had the right result ?

pierluca (commented):
Much nicer ! 
Still a few kinks to be ironed out, but we're getting there.
By the way, the `JsonConverter` class really deserves a better name 😅

GabrielFleischer (approved):
LGTM

PR #980, opened 2022-05-02
Fix Crash on LAO launch with BE1

Warhedgehog:
Solves a bug I introduced with #955. The current implementation starts the LAO with some of the the attributes being null. This fixes it by only launching when the catch up is received.

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/LaoDetailViewModel.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/LaoDetailViewModel.java
@@ -1181,7 +1181,7 @@ public class LaoDetailViewModel extends AndroidViewModel
                 lao -> {
                   Log.d(TAG, "got an update for lao: " + lao.getName());
                   mCurrentLao.postValue(lao);
-                  boolean isOrganizer = lao.getOrganizer().equals(keyManager.getMainPublicKey());
+                  boolean isOrganizer = keyManager.getMainPublicKey().equals(lao.getOrganizer());
                   mIsOrganizer.setValue(isOrganizer);
                 }));
   }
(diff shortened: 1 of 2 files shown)


fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/LaoDetailViewModel.java
```diff
@@ -1181,7 +1181,7 @@ public void subscribeToLao(String laoId) {
                 lao -> {
                   Log.d(TAG, "got an update for lao: " + lao.getName());
                   mCurrentLao.postValue(lao);
-                  boolean isOrganizer = lao.getOrganizer().equals(keyManager.getMainPublicKey());
+                  boolean isOrganizer = keyManager.getMainPublicKey().equals(lao.getOrganizer());
```
GabrielFleischer, line 1184: What does this affect ? lao.getOrganizer() should not be null anyway.

GabrielFleischer (approved):
I was certain you had done this in the last PR 😅 

Good Job

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/LaoDetailViewModel.java
Warhedgehog: Well this was the reason it crashed instead of gracefully, just waiting for the catchup to complete. Now it cannot be a problem anymore so it's not needed.

Tesa27 (approved):
LGTM

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/LaoDetailViewModel.java
GabrielFleischer: Oh ! I see. Ok

GabrielFleischer (approved):
LGTM

PR #987, opened 2022-05-05
Self contained tests for roll call

Tesa27:
This PR addresses self contained tests for already existent Roll Call karate tests, makes them more readable and enhances the building blocks addressed in PR [#979](https://github.com/dedis/popstellar/pull/979)


GabrielFleischer (changes requested):
This is again a big step forward.

I just have some small questions and requests, I let you go through them.

tests/karate/src/test/java/be/rollCall/closeRollCall.feature
```diff
@@ -13,40 +13,66 @@ Feature: Close a Roll Call
...
+          "object": "roll_call",
+          "action": "close",
+          "update_id": '#(getRollCallCloseValidUpdateId)',
+          "closes": '#(getRollCallCloseValidId)',
+          "closed_at": 1633099135,
+          "attendees": ['#(getAttendee)']
+        }
+      """
+    When frontend.publish(JSON.stringify(validCloseRollCall), laoChannel)
+    And json answer = frontend.getBackendResponseWithBroadcast()
```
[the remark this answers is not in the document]
Tesa27: Well by checking the id and checking if the answer contains the "result" keyword that should be enough to check if the answer is the expected one I think. As for the id I was indeed not checking it's validity. It should be fixed now

tests/karate/src/test/java/be/rollCall/createRollCall.feature
```diff
@@ -15,104 +15,194 @@ Feature: Create a Roll Call
...
+          "id": '#(getRollCallValidId)',
+          "name": "Roll Call ",
+          "creation": 1633098853,
+          "proposed_start": 1633099125,
+          "proposed_end": 1633099140,
+          "location": "EPFL",
+          "description": "Food is welcome!"
+        }
+      """
+    * frontend.setNonAttendeeAsSender()
```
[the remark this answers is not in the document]
Tesa27: I tried giving it a better name now. But nonetheless with the test description it is quite descriptive I think, but I am maybe not objective enough since I worked on it.

tests/karate/src/test/java/common/net/MultiMsgWebSocketClient.java
```diff
@@ -13,15 +13,15 @@ public class MultiMsgWebSocketClient extends WebSocketClient {
 
   private final MessageQueue queue;
   private final Logger logger;
-  private JsonConverter jsonConverter;
-  private final String nonAttendeePk = "oKHk3AivbpNXk_SfFcHDaVHcCcY8IBfHE7auXJ7h4ms=";
-  private final String nonAttendeeSkHex = "0cf511d2fe4c20bebb6bd51c1a7ce973d22de33d712ddf5f69a92d99e879363b";
+  private JsonConverter jsonConverter = new JsonConverter();
+  private static final String nonAttendeePk = "oKHk3AivbpNXk_SfFcHDaVHcCcY8IBfHE7auXJ7h4ms=";
+  private static final String nonAttendeeSkHex = "0cf511d2fe4c20bebb6bd51c1a7ce973d22de33d712ddf5f69a92d99e879363b";
+  private int idAssociatedWithSentMessage;
```
GabrielFleischer, line 19: What if we send multiple messages ?

How would you handle this case with the current API ?

tests/karate/src/test/java/common/net/MultiMsgWebSocketClient.java
```diff
@@ -64,11 +64,13 @@ public String getBackendResponseWithBroadcast(){
     String answer1 = getBuffer().takeTimeout(5000);
     String answer2 = getBuffer().takeTimeout(5000);
     String result = answer1.contains("result") ? answer1 : answer2;
+    assert (result.contains("\"id\":" + idAssociatedWithSentMessage));
```
GabrielFleischer, line 67: Beware, in the json specification, nothing says that a json should have the form `"id":12`, it could as well be `"id" : 13` and your code would break.

This is not especially important in this case, but the best practice is to convert the string to json and do assertions on that object.

GabrielFleischer (changes requested):
It's better, but not yet perfect. 

Close to, but not yet.

tests/karate/src/test/java/common/net/MultiMsgWebSocketClient.java
Tesa27: Yes you are right, I applied some changes

tests/karate/src/test/java/common/net/MultiMsgWebSocketClient.java
Tesa27: Agree

GabrielFleischer (commented):
I am still not completely sold by your implementation

tests/karate/src/test/java/common/net/MultiMsgWebSocketClient.java
```diff
@@ -44,24 +49,45 @@ public MessageBuffer getBuffer() {
...
+        int idResult = resultJson.get("id");
+        if (idData == idResult){
+          idAssociatedWithSentMessages.remove(data);
+          return answer;
+        }else{
+          idAssociatedWithAnswers.put(idResult, answer);
+        }
+      }
+      answer = getBuffer().takeTimeout(5000);
+    }
```
GabrielFleischer, line 88: There is a function defined in the buffer that lets you retrieve a message based on a specific filter.

This might be useful here.

(Some filters are already defined in a class)

GabrielFleischer (approved):
Looks good to me 👍

It can still be refined, but I think this is suitable to be merged.

tests/karate/src/test/java/common/net/MultiMsgWebSocketClient.java
Tesa27: I have an upcoming PR for election self contained tests, I will address this refactoring there, thanks for pointing it out.

PR #994, opened 2022-05-08
Fix bug in RollCallOpened

Tyratox:
This PR fixes smaller issues in RollCallOpened:
- Check all scanned tokens whether they are valid (base64 strings of a given length). (Prevents the lao connection qr code from being added as a pop token)
- Add token of the organizer if it failed the first time, setup his wallet afterwards and then re-opens the rollcall
- Deduplicate pop tokens when merging the newly scanned ones (after re-opening) with the old ones
- Remove inline style


Xelowak (approved):
Thanks for the small changes!

PR #998, opened 2022-05-09
Fix some styling related issues

Tyratox:
This PR takes care of some styling related issues:
- In 3 places, styles were passed as navigation parameters. This is very counterintuitive, styles should be static.
- Removed all inline styling
- Changes linter rule [`react-native/no-inline-styles`](https://github.com/Intellicode/eslint-plugin-react-native) from `warn` to `error`
- Enabled additional linter rules from [`eslint-plugin-react-native`](https://github.com/Intellicode/eslint-plugin-react-native) to detect unused stylesheet rules, color literals, raw text and single element style arrays


fe1-web/src/features/evoting/screens/CreateElection.tsx
```diff
@@ -46,6 +47,16 @@ const EMPTY_QUESTION: NewQuestion = {
...
+  view: {
+    ...Views.base,
+    flexDirection: 'row',
+    zIndex: 3,
+  } as ViewStyle,
+  viewVertical: {
+    ...Views.base,
+    flexDirection: 'column',
+    zIndex: 3,
+  } as ViewStyle,
```
Xelowak, line 59: Maybe this style relative to events could be stored somewhere instead of being copy/pasted 3 times.

Xelowak (commented):
Thanks a lot for taking time to handle this problem!

fe1-web/src/features/evoting/screens/CreateElection.tsx
Tyratox: Yes that is true but I plan to change the UI a bit around next week so I thought I would tackle this then 😅 But since it is not much effort I'll quickly do it :)

Xelowak (approved):

PR #1002, opened 2022-05-09
Add support for the implementation of the greeting message in scala 

Ajkunas:
Implementation of the greeting message based on the specification protocol of PR #917

### DIFF
--- a/be2-scala/src/main/scala/ch/epfl/pop/model/objects/LaoData.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/model/objects/LaoData.scala
@@ -33,14 +33,20 @@ final case class LaoData(
     that.toJson.toString
   }
 
-  def updateWith(message: Message, address: String): LaoData = {
+  def updateWith(message: Message, adr: Option[String]): LaoData = {
     message.decodedData.fold(this) {
-      case call: CloseRollCall => LaoData(owner, call.attendees, privateKey, publicKey, witnesses, address)
-      case lao: CreateLao => LaoData(lao.organizer, lao.organizer :: Nil, privateKey, publicKey, lao.witnesses, address)
+      case call: CloseRollCall => adr match {
+        case Some(str) => LaoData(owner, call.attendees, privateKey, publicKey, witnesses, str)
+        case _ => LaoData(owner, call.attendees, privateKey, publicKey, witnesses, address)
+      }
+
+      case lao: CreateLao => adr match {
+        case Some(str) => LaoData(lao.organizer, lao.organizer :: Nil, privateKey, publicKey, lao.witnesses, str)
+        case _ => LaoData(lao.organizer, lao.organizer :: Nil, privateKey, publicKey, lao.witnesses, address)
+      }
       case _ => this
     }
   }
-
 }
 
 object LaoData extends Parsable {
--- a/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/LaoHandler.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/LaoHandler.scala
@@ -28,6 +28,7 @@ case object LaoHandler extends MessageHandler {
         val laoChannel: Channel = Channel(s"${Channel.ROOT_CHANNEL_PREFIX}${data.id}")
         val socialChannel: Channel = Channel(s"$laoChannel${Channel.SOCIAL_MEDIA_CHIRPS_PREFIX}")
         val reactionChannel: Channel = Channel(s"$laoChannel${Channel.REACTIONS_CHANNEL_PREFIX}")
+        //we get access to the canonincal address of the server
         val config = ServerConf(appConf)
         val address: String = f"ws://${config.interface}:${config.port}/${config.path}"
 
--- a/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/LaoHandler.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/LaoHandler.scala
@@ -30,7 +30,7 @@ case object LaoHandler extends MessageHandler {
         val reactionChannel: Channel = Channel(s"$laoChannel${Channel.REACTIONS_CHANNEL_PREFIX}")
         //we get access to the canonical address of the server
         val config = ServerConf(appConf)
-        val address: Option[String] = Option(f"ws://${config.interface}:${config.port}/${config.path}")
+        val address: Option[String] = Some(f"ws://${config.interface}:${config.port}/${config.path}")
 
         val combined = for {
           // check whether the lao already exists in db
--- a/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/LaoHandler.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/LaoHandler.scala
@@ -30,7 +30,7 @@ case object LaoHandler extends MessageHandler {
         val reactionChannel: Channel = Channel(s"$laoChannel${Channel.REACTIONS_CHANNEL_PREFIX}")
         //we get access to the canonical address of the server
         val config = ServerConf(appConf)
-        val address: String = f"ws://${config.interface}:${config.port}/${config.path}"
+        val address: Option[String] = Option(f"ws://${config.interface}:${config.port}/${config.path}")
 
         val combined = for {
           // check whether the lao already exists in db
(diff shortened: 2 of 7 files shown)


lourkeur (dismissed):
Looks ok on the Scala side. Might want a TS specialist to take a look.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/LaoHandler.scala
```diff
@@ -14,15 +17,20 @@ import scala.util.{Failure, Success}
...
+    val params: Option[Message] = rpcMessage.getParamsMessage
+    params match {
       case Some(message: Message) =>
         val data: CreateLao = message.decodedData.get.asInstanceOf[CreateLao]
 
         // we are using the lao id instead of the message_id at lao creation
         val laoChannel: Channel = Channel(s"${Channel.ROOT_CHANNEL_PREFIX}${data.id}")
         val socialChannel: Channel = Channel(s"$laoChannel${Channel.SOCIAL_MEDIA_CHIRPS_PREFIX}")
         val reactionChannel: Channel = Channel(s"$laoChannel${Channel.REACTIONS_CHANNEL_PREFIX}")
+        //we get access to the canonincal address of the server
```
tuomas-p, line 31:
```suggestion
        //we get access to the canonical address of the server
```

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/LaoHandler.scala
```diff
@@ -1,10 +1,13 @@
...
+import ch.epfl.pop.config.RuntimeEnvironment.appConf
+import ch.epfl.pop.config.ServerConf
+import ch.epfl.pop.json.MessageDataProtocol.GreetLaoFormat
 import ch.epfl.pop.model.network.JsonRpcRequest
 import ch.epfl.pop.model.network.method.message.Message
 import ch.epfl.pop.model.network.method.message.data.ObjectType
-import ch.epfl.pop.model.network.method.message.data.lao.{CreateLao, StateLao}
-import ch.epfl.pop.model.objects.{Channel, DbActorNAckException, Hash}
+import ch.epfl.pop.model.network.method.message.data.lao.{CreateLao, GreetLao, StateLao}
+import ch.epfl.pop.model.objects.{Base64Data, Channel, DbActorNAckException, Hash, Signature}
```
tuomas-p, line 10: is the Signature import used somewhere?

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
```diff
@@ -84,9 +84,10 @@ class RollCallHandler(dbRef: => AskableActorRef) extends MessageHandler {
                 rpcRequest.id
               ))
               case Some(_) =>
+                val laoData: LaoData = readData(rpcRequest.getParamsChannel)
                 val combined = for {
                   _ <- dbActor ? DbActor.ReadLaoData(rpcRequest.getParamsChannel)
-                  _ <- dbActor ? DbActor.WriteLaoData(rpcRequest.getParamsChannel, message)
+                  _ <- dbActor ? DbActor.WriteLaoData(rpcRequest.getParamsChannel, message, laoData.address)
```
tuomas-p, line 90: This is an optional comment but I would avoid an additional DbActor request if possible. A possibility to be able to keep using WriteLaoData without the extra read, but with the additional address modification functionality could be to have an Option with a String (only modify the address in the case Some(str)). This way, you could simply place a None here

tuomas-p: The approach is not wrong though, just a bit overkill imo

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/MessageHandler.scala
```diff
@@ -58,4 +59,33 @@ trait MessageHandler extends AskPatternConstants {
...
+
+  /**
+   * Helper function for broadcasting messages to subscribers
+   *
+   * @param rpcMessage       : message for which we want to generate the broadcast
+   * @param channel          : the Channel in which we read the data
+   * @param broadcastData    : the message data we broadcast converted to Base64Data
+   * @param broadcastChannel : the Channel in which we broadcast
+   */
+  def dbBroadcast(rpcMessage: JsonRpcRequest, channel: Channel, broadcastData: Base64Data, broadcastChannel: Channel): GraphMessage = {
```
tuomas-p, line 71: I believe (feel free to correct me if I'm wrong) that we could simply use the broadcastChannel as an argument to get the LaoData with the ReadLaoData, since they are in the same LAO. This would remove the need for the channel argument

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/LaoValidator.scala
```diff
@@ -70,6 +70,32 @@ case object LaoValidator extends MessageDataContentValidator {
...
+      case Some(message: Message) =>
+        val data: GreetLao = message.decodedData.get.asInstanceOf[GreetLao]
+
+        val channel: Channel = rpcMessage.getParamsChannel
+
+        val expectedLaoId: Hash = rpcMessage.extractLaoId
+
+        if (expectedLaoId != data.lao) {
+          Right(validationError("unexpected id"))
+        } else if (data.frontend !=  message.sender) {
```
tuomas-p, line 86:
```suggestion
        } else if (data.frontend != message.sender) {
```

be2-scala/src/main/scala/ch/epfl/pop/storage/DbActor.scala
```diff
@@ -83,14 +83,14 @@ final case class DbActor(
   }
 
   @throws [DbActorNAckException]
-  private def writeLaoData(channel: Channel, message: Message): Unit = {
+  private def writeLaoData(channel: Channel, message: Message, address: String): Unit = {
```
tuomas-p, line 86: same comment as above, you could use an Option for address (should be changed in updateWith too then)

be2-scala/src/main/scala/ch/epfl/pop/model/objects/LaoData.scala
```diff
@@ -18,23 +18,25 @@ import com.google.crypto.tink.subtle.Ed25519Sign
...
+                          witnesses: List[PublicKey],
+                          address: String
                         ) {
   def toJsonString: String = {
     val that: LaoData = this // tricks the compiler into inferring the right type
     that.toJson.toString
   }
 
-  def updateWith(message: Message): LaoData = {
+  def updateWith(message: Message, address: String): LaoData = {
```
tuomas-p, line 36: referring to comments below, could be cleaner with an Option

tuomas-p: Also, I don't know whether this could be an issue, but currently, you will not be able to change the address unless the associated message is a CloseRollCall or CreateLao (from what I understood, it should only happen during Lao creation so it would not be a problem but still wanted to point this out)

tuomas-p (changes requested):
Good job! Some (minor) remarks, if you want to discuss the use of an Option for the LaoData update, feel free to do so, the rest looks good to me unless the TS needs validation

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/MessageHandler.scala
Ajkunas: True for now considering the changes for the SocialMedia and Lao, but for my other PR concerning the message KeyElection, we need to read in the laoChannel and the broadcast in the ElectionChannel. I preferred to do the change directly here eventhough it is not needed here.

be2-scala/src/main/scala/ch/epfl/pop/model/objects/LaoData.scala
Ajkunas: Yes it is true, but as you said it sould not be changed in other handlers except in the lao, since it is where the greeting message is sent. So I thought that it should stay as it was: only concerning the rollcall and election

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
Ajkunas: You are right. It is safer

be2-scala/src/test/scala/ch/epfl/pop/storage/DbActorSuite.scala
```diff
@@ -285,7 +285,7 @@ class DbActorSuite extends TestKit(ActorSystem("DbActorSuiteActorSystem")) with
     // arrange
     val messageRollCall: Message = MessageExample.MESSAGE_CLOSEROLLCALL
     val messageLao: Message = MessageExample.MESSAGE_CREATELAO_SIMPLIFIED
-    val address: String = "ws://popdemo.dedis.ch"
+    val address: Option[String] = Option("ws://popdemo.dedis.ch")
```
tuomas-p, line 288: Just as a detail, using Option works without issues, as it yields Some(x) for some value x and None for a null input, but using Some(x) directly is also a possibility and can be clearer in the code

tuomas-p (approved):
Good job!

Issue #988, opened 2022-05-06
check range of transaction value


lourkeur:
As discussed in the meeting, we want to define the range for transaction amounts to `[0; 2^53[` and add overflow checks.

PR #938, opened 2022-04-08
Digital Cash - update json schema + doc

MaximeZmt:
It contains the implementation of the transaction_id field, the update of the JSON schema, and some corrections of previous pr.

PR #1003, opened 2022-05-09
add safe bounds for transaction amount

lourkeur:
Quick and easy first step towards #988. This should affect both back-ends. Front-ends are not ready yet.

re BE1: there is no difference on Intel 64, but there's a discrepancy on 32-bits platforms, e.g. `GOARCH=386`, so for the sake of completeness I fixed it.


be1-go/channel/digitalCash/mod_test.go
```diff
@@ -235,6 +235,63 @@ func Test_SendReaction(t *testing.T) {
 	require.NoError(t, channel.Publish(message, socket.ClientSocket{}))
 }
 
+// Tests that the channel works correctly when it receives a large transaction
+func Test_SendTransactionMaxAmount(t *testing.T) {
+	// Create the hub
+
+	var laoID = "fzJSZjKf-2cbXH7kds9H8NORuuFIRLkevJlN7qQemjo="
+	var sender = "M5ZychEi5rwm22FjwjNuljL1qMJWD2sE7oX9fcHNMDU="
+	var digitalCashChannelName = "/root/" + laoID + "/digitalCash/transaction"
```
MaximeZmt, line 244:
```suggestion
	var digitalCashChannelName = "/root/" + laoID + "/coin"
```
Update according to the specification

MaximeZmt (commented):
LGTM 👍!
Just found a small modification that should be done to match the specification

be1-go/channel/digitalCash/mod_test.go
lourkeur: I can put it here but it would make more sense to put it in #938

lourkeur: Hold on my bad I misread the diff

MaximeZmt (approved):
👍

tuomas-p (approved):
The changes in the Scala part look good to me at least, however the unit tests fail

tuomas-p (changes requested):
Unit tests fail

lourkeur:
Cant seem to reproduce it locally hmm, le what moment...

nkcr (commented):
The test is only checking the happy path. You should also check that a transaction with MAX+1 and MIN-1 fails accordingly.

tuomas-p (approved):
Looks good to me Scala-wise

nkcr (approved):
LGTM 🚀

Issue #942, opened 2022-04-08
[BUG] Vote id not checked properly on go backend


Tesa27:
If vote_id is complaint with the json shchema the go backend will not check it's construction:

Example:
-  if `vote id = saofnaosidasdkaskdodsssa= `(which is not complaint with JsonSchema) -> backend rejects it
- if `vote id = p_EYbHyMv6sopI5QhEXBf40MO_eNoq7V_LygBd4c9RA=`  (which is also a valid id of an lao) backend will accept it

Issue #929, opened 2022-04-06
[BUG] question id is never checked on the backend go


Tesa27:
When sending a cast vote message to the backend there is a question id field that serves as an integrity check, for now the go backend will accept any value provided for this  field, which should not be the case, the backend needs to reject any message that has an invalid question id.

PR #1012, opened 2022-05-10
Added vote id verification on be1

Robingoumaz:
This PR is a fix for #942 and #929 where vote IDs and question IDs were not checked in the verification of the message.

### DIFF
--- a/be1-go/channel/election/verification.go
+++ b/be1-go/channel/election/verification.go
@@ -132,10 +132,10 @@ func (c *Channel) verifyMessageCastVote(castVote messagedata.VoteCastVote) error
 		sort.Ints(vote.Vote)
 		sortedIndex := arrayToString(vote.Vote, ",")
 
-		hash := messagedata.Hash("Vote", electionID, string(qs.id), sortedIndex)
+		hash := messagedata.Hash("Vote", electionID, string(qs.ID), sortedIndex)
 
 		if vote.ID != hash {
-			return xerrors.Errorf("vote ID of vote %d is %s should be %s", i, vote.ID, hash)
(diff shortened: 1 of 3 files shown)


be1-go/channel/election/verification.go
```diff
@@ -122,6 +123,22 @@ func (c *Channel) verifyMessageCastVote(castVote messagedata.VoteCastVote) error
...
+			return xerrors.Errorf("no Question with question ID %s exists", vote.Question)
+		}
+
+		sort.Ints(vote.Vote)
+		sortedIndex := arrayToString(vote.Vote, ",")
+
+		hash := messagedata.Hash("Vote", electionID, string(qs.id), sortedIndex)
+
+		if vote.ID != hash {
+			return xerrors.Errorf("vote ID of vote %d is %s should be %s", i, vote.ID, hash)
```
MaximeZmt, line 138: Isn't it weird to have sometimes id in lowercase and sometimes ID in uppercase? 

I know that it isn't probably you who wrote that, but if it isn't too complicated I think it could be great to be consistent.
What do you think?

MaximeZmt (commented):
Nice work 👍  Just one quick question below

be1-go/channel/election/verification.go
Warhedgehog, line 138: Personally, I wonder if it would not be problematic to display the id we expect

Warhedgehog (changes requested):
Just my sole concern. Good fix otherwise 👍

Warhedgehog (approved):
LGTM

MaximeZmt (approved):
LGTM 👍

lhmerino (approved):
LGTM. Merge at will.

PR #1017, opened 2022-05-11
Remove all inter-feature dependencies for the 'rollCall' feature

Tyratox:
Also in this PR is the fix where the RollCall messages no longer rely on `OpenedLaoStore` to retrieve the current laoId and verify the Hash. This was problematic since it crashed if a message was built without being connected to a lao (see #969) and would also be problematic if we connect to more than one LAO.

Now the channel is attached to **all** messages and **all** messages have the `laoId` getter. This allows the `laoId` to be additionally passed to `buildMessageData` and  thus to the `build` functions (`fromJSON`). 

Note that this comes with the "problem" that we cannot handle messages that could theoretically be sent on channels that are not subchannels of a lao. This was a problem before since `ExtendedMessage` also had this `laoId` getter which would would just errored if it was called on a message not sent on a lao sub channel.

Thus this PR makes the `laoId` getter field return `Hash | undefined`, i.e. `undefined` in these cases and the different functions just have to make sure it is defined by themselves. i.e. the MessageHandlers & the `fromJson` function of messages that are always sent on a lao chan just check if it is undefined and throw an error in this case.


pierluca (dismissed):
LGTM, just minor remarks

fe1-web/src/core/network/ingestion/ExtendedMessage.ts
```diff
@@ -41,45 +41,51 @@ export class ExtendedMessage extends Message implements ProcessableMessage {
...
+    this.channel = msg.channel;
 
     this.receivedAt = msg.receivedAt || Timestamp.EpochNow();
-    this.receivedFrom = msg.receivedFrom;
     this.processedAt = msg.processedAt;
-    this.channel =
-      msg.channel ||
-      (() => {
-        throw new Error('channel not defined');
-      })();
```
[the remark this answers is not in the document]
Tyratox: Yes but I had to move it before the `super` call since the `Message` constructor also requires it (but doesn't store the value) :)

https://github.com/dedis/popstellar/blob/work-fe1-nico-separate-rollcall-feature/fe1-web/src/core/network/ingestion/ExtendedMessage.ts#L57-L63

fe1-web/src/core/network/jsonrpc/messages/Message.ts
```diff
@@ -132,26 +130,31 @@ export class Message {
...
-
     const jsonData = msg.data.decode();
     const dataObj = JSON.parse(jsonData);
-    this.#messageData = messageRegistry.buildMessageData(dataObj);
+
+    let laoId: Hash | undefined;
+    try {
+      laoId = getLaoIdFromChannel(channel);
+    } catch {
+      laoId = undefined;
```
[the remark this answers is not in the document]
Tyratox: I am not so sure, there will be some false positives because it would then warn every time for `lao#create` messages, right? (They are sent on the root channel)

fe1-web/src/core/objects/Channel.ts
```diff
@@ -6,12 +6,12 @@ export const ROOT_CHANNEL: Channel = '/root';
 
 export function getLaoIdFromChannel(ch: Channel): Hash {
   if (!ch || !ch.startsWith(ROOT_CHANNEL)) {
-    throw new Error('Invalid channel path');
+    throw new Error(`Invalid channel path: '${ch}'`);
```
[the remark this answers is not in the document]
Tyratox: Yeah that is a good idea! 👍 :)

0xdieman (dismissed):
I agree with your code ( === LGTM )

fe1-web/src/core/network/jsonrpc/messages/Message.ts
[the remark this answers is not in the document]
Tyratox: Yes exactly :) At the moment this is only the case for the first message to create the lao, before there is no lao channel so yeah 😅

fe1-web/src/features/evoting/network/ElectionHandler.ts
```diff
@@ -70,7 +70,8 @@ export const handleElectionOpenMessage =
     console.log('Handling Election open message');
     if (
       msg.messageData.object !== ObjectType.ELECTION ||
-      msg.messageData.action !== ActionType.OPEN
+      msg.messageData.action !== ActionType.OPEN ||
+      !msg.laoId
     ) {
       console.warn('handleElectionOpenMessage was called to process an unsupported message', msg);
```
Xelowak, line 76: I'm not sure this error message is very clear in the case where `msg.laoId` is `undefined`. Maybe another if printing a different error message could be nice, but I'm not sure if a message can be sent without `laoId`.
The same applies for all error in this file.

fe1-web/src/features/meeting/network/MeetingHandler.ts
```diff
@@ -20,14 +20,14 @@ export const handleMeetingCreateMessage =
   (msg: ProcessableMessage): boolean => {
     if (
       msg.messageData.object !== ObjectType.MEETING ||
-      msg.messageData.action !== ActionType.CREATE
+      msg.messageData.action !== ActionType.CREATE ||
+      !msg.laoId
     ) {
       console.warn('handleMeetingCreateMessage was called to process an unsupported message', msg);
```
Xelowak, line 26: Same comment.

fe1-web/src/features/rollCall/hooks/__tests__/RollCallHooks.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+mockStore.dispatch(addEvent(mockLaoId, mockRollCallState));
+
+const wrapper = ({ children }: { children: React.ReactChildren }) => (
+  <Provider store={mockStore}>
+    <FeatureContext.Provider value={contextValue}>{children}</FeatureContext.Provider>
+  </Provider>
+);
+
+describe('RollCallHooks', () => {
+  describe('RollCallHook.useCurrentLaoId', () => {
```
Xelowak, line 39:
```suggestion
  describe('useCurrentLaoId', () => {
```

fe1-web/src/features/rollCall/hooks/__tests__/RollCallHooks.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+    it('should return the current lao id', () => {
+      const { result } = renderHook(() => RollCallHooks.useCurrentLaoId(), { wrapper });
+      expect(result.current).toEqual(mockLaoIdHash);
+    });
+  });
+
+  describe('RollCallHook.useEventSelector', () => {
```
Xelowak, line 46:
```suggestion
  describe('useEventSelector', () => {
```

fe1-web/src/features/rollCall/hooks/__tests__/RollCallHooks.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+        {
+          wrapper,
+        },
+      );
+      // the 'end' field
+      expect(result.current?.toState()).toEqual(mockRollCallState);
+    });
+  });
+
+  describe('RollCallHook.useGenerateToken', () => {
```
Xelowak, line 75:
```suggestion
  describe('useGenerateToken', () => {
```

fe1-web/src/features/rollCall/hooks/__tests__/RollCallHooks.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+    it('should return the correct function', () => {
+      const { result } = renderHook(() => RollCallHooks.useGenerateToken(), { wrapper });
+      expect(result.current).toBe(mockGenerateToken);
+    });
+  });
+
+  describe('RollCallHook.hasSeed', () => {
```
Xelowak, line 82:
```suggestion
  describe('hasSeed', () => {
```

Xelowak (approved):
Really nice work!
Thanks a lot for working on removing the inter-feature dependencies. I've made a couple of comments but I approve anyway, as I don't want them to be blocking (they're not that important). If you still make changes in this PR, feel free to ask me for a re-approval on Slack.

Xelowak (approved):
Thanks for the changes!

PR #1018, opened 2022-05-13
Work karate tesa digital cash maxime

Tesa27:
This PR is open just to notify which go files should be changed so that the "coin" channel can be accessed since it is not the case for now.


Tesa27:
I am closing this PR, it was just to propose some ideas for the implementation of the digital cash channel for @MaximeZmt

PR #1019, opened 2022-05-14
fix coinbase example

lourkeur:


MaximeZmt:
Thanks for your help 👍

PR #1020, opened 2022-05-15
Remove all inter-feature dependencies for the 'events' feature

Tyratox:
This PR removes the inter-feature dependencies from the 'events' feature.

One of the major changes is that the EventReducer no longer stores all information regarding the different event types but just information about what type it is, its id and a start and end time. The different features are then required to handle everything else by themselves such as storing the concrete state for the different events.

Just as an information: This PR was horrible to do because a lot of changes in many places were required. I assume reviewing it will be even worse so yeah.. Sorry for that but I did not manage to split it into multiple PRs that make sense by themselves.


fe1-web/src/features/wallet/objects/Token.ts
```diff
@@ -48,7 +48,7 @@ export function generateToken(laoId: Hash, rollCallId: Hash | undefined): Promis
 /**
  * Retrieve the latest PoP token associated with the current LAO from the store.
  * @param getCurrentLao A function returning the current lao
- * @param getCurrentLao A function returning the current lao
+ * @param getRollCallById A function to get a roll call by its id
```
pierluca, line 51: nice catch !

fe1-web/src/features/wallet/objects/Token.ts
```diff
@@ -68,10 +68,16 @@ export const getCurrentPopTokenFromStore =
...
+    if (!rollCall) {
+      throw new Error(`Cannot retrieve pop token: did not find a roll call with id ${rollCallId}`);
+    }
+
     const token = await generateToken(lao.id, rollCallId);
-    if (rollCall.containsToken(token)) {
-      return token;
+
+    if (!rollCall.containsToken(token)) {
+      throw new Error('Cannot retrieve pop token: roll call does not contain this token');
```
pierluca, line 79: outside the scope of this PR: typed errors would be nice here, to distinguish between line 73, a "this-should-never-happen-and-all-our-expectations-have-been-violated" error, and line 79, a "tough-luck-you-have-no-token" kind of error

pierluca (commented):
Phew, this was a long one ! Pretty good overalls, only minor remarks, though I haven't been able to review everything in-depth.

@Xelowak I'll let you have a more in-depth check of the reducers and their associated tests (and approve if you're okay with them), I haven't reviewed those in detail.

fe1-web/src/features/wallet/interface/Feature.ts
```diff
@@ -37,9 +24,12 @@ export namespace WalletFeature {
     CLOSED,
   }
 
-  export interface RollCallState extends EventState {}
+  export interface RollCallState {}
```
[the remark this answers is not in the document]
Tyratox: No we don't, you are right :)

fe1-web/src/features/wallet/objects/DummyWallet.ts
```diff
@@ -4,26 +4,27 @@ import { EventTags, Hash, PublicKey, Timestamp } from 'core/objects';
 import { dispatch } from 'core/redux';
 import { addEvent, removeEvent } from 'features/events/reducer';
 import { connectToLao, disconnectFromLao, removeLao } from 'features/lao/reducer';
-import { EventTypeRollCall, RollCall, RollCallStatus } from 'features/rollCall/objects';
+import { RollCall, RollCallStatus } from 'features/rollCall/objects';
```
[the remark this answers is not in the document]
Tyratox: 👍

fe1-web/src/features/wallet/objects/Token.ts
Tyratox: Yes I agree!

fe1-web/src/features/rollCall/hooks/RollCallHooks.ts
```diff
@@ -53,4 +40,49 @@ export namespace RollCallHooks {
...
+        [laoId: string]: { [rollCallId: string]: RollCall };
+      } = {};
+
+      for (const laoId of Object.keys(eventStatesByLaoId)) {
+        map[laoId] = {};
+
+        for (const rollCallId of Object.keys(eventStatesByLaoId[laoId])) {
+          map[laoId][rollCallId] = rollCallById[rollCallId];
+        }
+      }
```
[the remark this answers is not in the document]
Tyratox: I rewrote the function using reduce and named functions :) I hope it is easier to read now?

fe1-web/src/features/evoting/reducer/ElectionReducer.ts
```diff
@@ -0,0 +1,114 @@
...
+
+export interface ElectionReducerState {
+  byId: Record<string, ElectionState>;
+  allIds: string[];
+}
+
+const initialState: ElectionReducerState = {
+  byId: {},
+  allIds: [],
+};
```
Xelowak, line 24: Is this related to any LAO or is it completely generic and stores all elections?
The same questions applies for each <Event>Reducer.

fe1-web/src/features/rollCall/hooks/RollCallHooks.ts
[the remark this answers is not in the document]
Xelowak: It is!

fe1-web/src/features/rollCall/reducer/RollCallReducer.ts
```diff
@@ -0,0 +1,219 @@
...
+          .reduce<Record<string, RollCall>>((map, rollCall) => {
+            map[rollCall.id.valueOf()] = rollCall;
+            return map;
+          }, {})
+      );
+    },
+  );
+
+/**
+ * Retrieves an rollcall by its id from the redux store
```
Xelowak, line 168:
```suggestion
 * Retrieves a rollcall by its id from the redux store
```

fe1-web/src/features/rollCall/reducer/__tests__/RollCallReducer.test.ts
```diff
@@ -0,0 +1,309 @@
...
+      ).toEqual({
+        byId: {
+          [mockRollCallState.id]: mockRollCallState,
+        },
+        allIds: [mockRollCallState.id],
+        idAlias: {},
+      } as RollCallReducerState);
+    });
+
+    it('throws an error if the store already contains an rollcall with the same id', () => {
```
Xelowak, line 50:
```suggestion
    it('throws an error if the store already contains a rollcall with the same id', () => {
```

Xelowak (commented):
Good job!
As pierluca asked me, I checked reducers and their tests and everything seems fine to me. I just had some questions for you.

fe1-web/src/features/evoting/reducer/ElectionReducer.ts
Tyratox: Yes this is supposed to be generic and store all elections. The ids of the different elections, roll calls, etc. are obtained by computing a hash function on some constants, the laoId and some other properties. Thus as long as the hash function is collision resistant (which it should be, otherwise we have problems with the security-related parts of the protocol), the ids will be unique. And as you said, if you would like to obtain all elections for a given lao, you can use the events reducer to obtain the set of ids and then load them from the election reducer. As far as I remember, this is only done for roll calls in the wallet and in all other placed we do not need a list of a given event type for a given lao id.

Xelowak: Okay, thanks for the explanation. This implementation seems fine to me.

Xelowak (approved):

Issue #1008, opened 2022-05-09
[BUG] The return error code must be -1,-2,-3,-4,-5 or -6


Tesa27:
```
// handleRootChannelPublishMesssage handles an incominxg publish message on the root channel.
func (h *Hub) handleRootChannelPublishMesssage(sock socket.Socket, publish method.Publish) error {
	jsonData, err := base64.URLEncoding.DecodeString(publish.Params.Message.Data)
	if err != nil {
		err := xerrors.Errorf("failed to decode message data: %v", err)
		sock.SendError(&publish.ID, err)
		return err
	}

	// validate message data against the json schema
	err = h.schemaValidator.VerifyJSON(jsonData, validation.Data)
	if err != nil {
		err := xerrors.Errorf("failed to validate message against json schema: %v", err)
		sock.SendError(&publish.ID, err)
		return err
	}

	// get object#action
	object, action, err := messagedata.GetObjectAndAction(jsonData)
	if err != nil {
		err := xerrors.Errorf("failed to get object#action: %v", err)
		sock.SendError(&publish.ID, err)
		return err
	}

	// must be "lao#create"
	if object != messagedata.LAOObject || action != messagedata.LAOActionCreate {
		err := answer.NewErrorf(publish.ID, "only lao#create is allowed on root, "+
			"but found %s#%s", object, action)
		sock.SendError(&publish.ID, err)
		return err
	}

	var laoCreate messagedata.LaoCreate

	err = publish.Params.Message.UnmarshalData(&laoCreate)
	if err != nil {
		h.log.Err(err).Msg("failed to unmarshal lao#create")
		sock.SendError(&publish.ID, err)
		return err
	}

	err = laoCreate.Verify()
	if err != nil {
		h.log.Err(err).Msg("invalid lao#create message")
		sock.SendError(&publish.ID, err)
	}

	err = h.createLao(publish.Params.Message, laoCreate, sock)
	if err != nil {
		h.log.Err(err).Msg("failed to create lao")
		sock.SendError(&publish.ID, err)
		return err
	}

	h.rootInbox.StoreMessage(publish.Params.Message)

	return nil
}
```

In this function the return code is always publish.ID which means nothing since it is not a defined error code

## Suggested fix
define a variable` id` and set it to `-4`  (which stands for invalid message field) and replace `sock.SendError(&publish.ID, err)` by `sock.SendError(&id, err)`

Robingoumaz:
[...]

PR #1021, opened 2022-05-15
Fix for issue #1008 + issue raised in go meeting 11.05 + refactor hub

Robingoumaz:
Fix for issue #1008 
Fixed issue where somewhere on the hub there was an error created not custom
Refactored a little the hub files.


be1-go/hub/standard_hub/message_handling.go
```diff
@@ -79,40 +79,38 @@ func (h *Hub) handleRootChannelPublishMesssage(sock socket.Socket, publish metho
...
 func (h *Hub) handleRootChannelBroadcastMesssage(sock socket.Socket,
 	broadcast method.Broadcast) error {
 
-	id := -1
-
 	jsonData, err := base64.URLEncoding.DecodeString(broadcast.Params.Message.Data)
 	if err != nil {
 		err := xerrors.Errorf("failed to decode message data: %v", err)
-		sock.SendError(&id, err)
+		sock.SendError(nil, err)
```
Warhedgehog, line 89: Why put a nil in this function ?

Robingoumaz: because broadcasts don't have a messageID. According to the comment of SendError() we can put nil when no messageID is available.
Also &id was a pointer to the literal 1, which is what SendError sends when given nil as messageID, I felt this implementation was more compliant with how the function was written.

Warhedgehog: Thanks for the clarification

be1-go/hub/standard_hub/message_handling.go
```diff
@@ -19,7 +19,7 @@ import (
 
 const publishError = "failed to publish: %v"
 
-// handleRootChannelPublishMesssage handles an incominxg publish message on the root channel.
+// handleRootChannelPublishMesssage handles an incoming publish message on the root channel.
 func (h *Hub) handleRootChannelPublishMesssage(sock socket.Socket, publish method.Publish) error {
```
Warhedgehog, line 23: It's not from you of course but it could be occasion to correct the typo in function name and comment

be1-go/hub/standard_hub/message_handling.go
```diff
@@ -79,40 +79,38 @@ func (h *Hub) handleRootChannelPublishMesssage(sock socket.Socket, publish metho
 	return nil
 }
 
-// handleRootChannelPublishMesssage handles an incominxg publish message on the root channel.
+// handleRootChannelPublishMesssage handles an incoming publish message on the root channel.
 func (h *Hub) handleRootChannelBroadcastMesssage(sock socket.Socket,
```
Warhedgehog, line 83: Here too

Warhedgehog (changes requested):
Just tiny changes on top of what you did. Great work 👍

Warhedgehog (approved):
LGTM. I don't think the code coverage should be taken into account since it is a fix

MaximeZmt (approved):
LGTM 👍

nkcr (approved):

nkcr (approved):

Tesa27:
I tested this implementation with the karate tests that where failing concerning the non-custom errors and they still fail so there must be some other place where the error code is not custom. I will look into this

PR #1024, opened 2022-05-16
Fix wrong error code returns

Tesa27:
- Whenever an invalid message was sent on the root channel, there was always an error code of -6 representing INTERNAL_SERVER_ERROR that was return by default since the errors returned when bad messages received on the root channel where not custom(the ones we use throughout the go backend) which was not recognized by the backend since it is not one of it's messages so an error code of -6 was imposed due to the fact that the server could not figure out which error it had received. This was due to whenever there is a check on messages that arrived on root the error returned was using the library 'xerror' instead of 'answer' which is custom made. 
- There was also a lao channel created due to a forgotten return of an error received on the root channel, which caused some confusion since an Error and a Result where sent by the backend on after another and this was captured by Karate tests, but could create a great vulnerability if invalid lao create messages still created lao channels even if proven invalid
- Lastly due to the previous point some tests where passing even if they should not since they where building the lao id in a wrong way, now these kind of issues are more easily detectable, but tests that were not robust enough had to be changed a bit.


be1-go/hub/standard_hub/message_handling.go
```diff
@@ -19,36 +19,33 @@ import (
 
 const publishError = "failed to publish: %v"
 
-// handleRootChannelPublishMesssage handles an incominxg publish message on the root channel.
+// handleRootChannelPublishMessage handles an incoming publish message on the root channel.
```
Robingoumaz, line 22: Some of the typos were already corrected in #1021, you might want to merge master first.

be1-go/hub/standard_hub/message_handling.go
```diff
@@ -19,36 +19,33 @@ import (
...
 func (h *Hub) handleRootChannelPublishMesssage(sock socket.Socket, publish method.Publish) error {
 	jsonData, err := base64.URLEncoding.DecodeString(publish.Params.Message.Data)
 	if err != nil {
-		err := xerrors.Errorf("failed to decode message data: %v", err)
-		sock.SendError(&publish.ID, err)
```
Robingoumaz, line 27: I think we still need to send the error to the socket, the returned error will just be logged

be1-go/hub/standard_hub/message_handling.go
```diff
@@ -19,36 +19,33 @@ import (
...
+		err := answer.NewErrorf(-4, "failed to decode message data: %v", err)
+
 		return err
 	}
 
 	// validate message data against the json schema
 	err = h.schemaValidator.VerifyJSON(jsonData, validation.Data)
 	if err != nil {
-		err := xerrors.Errorf("failed to validate message against json schema: %v", err)
-		sock.SendError(&publish.ID, err)
```
Robingoumaz, line 35: same

be1-go/hub/standard_hub/message_handling.go
```diff
@@ -19,36 +19,33 @@ import (
...
+		err := answer.NewErrorf(-4, "failed to validate message against json schema: %v", err)
 		return err
 	}
 
 	// get object#action
 	object, action, err := messagedata.GetObjectAndAction(jsonData)
 	if err != nil {
-		err := xerrors.Errorf("failed to get object#action: %v", err)
-		sock.SendError(&publish.ID, err)
```
Robingoumaz, line 43: same

be1-go/hub/standard_hub/message_handling.go
```diff
@@ -19,36 +19,33 @@ import (
...
+		err := answer.NewErrorf(-4, "failed to get object#action: %v", err)
 		return err
 	}
 
 	// must be "lao#create"
 	if object != messagedata.LAOObject || action != messagedata.LAOActionCreate {
-		err := answer.NewErrorf(publish.ID, "only lao#create is allowed on root, "+
+		err := answer.NewErrorf(-4, "only lao#create is allowed on root, "+
 			"but found %s#%s", object, action)
-		sock.SendError(&publish.ID, err)
```
Robingoumaz, line 51: same

be1-go/hub/standard_hub/message_handling.go
```diff
@@ -57,20 +54,18 @@ func (h *Hub) handleRootChannelPublishMesssage(sock socket.Socket, publish metho
 	err = publish.Params.Message.UnmarshalData(&laoCreate)
 	if err != nil {
 		h.log.Err(err).Msg("failed to unmarshal lao#create")
-		sock.SendError(&publish.ID, err)
```
Robingoumaz, line 60: same

be1-go/hub/standard_hub/mod_test.go
```diff
@@ -138,20 +133,15 @@ func Test_Create_LAO_Bad_MessageID(t *testing.T) {
...
+	laoID := messagedata.Hash(base64.URLEncoding.EncodeToString(keypair.publicBuf), fmt.Sprintf("%d", now), name)
 
 	data := messagedata.LaoCreate{
 		Object:    messagedata.LAOObject,
 		Action:    messagedata.LAOActionCreate,
 		ID:        laoID,
 		Name:      name,
-		Creation:  123,
-		Organizer: base64.URLEncoding.EncodeToString([]byte("XXX")),
+		Creation:  now,
```
Robingoumaz, line 143: same

be1-go/hub/standard_hub/mod_test.go
```diff
@@ -221,20 +211,15 @@ func Test_Create_LAO_Bad_Signature(t *testing.T) {
...
```
Robingoumaz, line 221: same

be1-go/hub/standard_hub/mod_test.go
```diff
@@ -303,20 +288,240 @@ func Test_Create_LAO(t *testing.T) {
...
+
+	hub.handleMessageFromClient(&socket.IncomingMessage{
+		Socket:  sock,
+		Message: publishBuf,
+	})
+
+	require.EqualError(t, sock.err, fmt.Sprintf("%v", sock.err))
+}
+
+func Test_Create_Invalid_Json_Schema(t *testing.T) {
```
Robingoumaz, line 353: json schemas are tested in the validation folder

be1-go/hub/standard_hub/mod_test.go
```diff
@@ -303,20 +288,240 @@ func Test_Create_LAO(t *testing.T) {
...
+
+	hub.handleMessageFromClient(&socket.IncomingMessage{
+		Socket:  sock,
+		Message: publishBuf,
+	})
+
+	assert.Contains(t, sock.err.Error(), "failed to validate message against json schema", "error message %s", "formatted")
+}
+
+func Test_Create_Invalid_Lao_Id(t *testing.T) {
```
Robingoumaz, line 425: if you want it may be easier if you read examples from files in popstellar/protocol/examples/messageData/lao_create
There are a lot of bad lao_create to fit your needs

Robingoumaz (changes requested):
Some small things

be1-go/message/messagedata/lao_create.go
```diff
@@ -40,25 +41,25 @@ func (message LaoCreate) Verify() error {
...
-		return xerrors.Errorf("lao creation is %d, should be minimum 0", message.Creation)
+		return answer.NewErrorf(-4, "lao creation is %d, should be minimum 0", message.Creation)
 	}
 
 	// verify organizer is base64URL encoded
 	_, err = base64.URLEncoding.DecodeString(message.Organizer)
 	if err != nil {
-		return xerrors.Errorf("lao organizer is %s, should be base64URL encoded", message.Organizer)
+		return answer.NewErrorf(-4, "lao organizer is %s, should be base64URL encoded", message.Organizer)
 	}
```
pierluca, line 56: Please use constants and avoid magic numbers, here and everywhere else 😉

be1-go/hub/standard_hub/message_handling.go
Tesa27: Will do thank you

be1-go/hub/standard_hub/message_handling.go
Tesa27: That is the thing I realized here, the Error will be sent to the socket either way this is done in the function `handleMessageFromClient` , I've had this error before that even when we get an error we receive another message afterwards this was due (like I explain in the description of this PR) to the fact that we send an error twice to the socket if an error happens on the root channel. If you take a look at how errors are handled in other channels lao,election... you will see that none of them actually send the Error to the socket but propagate it where the `handleMessageFromClient` happens and there we send the error to the socket. We could also delete the error sending there and send it beforehand but I am not sure that it will be the right approach

be1-go/hub/standard_hub/mod_test.go
```diff
@@ -53,20 +53,15 @@ func Test_Create_LAO_Bad_Key(t *testing.T) {
...
+	laoID := messagedata.Hash(base64.URLEncoding.EncodeToString(wrongKeypair.publicBuf), fmt.Sprintf("%d", now), name)
 
 	data := messagedata.LaoCreate{
 		Object:    messagedata.LAOObject,
 		Action:    messagedata.LAOActionCreate,
 		ID:        laoID,
 		Name:      name,
-		Creation:  123,
-		Organizer: base64.URLEncoding.EncodeToString([]byte("XXX")),
+		Creation:  now,
```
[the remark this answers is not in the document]
Tesa27: Could be yeah, but like this we ensure that the test will work under different values so it is more robust in my opinion

be1-go/hub/standard_hub/mod_test.go
Tesa27: Thank you for noting

be1-go/hub/standard_hub/mod_test.go
Tesa27: I think having self contained tests is more explicit and shows where the invalid field is and how it is computed that's what I did with karate as well, but if you think a different approach is better I am happy to discuss it

be1-go/message/answer/error.go
```diff
@@ -44,3 +44,43 @@ func NewInvalidObjectError(action string) *Error {
 		Description: fmt.Sprintf("invalid object: %s", action),
 	}
 }
+
+// NewInvalidResourceError returns an error with -2 for an object with invalid ressources
+func NewInvalidResourceError(description string) *Error {
+	return &Error{
+		Code:        -2,
+		Description: description,
```
Robingoumaz, line 52: Maybe just wrap it in a message like the first two

Robingoumaz (changes requested):
Nice

Robingoumaz (dismissed):

be1-go/message/answer/error.go
```diff
@@ -44,3 +44,43 @@ func NewInvalidObjectError(action string) *Error {
 		Description: fmt.Sprintf("invalid object: %s", action),
 	}
 }
+
+// NewInvalidResourceError returns an error with -2 for an object with invalid ressources
```
nkcr, line 48: You should have a generic error function, and then specialized error functions that use it. You can also make those functions accept a format and arbitrary args, in go we then end those functions with `f`: 

```go
func NewError(code int, desc string) {
	return &Error{
		Code:       code,
		Description: desc,
	}
}

func NewErrorf(code int, format string, a ...interface{}) {
  return NewError(code, fmt.Sprintf(format, ...a))
}

func NewInvalidResourceErrorf(format string, a ...interface{}) {
  return NewErrorf(-2, format, ...a)
}

...
```

nkcr: Also please format comments at 80 chars length.

nkcr (commented):
Thanks for spotting that 👍

be1-go/message/answer/error.go
```diff
@@ -30,17 +30,64 @@ func NewErrorf(code int, format string, values ...interface{}) *Error {
 }
 
 // NewInvalidActionError returns an error with the code -1 for an invalid action.
-func NewInvalidActionError(action string) *Error {
-	return &Error{
-		Code:        -1,
-		Description: fmt.Sprintf("invalid action: %s", action),
-	}
+func NewInvalidActionError(action string, values ...interface{}) *Error {
```
nkcr, line 33: this is to stay consistent with the `Sprintf` function. 

```suggestion
func NewInvalidActionErrorf(format string, a ...interface{}) *Error {
```

be1-go/hub/standard_hub/mod.go
```diff
@@ -533,24 +534,24 @@ func (h *Hub) createLao(msg message.Message, laoCreate messagedata.LaoCreate,
 	laoChannelPath := rootPrefix + laoCreate.ID
 
 	if _, ok := h.channelByID[laoChannelPath]; ok {
-		return answer.NewErrorf(-3, "failed to create lao: duplicate lao path: %q", laoChannelPath)
+		return answer.NewDuplicateResourceError(fmt.Sprintf("failed to create lao: duplicate lao path: %q", laoChannelPath))
```
nkcr, line 537: This is how you should use your updated error functions:

```suggestion
		return answer.NewDuplicateResourceErrorf("failed to create lao: duplicate lao path: %q", laoChannelPath)
```

nkcr (approved):

Issue #967, opened 2022-04-27
BE2-SCALA : Open rollcall and close rollcall fix validators


rombirli:
We can open unclosed rollcall, close closed rollcall.... We should fix it in rollCallValidator.scala

lourkeur:
Test case courtesy of @Tesa27

1.
```
{
"object":"lao",
 "action":"create",
 "name":"LAO",
 "creation":1633035721,
 "organizer":"J9fBzJV70Jk5c-i3277Uq4CmeL4t53WDfUghaK0HpeM=",
 "witnesses":[],
 "id":"p_EYbHyMv6sopI5QhEXBf40MO_eNoq7V_LygBd4c9RA="
}
```
2.
```
{
  "object": "roll_call",
  "action": "open",
  "update_id": "l2OYtZueg1xkjvh3RCWw0nSZrrPNThuaz3U3ys7MjHI=",
  "opens": "VSsRrcHoOTQJ-nU_VT_FakiMkezZA86z2UHNZKCxbN8=",
  "opened_at": 1633099127
}
```

Expected: validation error for opening a non-existent rollcall
Actual: no error

PR #1025, opened 2022-05-17
[BUG FIX] Opening and closing RollCalls and checks for question ids and vote ids in Be2 Scala

Ajkunas:
Fixed bug mentioned in issue #967 concerning the RollCall validators and RollCall handlers
Added checks for question ids and vote ids in the Election validators


Tesa27 (approved):
Karate confirms the check for election ids and vote ids, the extra checks for closing a closed roll call and opening an opened roll call also seem to work. To be done in an upcoming PR is to check that closing an unopened roll call and opening a non created roll call should be considered as invalid by the backend too. Good job overall!

lourkeur (approved):
Reviewed off-line due to web interface issues, LGTM

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/ElectionValidator.scala
```diff
@@ -51,6 +51,10 @@ sealed class ElectionValidator(dbActorRef: => AskableActorRef) extends MessageDa
 
         val sender: PublicKey = message.sender
 
+        //checks if the question ids are valid wrt protocol
+        def validateQuestionId(questions: List[ElectionQuestion], election_id: Hash): Boolean =
+          questions.forall(q => q.id == Hash.fromStrings("Question", election_id.toString, q.question))
```
RaulinN, line 56: Would be nice to have this as a constant in one of the related classes. Same for line 157 with "Vote" =)

RaulinN (approved):
Nice job! I was confused in some commits because it looks like there are changes than Lourkeur did in a previous PR 🤔. Other than that, nice work

PR #1026, opened 2022-05-17
Fix roll call drop down and UI enhancement

0xdieman:
In this PR:

- Fixed the roll call dropdown which errored in the console
- Added a `RoundIconButton`, that uses the google material icons
- Used this button as back buttons for the wallet setup and to logout
- Added some rendering tests for the components


pierluca (commented):
Had a super quick look, not bad at all ! 👍
I'll let @Tyratox and @Xelowak do a proper review 😄

fe1-web/src/core/components/RoundIconButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+    shadowColor: 'gray',
+    shadowOffset: {
+      width: 2,
+      height: 2,
+    },
+    borderRadius: 30,
+    backgroundColor: '#2196F3',
+    display: 'flex',
+    justifyContent: 'center',
+    alignItems: 'center',
```
Tyratox, line 19: In general we should try to avoid constants in styles as much as possible and refer to global variables instead. I am doing some work on a PR regarding this since it is the case all over the application 😅 But it would be cool if you could for example reference a color define in `core/styles/colors` as it will make UI changes much much easier :)

fe1-web/src/core/components/RoundIconButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+ * you can look for icon and their names here : https://fonts.google.com/icons
+ */
+const RoundIconButton = (props: IPropTypes) => {
+  const { name, onClick } = props;
+  return (
+    <TouchableOpacity style={styles.roundButton} onPress={onClick}>
+      <Icon name={name} type="material" color="white" size={25} />
+    </TouchableOpacity>
+  );
+};
```
Tyratox, line 35: This is design choice but I would argue it might make sense to define several Icon components. Your version is much more general but this "allows" the usage of many many different icons. So maybe not export this component directly but rather export `SettingsIcon`, `BackIcon`, etc. that fix the icon used (i.e. the "name"). This will make it harder to use arbitrary icons / easier to use the existing ones. We should try to keep the used icons consistent.

@pierluca @Xelowak  What do you think?

Tyratox (dismissed):
I general the PR seems to be very nice! I left some picky comments that are not specific to this PR but rather the direction we want to go with the design in general :)

There are a few things I do not like regarding the UI but I think this is okay. I am working on several UI changes at the moment and will try to incorporate your changes and try to make all of it look consistent :) Also I think discussing this in person is much easier!

fe1-web/src/core/components/RoundIconButton.tsx
Xelowak: I agree. This is cleaner and will be much easier to keep consistent.

fe1-web/src/core/components/RoundIconButton.tsx
Xelowak: I agree with you. Having icons that are directly mapped to their functionality is easier to understand, and it remains consistent easier throughout the application.

fe1-web/src/features/wallet/components/__tests__/RollCallTokensDropDown.test.tsx
```diff
@@ -0,0 +1,31 @@
+import { render } from '@testing-library/react-native';
+import React from 'react';
+
+import { mockKeyPair, mockLao, mockRC } from '__tests__/utils';
+
+import { PopToken } from '../../../../core/objects';
+import { RollCallToken } from '../../objects/RollCallToken';
```
Xelowak, line 7: It might be better to use absolute paths for these given the amount of `../` (especially `PopToken`).

fe1-web/src/features/wallet/components/__tests__/RollCallTokensDropDown.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+  it('renders correctly', () => {
+    const { toJSON } = render(
+      <RollCallTokensDropDown
+        rollCallTokens={mockArray}
+        selectedTokenIndex={0}
+        onIndexChange={() => {}}
+      />,
+    );
+    expect(toJSON()).toMatchSnapshot();
+  });
```
Xelowak, line 30: This is not important at all, but be aware that you can do this instead of the curly braces notation (which is commonly used when getting multiples elements).
```suggestion
  it('renders correctly', () => {
    const component = render(
      <RollCallTokensDropDown
        rollCallTokens={mockArray}
        selectedTokenIndex={0}
        onIndexChange={() => {}}
      />
    ).toJSON();
    expect(component).toMatchSnapshot();
  });
```

Xelowak (commented):
Good job! I left some tiny comments, but it should be ready to merge very soon.

Xelowak (commented):
Good job! We can merge it as soon as your tests pass

Tyratox (dismissed):

Xelowak (approved):
LGTM!

PR #1037, opened 2022-05-20
Update election#cast_vote implementation

Tyratox:
This PR updates the implementation to reflect the specification changes discussed on slack

> Hello,
> Given the current specification of the secret-ballot cast_vote the indexes are encrypted and then sent in an array (it is now an array of encrypted strings). I found that it could be a vulnerability when we allow multiple choice questions because if only one attendee sends two choices, we can then link the votes to the sender even with shuffling.
> There are multiple solution to this problem (that I could find),
> We can disallow multiple choice in secret ballot, update the specs and continue to run like this. (The problem with this is that for future/current pop students that will want to add new voting methods, they will have to redo the implementation that we are doing now)
> We can still have an array of plain text indices but then encrypt the whole vote and encode it in base64. (The problem with this is that the encryption that we use has an upper limit to the size of the data to be encrypted and with multiple choice the data could grow without limit and if you started implementing the other way you would have to change your implementation)
> What do you think is the right way to go ?

Where we decided for option 1.

### DIFF
--- a/fe1-web/src/features/events/objects/__tests__/LaoEventBuilder.test.ts
+++ b/fe1-web/src/features/events/objects/__tests__/LaoEventBuilder.test.ts
@@ -56,7 +56,7 @@ describe('LaoEventBuilder', () => {
     const registeredVotes: RegisteredVote = {
       createdAt: 1520255700,
       sender: 'Sender1',
-      votes: [{ id: 'v1', question: 'q1', vote: [0] }],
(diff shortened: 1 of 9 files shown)


Xelowak (approved):
LGTM! You might have to make sure that it will not break the system before merging, though. I mean, you should probably create a PR that changes this in all sub-systems at the same time.

Tyratox:
Yes, we will merge together with the secret ballot changes :)

PR #1039, opened 2022-05-20
Fe 2 - Digital Cash - Implementation of the json schema 

zoemarin:
create the json object and the test and take into account the names

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/Action.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/Action.java
@@ -12,7 +12,6 @@ public enum Action {
   SETUP("setup"),
   UPDATE("update_properties"),
   STATE("state"),
-  GREET("greet"),
   WITNESS("witness"),
   OPEN("open"),
   REOPEN("reopen"),
@@ -29,7 +28,8 @@ public enum Action {
   ADD("add"),
   NOTIFY_ADD("notify_add"),
   DELETE("delete"),
-  NOTIFY_DELETE("notify_delete");
+  NOTIFY_DELETE("notify_delete"),
+  POST_TRANSACTION("post_transaction");
 
   private static final List<Action> ALL = Collections.unmodifiableList(Arrays.asList(values()));
   private final String action;
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/Action.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/Action.java
@@ -12,6 +12,7 @@ public enum Action {
   SETUP("setup"),
   UPDATE("update_properties"),
   STATE("state"),
+  GREET("greet"),
   WITNESS("witness"),
   OPEN("open"),
   REOPEN("reopen"),
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/digitalcash/Output.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/digitalcash/Output.java
@@ -5,7 +5,7 @@ import com.google.gson.annotations.SerializedName;
 import java.util.Objects;
 
 // Object representing an output of this transaction
-public class Output {
+public final class Output {
   @SerializedName("value")
   private final long value; // the value of the output transaction, expressed in miniLAOs
 
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/digitalcash/Output.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/digitalcash/Output.java
@@ -1,0 +1,49 @@
+package com.github.dedis.popstellar.model.network.method.message.data.digitalcash;
+
+import com.google.gson.annotations.SerializedName;
+
+import java.util.Objects;
+
+// Object representing an output of this transaction
+public class Output {
+  @SerializedName("value")
+  private final int value; // the value of the output transaction, expressed in miniLAOs
+
+  @SerializedName("script")
+  private final Script_output script; // The script describing the TxOut unlock mechanism
+
+  /**
+   * @param value the value of the output transaction, expressed in miniLAOs
+   * @param script The script describing the TxOut unlock mechanism
+   */
+  public Output(int value, Script_output script) {
+    this.value = value;
+    this.script = script;
+  }
+
+  public int get_value() {
+    return value;
+  }
+
+  public Script_output get_script() {
+    return script;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) return true;
+    if (o == null || getClass() != o.getClass()) return false;
+    Output txOut = (Output) o;
+    return value == txOut.value && script.equals(txOut.script);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(get_value(), get_script());
+  }
+
+  @Override
+  public String toString() {
+    return "output{" + "value=" + value + ", script=" + script + '}';
+  }
+}
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/digitalcash/Output.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/digitalcash/Output.java
@@ -7,7 +7,7 @@ import java.util.Objects;
 // Object representing an output of this transaction
 public class Output {
   @SerializedName("value")
-  private final int value; // the value of the output transaction, expressed in miniLAOs
+  private final long value; // the value of the output transaction, expressed in miniLAOs
 
   @SerializedName("script")
   private final Script_output script; // The script describing the TxOut unlock mechanism
@@ -16,12 +16,12 @@ public class Output {
    * @param value the value of the output transaction, expressed in miniLAOs
    * @param script The script describing the TxOut unlock mechanism
    */
-  public Output(int value, Script_output script) {
+  public Output(long value, Script_output script) {
     this.value = value;
     this.script = script;
   }
 
-  public int get_value() {
+  public long get_value() {
     return value;
   }
 
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/digitalcash/Output.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/digitalcash/Output.java
@@ -10,22 +10,22 @@ public class Output {
   private final long value; // the value of the output transaction, expressed in miniLAOs
 
   @SerializedName("script")
-  private final Script_output script; // The script describing the TxOut unlock mechanism
+  private final ScriptOutput script; // The script describing the TxOut unlock mechanism
 
   /**
    * @param value the value of the output transaction, expressed in miniLAOs
    * @param script The script describing the TxOut unlock mechanism
    */
-  public Output(long value, Script_output script) {
+  public Output(long value, ScriptOutput script) {
     this.value = value;
     this.script = script;
   }
 
-  public long get_value() {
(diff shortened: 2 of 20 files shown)


fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/Action.java
```diff
@@ -12,7 +12,6 @@ public enum Action {
   SETUP("setup"),
   UPDATE("update_properties"),
   STATE("state"),
-  GREET("greet"),
```
Tesa27, line 15: Why are you deleting this, I think this is used for election but @ahominal will know more than me

Tesa27 (changes requested):
Looks good to me just a small question

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/Action.java
MaximeZmt: That was not intended (done by mistake). I've put it back.

Tesa27 (approved):

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/digitalcash/Output.java
```diff
@@ -0,0 +1,49 @@
...
+  /**
+   * @param value the value of the output transaction, expressed in miniLAOs
+   * @param script The script describing the TxOut unlock mechanism
+   */
+  public Output(int value, Script_output script) {
+    this.value = value;
+    this.script = script;
+  }
+
+  public int get_value() {
```
lourkeur, line 24:
```suggestion
  private final long value; // the value of the output transaction, expressed in miniLAOs

  @SerializedName("script")
  private final Script_output script; // The script describing the TxOut unlock mechanism

  /**
   * @param value the value of the output transaction, expressed in miniLAOs
   * @param script The script describing the TxOut unlock mechanism
   */
  public Output(long value, Script_output script) {
    this.value = value;
    this.script = script;
  }

  public long get_value() {
```
Since we have to support amounts bigger than `2^31`

ahominal (approved):
Seems good for me, be sure to test it with both backends before merge!

jbsv:
please fix the code smell

lourkeur (dismissed):
Good to go as far as I am concerned

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/security/PublicKey.java
```diff
@@ -33,4 +37,21 @@ public boolean verify(Signature signature, Base64URLData data) {
...
     }
   }
-}
+
+  /**
+   * Function that compute the hash of a public key
+   *
+   * @return String which correspond to the SHA256 Hash
+   */
+  public String computeHash() {
```
lourkeur, line 46: This is not complaint with the spec, see #1042 or the pubkey hash notebook

lourkeur (changes requested):

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/digitalcash/Transaction.java
```diff
@@ -0,0 +1,130 @@
...
+  /**
+   * Transaction constructor
+   * @param version The version of the transaction inputs
+   * @param inputs [Array[Objects]] array of output transactions to use as inputs
+   * @param outputs [Array[Objects]] array of outputs from this transactions
+   * @param lockTime TimeStamp
+   */
+  public Transaction(int version, List<Input> inputs, List<Output> outputs, long lockTime) {
+    this.version = version;
+    this.inputs = Collections.checkedList(inputs, Input.class);
```
lourkeur, line 37: Why do we need a mutable checked list here? wouldn't it make more sense to simply have an immutable list? This class is immutable, right? (I mean the signatures are mutable but everything else is immutable)

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/digitalcash/Transaction.java
```diff
@@ -0,0 +1,130 @@
...
+    // change the sig for all the inputs
+
+  }
+
+  public int getVersion() {
+    return version;
+  }
+
+  public List<Input> getInputs() {
+    return Collections.checkedList(inputs, Input.class);
```
lourkeur, line 49: Why do we need two layers of checked list?

fe2-android/app/src/test/java/com/github/dedis/popstellar/model/objects/security/KeyPairTest.java
```diff
@@ -44,4 +44,19 @@ public void getPublicKeyReturnsRightValue() {
 
     assertEquals(publicKey, keyPair.getPublicKey());
   }
+
+  @Test
+  public void pubKeyHash() {
+    // Tested with value in keypair.json (see #1042)
```
lourkeur, line 50: ideally it would be better to pull the value from the file programmatically. (ofc requires merging in #1043)

lourkeur (dismissed):
LGTM, all my remarks are optional and I am highly uninformed about Android development practices.

fe2-android/app/src/test/java/com/github/dedis/popstellar/model/objects/security/KeyPairTest.java
MaximeZmt: All others test are hardcoded yet, I mean the example will still true, so as there are not a lot of value I don't think this is a problem.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/digitalcash/Transaction.java
MaximeZmt: Will check later if there are any differences with other parts of the codebase for Object Array in JSON

fe2-android/app/src/test/java/com/github/dedis/popstellar/model/objects/security/KeyPairTest.java
lourkeur: ok fair enough. We aren't in the "ideally" realm anyways.

Tesa27 (dismissed):
Looks fine to me

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/digitalcash/Transaction.java
```diff
@@ -0,0 +1,130 @@
...
+      collectTransaction.add(currentTxin.getTxOutHash());
+      // TxOutIndex
+      collectTransaction.add(String.valueOf(currentTxin.getTxOutIndex()));
+    }
+
+    // lock_time
+    collectTransaction.add(String.valueOf(lockTime));
+    // Outputs
+    for (int i = 0; i < outputs.size(); i++) {
+      Output currentTxout = outputs.get(i);
```
GabrielFleischer, line 85: Here too

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/digitalcash/ScriptInput.java
```diff
@@ -0,0 +1,75 @@
...
+
+import java.util.Objects;
+
+// The script describing the unlock mechanism
+public class ScriptInput {
+  @SerializedName("type")
+  private final String type; // The script describing the unlock mechanism
+
+  @SerializedName("pubkey")
+  private final String pubKeyRecipient; // The recipient’s public key
```
GabrielFleischer, line 13: Would it be possible to ues the PublicKey type here ?
It would make it much more readable.

(The is already a rule to parse `base64 strings` -> `Public Key` for the json serialization)

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/digitalcash/ScriptInput.java
```diff
@@ -0,0 +1,75 @@
...
+
+  @SerializedName("sig")
+  private String sig; // Signature on all txins and txouts using the recipient's private key
```
GabrielFleischer, line 16: Same here but with a Signature type.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/digitalcash/Transaction.java
MaximeZmt: removed

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/digitalcash/Transaction.java
MaximeZmt: would be removed

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/digitalcash/Transaction.java
```diff
@@ -0,0 +1,128 @@
...
+   * Transaction constructor
+   * @param version The version of the transaction inputs
+   * @param inputs [Array[Objects]] array of output transactions to use as inputs
+   * @param outputs [Array[Objects]] array of outputs from this transactions
+   * @param lockTime TimeStamp
+   */
+  public Transaction(int version, List<Input> inputs, List<Output> outputs, long lockTime) {
+    this.version = version;
+    this.inputs = Collections.unmodifiableList(inputs);
+    this.outputs = Collections.unmodifiableList(outputs);
```
lourkeur, line 38: Typically, you would also defensively copy the list to avoid aliasing. That's a clean code thing, not a bug per se.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/digitalcash/Transaction.java
```diff
@@ -0,0 +1,128 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * The transaction object
+ */
+public class Transaction {
```
lourkeur, line 15:
```suggestion
public final class Transaction {
```
I think all classes in this PR could be final, assuming the back-end goes by the "Classes should be final by default" rule.

lourkeur (approved):
LGTM just nitpicks

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/digitalcash/Transaction.java
MaximeZmt: I am not sure to understand what you mean ? You think that instead of doing an unmodifiableList I should do a defensive copy ? 
I've looked for other json array implementation in the code (for instance ElectionResult or ElectionResultQuestion) but none of them does a defensive copy. 😅

lourkeur: You should do both.

Aliasing means there exists more than one pointer to the same object. In this case, you could have this:
```java
List<Inputs> aliased = ...;
var tx = new Transaction(..., aliased, ...);
aliased.add(...); // whoops
tx.getInputs(); // tx has been sneakily modified
```
So if we want immutability, we should defensive copy.

That said, feel free to align with the rest of the codebase and leave this concern out of scope.

GabrielFleischer (changes requested):
The test class names for Script input and output are partially in snake case.

Could you please address that ?

jbsv (approved):

GabrielFleischer (approved):
LGTM

PR #1042, opened 2022-05-20
[BE2] implement pubkey hash

lourkeur:
contains #1043. I did not use the `Hash` type because to me it implies that it's a `HashLen` hash, but maybe that's incorrect.


MaximeZmt (approved):
From a point of view of someone in Digital Cash team, everything looks good 👍

tuomas-p (approved):
As we discussed, looks good

PR #1044, opened 2022-05-20
Open election tests

Tesa27:
- This PR addresses first of all tests for open election message and adapts the flow from election setup to election end in the current karate tests.
- It also constructs an API that build different ids ( here vote ids, election ids and question ids) since we do not want to have them constant and not know how they where computed, this way if something changes in the computation it will be much easier to adapt, and the errors if they happen are easier to spot.


tests/karate/src/test/java/be/constants.feature
```diff
@@ -126,10 +158,30 @@ Feature: Constants
...
+      """
+        function(){
+          return createIsThisProjectFunVoteId("0")
+        }
+      """
+    * def createIsThisProjectFunVoteIdVoteNo =
       """
         function(){
-          return "d60B94lVWm84lBHc9RE5H67oH-Ad3O1WFflK3NSY3Yk="
+          return createIsThisProjectFunVoteId("1")
```
GabrielFleischer, line 184: I am not particularly up to date with the spec, but what is the 1 supposed to mean ?

tests/karate/src/test/java/be/election/electionOpen.feature
```diff
@@ -0,0 +1,122 @@
+
+CODE CLIMATE OPTIONS
+Show issues
+Show test coverage
+Highlight covered lines
```
GabrielFleischer, line 5: What is this ?

GabrielFleischer (dismissed):
LGTM
I just have a small question o the election. But this is just my lack of knowledge concerning the election protocol.

tests/karate/src/test/java/be/constants.feature
Tesa27: 1 is the index of the corresponding ballot option

Tesa27: Here 0 or 1 would correspond for voting yes or no respectively which are the ballot options available

tests/karate/src/test/java/be/election/electionOpen.feature
Tesa27: I do not have a slightest idea, I removed it now

GabrielFleischer (approved):
LGTM

PR #1046, opened 2022-05-21
created simple test that tests validity of roll call( sample tests)

Tesa27:
- Basic roll call create test


Tesa27:
Just a PR to test how to test validators, but since #1025 has all the roll call refactoring and testing the validator

PR #1048, opened 2022-05-22
Fe 2 - Digital Cash - Function in the viewmodel from digital cash that handle transaction

zoemarin:
Create some post transaction in the viewmodel and make some change in the send fragment 

Contain also the fragment pr !


ahominal:
Has this PR code from another PR that have already been opened ?

PR #1050, opened 2022-05-22
Enhanced election validator tests

Tesa27:
- Not the entire election validator was tested, especially the one with the standard actor
- election end and cast vote if invalid id are not by tests yes


tuomas-p (approved):
All good for me!

PR #1052, opened 2022-05-23
JAVA Frontend support for encryption

ahominal:
- Added support for encryption in while casting
- Added generation for ed25519 key pairs (for future  use)
- Added encryption using ed25519 curve under ElGamal
- Added decryption using ed25519 curve under ElGamal
- Force the user to cast only one vote at a time given new specifications


fe2-android/app/build.gradle
```diff
@@ -124,10 +123,12 @@ dependencies {
 
     // Security
     implementation "androidx.security:security-crypto:1.1.0-alpha03"
-    implementation 'net.i2p.crypto:eddsa:0.3.0'
+    implementation 'net.i2p.crypto:eddsa:0.2.0'
     implementation 'org.bouncycastle:bcpkix-jdk15on:1.65'
     implementation 'io.github.novacrypto:BIP39:2019.01.27'
     implementation 'com.google.crypto.tink:tink-android:1.5.0'
+    implementation 'ch.epfl.dedis:cothority:3.3.1'
```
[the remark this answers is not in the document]
ahominal: well, if you try with the 'net.i2p.crypto:eddsa:0.3.0' version it will almost surely crash!

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/election/CastVote.java
```diff
@@ -1,16 +1,16 @@
...
 import com.github.dedis.popstellar.model.network.method.message.data.Objects;
 import com.google.gson.annotations.SerializedName;
 
 import java.time.Instant;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.List;
 
-public class CastVote extends Data {
+public class CastVote <E> extends Data {
```
zoemarin, line 14: What does exactly this <E> mean ?

ahominal: It's a generic type, CastVote can be either encrypted or not. Type is specified when the class is instantiated

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/election/ElectionVote.java
```diff
@@ -34,15 +28,13 @@ public class ElectionVote {
    */
   public ElectionVote(
       String questionId,
-      List<Integer> vote,
+      Integer vote,
```
zoemarin, line 31: Why do we updated this in this new json schema ?

zoemarin (commented):
Seems fine to me :)

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Election.java
```diff
@@ -22,17 +28,23 @@
 public class Election extends Event {
 
   private Channel channel;
-  private String id;
+  private static String id;
```
Warhedgehog, line 31: Why is this field static ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Election.java
```diff
@@ -133,19 +138,33 @@ public Map<MessageID, PublicKey> getMessageMap() {
     return messageMap;
   }
 
-  public void putVotesBySender(PublicKey senderPk, List<ElectionVote> votes) {
+  public void putOpenBallotVotesBySender(PublicKey senderPk, List<ElectionVote> votes) {
```
Warhedgehog, line 141: I prefer plain as the antonym of encrypted, it's standard in cryptorgraphy discussions. Except if you agreed on all subsystsms to use open ofc

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Election.java
```diff
@@ -264,50 +289,111 @@ public static String generateElectionQuestionId(String electionId, String questi
    *     Hash('Vote'||election_id||question_id||(vote_index(es)|write_in))
    */
   public static String generateElectionVoteId(
-      String electionId,
-      String questionId,
-      List<Integer> voteIndex,
-      String writeIn,
-      boolean writeInEnabled) {
+          String electionId,
```
Warhedgehog, line 292: Same as above about formatting

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/security/Ed25519/ElectionPublicKey.java
```diff
@@ -0,0 +1,133 @@
...
+    public String encrypt(@NonNull byte[] message) {
+        if (message.length > 29) {
+            throw new IllegalArgumentException("The message should contain at maximum 29 bytes");
+        }
+        // Follows this implementation:
+        // https://github.com/dedis/cothority/blob/0299bcd78bab22bde6d6449b1594613987355535/evoting/lib/elgamal.go#L15-L23
+        // An example for embedding was found here:
+        // https://github.com/dedis/cothority/blob/0299bcd78bab22bde6d6449b1594613987355535/external/js/kyber/spec/group/edwards25519/point.spec.ts#L203
+
+        // M := cothority.Suite.Point().Embed(message, random.New())
```
Warhedgehog, line 96: Why is there what appears to be Go code commented here ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/election/ElectionEncryptedVote.java
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  @NonNull
+  public String getQuestionId(){
+    return questionId;
+  }
+
+  @Override
+  public int hashCode() {
+    return java.util.Objects.hash(getId(), getVote(), getQuestionId());
```
Tesa27, line 58: `Objectd.hash` seems more explicit to me, `java.util `makes this statement too crowded

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/election/ElectionVote.java
ahominal: The vote contain now an integer if OPEN_BALLOT, or a String id SECRET_BALLOT. The vote is unique, however, the list is kept we add vote with multiple options.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/election/ElectionVote.java
```diff
@@ -3,25 +3,19 @@
...
+   // Id of the object ElectionVote : Hash(“Vote”||election_id|| || JsonTestUtils.testData(castOpenVote);
   @SerializedName(value = "question")
   private final String questionId; // id of the question
 
-  private final List<Integer> vote; // list of indexes for the votes
-  private final boolean
-      writeInEnabled; // represents a boolean to know whether write_in is allowed or not
-
-  @SerializedName(value = "write_in")
-  private final String writeIn; // If write in is enabled this represents the writeIn string
```
Tesa27, line 24: Will this evict the `write_in `field from the json? Could fail json schema validation on the backend potentially since the `write_in` is in the specifications for now. A solution is to keep it and not use it or update the specification and ensure this does not cause problems with the json validator on the backend

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/LaoDetailViewModel.java
```diff
@@ -361,18 +363,30 @@ public void sendVote(List<ElectionVote> votes) {
...
-      CastVote castVote = new CastVote(votes, election.getId(), lao.getId());
-      Channel electionChannel = election.getChannel();
+      CastVote<ElectionVote> castOpenBallotVote = null;
+      CastVote<ElectionEncryptedVote> castEncryptedVote = null;
+
+      // Construct the cast vote depending if the messages need to be encrypted or not
+      if (election.getElectionVersion() == ElectionVersion.OPEN_BALLOT) {
+        castOpenBallotVote = new CastVote<>(votes, election.getId(), lao.getId());
+      } else {
+        Log.d(TAG, "Encrypting the vote ...");
```
Warhedgehog, line 373: Small detail but these logs are for debugging purposes. This looks more like communication to the user, in which case you should use a toast

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/election/ElectionVote.java
Tesa27: Same question, would this be a problem if we have multiple questions?

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/event/election/fragments/CastVoteFragment.java
```diff
@@ -45,10 +45,11 @@ public class CastVoteFragment extends Fragment {
         for (int i = 0; i < electionQuestions.size(); i++) {
           ElectionQuestion electionQuestion = electionQuestions.get(i);
           List<Integer> votes = mLaoDetailViewModel.getCurrentElectionVotes().getValue().get(i);
+          // Only one vote should be selected.
           ElectionVote electionVote =
               new ElectionVote(
                   electionQuestion.getId(),
-                  votes,
+                  votes.get(0),
```
Warhedgehog, line 52: Why do you get only the vote of the 1 st question ?

Warhedgehog (changes requested):
This is a partial review from me, in particular I haven't spent the time needed to understand the encryption part and the tests. 
From what I've seen this seems good work ! A few things I'd like discussed/changed

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Election.java
```diff
@@ -189,8 +208,14 @@ public String computerRegisteredVotes() {
...
-        listOfVoteIds.add(vote.getId());
+      if (getElectionVersion() == ElectionVersion.OPEN_BALLOT) {
+        for (ElectionVote vote : openVoteMap.get(senderPk)) {
+          listOfVoteIds.add(vote.getId());
+        }
+      } else {
+        for (ElectionEncryptedVote vote : encryptedVoteMap.get(senderPk)) {
+          listOfVoteIds.add(vote.getId());
+        }
       }
```
Tesa27, line 219:
```suggestion
     Map<PublicKey, List<ElectionVote>> voteMap = getElectionVersion() == ElectionVersion.OPEN_BALLOT ? openVoteMap :  encryptedVoteMap;
        for (ElectionVote vote :  voteMap.get(senderPk)) {
          listOfVoteIds.add(vote.getId());
        }
```

Tesa27: I think it is pretty messy the way it is done right now, here is a suggestion to make it more clean

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/security/Ed25519/ElectionPrivateKey.java
```diff
@@ -0,0 +1,102 @@
...
+        ElectionPrivateKey that = (ElectionPrivateKey) o;
+        return that.getPrivateKey().equals(getPrivateKey());
+    }
+
+    /**
+     * Added this hash behavior by default
+     */
+    @Override
+    public int hashCode() {
+        return java.util.Objects.hash(getPrivateKey().toString());
```
Tesa27, line 52: omit `java.util` in my opinion

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/election/ElectionEncryptedVote.java
```diff
@@ -0,0 +1,89 @@
...
+  // Id of the question
+  @SerializedName(value = "question")
+  private final String questionId;
+
+  // Vote array containing index corresponding to ballot options
+  private final String vote;
+
+  /**
+   * @param questionId       id of the question
+   * @param encryptedVote    encrypted uniaue indice of the chosen vote
```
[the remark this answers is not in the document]
ahominal: First for the generic type, this is the only solution i found that is compatible with the serialization. I didn't manage to find a solution to properly serialize and deserialize it. On top of this, cast vote needs to have some functions to be implemented that are used by the rest of the code. If i make them abstract, there conflicts everywhere on the code (type errors) and I couldn't  find a solution. This is the 'easiest' solution I found.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/election/ElectionVote.java
```diff
@@ -3,25 +3,19 @@
 import com.github.dedis.popstellar.model.objects.Election;
 import com.google.gson.annotations.SerializedName;
 
-import java.util.Arrays;
-import java.util.List;
-
-public class ElectionVote {
+public class ElectionVote{
```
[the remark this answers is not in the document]
ahominal: It should be, but apparently not. good that you noticed

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/security/Ed25519/ElectionPublicKey.java
```diff
@@ -0,0 +1,133 @@
...
+
+    /**
+     * Encrypts with ElGamal under Ed25519 curve
+     *
+     * @param message message a 2 byte integer corresponding to the chosen vote
+     * @return Returns 2 32 byte strings which should be appended
+     * together and base64 encoded. For clarification, the first 32 bytes is point K and the second 32 bytes is point C.
+     */
+    public String encrypt(@NonNull byte[] message) {
+        if (message.length > 29) {
```
Tesa27, line 88: why 29?

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/security/Ed25519/ElectionPublicKey.java
ahominal: Originally, the library is written in go. The Java library is just a translation of the go library. I took the examples from the go library

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/LaoDetailViewModel.java
ahominal: This seems more logical! dont know why I put a log there

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/event/election/fragments/CastVoteFragment.java
ahominal: because now the vote is unique, even if the user selects multiple questions, it will discard them and take only one vote, the first.

Tesa27 (changes requested):
Good job overall, there are some comments here and there.
I would suggest to have a test where you encrypt a plaintext than decrypt the cyphertext and check that it corresponds to the original plaintext.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/security/Ed25519/ElectionPublicKey.java
Warhedgehog: My question is why did you put the go code in comments here. IMO, providing the link is sufficient

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Election.java
ahominal: It won't work, because openVoteMap is List<ElectionVote> and encryptedVoteMap is List<ElectionEncryptedVote> but thanks for suggestions !

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/election/ElectionEncryptedVote.java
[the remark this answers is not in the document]
Warhedgehog: Okay, I think we'll need the TA's advice here

Warhedgehog:
Also there seems to be an issue with the Go schema validators that would need fixed before merging

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/security/Ed25519/ElectionPublicKey.java
ahominal: I found this number on the library, if you  have higher number, it will not properly embed.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Election.java
ahominal: Good question, ill change it.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/security/Ed25519/ElectionPublicKey.java
ahominal: Ohh, just that the user can compare the java and the go code, it seems more straightforward (at least for me)

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Election.java
Tesa27: l.44-46 says that `openVoteMap `and` encryptedVoteMap` are Maps, and calling a variable that ends with Map in the naming is bad practice :) 
```
  private final Map<PublicKey, List<ElectionVote>> openVoteMap;
  // Map that associates each sender pk to their encrypted votes
  private final Map<PublicKey, List<ElectionEncryptedVote>> encryptedVoteMap;
```

Tesa27:
> Also there seems to be an issue with the Go schema validators that would need fixed before merging

 I think this is due to the` write_in` field bein removed

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/event/election/fragments/CastVoteFragment.java
Warhedgehog: Just to clarifiy to be sure there is no misunderstanding : 

The user can't select multiple questions, he can vote on (for now) 1 ballot option per question. The organizer when he sets up the election sets the number of questions (>=1) and the number (>=2) and content of ballots options. Of course for the plurality vote the user cannot vote for more than one option per question. 
The multiple question thing is like having the possibility to vote for national and regional elections in the same event. Each question (in this case could regional or national) has its set of ballots options, and the user can vote for one option of each

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Election.java
ahominal: what do you mean by antonym?

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/security/Ed25519/ElectionPublicKey.java
Tesa27: I would say that the Java programmer should not even be aware how the Go code is structured since for the developer the backend should be taken as a black box, so this Go code should be removed as Johann suggested

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/event/election/fragments/CastVoteFragment.java
ahominal: I specifically it for the new cast vote protocol here https://github.com/dedis/popstellar/pull/918. The ElectionVote now contains only one integer instead of of list of integers. Do you think my implementation has a problem ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/election/ElectionVote.java
ahominal: The write in is only used while encrypting the hash. If i use it, it will create problem while I serialize my class.  I think they didnt want to keep it in the protocol since it is probably not a big priority.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Election.java
ahominal: oh I see, i refactored the name !

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/election/ElectionVote.java
Tesa27: I agree, write_in doe snot bring anything in the current state of the project bit is still checked in the validator of json schemas on the backends and if write_in is not find any cast vote message will not adhere to the schema and will be rejected by the backend right? If you already addressed this even better, but than the specs should be updated

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/event/election/fragments/CastVoteFragment.java
Warhedgehog: I think so. It is more subtle than I realised at first. It is probably that the current implementation requires more adaptations to fit the new protocol.  I think that we should discuss in more details. I don't have time until tomorrow afternoon unfortunately.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Election.java
Warhedgehog: Contrary to. like using plain (and not open) as opposed to encrypted. Like encryptedVote - plainVote (instead of openVote)

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Election.java
Tesa27: I might not see what you are refering to but at line 44 and 46 I say that the 2 variables `openVoteMap` and `encryptedVoteMap` are declared as maps, so no need to rename IMO since the list is obtainged with `openVoteMap.get(senderPk)` for instance

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/election/ElectionVote.java
ahominal: Currently, the Json scheme doesnt need the write in, so the backend shouldnt reject it.

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/event/election/fragments/CastVoteFragment.java
ahominal: i will be glad to discuss it yeah !

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Election.java
ahominal: We already chose the OPEN_BALLOT for describing an open_ballot election. I thought that using the same convention would lead to less confusion.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/election/ElectionVote.java
Tesa27: That's good news!

ahominal: i mark it as resolved

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/LaoDetailViewModel.java
```diff
@@ -357,18 +358,30 @@ public void sendVote(List<ElectionVote> votes) {
...
-                  () -> {
-                    Log.d(TAG, "sent a vote successfully");
+              networkManager
+                      .getMessageSender()
+                      .publish(token, electionChannel,
+                              election.getElectionVersion() == ElectionVersion.OPEN_BALLOT ? castOpenBallotVote : castEncryptedVote)
+                      .doFinally(this::openLaoDetail)
+                      .subscribe(
+                              () -> {
+                                Log.d(TAG, "sent a vote successfully");
```
GabrielFleischer, line 384: This is weird.

The control flow is hard to understand.
Could you try to create sub-functions such that the code is easier to read ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Election.java
```diff
@@ -266,50 +269,126 @@ public static String generateElectionQuestionId(String electionId, String questi
...
+      if (getElectionVersion() == ElectionVersion.OPEN_BALLOT) {
+        for (ElectionVote vote : openVoteByPublicKey.get(senderPk)) {
+          listOfVoteIds.add(vote.getId());
+        }
+      } else {
+        for (ElectionEncryptedVote vote : encryptedVoteByPublicKey.get(senderPk)) {
+          listOfVoteIds.add(vote.getId());
+        }
+      }
+    }
```
GabrielFleischer, line 328: There is a bit of code duplication here.

I'm certain you can find a way to modularize this.

fe2-android/app/src/test/java/com/github/dedis/popstellar/model/network/method/message/data/election/CastVoteTest.java
```diff
@@ -21,43 +27,94 @@ public class CastVoteTest {
   private final String electionId = " myElection";
   private final boolean writeInEnabled = false;
   private final String write_in = "My write in ballot option";
+  Gson GSON = JsonModule.provideGson(DataRegistryModule.provideDataRegistry());
```
GabrielFleischer, line 30: This variable should be `private final` and probably `static`

fe2-android/app/src/test/java/com/github/dedis/popstellar/model/objects/security/Ed25519/ElectionKeyPairTest.java
```diff
@@ -0,0 +1,50 @@
...
+      // Decrypt
+      byte[] decryptedData = electionPrivateKey.decrypt(encryptedData);
+      Log.d(
+          "Private base64 encoded key : ",
+          new Base64URLData(electionPrivateKey.toString()).getData().toString());
+      Log.d(
+          "Public base64 encoded key : ",
+          new Base64URLData(electionPublicKey.toString()).getData().toString());
+      // Pad the decrypted data and observe the result
+      int decryptedINt = ((decryptedData[1] & 0xff) << 8) | (decryptedData[0] & 0xff);
```
GabrielFleischer, line 45: There is a typo here :`
```suggestion
      int decryptedInt = ((decryptedData[1] & 0xff) << 8) | (decryptedData[0] & 0xff);
```

GabrielFleischer (changes requested):
This looks great !

Please fix the 3 codesmells left and reformat the code.

I also left some comments that you should take into account. Especially, I do think that the way you deserialize a Json is not particualry robust. I would like you to fix that.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Election.java
```diff
@@ -180,26 +192,17 @@ public long getEndTimestamp() {
...
-      return "";
-    } else {
-      return Hash.hash(listOfVoteIds.toArray(new String[0]));
+    if (votes == null || votes.isEmpty()) {
+      throw new IllegalArgumentException("Encrypted votes cannot be null or empty");
     }
+    // The list must be sorted by order of question ids
+    List<ElectionEncryptedVote> votesCopy = new ArrayList<>(votes);
+    votesCopy.sort(Comparator.comparing(ElectionEncryptedVote::getQuestionId));
+    encryptedVoteByPublicKey.put(senderPk, votesCopy);
```
Tyratox, line 205: Sorry for the misunderstanding, it should be ordered by `vote_id`, not question id.

As described here https://github.com/dedis/popstellar/pull/1081/files#diff-b577fe0224ef66435ee10949840f99a9f91194dc18e0138c16fe15f5c8190500

or implemented here:

https://github.com/dedis/popstellar/pull/1081/files#diff-66a954b5689d1a6ede24f210ef79643d6fa6c904997e5d07634b6a58abe4997d

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/serializer/JsonDataSerializer.java
```diff
@@ -51,7 +56,11 @@ public Data deserialize(JsonElement json, Type typeOfT, JsonDeserializationConte
               + action.getAction()
               + ") does not exists in the protocol");
     }
-
+    // Due to the generic type of the vote, CastVote message needs a special
+    // deserialization
+    if (action == Action.CAST_VOTE) {
+      return customCastVoteDeserializer(json, context);
+    }
```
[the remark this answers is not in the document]
ahominal: I tried this way, and i get infinite loop while trying to call context.deserialize(json, clazz.get());, i really don't know how is it achievable :(

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Election.java
ahominal: Hmm, i don't know how to achieved, i could write two functions but I dont know if thats very useful

GabrielFleischer (approved):
LGTM

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/serializer/JsonDataSerializer.java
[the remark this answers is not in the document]
GabrielFleischer: Well, let's skip that for now

Warhedgehog (approved):
Approving the pop party

PR #1055, opened 2022-05-24
[BE2-SCALA] ElGamal encryption and decryption

rombirli:
- Encryption and decryption seems to work


be2-scala/src/main/scala/ch/epfl/pop/model/objects/PublicKey.scala
```diff
@@ -1,5 +1,37 @@
...
 final case class PublicKey(base64Data: Base64Data) {
+  val asPoint = new Ed25519Point(base64Data.decode())
+
+
+  def encrypt(messageB64: Base64Data): Base64Data = {
+    val message = messageB64.decode()
+    if (message.length > 29) throw new IllegalArgumentException("The message should contain at maximum 29 bytes")
+    val M = Ed25519Point.embed(message)
+    val seed = new Array[Byte](Ed25519.field.getb / 8)
+    new Random().nextBytes(seed)
```
lourkeur, line 16: >Instances of java.util.Random are not cryptographically secure. Consider instead using [SecureRandom](https://docs.oracle.com/javase/8/docs/api/java/security/SecureRandom.html) to get a cryptographically secure pseudo-random number generator for use by security-sensitive applications.

rombirli: resolved

PR #1058, opened 2022-05-25
Support for encrypted vote casting

Robingoumaz:
Added support for encrypted vote casting :

- Change to vote_cast_vote messagedata to support both int and string for actual votes
- Change to valid_votes to accept both int index and encrypted string index
- Added decryption algorithm to decrypt tallied votes
- Added encryption algorithm to test decryption


MaximeZmt (commented):
I've done a first check for the go files, I just left one comment below.

be1-go/message/messagedata/vote_cast_vote.go
```diff
@@ -15,9 +20,38 @@ type VoteCastVote struct {
...
+	ID       string
+	Question string
+	Vote     interface{}
+}
+
+type tempVote struct {
+	ID       string      `json:"id"`
+	Question string      `json:"question"`
+	Vote     interface{} `json:"vote"`
+}
```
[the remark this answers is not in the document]
Robingoumaz: We need both because I changed the unmarshalling function and it without the tempVote it would go into an infinite loop. I should probably add a comment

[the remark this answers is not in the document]
MaximeZmt: Ok, yes it may be a good idea to comment it.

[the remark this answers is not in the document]
MaximeZmt: Also, why you don't have the `json: ??? ` for the all Vote fields ? 😅 
Finally,  maybe also write "TempVote" to follow the codebase naming convention.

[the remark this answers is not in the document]
MaximeZmt: oh ok didn't know, I understand now.
I think you should really comment it as it differs from other parts of the code. Otherwise looks good 👍

Warhedgehog (commented):
Great work. As it is not trivial more comments would be nice

be1-go/channel/election/mod.go
```diff
@@ -105,8 +105,8 @@ type validVote struct {
 	// voteTime represents the time of the creation of the vote
 	voteTime int64
 
-	// indexes represents the indexes of the ballot options
-	indexes []int
```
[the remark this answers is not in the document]
Robingoumaz: It is due to a change in the specifications.
We encrypted votes by encrypting the index inside the array, but that gave away information on the sender in the case of multiple choice questions. 
We cannot encrypt the array in itself at least with the current specification because it could grow arbitrarly in size and overflow the 64 bytes we have for the vote size.

Tyratox:
Works with fe-1 `work-fe1-evoting-request-election-key` ✅

be1-go/channel/election/mod.go
```diff
@@ -681,6 +684,70 @@ func gatherResults(questions map[string]*question,
 	return resultElection, nil
 }
 
+// getVoteIndex extracts the index of the vote from the validVotes
+func (c *Channel) getVoteIndex(vote validVote) (int, error) {
+	switch m := c.electionType; m {
```
nkcr, line 689: Avoid doing it in one line, for clarity: 

```go
elecType := c.electionType
switch elecType {
```

be1-go/channel/election/mod.go
```diff
@@ -681,6 +684,70 @@ func gatherResults(questions map[string]*question,
...
+	// open ballot votes have the index in plain text
+	case messagedata.OpenBallot:
+		index, _ := vote.index.(int)
```
nkcr, line 692: what if the cast fails ? You should check :

```go
index, ok := ...
if !ok {
  return xerrors....
```

nkcr: same bellow

be1-go/channel/election/mod.go
```diff
@@ -681,6 +684,70 @@ func gatherResults(questions map[string]*question,
...
+		return index, nil
+	// secret ballot votes must be decrypted to get the index
+	case messagedata.SecretBallot:
+		temp, _ := vote.index.(string)
+		index, err := c.decryptVote(temp)
+		if err != nil {
+			// logs a warning because we don't stop the tallying for non-conform votes
+			// we just disregards them
+			c.log.Warn().Msgf("failed to decrypt a vote: %v", err)
+			return index, err
```
nkcr, line 702: we always wrap errors

be1-go/channel/election/mod.go
```diff
@@ -681,6 +682,57 @@ func gatherResults(questions map[string]*question,
...
+func (c *Channel) decryptVote(vote string) (int, error) {
+	votebuf, err := base64.URLEncoding.DecodeString(vote)
+	if err != nil {
+		return -1, answer.NewErrorf(-4, "vote %s is not base64 encoded", vote)
+	}
+
+	K := crypto.Suite.Point()
+	C := crypto.Suite.Point()
+
+	err = K.UnmarshalBinary(votebuf[:32])
```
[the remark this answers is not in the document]
nkcr: good comment !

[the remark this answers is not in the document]
nkcr: Additionally, there should be a check that `len(voteBuf) == 64`

be1-go/channel/election/mod.go
```diff
@@ -681,6 +684,70 @@ func gatherResults(questions map[string]*question,
...
+	S := crypto.Suite.Point().Mul(c.secElectionKey, K)
+	data, err := crypto.Suite.Point().Sub(C, S).Data()
+	if err != nil {
+		return -1, answer.NewErrorf(-4, "vote data is invalid")
+	}
+
+	var index int16
+
+	// interprets the data as a big endian int
+	buf := bytes.NewReader(data)
```
nkcr, line 742: `index := binary.BigEndian.Uint16(data)` is a bit more compact

be1-go/channel/election/mod.go
nkcr: or directly `switch c.electionType` if you don't need it inside your cases

be1-go/channel/election/verification.go
```diff
@@ -254,6 +246,44 @@ func verifyRegisteredVotes(electionEnd messagedata.ElectionEnd,
...
+		if !ok {
+			return answer.NewErrorf(-4, "votes in secret ballot should be string")
+		}
+
+		temp, err := base64.URLEncoding.DecodeString(vs)
+		if err != nil {
+			return answer.NewErrorf(-4, "vote should be base64 encoded")
+		}
+
+		if l := len(temp); l != 64 {
```
nkcr, line 274: we avoid initialization in "if" statement for clarity. Do it in two steps.

be1-go/message/messagedata/vote_cast_vote.go
```diff
@@ -13,11 +18,44 @@ type VoteCastVote struct {
...
 
-// Vote defines a vote of a cast vote
+// Vote defines a vote of a cast vote, Vote.Vote can be an int or a string
 type Vote struct {
-	ID       string `json:"id"`
-	Question string `json:"question"`
-	Vote     []int  `json:"vote"`
+	ID       string
+	Question string
+	Vote     interface{}
```
nkcr, line 25: there you should explain the possible values of `Vote`

be1-go/message/messagedata/vote_cast_vote.go
```diff
@@ -13,11 +18,44 @@ type VoteCastVote struct {
...
+	// checking that Vote.Vote is either an int or a string
+	switch t := vTemp.Vote.(type) {
+	// json unmarshalls numbers only into float64, we have to make additional checks
+	case float64:
+		i := int(t)
+		if float64(i) != t {
+			return answer.NewErrorf(-4, "invalid vote type, should be int but was float")
+		}
+		*v = Vote{vTemp.ID, vTemp.Question, i}
+		return nil
```
nkcr, line 52: not needed

be1-go/message/messagedata/vote_cast_vote.go
```diff
@@ -13,11 +18,44 @@ type VoteCastVote struct {
...
+	case string:
+		*v = Vote(vTemp)
+		return nil
```
nkcr, line 55: not needed

be1-go/channel/election/mod.go
Robingoumaz: I already check during the verification of the cast_votes, should I check it here too ?

be1-go/channel/election/mod.go
Robingoumaz: I tried with it, but it panics if the data is empty.

nkcr (approved):

Issue #982, opened 2022-05-03
[BUG] election result graph is confusing


lourkeur:
### Description (Actual behavior)
When election results are displayed, the graph does not start at 0 and some numbers are repeated.
![image](https://user-images.githubusercontent.com/15657735/166481242-329d4d63-14a3-4b17-9b28-03ec1bffc4ac.png)

### Expected behavior
A user-friendly histogram

### How to reproduce
1. Hold an election in a LAO
2. Look at results in FE-1

### Version & Environment
This bug was reproduced on:
- 8177a176492daa6097a4f533ae3c05840f968344
- 2020-05-03 PoP party

##### Front-ends:
- [x] Fe1-Web (please include browser's names & version)
- [ ] Fe2-Android (please specify if phone or emulation, and Android version)
- [ ] Not applicable

##### Back-ends:
- [ ] Be1-Go
- [ ] Be2-Scala
- [x] Not applicable

##### Environment (as applicable):
 - Device: HP 470 G7 Laptop
 - OS: NixOS Linux unstable
 - Browser Firefox 99.0.1
 - 
### Workaround
You have to read the graph closely

### Impact
Not user-friendly

PR #1061, opened 2022-05-28
Changes to the UI: Updated LAO screens including Identity, Notification and Events

Tyratox:
This reworks most of the remaining screens that were not covered by #1038.
There is still a lot of work that needs to be done but this PR tries to make the UI consistent and somewhat clean. Making it fancy would require more work.

## Lao List
Screenshot 2022-05-28 at 10 22 39

## Lao Home
Screenshot 2022-05-28 at 10 22 48

Screenshot 2022-05-28 at 10 22 51

## Notifications
Screenshot 2022-05-28 at 10 23 00

## Events
Screenshot 2022-05-28 at 10 24 05

Screenshot 2022-05-28 at 10 23 21
Screenshot 2022-05-28 at 10 23 35
Screenshot 2022-05-28 at 10 23 57

Screenshot 2022-05-28 at 10 26 43

## Roll Call
Screenshot 2022-05-28 at 10 26 54
Screenshot 2022-05-28 at 10 26 57
Screenshot 2022-05-28 at 10 27 00

 
Screenshot 2022-05-28 at 10 27 32
Screenshot 2022-05-28 at 10 27 37

Screenshot 2022-05-28 at 10 27 43

## Election
Screenshot 2022-05-28 at 10 27 59
Screenshot 2022-05-28 at 10 28 01

Screenshot 2022-05-28 at 10 28 39
Screenshot 2022-05-28 at 10 28 45


fe1-web/.eslintrc.js
```diff
@@ -103,7 +103,5 @@ module.exports = {
     // enforce the use of color variables. this should make programmers use the defined color palette
     // or extend it if really necessary but not add new colors for every new component
     "react-native/no-color-literals": 'error',
-    // react native cannot handle text outside of <Text/> components. this only works on the web
-    "react-native/no-raw-text": 'error',
```
Tyratox, line 107: Unfortunately this leads to false positives if some other component already wraps its children in a <Text/>.

fe1-web/src/features/rollCall/screens/ViewSingleRollCall.tsx
```diff
@@ -0,0 +1,350 @@
+import { CompositeScreenProps, useRoute } from '@react-navigation/core';
+import { useNavigation } from '@react-navigation/native';
```
Tyratox, line 2: I think this file should in a future PR be split as in the case of the election

fe1-web/src/features/notification/screens/NotificationScreen.tsx
```diff
@@ -98,58 +51,132 @@ const NotificationScreen = () => {
...
+                <View style={List.icon}>
+                  <NotificationType.Icon size={Icon.size} color={Color.primary} />
+                </View>
+                <ListItem.Content>
+                  <ListItem.Title style={Typography.base}>{notification.title}</ListItem.Title>
+                  <ListItem.Subtitle style={Typography.small}>
+                    <ReactTimeago date={notification.timestamp * 1000} />
+                  </ListItem.Subtitle>
+                </ListItem.Content>
+              </ListItem>
```
Xelowak, line 103: I feel like it is very similar for both notification types. It could be nice to create a function to remove this duplication if it doesn't introduce too many arguments.

fe1-web/src/features/rollCall/screens/ViewSingleRollCall.tsx
Xelowak: Yes that's a great idea given the number of lines.

fe1-web/src/features/rollCall/screens/ViewSingleRollCall.tsx
```diff
@@ -0,0 +1,356 @@
...
+        screen: STRINGS.navigation_lao_events,
+        params: {
+          screen: STRINGS.navigation_lao_events_open_roll_call,
+          params: { rollCallId: rollCall.id.toString() },
+        },
+      });
+    } else {
+      makeToastErr('Unable to scan attendees, the event does not have an idAlias');
+      console.debug('Unable to scan attendees, the event does not have an idAlias');
+    }
```
Xelowak, line 297: Since these strings will be displayed in the UI (toast message), I guess you can add them in `string.ts`.

fe1-web/src/features/rollCall/screens/__tests__/ViewSingleRollCall.test.tsx
```diff
@@ -0,0 +1,150 @@
...
+  [ROLLCALL_FEATURE_IDENTIFIER]: {
+    useCurrentLaoId: () => mockLaoIdHash,
+    makeEventByTypeSelector,
+    generateToken,
+    hasSeed: () => getWalletState(mockStore.getState()).seed !== undefined,
+  } as RollCallReactContext,
+};
+
+jest.spyOn(Date.prototype, 'toLocaleDateString').mockReturnValue('2022-05-28');
+jest.spyOn(Date.prototype, 'toLocaleTimeString').mockReturnValue('00:00:00');
```
Xelowak, line 76: For your information, you can do this by setting the local time of the system, like in [here](https://github.com/dedis/popstellar/blob/master/fe1-web/src/features/social/components/__tests__/ChirpCard.test.tsx).

Xelowak (commented):
Thanks a lot for all of this! I added a few comments but overall this is great work.

I'd suggest adding descriptive comments at the beginning of all files you added (like `ElectionListItem`), so that it's easier to understand what it does and how to re-use it if needed. It can be relevant in files that contain multiple components, too.

I've not tested the UI myself so I trust you about the fact that everything works fine. However, I saw on your 5th screenshot (about events) that there is a dollar sign before the date and I imagine it's not intended.

Maybe it could be nice to check if you can create private functions or components in some classes/components, because I feel like the code you wrote (especially those `.map`) is very similar in certain places.

Given the amount of code to review, I'm quite sure that I didn't catch everything but since the quality of your code is really good it should be ok :). If something breaks, you still have time to fix it until the final deadline anyway.

fe1-web/src/features/rollCall/screens/__tests__/ViewSingleRollCall.test.tsx
Tyratox: I don't think this will work. While this will set the time, it will not make the output consistent. `toLocale...` functions depend on the browser / node settings which means the CI can suddenly output something else and the test will break.

fe1-web/src/features/evoting/components/ElectionTerminated.tsx
```diff
@@ -1,19 +1,66 @@
...
+import { ListItem } from 'react-native-elements';
 
-import { TimeDisplay } from 'core/components';
-import { Typography } from 'core/styles';
+import ScreenWrapper from 'core/components/ScreenWrapper';
+import { List, Typography } from 'core/styles';
 
 import { Election } from '../objects';
 
 const ElectionTerminated = ({ election }: IPropTypes) => {
```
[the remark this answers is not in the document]
Tyratox: Yes I think a lot of the duplication will go away by creating a ElectionQuestion component 👍

fe1-web/src/features/evoting/screens/ViewSingleElection.tsx
```diff
@@ -0,0 +1,102 @@
...
+  const election = useSelector(selectElection);
+
+  if (!election) {
+    throw new Error(`Could not find a roll call with id ${electionId}`);
+  }
+
+  return <Text style={Typography.topNavigationHeading}>{election.name}</Text>;
+};
+
+export const ViewSinglRollCallScreenRightHeader = () => {
```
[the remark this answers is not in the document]
Tyratox: Oh oops, thank you!! 😬

fe1-web/src/features/home/interface/Configuration.ts
```diff
@@ -86,14 +86,11 @@ export interface HomeInterface extends FeatureInterface {
     /**
      * Given the lao server address and the lao id, this computes the data
```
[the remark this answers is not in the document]
Tyratox: 👍

fe1-web/src/features/home/objects/ConnectToLao.ts
```diff
@@ -37,6 +37,16 @@ export class ConnectToLao {
 
     return new ConnectToLao({
       ...obj,
+      /* in the future the qr code should encode a list of servers */
+      servers: [obj.server],
```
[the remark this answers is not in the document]
Tyratox: Hmmm..  Apparently QR codes can hold up to 3 KB of information. I guess we would need to compute the size in bytes to throw an accurate warning as it will also depend on the length of the server addresses.. I think this should be doable using Buffers 🤔

fe1-web/src/features/notification/screens/NotificationScreen.tsx
Tyratox: I'll try it :)

fe1-web/src/features/home/objects/ConnectToLao.ts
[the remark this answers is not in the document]
Tyratox: Hmm okay apparently it is not so easy because it also depends on the error correction level and the version used 😬

Tyratox:
Thank you for the review! 😁

> I'd suggest adding descriptive comments at the beginning of all files you added (like ElectionListItem), so that it's easier to understand what it does and how to re-use it if needed. It can be relevant in files that contain multiple components, too.

On it!

> I've not tested the UI myself so I trust you about the fact that everything works fine. However, I saw on your 5th screenshot (about events) that there is a dollar sign before the date and I imagine it's not intended.

Yes, I realized this too. The screenshots are not up to date since there were a few more issues I noticed. 😅

> Maybe it could be nice to check if you can create private functions or components in some classes/components, because I feel like the code you wrote (especially those .map) is very similar in certain places.

I just spent 1.5 hours trying to generalize the list component but unfortunately it is not trivial at all without sacrificing generality. (I wanted to get rid of the `List.getListItemStyles(..)` but this requires a general component to know the number of times which means we have to pass an array of components to it instead of of simple react children. This then again means we cannot apply the `key` property without a lot of react trickery and magic which probably is not stable 😅)

> Given the amount of code to review, I'm quite sure that I didn't catch everything but since the quality of your code is really good it should be ok :). If something breaks, you still have time to fix it until the final deadline anyway.

Haha yes! :) I will make sure to test all actions in the UI before the deadline!!

Tyratox:
I just added comments to these files :)

Tyratox:
Will fix #982 by replacing the graph with showing the number of votes for each option. In the future it is possible to re-introduce a graph but readability should be considered this time around.

Xelowak (approved):
I've had a quick look at everything you changed and it seems fine to me.

Regarding the function that I suggested to add, even if it's not possible to re-use them across files, maybe it could be nice to split a little bit the code when the return is large ( like in `ElectionOpened`). It would make it clearer to understand which part renders which part of the component/screen instead of having a lot of code directly in the return. 

I'm already approving if you want to treat it later, but I think it would be great.
You still have failing tests though, but I don't know why.

PR #1062, opened 2022-05-28
BUG FIX: RollCall validator & handler

Ajkunas:
Continuity of the PR #1025:
- Fixed issue of opening a non created roll call OR non closed roll call, i.e, we cannot open an already opened roll call
- Fixed issue of closing a non opened roll call
- We can now check the validity of the "opens" and "closes" attribute of OpenRollCall and CloseRollCall


be2-scala/src/main/scala/ch/epfl/pop/storage/DbActor.scala
```diff
@@ -363,6 +415,31 @@ object DbActor {
...
+   *
+   * @param laoId    the channel we need the Rollcall's data for
+   * @param message  rollcall message sent through the channel
+   */
+  final case class WriteRollCallData(laoId: Hash, message: Message) extends Event
+
+  /**
+   * Request to create election data in the db with an id and a keypair
+   *
+   * @param laoId    unique id of the lao in which aret the rollcall messages
```
tuomas-p, line 437:
```suggestion
   * @param laoId    unique id of the lao in which the rollcall messages are
```

be2-scala/src/main/scala/ch/epfl/pop/storage/DbActor.scala
```diff
@@ -363,6 +415,31 @@ object DbActor {
...
+  /**
+   * Request to read the rollcallData of the LAO, with key laoId
+   *
+   * @param laoId    the channel we need the Rollcall's data for
+   * @param message  rollcall message sent through the channel
+   */
+  final case class WriteRollCallData(laoId: Hash, message: Message) extends Event
+
+  /**
+   * Request to create election data in the db with an id and a keypair
```
tuomas-p, line 435: rollcall data right?

be2-scala/src/main/scala/ch/epfl/pop/storage/DbActor.scala
```diff
@@ -363,6 +415,31 @@ object DbActor {
...
+  /**
+   * Request to read the rollcallData of the LAO, with key laoId
+   *
+   * @param laoId
+   * the channel we need the Rollcall's data for
+   */
+  final case class ReadRollCallData(laoId: Hash) extends Event
+
+  /**
+   * Request to read the rollcallData of the LAO, with key laoId
```
tuomas-p, line 427:
```suggestion
   * Request to write the rollcallData of the LAO, with key laoId
```

tuomas-p (changes requested):
Apart from a few details, PR looks good otherwise, however the issue of the attendees in the LaoData needs to be discussed

be2-scala/src/main/scala/ch/epfl/pop/model/objects/LaoData.scala
```diff
@@ -36,8 +36,8 @@ final case class LaoData(
   def updateWith(message: Message, adr: Option[String]): LaoData = {
     message.decodedData.fold(this) {
       case call: CloseRollCall => adr match {
-        case Some(str) => LaoData(owner, call.attendees, privateKey, publicKey, witnesses, str)
-        case _ => LaoData(owner, call.attendees, privateKey, publicKey, witnesses, address)
+        case Some(str) => LaoData(owner, call.attendees ::: attendees, privateKey, publicKey, witnesses, str)
```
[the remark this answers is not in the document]
Ajkunas: Yes, so found what was the problem. I simply added the `owner` to the list of attendees after a CloseRollCall

be2-scala/src/main/scala/ch/epfl/pop/model/objects/RollCallData.scala
```diff
@@ -0,0 +1,40 @@
+package ch.epfl.pop.model.objects
+
+import ch.epfl.pop.json.MessageDataProtocol._
+import ch.epfl.pop.model.network.Parsable
+import ch.epfl.pop.model.network.method.message.Message
+import ch.epfl.pop.model.network.method.message.data.ActionType.ActionType
+import ch.epfl.pop.model.network.method.message.data.rollCall.{CloseRollCall, CreateRollCall, OpenRollCall, ReopenRollCall}
+import spray.json._
+
+final case class RollCallData(update_id: Hash, state: ActionType) {
```
lourkeur, line 10: can you rename this field to `update_id`? This is a code smell

be2-scala/src/main/scala/ch/epfl/pop/model/objects/RollCallData.scala
```diff
@@ -0,0 +1,40 @@
...
+}
+
+object RollCallData extends Parsable {
+  def apply(
+             update_id: Hash,
+             state: ActionType
+           ): RollCallData = {
+    new RollCallData(update_id, state)
+  }
+
```
lourkeur, line 34:
```suggestion
```

This is equivalent to the auto-generated factory method from `case`.

be2-scala/src/main/scala/ch/epfl/pop/model/objects/RollCallData.scala
Ajkunas: I do not understand: isn’t it already already `update_id` ?

tuomas-p (changes requested):
The CI does not pass because of complexity, this will have to be refactored

lourkeur (changes requested):

be2-scala/src/main/scala/ch/epfl/pop/model/objects/RollCallData.scala
```diff
@@ -0,0 +1,32 @@
...
+  def updateWith(message: Message): RollCallData = {
+    message.decodedData.fold(this) {
+      case create: CreateRollCall => RollCallData(create.id, create.action)
+      case open: OpenRollCall => RollCallData(open.update_id, open.action)
+      case reopen: ReopenRollCall => RollCallData(reopen.update_id, reopen.action)
+      case close: CloseRollCall => RollCallData(close.update_id, close.action)
+      case _ => this
+    }
```
lourkeur, line 23: Took me a couple seconds to figure what that does. `Option.fold` works, but It seems like it would be clearer to just do

```suggestion
    message.decodedData match {
      case Some(create: CreateRollCall) => RollCallData(create.id, create.action)
      case Some(open: OpenRollCall) => RollCallData(open.update_id, open.action)
      case Some(reopen: ReopenRollCall) => RollCallData(reopen.update_id, reopen.action)
      case Some(close: CloseRollCall) => RollCallData(close.update_id, close.action)
      case _ => this
    }
```

be2-scala/src/main/scala/ch/epfl/pop/model/objects/LaoData.scala
```diff
@@ -36,8 +36,8 @@ final case class LaoData(
   def updateWith(message: Message, adr: Option[String]): LaoData = {
     message.decodedData.fold(this) {
       case call: CloseRollCall => adr match {
-        case Some(str) => LaoData(owner, call.attendees, privateKey, publicKey, witnesses, str)
-        case _ => LaoData(owner, call.attendees, privateKey, publicKey, witnesses, address)
+        case Some(str) => LaoData(owner, owner :: call.attendees, privateKey, publicKey, witnesses, str)
+        case _ => LaoData(owner, owner :: call.attendees, privateKey, publicKey, witnesses, address)
```
lourkeur, line 40: so why did that change? It makes sense that the owner is an attendee, (although I think front ends do it anyway) but what does this have to do with the bug?

fe1-web/package.json
```diff
@@ -45,7 +45,7 @@
     "color": "^3.1.3",
     "ed25519-hd-key": "^1.2.0",
     "expect": "^26.6.2",
-    "expo": "^42.0.0",
```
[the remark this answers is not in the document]
lourkeur: I guess not then, let's see what Nico says

lourkeur (changes requested):

fe1-web/package.json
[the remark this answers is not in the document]
lourkeur: Actually yes, we should not be making changes to fe1 in this PR, period.

be2-scala/src/main/scala/ch/epfl/pop/model/objects/LaoData.scala
Ajkunas: I mentioned it on Slack, but briefly, the back end was always throwing an error after trying to close a re opened RollCall: "invalid keys attendee", because, in the updateWith function of LaoData, we did not include the PoP token of the LAO (aka the owner). With a first close RollCall (after opening it for the first time), the error did not occur, since the LaoData was correctly already included the owner in the attendees field (done during CreateLao). 
Well, it seems as the `attendees` field in a RollCall message coming from the front end do not include the owner (PoP token of LAO), but only the generated PublicKey by the RollCall. Therefore, I decided to add the owner into the attendees List after a rollcall close process.

be2-scala/src/main/scala/ch/epfl/pop/model/objects/RollCallData.scala
Ajkunas: I wanted to be consistent with the implementation of the updateWith function of the `LaoData` (which uses the fold function), but if it seems cleaner, then I can change that.

be2-scala/src/main/scala/ch/epfl/pop/model/objects/LaoData.scala
lourkeur: Oh yeah that part is just below. That makes perfect sense, thank you!

lourkeur (approved):
LGTM

As mentioned on Slack, we have permission to ignore the cognitive complexity warning

be2-scala/src/main/scala/ch/epfl/pop/storage/DbActor.scala
```diff
@@ -199,6 +200,31 @@ final case class DbActor(
...
   }
 
+  //generates the key of the RollCallData to store in the database
+  private def generateRollCallDataKey(laoId: Hash): String = {
+    s"${ROLL_CALL_DATA_PREFIX}${laoId.toString}"
+  }
+
+  @throws [DbActorNAckException]
+  private def readRollCallData(laoId: Hash): RollCallData = {
+    Try(storage.read(s"${ROLL_CALL_DATA_PREFIX}${laoId.toString}")) match {
```
RaulinN, line 210: can you use generateRollCallDataKey since you defiend it in this case? So that if we change it for some reason, no need to modify multiple instances :)

be2-scala/src/main/scala/ch/epfl/pop/storage/DbActor.scala
```diff
@@ -199,6 +200,31 @@ final case class DbActor(
...
+    }
+  }
+
+  @throws [DbActorNAckException]
+  private def writeRollCallData(laoId: Hash, message: Message): Unit = {
+    this.synchronized {
+      val rollcallData: RollCallData = Try(readRollCallData(laoId)) match {
+        case Success(data) => data
+        case Failure(_) => RollCallData(null, null)
+      }
```
RaulinN, line 223: ℹ️ In general, never use `null`  as "filler parameters" in scala ^^. If you really need to do this, I'd suggest Options or dummy values (which are not null)

Note: I believe that it's ok in this case since you create it and then fill it directly no matter what happens

be2-scala/src/main/scala/ch/epfl/pop/storage/DbActor.scala
```diff
@@ -295,6 +321,20 @@ final case class DbActor(
...
 
+    case ReadRollCallData(laoId) =>
+      log.info(s"Actor $self (db) received an ReadRollcallData request for lao '$laoId'")
+      Try(readRollCallData(laoId)) match {
+        case Success(rollcallData) => sender() ! DbActorReadRollCallDataAck(rollcallData)
+        case failure => sender() ! failure.recover(Status.Failure(_))
+      }
+
+    case WriteRollCallData(laoId, message) =>
+      log.info(s"Actor $self (db) received a WriteLaoData request for LAO id $laoId")
```
RaulinN, line 332: string is not accurate

RaulinN (changes requested):

RaulinN:
Nice work! Left a few minor comments and suggestions :)

RaulinN (approved):
👍

PR #1070, opened 2022-06-02
Draft : First attemped to create the option (with backendcommunication)

zoemarin:


fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/digitalcash/Transaction.java
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    while (ite_output.hasNext()) {
+      Output current = ite_output.next();
+      sig.add(String.valueOf(current.getValue()));
+      sig.add(current.getScript().getType());
+      sig.add(current.getScript().getPubkeyHash());
+    }
+    return Hash.hash(sig.toArray(new String[0]));
+  }
```
MaximeZmt, line 134: Need to add it back

PR #1076, opened 2022-06-05
Checking if election open is not before election setup create time

Tesa27:
In the current state of the system we could have an election open time that is prior to election creation time, which does not make sense timewise and thus this PR inserts a check that will prevent a misuse of this flaw.

### DIFF
--- a/be1-go/channel/election/verification_test.go
+++ b/be1-go/channel/election/verification_test.go
@@ -103,7 +103,8 @@ func TestVerify_ElectionOpen_already_closed(t *testing.T) {
 }
 
 func TestVerify_ElectionOpen_created_time_less_than_create_time_setup(t *testing.T) {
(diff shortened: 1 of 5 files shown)


Robingoumaz (dismissed):

be1-go/channel/election/verification_test.go
```diff
@@ -102,6 +102,26 @@ func TestVerify_ElectionOpen_already_closed(t *testing.T) {
 	require.Error(t, err)
 }
 
+func TestVerify_ElectionOpen_created_time_less_than_create_time_setup(t *testing.T) {
+	// create the opened election channel with election open time less than election creation time
```
nkcr, line 106: Comments must be formatted at 80 chars:
```suggestion
  // create the opened election channel with election open time less than
  // election creation time
```

PR #1077, opened 2022-06-06
[BE2] validate sum of transaction outputs

lourkeur:
There's a transparent type alias for Uint53 along with a summation operation.

We check sum of outputs because it's possible in a stateless way. Conveniently, it's the one you should check to avoid ex-nihilo money bugs.

I also found out that some of the negative test cases fail for the wrong reasons so I fixed that.

### DIFF
--- a/be2-scala/src/main/scala/ch/epfl/pop/model/objects/Uint53.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/model/objects/Uint53.scala
@@ -1,0 +1,42 @@
+package ch.epfl.pop.model.objects
+
+
+class ArithmeticOverflowError(msg: String) extends Error(msg)
+
+
+/**
+ * Operations on 53-bits non-negative integers.
+ *
+ * These values can be handled without loss of precision in IEEE754 64-bit
+ * floating numbers, and thus in JSON.
+ */
+object Uint53 {
+  /** Internal representation in this implementation */
+  type Repr = Long
+
+  /** Smallest allowed value */
+  final val MinValue: Uint53 = 0
+
+  /** Largest allowed value */
+  final val MaxValue: Uint53 = 0x1FFFFFFFFFFFFFL
+
+  /** Whether the value is in-range */
+  def inRange(v: Uint53) = v >= MinValue && v <= MaxValue
+
+  /** Compute the sum of many Uint53 values, or indicate an overflow if
+   *  the sum is too large.
+   *
+   *  Per the Either documentation, "Convention dictates that Left is used for
+   *  failure and Right is used for success."
+   */
+  def safeSum(seq: IterableOnce[Uint53]): Either[ArithmeticOverflowError, Uint53] = {
+    var acc = 0L
+    for (v <- seq.iterator) {
+      require(inRange(v), s"value $v out of range for uint53")
+      acc += v
+      if (acc > MaxValue)
+        return Left(new ArithmeticOverflowError("uint53 addition overflow"))
+    }
+    Right(acc)
+  }
+}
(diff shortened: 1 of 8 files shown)


be2-scala/src/main/scala/ch/epfl/pop/model/objects/Uint53.scala
```diff
@@ -0,0 +1,38 @@
...
+
+  /** Compute the sum of many Uint53 values, or indicate an overflow if
+   *  the sum is too large. */
+  def safeSum(seq: IterableOnce[Uint53]): Either[ArithmeticOverflowError, Uint53] = {
+    var acc = 0L
+    for (v <- seq.iterator) {
+      require(inRange(v), s"value $v out of range for uint53")
+      acc += v
+      if (acc > MaxValue)
+        return Left(new ArithmeticOverflowError("uint53 addition overflow"))
```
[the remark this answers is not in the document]
lourkeur: It cannot since the allowed range is unsigned.  This is enforced with `require`.

Ajkunas (approved):
LGTM

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/CoinValidator.scala
```diff
@@ -24,7 +24,10 @@ case object CoinValidator extends MessageDataContentValidator {
         } else if (!data.transaction.checkSignatures()) {
           Right(validationError("bad signature"))
         } else {
-          Left(rpcMessage)
+          data.transaction.sumOutputs() match {
+            case Left(err) => Right(validationError(err.getMessage()))
+            case Right(_) => Left(rpcMessage)
```
Tesa27, line 29: Hmmm maybe I am missing something here but shouldn't Left and Right be interchanged here?

protocol/examples/messageData/coin/post_transaction_bad_signature.json
```diff
@@ -1,7 +1,7 @@
 {
     "object": "coin",
     "action": "post_transaction",
-    "transaction_id": "ru0acQHE0QEqkA1IgdjTlqIwYxyq3YgN1M6pj9A2FF8=",
+    "transaction_id": "R5bU6MLWtJasGCJSnV73GBsinYajGsuai0sTNau8nv8=",
```
Tesa27, line 4: Could this influence the tests in Go? If so they shouldn't be changed there as well (if not already changed  in the other PR of course)

Tesa27 (changes requested):
Looks good just a few potential remarks, not sure if they are valid.

protocol/examples/messageData/coin/post_transaction_bad_signature.json
lourkeur: let me double check

lourkeur: It just does `require.Error(t, err)` so it's actually unspecific just like the old Scala one

Tesa27 (dismissed):
LGTM

be2-scala/src/main/scala/ch/epfl/pop/model/objects/Uint53.scala
```diff
@@ -0,0 +1,38 @@
...
+
+  /** Largest allowed value */
+  final val MaxValue: Uint53 = 0x1FFFFFFFFFFFFFL
+
+  /** Whether the value is in-range */
+  def inRange(v: Uint53) = v >= MinValue && v <= MaxValue
+
+  /** Compute the sum of many Uint53 values, or indicate an overflow if
+   *  the sum is too large. */
+  def safeSum(seq: IterableOnce[Uint53]): Either[ArithmeticOverflowError, Uint53] = {
```
tuomas-p, line 28: I would document here that the Left/Right convention is the general one and not the "wrong" one used in the project, just to avoid confusion

tuomas-p (changes requested):
LGTM code-wise, I'd just add a comment for future generations about the use of Left/Right

tuomas-p (approved):
All good now thanks!

PR #1079, opened 2022-06-06
Digital Cash wallet transactions

0xdieman:
This PR adds the transactions feature from the digital cash specification to the front end 1.

The `DigitalCashMessageApi` functions will do basic checks concerning the balance and the availability of the transaction inputs, but the rest of the checks will be performed by `DigitalCashHandler` upon receiving a transaction.

I decided to split the signatures and inputs/outputs validity check from the `Transaction` creation process because it's easier to test and it makes the Transaction object more flexible.


fe1-web/src/core/network/jsonrpc/messages/MessageRegistry.ts
```diff
@@ -85,6 +86,9 @@ export class MessageRegistry {
 
     // Reactions
     [k(REACTION, ADD), { signature: POP_TOKEN }],
+
+    // Coin
+    [k(COIN, POST_TRANSACTION), { signature: KEYPAIR }],
```
Tyratox, line 91: Are transactions signed using the general public key and not a pop token?

fe1-web/src/features/wallet/network/__tests__/DigitalCashMessageApi.test.tsx
```diff
@@ -0,0 +1,132 @@
...
+      await requestSendTransaction(
+        mockPopToken,
+        mockKeyPair.publicKey,
+        mockTransactionValue,
+        new Hash(mockLaoId),
+      );
+      fail();
+    } catch (e) {
+      expect(e).toBeInstanceOf(Error);
+    }
```
Tyratox, line 102: expect(...).rejects.toBeInstanceOf() is a bit shorter :)

fe1-web/src/core/network/jsonrpc/messages/MessageRegistry.ts
Tyratox: I guess this is now clear, sorry for that :)

Tyratox:
Also make sure to check the code smells. You can fix the non-null assertion issue by adding a null check and throwing an error if it is null

fe1-web/src/features/index.ts
```diff
@@ -73,11 +73,13 @@ export function configureFeatures() {
 
   const walletComposition = wallet.compose({
     keyPairRegistry,
+    messageRegistry,
     getCurrentLao: laoConfiguration.functions.getCurrentLao,
     useCurrentLaoId: laoConfiguration.hooks.useCurrentLaoId,
     getEventById: eventConfiguration.functions.getEventById,
     getRollCallById: rollCallConfiguration.functions.getRollCallById,
     useRollCallsByLaoId: rollCallConfiguration.hooks.useRollCallsByLaoId,
+    getLaoOrganizer: laoConfiguration.functions.getLaoOrganizer,
```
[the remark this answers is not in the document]
0xdieman: 👍 Thanks !

fe1-web/src/features/wallet/components/RoundIconButton.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+const RoundIconButton = (props: IPropTypes) => {
+  const { name, onClick } = props;
+  return (
+    <TouchableOpacity style={styles.roundButton} onPress={onClick}>
+      <Icon name={name} type="material" color="white" size={30} />
+    </TouchableOpacity>
+  );
+};
+
```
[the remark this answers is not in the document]
0xdieman: That's non intended, I'll fix that

fe1-web/src/features/wallet/network/DigitalCashMessageApi.ts
```diff
@@ -0,0 +1,76 @@
...
+  if (transactionStates.length === 0) {
+    throw new Error(makeErr('no transaction out were found for this public key'));
+  }
+
+  const balance = getBalance(laoId.valueOf(), from.publicKey.valueOf());
+
+  if (amount < 0 || amount > balance) {
+    throw new Error(makeErr('balance is not sufficient to send this amount'));
+  }
+
```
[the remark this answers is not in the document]
0xdieman: As I said in the description, I thought that it was easier to test and let more flexibility for later to the transaction creation and verification

fe1-web/src/features/wallet/network/__tests__/DigitalCashMessageApi.test.tsx
0xdieman: Seems like 
`await expect(
      await requestSendTransaction(
        mockPopToken,
        mockKeyPair.publicKey,
        -mockTransactionValue,
        new Hash(mockLaoId),
      ),
    ).rejects.toBeInstanceOf(Error);`

Does not work as expected, am I doing something wrong ?

fe1-web/src/features/wallet/objects/transaction/Transaction.ts
```diff
@@ -0,0 +1,387 @@
...
+    const inputs: Omit<TransactionInputState, 'script'>[] = Transaction.getInputsInToSign(
+      from.publicKey.valueOf(),
+      inputTransactions,
+    );
+
+    // Concatenate the data to sign
+    const dataString = Transaction.concatenateTxData(inputs, outputs);
+
+    // Sign with the popToken
+    const signature = from.privateKey.sign(Base64UrlData.encode(dataString));
```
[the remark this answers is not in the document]
0xdieman: The general public key ? Are you talking about the `KeyPairStore.get()` in WalletHome ?

fe1-web/src/features/wallet/objects/transaction/Transaction.ts
```diff
@@ -0,0 +1,387 @@
...
+          (output) => output.script.publicKeyHash.valueOf() === Hash.fromPublicKey(pk).valueOf(),
+        )
+        .map((output, index) => {
+          return {
+            txOutHash: tr.transactionId!.valueOf(),
+            txOutIndex: index,
+          };
+        }),
+    );
+  };
```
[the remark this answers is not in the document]
0xdieman: True, but this does not seem to work here I can't understand why
```suggestion
        .map((output, index) => {
            txOutHash: tr.transactionId!.valueOf(),
            txOutIndex: index,
        }),
```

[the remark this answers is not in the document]
0xdieman: True, but this does not seem to work and I can't understand why
```suggestion
  ): Omit<TransactionInputState, 'script'>[] => {
    return transactions.flatMap((tr) =>
      tr.outputs
        .filter(
          (output) => output.script.publicKeyHash.valueOf() === Hash.fromPublicKey(pk).valueOf(),
        )
        .map((output, index) => {
            txOutHash: tr.transactionId!.valueOf(),
            txOutIndex: index,
        }),
    );
  };
```

fe1-web/src/features/wallet/objects/transaction/Transaction.ts
```diff
@@ -0,0 +1,387 @@
...
+    const encodedData = Base64UrlData.encode(
+      Transaction.concatenateTxData(
+        this.inputs.map((input) => input.toState()),
+        this.outputs.map((output) => output.toState()),
+      ),
+    );
+
+    let totalInputAmount = 0;
+
+    const inputsAreValid = !this.inputs.some((input) => {
```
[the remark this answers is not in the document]
0xdieman: Hahaha that's the method I was looking for !!

fe1-web/src/features/wallet/reducer/DigitalCashReducer.ts
```diff
@@ -0,0 +1,136 @@
...
+  /**
+   * Transactions by their transaction hash (transaction id)
+   */
+  transactionsByHash: Record<string, TransactionState>;
+
+  /**
+   * A mapping between public key hashes and a set of the transactions which contain this hash
+   * in one or more of their TxOuts
+   */
+  transactionsByPubHash: Record<string, TransactionState[]>;
```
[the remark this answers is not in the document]
0xdieman: I'm not sure anymore, and I agree with your idea !

fe1-web/src/features/wallet/screens/WalletHome.tsx
```diff
@@ -109,6 +136,41 @@ const WalletHome = () => {
...
+            }}
+          />
+        </View>
+      )}
+      <SendModal
+        modalVisible={sendModalVisible}
+        setModalVisible={setSendModalVisible}
+        send={(receiver: string, amount: number, isCoinbase: boolean) => {
+          if (isCoinbase) {
+            requestCoinbaseTransaction(KeyPairStore.get(), new PublicKey(receiver), amount, laoId!)
```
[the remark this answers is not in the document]
0xdieman: I did that by having the new UI in mind, but now I also feel like it's better if any code merged to master is reliable, so I'll fix that asap anyway.

fe1-web/src/resources/strings.ts
```diff
@@ -256,8 +256,17 @@ namespace STRINGS {
   export const lao_id = 'LAO ID';
   export const lao_name = 'LAO name';
   export const roll_call_name = 'Roll Call name';
+
+  /* --- Digital Cash Strings --- */
+  export const script_type = 'Pay-to-Pubkey-Hash';
+  export const coinbase_hash = 'AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=';
```
lourkeur, line 262: Nitpick: this file looks to me to be human-readable interface text. It might be that translators even look at it. If that's the case, then those two constants would be more at home in `const.ts` because they are not meant to be understood by a human or translated.

fe1-web/src/features/wallet/objects/transaction/Transaction.ts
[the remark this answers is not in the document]
lourkeur: probably the braces. `() => { x; }` may not be the way.

fe1-web/src/__tests__/utils/TestUtils.ts
```diff
@@ -75,3 +78,61 @@ export const mockJsonRequest: Partial<JsonRpcRequest> = {
...
+  '1E7yPEY7SwZ1NNxi_jN85-IvThR7GzUFs8-Lfwao44waoCf-qj5xLKnpzsc8yP0cFr5UNOIzsgKUYaSzv1gpDQ==';
+export const mockCBHash = 'FhlMNTEOqOzkKbe8RH00fmF-Op0S_ipowEn0nj402Ts=';
+export const mockCoinbaseTransactionJSON: TransactionJSON = {
+  version: 1,
+  inputs: [
+    {
+      tx_out_hash: STRINGS.coinbase_hash,
+      tx_out_index: 0,
+      script: {
+        type: 'Pay-to-Pubkey-Hash',
```
lourkeur, line 96: I know the google doc is out of sync, but all 3 other codebases and `protocol/` use `'P2PKH'`. There will be compatibility issues if we don't fix it here.

lourkeur (changes requested):

fe1-web/src/features/wallet/network/DigitalCashMessageApi.ts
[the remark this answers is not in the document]
Tyratox: Sorry, must have missed this 👍

fe1-web/src/features/wallet/network/__tests__/DigitalCashMessageApi.test.tsx
Tyratox: Should be `await expect(somePromise).rejects.toBeInstanceOf(Error);` so without the await inside the expect() :)

fe1-web/src/features/wallet/objects/transaction/Transaction.ts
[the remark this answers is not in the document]
Tyratox: No, what I mean is that you set `[k(COIN, POST_TRANSACTION), { signature: KEYPAIR }],` instead of `[k(COIN, POST_TRANSACTION), { signature: POP_TOKEN }],` in the message registry.

fe1-web/src/__tests__/utils/TestUtils.ts
0xdieman: I'll change that, but when changing the protocol the google doc should also be updated

fe1-web/src/features/wallet/objects/transaction/Transaction.ts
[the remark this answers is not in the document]
Tyratox: (Which is fine since here you add a pop token signature, was just confused at first why you are doing it this way and I then assumed it would be because `[k(COIN, POST_TRANSACTION), { signature: POP_TOKEN }]` will not allow you to use a specific pop token)

fe1-web/src/features/wallet/objects/transaction/Transaction.ts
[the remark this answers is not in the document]
Tyratox: Ahh yes, this is because

`() => {}` is the syntax for a function where a function body is expected in the curly braces. If you want to directly return an object (which also uses the curly brace syntax) you have to add parenthesis:


`() => ({someKey: "someValue"})`

Btw, you can also use this syntax at the top level, so instead of

```
): Omit<TransactionInputState, 'script'>[] => {
return transactions.flatMap((tr) => ...
```

you could write

```
): Omit<TransactionInputState, 'script'>[] => transactions.flatMap((tr) => ...
```

fe1-web/src/resources/strings.ts
Tyratox: I agree, it might be a good time starting with this file. But to be fair, in the beginning of the semester and probably still at the moment `strings.ts` contains some constants mixed into the UI strings ... 😂

So while it is not optimal, it is consistent with the remaining code 😂

fe1-web/src/features/wallet/network/__tests__/DigitalCashMessageApi.test.tsx
0xdieman: Still does not catch the error :(

fe1-web/src/features/wallet/objects/transaction/Transaction.ts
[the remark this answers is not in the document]
0xdieman: Oh okay I missed the 'the message' part, in fact I made it this way, because at first the coinbase transaction messages would have needed a signature from the organizer, but it's not the case anymore (the organizer's signature is still needed but inside the transaction itself), so it does not really matter now

fe1-web/src/features/wallet/network/__tests__/DigitalCashMessageApi.test.tsx
Tyratox: Huh... What is `fail()` doing in the above code btw?

fe1-web/src/features/wallet/objects/transaction/Transaction.ts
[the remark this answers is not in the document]
0xdieman: That was it thanks !

fe1-web/src/features/wallet/network/__tests__/DigitalCashMessageApi.test.tsx
0xdieman: If the request does not launch an error, the test should fail

fe1-web/src/features/wallet/reducer/DigitalCashReducer.ts
```diff
@@ -0,0 +1,137 @@
...
+          allTransactionsHash: [],
+          transactionsByHash: {},
+          transactionsByPubHash: {},
+        };
+      }
+
+      const laoState: DigitalCashReducerState = state.byLaoId[laoId];
+
+      laoState.transactionsByHash[transactionHash] = transactionState;
+      laoState.allTransactionsHash.push(transactionHash);
```
lourkeur, line 84: my 2 miniLAOs: `allTransactionsHash` is equal to the set of all keys in `transactionsByHash` so it's still duplicative. Just like `allTransactions` was duplicative of the set of its values. (This could be me being bad at TS)

Tyratox: Yes this is true, it `allTransactions` is equal to `Object.keys(transactionsByHash)` and could in theory be removed. But then this is a pattern used all over the codebase and in general with redux:

https://stackoverflow.com/a/61959401/2897827

0xdieman: Yes, it now all depends on performance:

-Is calling Object.keys(transactionsByHash) each time much slower than having the object `allTransactionsHash`
(of course this might only be an issue when we reach a consequent number of transactions)

Order (we would use this for the transaction history) seems [to not be an issue here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys):

> The ordering of the properties is the same as that given by looping over the properties of the object manually.

This seems to mean that it would give us the order in which we added each keys.

What do you think @Xelowak @Tyratox ?

fe1-web/src/features/wallet/network/__tests__/DigitalCashMessageApi.test.tsx
0xdieman: It is the jest fail function:
> Fails a test when called within one.

I just tried it and it makes the test fail even inside a try catch

fe1-web/src/features/wallet/reducer/DigitalCashReducer.ts
0xdieman: @Tyratox Seems like I type too slowly

lourkeur: yep sounds good to me

fe1-web/src/core/objects/Hash.ts
```diff
@@ -44,6 +45,25 @@ export class Hash extends Base64UrlData {
...
+    const hash = sha256.create();
+
+    // We need the raw public key
+    const decodedPK = Base64UrlData.fromBase64(publicKey.valueOf()).toBuffer();
+
+    const bString = hash.update(decodedPK).array();
+    // Taking only the first 20 bytes
+    const str = String.fromCharCode(...bString.slice(0, 20));
+
+    return new Hash(Base64UrlData.encode(str, 'binary').valueOf());
```
lourkeur, line 64: If it works it works, but it seems a bit convoluted and fragile to stuff your bytes into a (UTF-16) string and then re-decode it as "binary". If we could go through `Buffer` instead.

fe1-web/src/features/wallet/network/DigitalCashMessageApi.ts
```diff
@@ -0,0 +1,76 @@
...
+  amount: number,
+  laoId: Hash,
+): Promise<void> {
+  const transactionStates = DigitalCashStore.getTransactionsByPublicKey(
+    laoId.valueOf(),
+    from.publicKey.valueOf(),
+  );
+
+  if (transactionStates.length === 0) {
+    throw new Error(makeErr('no transaction out were found for this public key'));
```
lourkeur, line 31: I noticed when testing manually that this exception isn't caught. (I get the metro screen) It would be better if it triggered a toast.

image

lourkeur (approved):
my other concerns have been addressed, just two  non-blockers

fe1-web/src/features/wallet/network/DigitalCashMessageApi.ts
0xdieman: Thanks, that was supposed to be caught !

fe1-web/src/features/wallet/network/__tests__/DigitalCashMessageApi.test.tsx
Tyratox: The problem was that the function did not return a rejecting promise but rather threw an error directly. Fixed it by making the function async :) (You could have also instead return Promise.reject(new Error()) )

fe1-web/src/core/objects/Hash.ts
0xdieman: True and done

fe1-web/src/features/wallet/screens/WalletHome.tsx
```diff
@@ -72,12 +104,65 @@ const WalletHome = () => {
...
+        },
+      );
+    } else {
+      requestSendTransaction(
+        tokens[selectedTokenIndex].token,
+        new PublicKey(receiver),
+        amount,
+        tokens[selectedTokenIndex].laoId,
+      ).then(
+        () => toast.show('Sent transaction'),
```
Xelowak, line 148:
```suggestion
        () => toast.show('Transaction sent'),
```

fe1-web/src/features/wallet/screens/WalletHome.tsx
```diff
@@ -72,12 +104,65 @@ const WalletHome = () => {
...
+        (reason) => {
+          toast.show('Failed to send transaction');
```
Xelowak, line 150: It could be nice to add all these toast message strings in `string.ts` (because they are display in the UI).

fe1-web/src/features/wallet/store/DigitalCashStore.ts
```diff
@@ -0,0 +1,29 @@
+import { Hash } from 'core/objects';
+import { getStore } from 'core/redux';
+
+import { TransactionState } from '../objects/transaction';
+import { getDigitalCashState } from '../reducer';
+
+export namespace DigitalCashStore {
```
Xelowak, line 7: Adding spaces could be nice to improve clarity.

Moreover, I think this namespace lacks some documentation (especially to say that it shouldn't be used outside of a React component).

fe1-web/src/features/wallet/store/__tests__/DigitalCashStore.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+        [mockKPHash.valueOf()]: [mockTransaction.transactionId],
+      },
+    },
+  },
+};
+
+jest.mock('features/wallet/reducer/DigitalCashReducer');
+
+beforeAll(() => {
+  (getDigitalCashState as jest.Mock).mockReturnValue(filledState);
```
Xelowak, line 38: I think this can be done at the beginning, outside of a `beforeAll`.

fe1-web/src/features/wallet/store/__tests__/DigitalCashStore.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+});
+
+describe('Digital Cash Store', () => {
```
Xelowak, line 41: It could be nice to add spaces between the `it`s.

fe1-web/src/features/wallet/objects/transaction/Transaction.ts
```diff
@@ -0,0 +1,381 @@
...
+  inputs: TransactionInputState[];
+  outputs: TransactionOutputState[];
+  lockTime: number;
+  transactionId?: string;
+}
+
+/**
+ * A coin transaction object
+ */
+export class Transaction {
```
Xelowak, line 29: I feel like this file contains a lot of lines (381). Maybe it could be split into 2 files, or reduced by using private functions for somewhat similar code (like `create` and `createCoinbase`).

fe1-web/src/resources/strings.ts
Xelowak: I agree it's better to put it in `const.ts`. Even if it's not the case everywhere, we should clearly separate UI strings from constants.

fe1-web/src/features/wallet/store/__tests__/DigitalCashStore.test.tsx
0xdieman: Noted !

fe1-web/src/core/components/RoundIconButton.tsx
```diff
@@ -54,6 +55,31 @@ export const LogoutRoundButton = (props: IButtonPropTypes) => {
   );
 };
 
+export const SendRoundButton = (props: IButtonPropTypes) => {
```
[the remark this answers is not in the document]
0xdieman: Nope, changed that

fe1-web/src/features/wallet/network/__tests__/DigitalCashMessageApi.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+    getTransactionsByPublicKeyMock.mockReturnValue([]);
+    await expect(
+      requestSendTransaction(
+        mockPopToken,
+        mockKeyPair.publicKey,
+        mockTransactionValue,
+        new Hash(mockLaoId),
+      ),
+    ).rejects.toBeInstanceOf(Error);
+  });
```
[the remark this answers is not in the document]
0xdieman: But this does not work as it returns a promise right ?

fe1-web/src/features/wallet/network/messages/__tests__/PostTransaction.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+    };
+    expect(() => new PostTransaction(object)).toThrow(ProtocolError);
+  });
+  it('should fail to create a message when undefined transaction_id', () => {
+    const object = {
+      transaction: postTransactionObject.transaction,
+    };
+    expect(() => new PostTransaction(object)).toThrow(ProtocolError);
+  });
+});
```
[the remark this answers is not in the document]
0xdieman: Also noted

fe1-web/src/features/wallet/objects/transaction/Transaction.ts
0xdieman: I tried to make it as small as possible but also to keep it modulable, maybe if we really want to reduce it we could split the verification process in another file, but I do not know if it is worth it

fe1-web/src/features/digital-cash/__tests__/utils.ts
```diff
@@ -0,0 +1,96 @@
...
+  DigitalCashReactContext,
+} from '../interface';
+import { TransactionJSON, TransactionState } from '../objects/transaction';
+
+export const mockRollCall = {
+  id: new Hash('rcid'),
+  name: 'rc',
+  attendees: [mockKeyPair.publicKey],
+  containsToken: () => true,
+} as DigitalCashFeature.RollCall;
```
Tyratox, line 17: For tests inter-feature dependencies are allowed so it would also be possible to just re-use the mock roll call from the roll call feature :)

fe1-web/src/resources/strings.ts
```diff
@@ -355,6 +375,54 @@ namespace STRINGS {
...
+
+  export const digital_cash_wallet_add_beneficiary = 'Add Beneficiary';
+  export const digital_cash_wallet_send_transaction = 'Send Transaction';
+
+  export const digital_cash_wallet_beneficiary = 'Beneficiary (-ies)';
+  export const digital_cash_wallet_beneficiary_placeholder =
+    'QVGlyoTAEYWxq3hSBuawE-lo3sHEyfIv8uizjQTzsIU=';
+
+  export const digital_cash_wallet_amount = 'Amount';
+  export const digital_cash_wallet_amount_placeholder = '42.50';
```
Tyratox, line 411: One last time, are floating point numbers allowed?

fe1-web/src/features/wallet/network/__tests__/DigitalCashMessageApi.test.tsx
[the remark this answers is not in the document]
Tyratox: Yep I thought so too since the function does not throw an error but returns a failing promise? 😅

Or are @0xdieman and me both missing something?

fe1-web/src/features/digital-cash/components/__tests__/TransactionHistory.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+import {
+  mockCBHash,
+  mockCoinbaseTransactionJSON,
+  mockTransactionHash,
+  mockTransactionState,
+  mockDigitalCashContextValue,
+} from '../../__tests__/utils';
+import TransactionHistory from '../TransactionHistory';
+
+jest.mock('features/digital-cash/hooks');
```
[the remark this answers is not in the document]
0xdieman: The `useTransactions` and `useTransactionsByHash` hooks make use of some state selectors which are independant of the context, and they are used in the component

fe1-web/src/features/digital-cash/network/DigitalCashMessageApi.ts
```diff
@@ -0,0 +1,76 @@
...
+  const transaction: Transaction = Transaction.create(from, to, balance, amount, transactionStates);
+
+  const postTransactionMessage = new PostTransaction({
+    transaction_id: transaction.transactionId,
+    transaction: transaction.toJSON(),
+  });
+
+  console.log(`Sending a transaction with id: ${transaction.transactionId.valueOf()}`);
+
+  return publish(getCoinChannel(laoId), postTransactionMessage);
```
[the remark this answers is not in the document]
0xdieman: Right

fe1-web/src/features/digital-cash/components/__tests__/TransactionHistory.test.tsx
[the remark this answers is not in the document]
Tyratox: Ah I see :) Another option would be to include a mock state but I guess the result is almost the same so 😊👍

0xdieman:
@Tyratox Think so too, not worth the component for the current usage

Xelowak (dismissed):
Approving blindly

fe1-web/src/features/digital-cash/objects/transaction/Transaction.ts
```diff
@@ -0,0 +1,388 @@
...
+          "The computed transaction hash does not correspond to the provided one when creating 'Transaction'",
+        );
+      }
+      this.transactionId = obj.transactionId;
+    }
+  }
+
+  /**
+   * Hashes the transaction to get its id
+   */
```
Tyratox, line 85: In general I think it would be great to link the specification for complex things like this but should be okay for this :)

Tyratox (approved):
LGTM

fe1-web/src/features/digital-cash/objects/transaction/Transaction.ts
```diff
@@ -0,0 +1,388 @@
...
+   * @param amount the amount to send to the receiver
+   * @param inputTransactions the transactions that contains the outputs to use as inputs in this transaction
+   */
+  public static create(
+    from: PopToken,
+    to: PublicKey,
+    currentBalance: number,
+    amount: number,
+    inputTransactions: TransactionState[],
+  ): Transaction {
```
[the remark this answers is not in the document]
Tyratox: But maybe you already added it somewhere else and I just missed it 😅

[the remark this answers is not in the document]
lourkeur: fwiw the json schema should in principle enforce non-floatness. If you kings want to program defensively feel free though.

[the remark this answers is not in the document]
Tyratox: I see! So only for the creation then (javascript only has a number type, not int/float)

Xelowak (approved):
Re-approving

PR #1080, opened 2022-06-06
Go CLI - add parameters to make pubkey optional 

MaximeZmt:


be1-go/hub/standard_hub/mod.go
```diff
@@ -548,7 +552,8 @@ func (h *Hub) createLao(msg message.Message, laoCreate messagedata.LaoCreate,
 		return answer.NewInvalidMessageFieldError("failed to unmarshal public key of the sender: %v", err)
 	}
 
-	if !h.GetPubKeyOwner().Equal(senderPubKey) {
+	//TODO
+	if !h.GetPubKeyOwner().Equal(senderPubKey) && h.isCheckingOrganizerKey {
```
Warhedgehog, line 556: I'd inverse them, so that the `!h.GetPubKeyOwner().Equal(senderPubKey)` is not even evaluated if we don't check the organizer

be1-go/channel/consensus/mod.go
```diff
@@ -277,6 +277,7 @@ func (c *Channel) processConsensusElect(message message.Message, msgData interfa
 
 	consensusInstance.createElectInstance(message.MessageID, c.hub.GetServerNumber())
 
+	// TODO
```
Warhedgehog, line 280: We will maybe have to pass the senderKey to the NewChannel like for the lao

Warhedgehog (commented):
Very nice work !

be1-go/hub/standard_hub/mod.go
```diff
@@ -92,6 +92,9 @@ type Hub struct {
 	// rootInbox and queries are used to help servers catchup to each other
 	rootInbox inbox.Inbox
 	queries   queries
+
+	// isCheckingOrganizerKey allows to enable or disable the security that allow only the organizer the LAO creation
```
Robingoumaz, line 96: maybe server owner is better, because if it is false everyone can be organizers

be1-go/cli/mod.go
```diff
@@ -47,34 +47,48 @@ func Serve(cliCtx *cli.Context, user string) error {
...
+	// TODO
+	var h *standard_hub.Hub
+	var err error
 
-	// create user hub
-	h, err := standard_hub.NewHub(point, clientServerAddress, log.With().Str("role", user).Logger(),
-		lao.NewChannel, hubType)
-	if err != nil {
-		return xerrors.Errorf("failed create the %s hub: %v", user, err)
+	if pk == "" {
```
Robingoumaz, line 63: Maybe you should also add a `Value: "",` to the publicKeyFlag in pop.go so that we don't have to specify --pk="" if we don't want to check the public key

be1-go/hub/standard_hub/mod.go
MaximeZmt: yes, you're right I will rename it to isCheckingOwnerKey

be1-go/cli/mod.go
MaximeZmt: After having checked there is no need to do that to have the behavior you've mentioned.
Currently if you specify any --pk it will require to be correct
But if you omit it, it will run it without it

Robingoumaz (approved):
LGTM 👍

Warhedgehog (dismissed):
lgtm

nkcr (approved):

PR #1081, opened 2022-06-06
Change computation of the election end hash

Tyratox:


Tyratox:
Just tested & works with `work-be2-scala-update-check-registerd-votes` ✅

Xelowak (approved):
LGTM!

PR #1082, opened 2022-06-06
Update of the check of the registered votes in EndElection in scala

Ajkunas:
To match the implementation of the front ends, we decided to change the protocol of the computation of the registered votes of EndElection: 

- Sorted by voteId uniquely 
- Updated the function getLastVote so that it returns only a List of CastVotes: the Message is not needed anymore


Tyratox:
Just tested & works with `work-fe1-nico-update-election-end-hash` ✅

Tyratox (approved):

PR #1091, opened 2022-06-07
Fix update vote

Robingoumaz:
When tested during pop meeting of the 7th of June, we realised that the go backend only counted the first votes of each sender instead of the last one.

- Changed the direction of the time check in update_vote
- Added a test for update_vote


Warhedgehog (approved):
LGTM

nkcr (approved):

PR #1098, opened 2022-06-08
Fix fe1 CI

Tyratox:
Hopefully fixes the CI. Not sure why this was not a problem before though.

Explanation: https://stackoverflow.com/a/66620869/2897827

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -55,7 +55,7 @@ jobs:
 
       - name: Setup npm dependencies
         run: |
-          npm install
+          npm install --legacy-peer-deps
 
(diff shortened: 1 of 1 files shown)


jbsv (approved):
The NPM default package version has probably changed. When CI runs, it dynamically installs   `npm` in the docker or VM where the CI script runs. That's why it silently went banana. Another way to fix it, is to upgrade all legacy dependencies (if possible). Yet another way to fix it, is to force installation of a given (and previous) version of npm. But we are not in the medical device nor avionics industry, so I'm happy with your fix. Astonishingly 1 file / 1 line this time changed this time.

PR #1099, opened 2022-06-08
[BE2] fix coin field names

lourkeur:
This was leftover when we did the big rename. Also avoided capitals.


Ajkunas (approved):
All is fine for me

RaulinN (approved):

PR #1108, opened 2022-06-13
Automatically append git commit hash

Tyratox:
Also adds `-dirty` if there are uncommitted changes.

A simple solution like `zip-build web-build -t fe1-$(git rev-parse HEAD)$(git diff --quiet || echo '-dirty').%EXT%` would  most likely only work on linux & mac os but not on windows as it uses bash features. Thus this PR adds an additional node js build script.

---

Since expo optimize will always update `fe1-web/docs/images/architecture-data-flow.png` if it remains non-optimized during the build process, this PR also includes an optimized version of this image. This way not all builds are automatically marked as dirty.


fe1-web/package.json
```diff
@@ -9,7 +9,7 @@
     "web": "expo start --web",
     "prebuild-web": "node node_modules/expo-optimize/build/index.js",
     "build-web": "expo build:web",
-    "postbuild-web": "zip-build web-build -t %NAME%_%TIMESTAMP%.%EXT%",
+    "postbuild-web": "zip-build web-build -t fe1.zip && node build/web-postbuild.js",
```
Xelowak, line 12: Did you erase the timestamp part? I think it was quite useful to be able to keep past versions of the build (and to locate them in time).

Xelowak (approved):
I just have a small remark, otherwise LGTM

fe1-web/package.json
Tyratox: That is true! Yes I needed to do this here so that the script knows the name of the file which should be renamed but I will add it back in `web-postbuild.js`.

Xelowak (approved):
Thanks for the changes!

PR #1110, opened 2022-06-13
Fix styling issues

Tyratox:
This PR fixes two styling issues:
- Lists not having the proper padding (i.e. they had too much padding on the right side making them not being aligned with the remaining content)
- Displayed toast messages prevented the user from clicking UI elements behind the toast container. The newest version sets the `pointerEvents` property: https://github.com/arnnis/react-native-toast-notifications/blob/66a275e03da7b82896c8116740cfd24b5ede3c96/src/toast-container.tsx#L124


0xdieman (approved):

Tyratox:
Unfortunately `box-none` does not mean what I thought it means.. See https://reactnative.dev/docs/view#pointerevents

PR #1111, opened 2022-06-13
[BE2] migrate to Scala 3

bbjubjub2494:
working on it

scapegoat is a compiler plugin so it isn't going to work trivially.


lourkeur:
Looks like we need to wait for the next Scala compiler release to get coverage info: https://github.com/lampepfl/dotty/pull/13880

be2-scala/src/main/scala/ch/epfl/pop/model/objects/ElectionChannel.scala
```diff
@@ -20,7 +20,7 @@ object ElectionChannel {
      * @tparam T type of data to extract from the election channel
      * @return Future of a list of tuple containing the message and the data extracted
      */
-    def extractMessages[T: Manifest](dbActor: AskableActorRef = DbActor.getInstance): Future[List[(Message, T)]] = {
+    def extractMessages[T: reflect.ClassTag](dbActor: AskableActorRef = DbActor.getInstance): Future[List[(Message, T)]] = {
```
rombirli, line 23: This generic type T was extending Manifest to accept all case class. We could now make it accept all case class extendding MessageData 
def extractMessages[T: MessageData]
for readability

lourkeur: T does not extend `Manifest` either way. This is the syntax for passing an implicit `Manifest[T]` to allow pattern matching to check for instances of `T`. Here the only difference it that `Manifest` is gone in Scala 3. Orthogonally to that, I agree that we could specify `T <: MessageData`, but that won't replace `Manifest`.

lourkeur: Can you review #1113 instead? That would be more practical

lourkeur:
Blocked by https://github.com/lampepfl/dotty/issues/15791

K1li4nL:
done in  #1742

PR #1113, opened 2022-06-13
[BE2] prepare for Scala 3 migration

bbjubjub2494:
this is a subset of #1111 that can be applied without actually switching to Scala 3. My hope is proactively doing these changes now will reduce friction when we are ready to switch.


be2-scala/src/test/scala/ch/epfl/pop/model/network/method/message/data/socialMedia/AddChirpSuite.scala
```diff
@@ -28,7 +29,7 @@ class AddChirpSuite extends FunSuite with Matchers {
 object AddChirpExample {
   final val TEXT: String = "text"
   final val PARENT_ID: Option[String] = None
-  final val TIMESTAMP = Timestamp(0)
+  final val TIMESTAMP: Timestamp = Timestamp(0)
```
rombirli, line 32: I think at some places it was unnecessary to add the typing

be2-scala/src/test/scala/ch/epfl/pop/model/objects/SignatureSuite.scala
```diff
@@ -4,27 +4,29 @@ import java.nio.charset.StandardCharsets
...
   }
   //Data used for testing signature
-  final val dataTest = Seq("PoP-scala", "HelloWorld", "Not true is false",
+  final val dataTest: Seq[String] = Seq("PoP-scala", "HelloWorld", "Not true is false",
     "😀", "OMEGA \u03A9", "\u03A8",
     "Non empty can be fully non empty",
     "Not false is true")
 
-  final val verify_pk = PublicKey(Base64Data.encode(tester.keyPair.getPublicKey))
+  final val verify_pk: PublicKey = PublicKey(Base64Data.encode(tester.keyPair.getPublicKey))
```
rombirli, line 29: Here too

rombirli (approved):
Looks really good overall

be2-scala/src/test/scala/ch/epfl/pop/model/network/method/message/data/socialMedia/AddChirpSuite.scala
lourkeur: I personally don't mind but the "migrate" plugin made that change.

be2-scala/src/test/scala/ch/epfl/pop/model/objects/SignatureSuite.scala
lourkeur: same

tuomas-p (dismissed):
LGTM, as this is for pure migration purposes, the coverage failure shouldn't be considered relevant imo

lourkeur:
@jbsv @pierluca I need either of you to force-merge this PR. The coverage check is in the way.

The rationale is that the changes are automated rewrites that are functionally equivalent. They are peppered all over the place so there's no clear behavior that I should go write a test about.

K1li4nL:
Done in #1742

PR #1116, opened 2022-06-13
Fix for unopened roll-call bug

### LINKED ISSUE #1114: [BUG] PoP token is non retrievable when there is an unopened roll-call
When the organizer creates a roll-call but does not open it, the pop token from subsequent roll-call is not retrieved and therefore prevents the user from using any feature that requires a pop token. That is most of them

### Expected behavior
We expect the user to be able to use the pop token from the lastly closed roll-call

### How to reproduce
1. Create a roll-call
2. Create open and close another rc afterwards
3. Create an election
4. See that user is unable to vote

### Version & Environment
This bug was reproduced on:
- Commit: [e6f7dc7](https://github.com/dedis/popstellar/commit/e6f7dc75ec530dc8fcb56ee8be52838011c7564f)

##### Front-ends:
- [ ] Not applicable

##### Back-ends:
- [ ] Be1-Go
- [ ] Be2-Scala
- [x] Not applicable

 - Device: [e.g. Pixel 4 emulator on Windows 10]

### Workaround
Always open and close created roll-call before creating another

### Possible root cause
My best guess is that the comparator logic used to determine which is the last roll-call is faulty and return the unopened roll-call as the last.

Warhedgehog:
this fixes #1114

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/security/KeyManager.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/security/KeyManager.java
@@ -74,7 +74,9 @@ public class KeyManager {
     keyPair = getKeyPair(keysetManager.getKeysetHandle());
   }
 
-  /** @return the device public key */
+  /**
+   * @return the device public key
+   */
   public PublicKey getMainPublicKey() {
     return keyPair.getPublicKey();
   }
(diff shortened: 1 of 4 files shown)


zoemarin (approved):
👍

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/security/KeyManager.java
```diff
@@ -113,6 +115,7 @@ public PoPToken getValidPoPToken(Lao lao) throws KeyException {
     // Find the latest closed RollCall and use the wallet to retrieve the key
     RollCall rollCall =
         lao.getRollCalls().values().stream()
+            .filter(RollCall::isClosed)
```
GabrielFleischer, line 118: Yes ! Nice catch !

GabrielFleischer (approved):
Great !

Nice catch

ahominal (approved):
Seems good !

PR #1117, opened 2022-06-14
Meeting tests

Tesa27:
Meeting objects were never tested on Scala, but this should be done if one day Meeting objects become more important so that we know that tests have to be adapted and avoid potential vulnerabilities and bugs.


lourkeur (dismissed):

tuomas-p (changes requested):
Minor change request for the inverted test names, otherwise good job!

be2-scala/src/test/scala/ch/epfl/pop/pubsub/graph/validators/MeetingValidatorSuite.scala
```diff
@@ -0,0 +1,256 @@
...
+    val messageStandardActor: GraphMessage = MeetingValidator.validateStateMeeting(STATE_MEETING_INVALID_START_RPC)
+    message shouldBe a[Right[_, PipelineError]]
+    messageStandardActor shouldBe a[Right[_, PipelineError]]
+    system.stop(dbActorRef.actorRef)
+  }
+
+  // Creation time > End Time
+   test("Creating and invalid meeting state with end time < creation time"){
+    val dbActorRef = mockDbWorkingSetup
+    println(dbActorRef)
```
lourkeur, line 219:
```suggestion
```

lourkeur (approved):

tuomas-p (approved):
Looks good to me now!

PR #1120, opened 2022-06-14
Fix end time displayed on roll call screen

Warhedgehog:
When the roll call is opened the android app would display its end the 1.1.1970. This fixes it


Tesa27 (approved):
LGTM

MaximeZmt (approved):
LGTM 👍

jbsv (approved):

PR #1123, opened 2022-06-14
data: fix out-of-spec signatures

lourkeur:
as mentioned on Slack, I messed up in #1000 an inserted some extra embellishment, shall we say, to the signature payload.

PR #1126, opened 2022-06-15
[BE1] add signature check for digital cash

MaximeZmt:
Add Signature check for back-end 1 for digital cash project.
It will do the same check then as the scala back-end.

Waiting that https://github.com/dedis/popstellar/pull/1123 is merged, then merging on master

### DIFF
--- a/be1-go/message/messagedata/coin_post_transaction.go
+++ b/be1-go/message/messagedata/coin_post_transaction.go
@@ -70,7 +70,7 @@ func (message PostTransaction) Verify() error {
 
 	err = message.verifyTransactionId()
 	if err != nil {
-		return xerrors.Errorf("failed too verify the transaction id: %v", err)
+		return xerrors.Errorf("failed to verify the transaction id: %v", err)
 	}
 
 	err = message.verifySignature()
--- a/be1-go/message/messagedata/coin_post_transaction.go
+++ b/be1-go/message/messagedata/coin_post_transaction.go
@@ -1,13 +1,13 @@
 package messagedata
 
 import (
+	"bytes"
 	"encoding/base64"
 	"go.dedis.ch/kyber/v3/sign/schnorr"
 	"golang.org/x/xerrors"
 	"popstellar/crypto"
 	"popstellar/message/answer"
 	"strconv"
-	"strings"
 )
 
 // PostTransaction defines a message data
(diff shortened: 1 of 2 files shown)


be1-go/message/messagedata/coin_post_transaction.go
```diff
@@ -66,6 +68,20 @@ func (message PostTransaction) Verify() error {
 			"encoded", message.TransactionID)
 	}
 
+	err = message.verifyTransactionId()
+	if err != nil {
+		return err
```
Robingoumaz, line 73: You should wrap the error

be1-go/message/messagedata/coin_post_transaction.go
```diff
@@ -66,6 +68,20 @@ func (message PostTransaction) Verify() error {
...
+	}
+
+	err = message.verifySignature()
+	if err != nil {
+		return err
```
Robingoumaz, line 78: Same here

Robingoumaz (changes requested):
Very good, just need to wrap two errors

lourkeur (approved):
LGTM

lourkeur (approved):

Robingoumaz (approved):

be1-go/message/messagedata/coin_post_transaction.go
```diff
@@ -113,6 +129,50 @@ func (message PostTransaction) Verify() error {
...
+		}
+
+		publicKeySender, err := base64.URLEncoding.DecodeString(inp.Script.PubKey)
+		if err != nil {
+			return xerrors.Errorf("failed to decode public key string: %v", err)
+		}
+
+		err = schnorr.VerifyWithChecks(crypto.Suite, publicKeySender, dataBytes, signatureBytes)
+		if err != nil {
+			return xerrors.Errorf("failed to verify signature : %v", err)
```
Tesa27, line 169: I think that a custom error from `answer` should be applied here instead of `xerrors`

MaximeZmt: Only Here ? or for all of them ? What would concretely differ from doing that?

be1-go/message/messagedata/coin_post_transaction.go
```diff
@@ -113,6 +129,50 @@ func (message PostTransaction) Verify() error {
 	return nil
 }
 
+func (message PostTransaction) verifySignature() error {
+	var sigComp []string
```
nkcr, line 133: what you need here is a `bytes.Buffer`:

```go
sigComps := new(bytes.Buffer)
...
sigComps.WriteString(value)
sigComps.WriteString(hash)
...
dataBytes := sigComps.Bytes()
```

Issue #1078, opened 2022-06-06
[BUG] Re-Subscriptions after re-connecting are not handled propertly


Tyratox:
### Description (Actual behavior)
fe1 does not re-subscribe to sub-channels of LAOs after closing/refreshing the app, re-opening it and re-connecting to a LAO. (Or the connection breaks). This behavior is fine assuming subscriptions are per public key but the backends both handle subscriptions by TCP socket which results in the subscription state being if a connection breaks.

### Expected behavior
Two options, either

1) The front-ends re-subscribe to all relevant channels after a connection breaks and when it connects to a LAO.
2) The backends keep track of the subscription state by publicKey and map the public keys to a list of sockets so that they can look up all subscribers (public keys) for a given channel and then obtain all sockets corresponding to these public keys.

### How to reproduce in fe1
1. Create a LAO
2. Create a roll call, open it, add yourself, close it
3. Create an election
4. Refresh the web page
5. Re-connect to the LAO
6. Open the election -> The UI will **not** update

### Version & Environment
This bug was reproduced on:
- Commit: 7075a627a0bb72bd855bb6707ed9c87ae338ee00

##### Front-ends:
- [x] Fe1-Web
- [ ] Fe2-Android (not sure, I don't have an android phone)
- [ ] Not applicable

##### Back-ends:
- [x] Be1-Go
- [x] Be2-Scala
- [ ] Not applicable

### Workaround
Make sure the connection never breaks and the page is not refreshed.

### Impact
All communication in sub-channels is not working after the connection breaks / is reset.

### What to do

I am not sure whether it is best to fix this on the front-end or the back-end. It almost seems as if a back-end would be easier to implement (just another redirection via the public keys for looking up the sockets) compared to the front-end where it would require major changes (at least for fe1). To me the seemingly easiest option would be to not store but rather re-build the application state from the stored messages which would then trigger all appropriate subscribe actions. Depending on the number of messages this will take some time though. The problematic part is that this also would need to happen in cases where the connection breaks because the internet was interrupted for a quick moment.

PR #1127, opened 2022-06-15
Fix UX issues

Tyratox:
This PR fixes several UX issues

- Adds the ability to go back from the connect & lao screen via the navigation bar
- Adds a more prominent X button to disconnect from a lao
- Prevents pop tokens from being "lost" when scanning tokens as the organizer. (The tokens were not saved if you scanned them, closed the scanner and re-opened the scanner)
- Show message when a given pop token is scanned twice
- Only show the camera reverse button if more than one video input device is detected
- Update the event list regularly
- Add workaround for https://github.com/dedis/popstellar/issues/1078. fe-1 now simply stores the set of all subscribed channels and re-subscribes when re-connecting.

It also fixes an issue with the tests where a given test was dependent on the CI current time 😅


fe1-web/src/features/home/screens/Home.tsx
```diff
@@ -1,20 +1,53 @@
...
+        // disconnect from this lao
+        // FIXME
+        // In the best case we would just call the function here
+        // unfortunately react-native-navigation sometimes still has lao screens loaded
+        // resulting in the application crashing
+        // setTimeout delays the function call but a this workaround is a bit brittle
+        // the cleanest solution would probably be to navigate to this screen
+        // if no an error indicating no active lao is thrown
+        // react error boundaries might help achieving this
+        // setTimeout(disconnectFromLao, 1000);
```
0xdieman, line 46: Is the `setTimeout` also supposed to be a comment ?

0xdieman (dismissed):
Yes 👍

fe1-web/src/features/home/screens/Home.tsx
Tyratox: Yes :( The problem is that I managed to make this „fix“ break several times when testing so I disabled it.
(By disconnecting from a LAO and re-connecting within this one second.) I think I could try to add a clearTimeout on unmount 👍

Tyratox: Just tested it and it does not work :(  Should I remove this code all together?

Tyratox: I just removed the code and created issue #1132

fe1-web/src/features/events/components/EventList.tsx
```diff
@@ -43,16 +43,47 @@ const EventList = () => {
...
+  const eventListSelector = useMemo(() => makeEventListSelector(laoId.valueOf()), [laoId]);
+  const events = useSelector(eventListSelector);
+
+  const [{ pastEvents, currentEvents, futureEvents }, setEvents] = useState<{
+    pastEvents: EventState[];
+    currentEvents: EventState[];
+    futureEvents: EventState[];
+  }>(() => {
+    const [newPastEvents, newCurrentEvents, newFutureEvents] = categorizeEventsByTime(
+      Timestamp.EpochNow().valueOf(),
```
pierluca, line 55: it seems a bit odd in the API that you'd pass a number, instead of a `Timestamp`. 
You might want to do the `valueOf` conversion internally.

fe1-web/src/features/home/interface/Configuration.ts
```diff
@@ -46,6 +53,11 @@ export interface HomeCompositionConfiguration {
    */
   useLaoList: () => HomeFeature.Lao[];
 
+  /**
+   * Gets the function to disconnect from the current lao
+   */
+  useDisconnectFromLao: () => () => void;
```
pierluca, line 59: would it be practical to disconnect by id , since we could be connected to multiple LAOs (in the future) ?

fe1-web/src/features/lao/components/LaoItem.tsx
```diff
@@ -50,13 +51,12 @@ const LaoItem = ({ lao, isFirstItem, isLastItem }: IPropTypes) => {
         getNetworkManager().connect(address),
       );
 
-      const channel = getLaoChannel(lao.id.valueOf());
```
pierluca, line 53: I can't comment a few lines above but I have to mention that `onPress` isn't a great name for a `onPress` callback.
The ListItem attribute already says that it's called on press, the function name should say what it actually does :-)

fe1-web/src/features/lao/components/LaoItem.tsx
```diff
@@ -50,13 +51,12 @@ const LaoItem = ({ lao, isFirstItem, isLastItem }: IPropTypes) => {
...
-      if (!channel) {
-        throw new Error('The given LAO ID is invalid');
-      }
-
-      // subscribe to the lao channel on the new connections
-      await subscribeToChannel(channel, connections);
+      // subscribe to the lao channel (or all previously subscribed to channels) on the new connection
+      await Promise.all(
+        lao.subscribed_channels.map((channel) =>
+          subscribeToChannel(lao.id, dispatch, channel, connections),
```
pierluca, line 57: I've skipped over it the first few times, waiting to see its usage, but I have to say that this new API does feel quite weird to me, as it leads to a lot of code being repeated here and there.
Can you walk me through why we need such a complex function for subscribe ?

In particular, I'm thinking as opposed to having simple APIs such as 
`connect(lao, dispatch)`
`resubscribe(lao, dispatch)`
that abstracts all of that away

fe1-web/src/features/lao/components/LaoItem.tsx
Tyratox: That is true :)

fe1-web/src/features/lao/components/LaoItem.tsx
Tyratox: Yes of course :) The main problem is #1132. To fix it I wanted to keep track of all subscribed channels for each LAO and thus `subscribeToChannel` must be able to to change the redux state (and thus have access to `dispatch`). Analogously it needs to know for which lao we are doing the subscribe. Lastly, the `channel` argument should be self-explanatory.

Now `dispatch` could be removed or made an optional argument if the function would simply use the global `dispatch` function. Removing it would require us to always configure the global store in tests while as an optional argument we could still use dependency injection in tests. But since we always have access to a dispatch function where `subscribeToChannel`, I thought we can also make it a required argument. What do you think?

Also the last argument, `connections`, already is optional. This is for the case where you have multiple connections for the same LAO (e.g. organizer and witnesses) but only want to send a message on a specific connection. (For example the subscribe & catchup only on the newly added connection to the new witness)

fe1-web/src/features/home/interface/Configuration.ts
Tyratox: Hmm I guess this is possible but I am not sure if it makes a lot of since at the moment. What would that function do? Read the currentId from the lao reducer, compare it to the passed one and only disconnect if it matches?

Also what id would we pass to the function? Just the currentId? Then it would seem a bit weird to me to start introducing partial support for this feature while the remaining codebase has no hints at that connections to multiple LAOs will ever be supported

pierluca (approved):
I'll get back to you on the comments, but I'm approving in the mean time ;)

Tyratox:
Restored the branch for potential further fixes :)

fe1-web/src/features/home/interface/Configuration.ts
pierluca: Fair enough :)

fe1-web/src/features/lao/components/LaoItem.tsx
pierluca: Makes sense, though I might have an idea for #1132 . I'll add it there

PR #1128, opened 2022-06-16
Dashboard + some more karate tests

Tesa27:
- This PR gives a good idea what our system guarantees and to which extent. It creates a dashboard with all the features we tested this semester and how well the system is performing in each subsystem for each feature.
- It also fixes some karate tests for election to address the change in specifications regarding the new use of the `version` field and the change of type for cast vote.
- Adds tests for the remaining roll call messages (open and close) to test that a non-organizer cannot close or open a roll call from the point of view of the back-end


GabrielFleischer (approved):
LGTM

PR #1130, opened 2022-06-18
Roll-call open front end integration test 

Warhedgehog:
This adds a karate test for roll call open and adds a check to the existing roll call create. 

- Roll-call id is now checked for create
- Test utils have been refactored
- roll-call open is checked for the following :
    - message sent is a publish
    - publish fields msg id and signature
    - update_id
    - message object and action
    - no subsequent message is sent


GabrielFleischer (approved):
LGTM

PR #1133, opened 2022-06-20
Election messages: timestamp checks

Tesa27:
Timestamps were not checked very well on the go back-end on election messages, meaning we could have an election open that has a creation time prior to election setup as allowed. This is a bogus behavior since it can in some cases lead to a replay attack.


be1-go/channel/election/verification.go
```diff
@@ -196,12 +200,17 @@ func (c *Channel) verifyMessageElectionEnd(electionEnd messagedata.ElectionEnd)
...
+		return answer.NewInvalidMessageFieldError("election end cannot have a creation time prior to election setup %d %d", electionEnd.CreatedAt, c.createdAt)
+	}
+
 	// verify order of registered votes
 	if len(electionEnd.RegisteredVotes) != 0 {
 		err := verifyRegisteredVotes(electionEnd, &c.questions)
 		if err != nil {
 			c.log.Err(err).Msgf("problem with registered votes: %v", err)
-			return nil
+			return err
```
nkcr, line 213: please wrap the error

nkcr (approved):

Issue #1132, opened 2022-06-20
[BUG] fe-1 does not properly disconnect from a LAO


Tyratox:
### Description (Actual behavior)
When disconnecting from a LAO, the user is navigated to the correct screen but  fe-1 does not unset `currentId` in the lao reducer.

### Expected behavior
The `currentId` in the lao reducer should be set to undefined after disconnecting.

### How to reproduce
1. Connect to a LAO
2. Disconnect from a LAO
3. Check the redux state of the lao reducer

### Version & Environment
This bug was reproduced on:
- Commit: 3cd501d7038af61d4d197607970869b9308a8cf4

##### Front-ends:
- [x] Fe1-Web (please include browser's names & version)
- [ ] Fe2-Android (please specify if phone or emulation, and Android version)
- [ ] Not applicable

##### Back-ends:
- [ ] Be1-Go
- [ ] Be2-Scala
- [x] Not applicable

### Workaround
Connect to a different lao

### Impact
Certain screens such as the wallet will behave differently (as if you were still connected to the previous LAO)

### Possible root cause
The issue is not easy to fix because of how react-navigation loads and unloads screens. If the value is immediately unset, the application crashes because the lao screen expects it to be set but is not instantly unloaded.
A workaround would be to delay this update but it is not clear how long: A long delay will make the application crash if you quickly re-connect to a second lao (because you are then "disconnected" while the lao navigation is open) and a short delay might not work on all devices. I currently do not see a nice way to fix it. Maybe react error boundaries would help catching and handling the exceptions correctly.

pierluca:
If I understand correctly the root cause analysis above, the issue is that you want to navigate before clearing the state, because clearing the state first and navigating afterwards might fail, due to the way react navigation works.
You don't want to use delays, because you (correctly) analyze that they might introduce their own set of (subtle) bugs.
Since we can't use delays, we need to do so in reaction to some event, and we know that changing page must happen before the unsetting, as the alternative leads to a crash.

Therefore, how about unsetting the current Lao ID when loading the correct page, triggering this logic by passing an extra argument to the page via react navigation ?  
I.e. we tell the application to change page on disconnect, passing an extra parameter, and as the page loads, it checks the parameter and clears the state.
What do you think? Is it feasible?

Tyratox:
Yes this would be feasible and I tried it here https://github.com/dedis/popstellar/commit/15df578b4c1a4ba412e544ced4baa1afda76f13e.

[...]

PR #1144, opened 2022-06-24
Add error boundary to handle #1132

Tyratox:


fe1-web/src/features/lao/navigation/LaoNavigation.tsx
```diff
@@ -51,49 +52,51 @@ const LaoNavigation: React.FC = () => {
...
-              tabBarIcon: tabBarIcon || undefined,
-              // hide the item if tabBarIcon is set to null
-              tabBarItemStyle: tabBarIcon === null ? { display: 'none' } : undefined,
-              headerShown,
-            }}
-          />
-        ),
-      )}
-    </OrganizationTopTabNavigator.Navigator>
+    <NoCurrentLaoErrorBoundary>
```
Xelowak, line 55: So if an error happens there, the whole component will be replaced by the simple text (in `NoCurrentLaoErrorBoundary.tsx`) saying that you're not connected to a LAO. Am I right?

fe1-web/src/resources/strings.ts
```diff
@@ -103,6 +103,7 @@ namespace STRINGS {
   export const lao_properties_current_connections = 'You are currently connected to';
   export const lao_properties_add_additional_connection = 'Add connection';
   export const lao_properties_disconnect = 'Disconnect';
+  export const lao_no_current = 'You are currently no connected to any LAO';
```
Xelowak, line 106:
```suggestion
  export const lao_no_current = 'You are currently not connected to any LAO';
```

Xelowak (commented):
I've asked you a simple question about the behaviour of `NoCurrentLaoErrorBoundary.tsx` and you've made a typo in your string, otherwise everything seems fine.
Thanks for the modification of `useCurrentLaoId` into `useAssertCurrentLaoId`, it's clearer this way.

By the way, don't hesitate to add documentation to `NoCurrentLaoErrorBoundary.tsx`.

fe1-web/src/features/lao/navigation/LaoNavigation.tsx
Tyratox: Exactly :) But yes I think you're right, I should add some documentation to the file :)

Tyratox: Unfortunately at the moment during development the error still triggers a popup but the UI does not break and the error should not show up in production. The reason is that it somehow is caught but then somehow still "thrown". I am not yet sure how to prevent this since this is an error we expect to happen.

Tyratox: Actually it seems to be intended this way 😅 https://stackoverflow.com/a/47400249/2897827

PR #1150, opened 2022-06-26
Fixing some scala tests(checking for now)

Tesa27:
Some LAO creation tests were passing but not testing the right functionality. Lao creation with wrong sender and empty LAO name were correctly failing but due to the id being wrongly computer and not because of the previously mentioned reasons, this PR fixes that and introduces an additional test for lao greet message.


tuomas-p (approved):
LGTM! Nice catches

PR #1151, opened 2022-06-28
Add feature READMEs

Tyratox:
This PR adds some short description of what each feature does and what its dependencies are.

### DIFF
--- a/fe1-web/src/features/home/README.md
+++ b/fe1-web/src/features/home/README.md
@@ -1,0 +1,9 @@
+# Home
+
+Provides UI while the user is not connected to a LAO. Currently this includes a screen for showing previously connected LAOs, the component for which is provided by the `lao` feature. Moreover there are is a screen for launching new LAOs, one enabling scanning QR codes for connecting to LAOs and one for manually connecting to a LAO.
+
+## Dependencies
+
+- Basic LAO functionality provided by the `lao` feature
+
+See `interface/Configuration.ts` for more details.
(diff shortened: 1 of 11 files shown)


fe1-web/src/features/home/README.md
```diff
@@ -0,0 +1,9 @@
+# Home
+
+Provides UI while the user is not connected to a LAO. Currently this includes a screen for showing previously connected LAOs, the component for which is provided by the `lao` feature. Moreover there are is a screen for launching new LAOs, one enabling scanning QR codes for connecting to LAOs and one for manually connecting to a LAO.
```
Xelowak, line 3: This line might be a bit too wide compared to other `README`s.

fe1-web/src/features/witness/README.md
```diff
@@ -0,0 +1,16 @@
+# Witnessing
+
+Should provide the witness functionality. Currently only supports witness
```
Xelowak, line 3: When you use "currently", it could be nice to add a timing context like the current date so that we know when it was written without having to check GitHub.

Xelowak (commented):
Thanks a lot for adding this! I made 2 tiny remarks, but it's great overall.

fe1-web/src/features/home/README.md
Tyratox: Opps, you are totally right. Forgot to run the re-wrapping function of the IDE 😅

fe1-web/src/features/witness/README.md
Tyratox: True 👍

Xelowak (approved):
That's perfect!

PR #1152, opened 2022-07-01
Updating Simple scenarios to respect self containment and deleting all json files

Tesa27:
We modularized the writing of karate tests by adding simple building blocks that can serve as a baseline for putting the backend in a certain state. The first approach had a lot of file reading in the tests and even when constructing simple scenarios that are these building blocks mentioned before. Now we found a way to have even the building blocks self contained and more readable so we do not require the json files with all the message data.


tests/karate/src/test/java/fe/utils/verification/MessageVerification.java
```diff
@@ -30,7 +30,8 @@ public boolean verifyMessageIdField(String message) {
     String signature = messageFieldJson.get(SIGNATURE);
     String msgId = messageFieldJson.get(MESSAGE_ID);
     try {
-      return msgId.equals(JsonConverter.hash(data.getBytes(), signature.getBytes()));
+      JsonConverter jsonConverter = new JsonConverter();
+      return msgId.equals(jsonConverter.hash(data.getBytes(), signature.getBytes()));
```
GabrielFleischer, line 34: What is the reason of this change ?

GabrielFleischer (commented):
There is a change whose purpose I am not certain to understand. Could you clarify it ?

tests/karate/src/test/java/fe/utils/verification/MessageVerification.java
Tesa27: The usage of the method hash always required an instance of `JsonConverter`. But @Warhedgehog changed the hash method to static, which made all the other usages where hash was invoked on a` JsonConverter` object illegal and there was quite a few of them. So I had to switch back to the non static usage of the hash method and update all the places where it was used which was only 3 times so changes are not drastic and the functionality still remains the same.

GabrielFleischer: Ok

GabrielFleischer (commented):
LGTM

GabrielFleischer (approved):
Woops, miss clicked

PR #1156, opened 2022-07-03
Fix fe1 CI fail

Warhedgehog:
Snapshots needed to be updated which made the fe1 CI fail


Xelowak (approved):
LGTM, but this has to be changed if it relies on current time

PR #1163, opened 2022-07-05
Election setup & open karate tests

Warhedgehog:
This adds karate integration tests for both election setup and open. 
It tests :
- Election setup with 1 question and 2 ballots : 
    - message sent is a publish
    - publish fields msg id and signature
    - election name
    - election id
    - message object and action
    - question id
    - question
    - ballots
    - voting method
    - no subsequent message is sent
- Election open :
    - message sent is a publish
    - publish fields msg id and signature
    - message object and action
    - no subsequent message is sent


tests/karate/src/test/java/fe/RollCall/rollCallCreation.feature
```diff
@@ -28,7 +28,7 @@ Feature: Create RollCall
     # Roll Call specific verification
     And match verificationUtils.getObject(create_rc_string) == constants.ROLL_CALL
     * match verificationUtils.getAction(create_rc_string) == constants.CREATE
-    * match (rollCallVerification.verifyRollCallName(create_rc_string, constants.RC_NAME)) == true
+    * match verificationUtils.getName(create_rc_string) == constants.RC_NAME
```
GabrielFleischer, line 31: Great change

tests/karate/src/test/java/fe/utils/verification/ElectionVerification.java
```diff
@@ -0,0 +1,92 @@
...
+  public boolean verifyElectionId(String message) {
+    Json setupMessageJson = getMsgDataJson(message);
+
+    String electionId = setupMessageJson.get(ID);
+    String createdAt = getStringFromIntegerField(setupMessageJson, CREATED_AT);
+    String laoId = setupMessageJson.get(LAO);
+    String electionName = setupMessageJson.get(NAME);
+
+    try {
+      JsonConverter jsonConverter = new JsonConverter();
```
GabrielFleischer, line 32: This definitely needs to be renamed 😅

tests/karate/src/test/java/fe/utils/web.feature
```diff
@@ -89,7 +99,7 @@ Feature: web test
     * script("setTimeout(() => document.evaluate('//div[text()=\\'Create Roll-Call\\']', document, null, XPathResult.FIRST_ORDERED_NODE_TYPE, null).singleNodeValue.click(), 1000)")
 
     # Provide roll call required information
-    And retry(5,1000).input(roll_call_title_selector, constants.RC_NAME)
+    And retry(5, 1000).input(roll_call_title_selector, constants.RC_NAME)
```
GabrielFleischer, line 102: If you want to write the retry() that way, you should change it everywhere.

(I think it is better like that 👍)

GabrielFleischer (approved):
Nice !

I left two comments. It can arguably be done in another PR as the change would have a wide effect on the codebase.

I let you choose.

tests/karate/src/test/java/fe/utils/verification/ElectionVerification.java
Warhedgehog: I could not agree more. I won't do it here though

tests/karate/src/test/java/fe/utils/web.feature
Warhedgehog: If I do it, it will be in a subsequent PR

PR #1174, opened 2022-07-21
Reformat Android project

GabrielFleischer:
Reformat the whole android project and change the rule regarding imports


GabrielFleischer:
It would be a good idea to try a reformat the project locally to make sure the rule are consistent among multiple setups

PR #1176, opened 2022-07-21
Reformat Android project

GabrielFleischer:
Reformat the whole android project and change the rule regarding imports

It would be a good idea to try a reformat the project locally to make sure the rule are consistent among multiple setups


pierluca:
@Warhedgehog Could you take care of the quick test that @GabrielFleischer is asking ? Thanks !

Warhedgehog:
On my end the imports are not the same as the code pulled from the branch. Static imports are on top on my end. Furthermore I do not have the nice grouping of pop related imports, test related imports, etc...
I've re-imported the import scheme provided in the code convention section of the documentation but it did not help

GabrielFleischer:
Ok, I think I need to add these specifications into the configuration file.

Thanks for finding that

GabrielFleischer:
I applied new rules to the configuration, it should be done

pierluca (approved):
All good for me. Let me know if I need to force-merge it to bypass the Go flakey test.

Warhedgehog (approved):
LGTM

PR #1177, opened 2022-07-21
Remove Livedata usage for events in the home activity 

GabrielFleischer:
This PR removes the usage of live data when dispatching events.

Currently, the logic on how to switch to a fragment/activity is held in the view model.
This is not the best way to do it and should be addressed later.


Warhedgehog:
I've not yet looked in depth but you might want to fix the little code smell

pierluca:
About that code smell, shouldn't we just switch that function to use a primitive boolean, instead of `java.lang.Boolean` ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeActivity.java
```diff
@@ -405,28 +165,45 @@ private void handleSocialMediaNavigation() {
...
   }
 
+  public void requestCameraPermission() {
+    // Setup result listener to open the connect tab once the permission is granted
+    getSupportFragmentManager()
+        .setFragmentResultListener(
+            CameraPermissionFragment.REQUEST_KEY, this, (k, b) -> openConnect());
+
+    mViewModel.openCameraPermission(getSupportFragmentManager(), getActivityResultRegistry());
+  }
```
pierluca, line 205: what happens if the camera permission is denied ?
I don't see a code path to deal with that situation / log it / warn the user in some way.

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeViewModel.java
```diff
@@ -162,34 +147,25 @@ protected void onCleared() {
...
-                      .subscribe(lao.getChannel())
-                      .subscribe(
-                          () -> {
-                            Log.d(TAG, "subscribing to LAO with id " + lao.getId());
-                            openLAO(lao.getId());
-                          },
-                          error ->
-                              ErrorUtils.logAndShow(
-                                  getApplication(), TAG, error, R.string.error_create_lao));
+                  Log.d(TAG, "subscribing to LAO with id " + lao.getId());
```
pierluca, line 167: once this is called, we're already subscribed, right ?

pierluca (commented):
Looks good, barring minor details.
I'm not too familiar with all the operations surrounding fragments though, @Warhedgehog will have a more informed opinion on that :-)

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeActivity.java
GabrielFleischer: It will reopen Connect which will redo the permission check and if the test fails it will re-ask the user.

It is basically a loop.

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeViewModel.java
GabrielFleischer: Yes, good point

GabrielFleischer: Fixed

GabrielFleischer:
> About that code smell, shouldn't we just switch that function to use a primitive boolean, instead of `java.lang.Boolean` ?

The thing is, LiveData produces Booleans which are Objects and can therefore be null.

I made a change that addresses the code smell.

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeActivity.java
Warhedgehog: Upon testing, I noticed that when denying, we go to the home fragment. In itself, it is not a problem but after twice denying, when clicking on the allow button (ours) it does not ask for permission but goes back to home like I denied permission

Moreover, I think the behaviour should be commented and if possible the logic behind. Imo it is not immediately clear from the code.

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeViewModel.java
```diff
@@ -199,147 +170,112 @@ public void launchLao() {
...
-    mOpenHomeEvent.postValue(new SingleEvent<>(true));
+  public void openLaunch(FragmentManager manager) {
+    setCurrentFragment(manager, R.id.fragment_launch, LaunchFragment::newInstance);
   }
 
   public void openConnecting(String laoId) {
-    mOpenConnectingEvent.postValue(new SingleEvent<>(laoId));
+    Intent intent = new Intent(getApplication().getApplicationContext(), ConnectingActivity.class);
+    intent.putExtra(Constants.LAO_ID_EXTRA, laoId);
+    getApplication().startActivity(intent);
```
Warhedgehog, line 210: Did you forget to pass the activity in the arguments like the other activity start ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeActivity.java
```diff
@@ -279,7 +129,9 @@ public boolean onCreateOptionsMenu(Menu menu) {
   @Override
   public boolean onOptionsItemSelected(MenuItem item) {
     if (item.getItemId() == R.id.settings) {
-      mViewModel.openSettings();
+      Intent intent = new Intent(this, SettingsActivity.class);
+      Log.d(HomeViewModel.TAG, "Trying to open settings");
+      startActivity(intent);
```
Warhedgehog, line 134: Most of the activity starting is done in ViewModel, is there a reason to do it here ?

Warhedgehog (changes requested):
Overall massive improvement of course. Couple things I'd like addressed 

- The new implementation of camera permission has a bug I think (double deny)
- There is some fragment and activity navigation in both the viewModel and the activity, should we not stick with just 1 ? Also some of the activity navigation methods are static and some not, would it not be better to be consistent ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeActivity.java
```diff
@@ -226,25 +89,12 @@ private void subscribeWalletEvents() {
   @Override
   protected void onResume() {
     super.onResume();
-    mViewModel.openHome();
+    mViewModel.openHome(getSupportFragmentManager());
```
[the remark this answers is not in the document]
GabrielFleischer: It would be more impactful and, to me, go outside the scope of the PR.

We could do a PR to move the activity and fragment start in their respective class and then apply the livedata change.

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeViewModel.java
GabrielFleischer: Yes 😅

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeActivity.java
GabrielFleischer: I missed this one

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeActivity.java
```diff
@@ -55,7 +46,7 @@ protected void onCreate(Bundle savedInstanceState) {
     super.onCreate(savedInstanceState);
     setContentView(R.layout.home_activity);
 
-    setupHomeFragment();
+    setCurrentFragment(getSupportFragmentManager(), R.id.fragment_home, HomeFragment::newInstance);
```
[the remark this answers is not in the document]
GabrielFleischer: I moved everything wher the call was made. That way the reader knows what is happening where the call is made.

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeViewModel.java
```diff
@@ -92,20 +74,14 @@ public HomeViewModel(
...
 
     mLAOs =
         LiveDataReactiveStreams.fromPublisher(
-            this.laoRepository.getAllLaos().toFlowable(BackpressureStrategy.BUFFER));
-  }
-
-  @Override
-  public void onPermissionGranted() {
-    openQrCodeScanning();
+            laoRepository.getAllLaos().toFlowable(BackpressureStrategy.BUFFER));
```
Warhedgehog, line 84: Why this change ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeViewModel.java
```diff
@@ -199,147 +166,44 @@ public void launchLao() {
...
-  }
-
   public void setLaoName(String name) {
     this.mLaoName.setValue(name);
   }
 
-  public void setIsWalletSetUp(Boolean isSetUp) {
+  public void setIsWalletSetUp(boolean isSetUp) {
     this.mIsWalletSetUp.setValue(isSetUp);
   }
```
Warhedgehog, line 190: I know this is not from you and not the point of the PR but would it not be better to use postValue ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeViewModel.java
```diff
@@ -199,147 +166,44 @@ public void launchLao() {
...
     wallet.logout();
     setIsWalletSetUp(false);
-    openWallet();
+  }
+
+  public void openConnecting(String laoId) {
+    Intent intent = new Intent(getApplication().getApplicationContext(), ConnectingActivity.class);
+    intent.putExtra(Constants.LAO_ID_EXTRA, laoId);
+    intent.addFlags(FLAG_ACTIVITY_NEW_TASK);
+    getApplication().startActivity(intent);
```
Warhedgehog, line 207: I'm not too fan about having some activity navigation in the viewModel but I have no better idea

Warhedgehog (approved):
LGTM. Couple questions/remark with which you do what you want :)

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeViewModel.java
GabrielFleischer: With the way the QR Scanner is done, there is no clean way to move this. So for now it will stay here.

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeViewModel.java
GabrielFleischer: Maybe, I am not certain of what the actual impact would be

PR #1179, opened 2022-07-23
Fix an issue with LAONetworkManager

GabrielFleischer:
Currently, the messages are sent when the Completable are created
The fix makes sure they are sent only when the subscription is done

This was not an issue, but with #1177 I started chaining requests, and in that scenario, we need to delay the sending of the second message.
What this simple fix achieves

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/remote/LAONetworkManager.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/remote/LAONetworkManager.java
@@ -182,36 +182,35 @@ public class LAONetworkManager implements MessageSender {
   }
 
   private Single<Answer> request(Query query) {
-    Single<Answer> answerSingle =
-        connection
-            .observeMessage() // Observe incoming messages
-            .filter(Answer.class::isInstance) // Filter the Answers
-            .map(Answer.class::cast)
-            // This specific request has an id, only let the related Answer pass
-            .filter(answer -> answer.getId() == query.getRequestId())
-            .doOnNext(answer -> Log.d(TAG, "request id: " + answer.getId()))
-            // Transform from an Observable to a Single
-            // This Means that we expect a result before the source is disposed and an error will be
-            // produced if no value is received.
-            .firstOrError()
-            // If we receive an error, transform the flow to a Failure
-            .flatMap(
-                answer -> {
-                  if (answer instanceof Error) {
-                    return Single.error(new JsonRPCErrorException((Error) answer));
-                  } else {
-                    return Single.just(answer);
-                  }
-                })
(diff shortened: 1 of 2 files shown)


Warhedgehog (dismissed):
Nice catch ! LGTM

fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/remote/LAONetworkManager.java
```diff
@@ -182,36 +182,35 @@ private void handleMessages(List<MessageGeneral> messages, Channel channel) {
...
+                .observeMessage() // Observe incoming messages
+                .filter(Answer.class::isInstance) // Filter the Answers
+                .map(Answer.class::cast)
+                // This specific request has an id, only let the related Answer pass
+                .filter(answer -> answer.getId() == query.getRequestId())
+                .doOnNext(answer -> Log.d(TAG, "request id: " + answer.getId()))
+                // Transform from an Observable to a Single
+                // This Means that we expect a result before the source is disposed and an error
+                // will be
+                // produced if no value is received.
```
pierluca, line 198:
```suggestion
                // will be produced if no value is received.
```

fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/remote/LAONetworkManager.java
```diff
@@ -182,36 +182,35 @@ private void handleMessages(List<MessageGeneral> messages, Channel channel) {
...
-            // Add a timeout to automatically dispose of the flow and end with a failure
-            .timeout(5, TimeUnit.SECONDS)
-            .cache();
-    // Only send the message after the single is created to make sure it is already waiting
-    // before the answer is received
-    connection.sendMessage(query);
-    return answerSingle;
+    // First send the message (embedded in the completable so that it is only sent upon subscription
+    return Completable.fromAction(() -> connection.sendMessage(query))
+        .andThen(
```
pierluca, line 187: I don't understand this and I'd like to :-) 

Aren't we first sending the query and only then doing the subscribing , in such a way that theoretically we could be "missing" the reply because we weren't yet subscribed to it ?

GabrielFleischer: Right, I recreated the old bug while fixing the new one.

GabrielFleischer: I pushed a potential fix, I tested it locally, and it seemed to work, but I am not certain

pierluca (approved):
This looks better, I feel even I understand it 😁

PR #1181, opened 2022-07-25
Refactor Connecting process

Warhedgehog:
Removes unnecessary verbose activity + viewModel + fragment by a standalone activity


fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/ConnectingActivity.java
```diff
@@ -0,0 +1,117 @@
...
+  private final CompositeDisposable disposables = new CompositeDisposable();
+
+  @Inject GlobalNetworkManager networkManager;
+
+  @Override
+  protected void onCreate(Bundle savedInstanceState) {
+    super.onCreate(savedInstanceState);
+    setContentView(R.layout.connecting_activity);
+    String channelId = (String) getIntent().getExtras().get(Constants.LAO_ID_EXTRA);
+    setupCancelButton();
```
pierluca, line 38: Is it standard practice to put private functions such as this one after protected ones ? 
It seems a bit unnatural for me to have to scroll down 50 lines to discover the 2 lines of code this call hides :sweat_smile:

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/ConnectingActivity.java
```diff
@@ -0,0 +1,117 @@
...
+    connectingText.setText(channelId);
+
+    // This object will allow us to get all the event of the underlying observable, even if they
+    // were emitted prior to the subscribe
+    ConnectableObservable<WebSocket.Event> replay =
+        networkManager.getMessageSender().getConnectEvents().replay();
+    disposables.add(
+        replay.subscribe(
+            v -> {
+              if (v.toString().contains("OnConnectionOpened")) {
```
pierluca, line 108: Is there no better way than a comparison with a hardcoded string?
Can we use types? Or at least a documented constant ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/ConnectingActivity.java
Warhedgehog: I don't know about any convention regarding the matter. But in the project, in activities this is consistent (at least mostly) as the onCreate is the first function usually. 
I could have organized this a tad better for github reviews though I will admit :). For IDE browsing it's not a problem IMO since ctrl click does wonders.
I'll rearrange the private methods a bit

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/ConnectingActivity.java
Warhedgehog: You are perfectly right !

pierluca (approved):
LGTM, do you plan on adding a Robolectric test, or is this going to be covered only by Karate ? 
i.e., should I force-merge? :-)

Warhedgehog:
I did not plan on adding a RoboElectric test as most of the behaviour is scarlet related. There is no karate test for now but that would be an important inclusion and not too difficult I think.
I've tested it manually only. I think adding a karate test would be in a future PR

PR #1182, opened 2022-07-26
ViewModel Refactoring

Warhedgehog:
This refactores the ViewModels of DigitalCash, SocialMedia and Settings. For Settings given the limited complexity I deemed a standalone Activity to be more concise

All the features refactored were manually tested


fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/digitalcash/DigitalCashActivity.java
```diff
@@ -39,25 +39,24 @@ protected void onCreate(Bundle savedInstanceState) {
     setupNavigationBar();
     setupBackButton();
 
+    subscribeToSelectItemEvents();
     setTheIntent();
```
pierluca, line 43: not your fault, but the `setTheIntent` function has terrible naming. We're not setting an intent, we're configuring the current view based on intent data. do you have any idea for a better name ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/digitalcash/DigitalCashActivity.java
```diff
@@ -39,25 +39,24 @@ protected void onCreate(Bundle savedInstanceState) {
     setupNavigationBar();
     setupBackButton();
 
+    subscribeToSelectItemEvents();
```
pierluca, line 42: we subscribe to select/choose item events , or subscribe to the select**ed** item's events ? 
if I understand correctly, the latter :)

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/digitalcash/DigitalCashActivity.java
```diff
@@ -233,59 +101,83 @@ public static DigitalCashViewModel obtainViewModel(FragmentActivity activity) {
...
+            } else if (id == R.id.digital_cash_send_menu) {
+              Log.d(TAG, "Opening send fragment");
+              openSendFragment();
+            } else if (id == R.id.digital_cash_receive_menu) {
+              Log.d(TAG, "Opening receive fragment");
+              openReceiveFragment();
+            } else if (id == R.id.digital_cash_issue_menu) {
+              Log.d(TAG, "Trying to open issue fragment");
+              handleOpenIssue();
+            }
```
pierluca, line 128: shouldn't this be a switch ? 😁

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/digitalcash/DigitalCashActivity.java
```diff
@@ -233,59 +101,83 @@ public static DigitalCashViewModel obtainViewModel(FragmentActivity activity) {
   }
 
   public void setupNavigationBar() {
-    bottomNavigationView = findViewById(R.id.digital_cash_nav_bar);
-    bottomNavigationView.setOnItemSelectedListener(
+    navbar = findViewById(R.id.digital_cash_nav_bar);
+    navbar.setOnItemSelectedListener(
         item -> {
```
pierluca, line 106: would it be helpful to define the content of the listener in its own method ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/digitalcash/DigitalCashActivity.java
```diff
@@ -233,59 +101,83 @@ public static DigitalCashViewModel obtainViewModel(FragmentActivity activity) {
...
+        R.id.fragment_digital_cash_receive,
+        DigitalCashReceiveFragment::newInstance);
+  }
+
+  private void handleOpenIssue() {
+    if (!mViewModel
+        .getCurrentLao()
+        .getValue()
+        .getOrganizer()
+        .equals(mViewModel.getKeyManager().getMainPublicKey())) {
```
pierluca, line 167: it might be more readable / expressive to use intermediate variables.

pseudo-code:
```
    organizerKey = mViewModel.getCurrentLao().getValue().getOrganizer()
    myKey = mViewModel.getKeyManager().getMainPublicKey()

    if (!myKey.equals(organizerKey)) {
```

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/digitalcash/DigitalCashSendFragment.java
```diff
@@ -80,7 +80,10 @@ public void onViewCreated(@NonNull View view, @Nullable Bundle savedInstanceStat
                           Collections.singletonMap(currentPublicKeySelected, currentAmount));
                       mViewModel.updateReceiptAddressEvent(currentPublicKeySelected);
                       mViewModel.updateReceiptAmountEvent(currentAmount);
-                      mViewModel.openReceipt();
+                      DigitalCashActivity.setCurrentFragment(
```
pierluca, line 83: This whole function is hideously large / messy.

Maybe not in this PR, but I'd strongly suggest refactoring it in a subsequent PR, if it's not planned already 😅

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/socialmedia/SocialMediaActivity.java
```diff
@@ -152,46 +99,80 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+            } else if (itemId == R.id.social_media_search_menu) {
+              Log.d(TAG, "Opening search");
+              openSocialMediaSearchFragment();
+            } else if (itemId == R.id.social_media_following_menu) {
+              Log.d(TAG, "Opening following");
+              openSocialMediaFollowingFragment();
+            } else if (itemId == R.id.social_media_profile_menu) {
+              Log.d(TAG, "Opening profile");
+              openSocialMediaProfileFragment();
+            }
```
pierluca, line 122: same thing, this looks like it should be a switch. Is there a technical reason it can't be ? 🤔

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/socialmedia/SocialMediaActivity.java
```diff
@@ -152,46 +99,80 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
 
-  public void setupSocialMediaHomeFragment() {
+  private void subscribeToLaoName() {
+    // Subscribe to "lao name" string
+    mViewModel
+        .getLaoName()
+        .observe(
+            this,
+            newLaoName -> {
+              if (newLaoName != null) {
```
pierluca, line 135: In general, since we're doing nothing in the null case, I'd suggest structuring the subscribers as follows:

```
if (X == null) {
  return;
}

// the rest , unindented
```

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/socialmedia/SocialMediaSendFragment.java
```diff
@@ -47,32 +49,17 @@ public void onViewCreated(@NonNull View view, @Nullable Bundle savedInstanceStat
...
-    mSocialMediaSendFragBinding.sendChirpButton.setOnClickListener(
-        v -> mSocialMediaViewModel.sendNewChirpEvent());
+    mSocialMediaSendFragBinding.sendChirpButton.setOnClickListener(v -> sendNewChirp());
   }
 
   private void sendNewChirp() {
-    // Trying to send a chirp when no LAO has been chosen in the application will not send it and
-    // make a toast appear
+    // Trying to send a chirp when no LAO has been chosen in the application will not send it, will
+    // make a toast appear and will log the error
```
pierluca, line 60:
```suggestion
    // Trying to send a chirp when no LAO has been chosen in the application will not send it, it will
    // make a toast appear and it will log the error
```

fe2-android/app/src/test/java/com/github/dedis/popstellar/ui/pages/digitalcash/DigitalCashPageObject.java
```diff
@@ -1,55 +1,55 @@
...
-  public static int digitalCashFragmentId() {
-    return R.id.fragment_digital_cash_home;
-  }
-
-  @IdRes
-  public static int digitalCashFragmentContainerId() {
-    return R.id.fragment_container_digital_cash;
-  }
-}
+package com.github.dedis.popstellar.ui.pages.digitalcash;
```
pierluca, line 1: how come the whole file changed? Did we switch newlines between LF and CRLF ?

GabrielFleischer (changes requested):
Great work overall ! I found some inconsistencies with what I am doing on my side, we should choose one way of doing and sticking to that.

Also, I have a proposition on the way we set the first fragment.

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/digitalcash/DigitalCashActivity.java
GabrielFleischer: Sadly not.

In java, switch cases must be constants.

But the ids in the R class are not defined as constant at compile time. So we need to use this ugly if/else

A possible solution would be to use an enum to represent the tabs. But I think we will do that in a later phase of the refactor.

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/socialmedia/SocialMediaActivity.java
GabrielFleischer: AS explained in the other comment, yes 😢

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/digitalcash/DigitalCashActivity.java
Warhedgehog: Just getIntentData would be already better and I'm sure there something more elegant. 
Nevertheless I hesitated before changing it and I decided not to because DigitalCash needs a complete "spring cleaning" with the most obvious being recurrent terrible naming (even java naming convention sometimes) and this is not the purpose of the current refactoring. 
So I would do it in a later PR if budget allows. If you disagree I will gladly do it on this one.

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/digitalcash/DigitalCashActivity.java
Warhedgehog: The latter :)

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/digitalcash/DigitalCashActivity.java
Warhedgehog: I would argue the only thing the method does is getting the navbar and adding a listener to it. IMO it is therefore fine to define said listener there directly.
Then again if you have strong feelings about it I have no problem changing it

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/digitalcash/DigitalCashActivity.java
Warhedgehog: Agreed !

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/digitalcash/DigitalCashSendFragment.java
Warhedgehog: As said above, the whole digital cash thing would need a refactor of its own; I could not agree more !

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/socialmedia/SocialMediaActivity.java
Warhedgehog: I'll change it. Out of curiosity and to avoid further "mistake", is it a standard somewhere or a personal preference ?

fe2-android/app/src/test/java/com/github/dedis/popstellar/ui/pages/digitalcash/DigitalCashPageObject.java
Warhedgehog: I have no idea, but I have noticed that some files are LF and some are CRLF and my Android Studio seems to adapt to the existing encoding file by file.
It would maybe be better to set a standard for the whole fe2 project and add the chosen format to the code convention document ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/settings/SettingsActivity.java
```diff
@@ -1,38 +1,77 @@
...
-    if (settingsFragment == null) {
-      settingsFragment = SettingsFragment.newInstance();
-    }
-    ActivityUtils.replaceFragmentInActivity(
-        getSupportFragmentManager(), settingsFragment, R.id.fragment_container_settings);
+  private void applyChanges() {
+    networkManager.connect(entryBoxServerUrl.getText().toString());
+    Intent intent = new Intent(this, HomeActivity.class);
+    Log.d(TAG, "Trying to open home");
+    startActivity(intent);
```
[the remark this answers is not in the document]
Warhedgehog: True I forgot to use the factory methods you added. Thanks for spotting it

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/socialmedia/SocialMediaActivity.java
```diff
@@ -152,46 +99,80 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
-          if (itemId == R.id.social_media_home_menu) {
-            mViewModel.openHome();
-          } else if (itemId == R.id.social_media_search_menu) {
-            mViewModel.openSearch();
-          } else if (itemId == R.id.social_media_following_menu) {
-            mViewModel.openFollowing();
-          } else if (itemId == R.id.social_media_profile_menu) {
-            mViewModel.openProfile();
+          if (itemId != mViewModel.getCurrentSelectedItem().getValue()) {
+            mViewModel.setCurrentSelectedItem(itemId);
```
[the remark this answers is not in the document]
Warhedgehog: The problem is that it would be executed twice when the user select a menu item themself. It would not be problematic most of the time I imagine but it's something we should avoid I think.
I'll add the comment I put in digital cash for the same check

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/socialmedia/SocialMediaHomeFragment.java
```diff
@@ -55,35 +51,11 @@ public void onViewCreated(@NonNull View view, @Nullable Bundle savedInstanceStat
...
-              if (chirpId != null) {
-                deleteChirp(chirpId);
-              }
-            });
   }
 
   private void setupSendButton() {
     mSocialMediaHomeFragBinding.socialMediaSendFragmentButton.setOnClickListener(
-        v -> mSocialMediaViewModel.openSend());
+        v -> mSocialMediaViewModel.openSend(requireActivity().getSupportFragmentManager()));
```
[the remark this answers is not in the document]
Warhedgehog: Definitely better 👍

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/socialmedia/SocialMediaViewModel.java
```diff
@@ -45,14 +45,7 @@ public class SocialMediaViewModel extends AndroidViewModel {
...
    */
-  private final MutableLiveData<SingleEvent<Boolean>> mOpenHomeEvent = new MutableLiveData<>();
-  private final MutableLiveData<SingleEvent<Boolean>> mOpenSendEvent = new MutableLiveData<>();
-  private final MutableLiveData<SingleEvent<Boolean>> mOpenSearchEvent = new MutableLiveData<>();
-  private final MutableLiveData<SingleEvent<Boolean>> mOpenFollowingEvent = new MutableLiveData<>();
-  private final MutableLiveData<SingleEvent<Boolean>> mOpenProfileEvent = new MutableLiveData<>();
-  private final MutableLiveData<SingleEvent<Boolean>> mSendNewChirpEvent = new MutableLiveData<>();
-  private final MutableLiveData<SingleEvent<MessageID>> mDeleteChirpEvent = new MutableLiveData<>();
-
+  private final MutableLiveData<Integer> mCurrentSelectedItem = new MutableLiveData<>(0);
```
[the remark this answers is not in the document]
Warhedgehog: True I'll change it for both social media and digital cash

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/digitalcash/DigitalCashActivity.java
Warhedgehog: Okay it triggered me as well so I changed just this one :)
Nevertheless the comment above subsist about digital cash in general

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/socialmedia/SocialMediaActivity.java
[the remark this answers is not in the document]
GabrielFleischer: Oh ok, tanks for the clarification

GabrielFleischer (approved):
LGTM

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/socialmedia/SocialMediaActivity.java
pierluca: It depends on the development environment.

In the automotive industry, for instance, there are some standards that require your functions to have just one exit point (i.e. one `return`), to ensure the flow is easy to follow and without surprises. The static analysis tools in those environments will look at the complexity of the "flow" as a quality criteria.

However, if you think of defensive programming, you'd want to validate all input and abort/exit as fast as possible when something is "off", like here a null value.
There's also another reason I tend to favor (if applicable) this style: it avoids crazy amounts of indentation.
Imagine you have the following (pseudo-)code
```
if x != null {
  if x.IsValid {
    // actual code in this function
  }
  else { 
    log.Error("X is invalid")
  }
}
```

the following is going to be easier to understand / maintain and, more importantly, *write correctly*
```
if x == null {
  return
}

if ! x.IsValid {
  log.Error("X is invalid")
  return
}

// the actual code in this function
```

this becomes especially true when you have more than just 1-2 checks.

fe2-android/app/src/test/java/com/github/dedis/popstellar/ui/pages/digitalcash/DigitalCashPageObject.java
pierluca: Agreed. My recommendation would be to use LF everywhere, and just ask Windows users to configure the relevant Git configurations (core.autocrlf , etc.) 
When there's no other PR open, I'd suggest you or @GabrielFleischer do a full `dos2unix` on the project codebase 😅

pierluca (approved):

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/socialmedia/SocialMediaActivity.java
Warhedgehog: Thanks for the explanation 👍

PR #1184, opened 2022-07-28
Add node version 18.x to CI

Xelowak:
Add node 18.x to CI (newer version) instead of version 14.x which caused issues.


pierluca:
@Xelowak have you tried re-running this after merging the changes in master? It might work now 😄

Tyratox:
The CI now runs with node v18 🥳

PR #1185, opened 2022-07-31
Improve Tab system of the Home Activity

GabrielFleischer:
Improve the tab system so that it is more robust, and it is easier to manually switch tabs.

The main idea is that the change of fragments are done from the navigation bar callback and to change tabs manually, we change the value of a live data that impacts the navigation bar.

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeActivity.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeActivity.java
@@ -145,7 +145,13 @@ public class HomeActivity extends AppCompatActivity {
 
   public void setupNavigationBar() {
     mViewModel.getCurrentTab().observe(this, tab -> navbar.setSelectedItemId(tab.getMenuId()));
-    navbar.setOnItemSelectedListener(item -> openTab(HomeTab.findByMenu(item.getItemId())));
+    navbar.setOnItemSelectedListener(
+        item -> {
+          HomeTab tab = HomeTab.findByMenu(item.getItemId());
+          boolean selected = openTab(tab);
+          if (selected) mViewModel.setCurrentTab(tab);
+          return selected;
+        });
     // Set an empty reselect listener to disable the onSelectListener when pressing multiple times
     navbar.setOnItemReselectedListener(item -> {});
   }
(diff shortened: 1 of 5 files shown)


fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeActivity.java
```diff
@@ -149,97 +144,99 @@ public static HomeViewModel obtainViewModel(FragmentActivity activity) {
...
+      showWalletWarning();
+      return false;
     }
-  }
 
-  public void openConnect() {
-    if (checkSelfPermission(Manifest.permission.CAMERA) == PackageManager.PERMISSION_GRANTED) {
+    // Check for the permission, if it is not granted, ask for it
+    if (checkSelfPermission(Manifest.permission.CAMERA) != PackageManager.PERMISSION_GRANTED)
+      requestCameraPermission();
```
pierluca, line 196: I'm not sure if this is something that needs to be edited in the `.editorconfig`, but please always have curly brackets {} for if/else clauses. It prevents subtle mistakes.

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeActivity.java
```diff
@@ -149,97 +144,99 @@ public static HomeViewModel obtainViewModel(FragmentActivity activity) {
...
+    return true;
   }
 
-  public void requestCameraPermission() {
+  private void requestCameraPermission() {
     // Setup result listener to open the connect tab once the permission is granted
     getSupportFragmentManager()
         .setFragmentResultListener(
-            CameraPermissionFragment.REQUEST_KEY, this, (k, b) -> openConnect());
+            CameraPermissionFragment.REQUEST_KEY, this, (k, b) -> openConnectTab());
```
pierluca, line 207: what happens if the permission is not granted?

pierluca (changes requested):
Looks good, barring a minor style change

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeActivity.java
GabrielFleischer: Probably yes, I'll look into that in another PR.

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeActivity.java
GabrielFleischer: The `openConnectTab()` function is called, which re-ask for permission if needed.

Ideally, we would just not ask for it again and use the (already in place) button in the scanning fragment.

pierluca (approved):

PR #1187, opened 2022-08-01
Home UI Cleanup

GabrielFleischer:
The PR aims to clean up a lot of different things that were added in the home UI over the years.

 - Rename different names such that they are more meaningful and have a reasonable size
 - Reorganize function in order to make more sense (to me, there are no convention regarding that, so it is mostly personal 😅)
 - Remove unnecessary complexities in the view model :
   - Remove the use of the Lao name livedata as we can just pass the value held in the text edit itself when needed
 - Create dedicated livedata for the social media availability such that its computation is done in the view model, not in the UI
 - Remove various outdated files and function calls


fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeViewModel.java
```diff
@@ -45,11 +45,11 @@ public class HomeViewModel extends AndroidViewModel implements QRCodeScanningVie
   private static final ScanningAction scanningAction = ScanningAction.ADD_LAO_PARTICIPANT;
 
   /** LiveData objects that represent the state in a fragment */
-  private final MutableLiveData<Boolean> mIsWalletSetUp = new MutableLiveData<>(false);
+  private final MutableLiveData<Boolean> isWalletSetup = new MutableLiveData<>(false);
 
-  private final MutableLiveData<String> mLaoName = new MutableLiveData<>();
   private final MutableLiveData<HomeTab> currentTab = new MutableLiveData<>(HomeTab.HOME);
-  private final LiveData<List<Lao>> mLAOs;
+  private final LiveData<List<Lao>> Laos;
```
Warhedgehog, line 51:
```suggestion
  private final LiveData<List<Lao>> laos;
```

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeViewModel.java
```diff
@@ -75,9 +75,10 @@ public HomeViewModel(
     this.wallet = wallet;
     this.networkManager = networkManager;
 
-    mLAOs =
+    Laos =
```
Warhedgehog, line 78:
```suggestion
    laos =
```

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeViewModel.java
```diff
@@ -75,9 +75,10 @@ public HomeViewModel(
...
         LiveDataReactiveStreams.fromPublisher(
             laoRepository.getAllLaos().toFlowable(BackpressureStrategy.BUFFER));
+    isSocialMediaEnabled = Transformations.map(Laos, laos -> laos != null && !laos.isEmpty());
```
Warhedgehog, line 81:
```suggestion
    isSocialMediaEnabled = Transformations.map(laos, laoSet -> laoSet != null && !laoSet.isEmpty());
```

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeViewModel.java
```diff
@@ -174,27 +173,27 @@ public LiveData<HomeTab> getCurrentTab() {
   }
 
   public LiveData<List<Lao>> getLAOs() {
-    return mLAOs;
+    return Laos;
```
Warhedgehog, line 176:
```suggestion
    return laos;
```

Warhedgehog (changes requested):
Great work ! Couple little things :)

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/LaunchFragment.java
```diff
@@ -30,34 +30,27 @@ public View onCreateView(
...
-    mLaunchFragBinding.setLifecycleOwner(getActivity());
-
-    return mLaunchFragBinding.getRoot();
-  }
-
-  @Override
-  public void onViewCreated(@NonNull View view, @Nullable Bundle savedInstanceState) {
-    super.onViewCreated(view, savedInstanceState);
+    viewModel = HomeActivity.obtainViewModel(requireActivity());
+    binding.setLifecycleOwner(getActivity());
```
pierluca, line 36: I'm being an Android-newbie here:
my understanding is that this stuff normally happens implicitly.
Why do we need to manually inflate and set the activity, in this context ?
Is it because it's the entry point ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/wallet/ContentWalletFragment.java
```diff
@@ -51,38 +51,37 @@ public View onCreateView(
...
-                (dialog, which) -> {
-                  mHomeViewModel.logoutWallet();
-                  WalletFragment.openWallet(
-                      getParentFragmentManager(), mHomeViewModel.isWalletSetUp());
-                });
-            builder.setNegativeButton(R.string.cancel, (dialog, which) -> dialog.dismiss());
-            logoutAlert = builder.create();
-            logoutAlert.show();
-          });
+      setupWalletContent();
```
pierluca, line 58: Just from a naming perspective, it seems weird to have the logic "if it's setup, set it up!" 
Can we distinguish between "setup" and "setup" ? 😅

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/wallet/ContentWalletFragment.java
```diff
@@ -51,38 +51,37 @@ public View onCreateView(
...
+    return mWalletContentBinding.getRoot();
+  }
+
+  private void setupWalletContent() {
+    mWalletContentBinding.logoutButton.setVisibility(View.VISIBLE);
+    mWalletContentBinding.logoutButton.setOnClickListener(
+        clicked -> {
+          if (logoutAlert != null && logoutAlert.isShowing()) {
+            logoutAlert.dismiss();
+          }
```
pierluca, line 70: why dismiss it if it's already shown, and then re-create it ? 
Why not just `return` ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeActivity.java
```diff
@@ -229,8 +205,7 @@ private void showWalletWarning() {
   }
 
   private void openSocialMediaTab() {
-    List<Lao> laos = mViewModel.getLAOs().getValue();
-    if (laos == null || laos.isEmpty()) {
+    if (!Boolean.TRUE.equals(viewModel.isSocialMediaEnabled().getValue())) {
```
pierluca, line 208: why do we work with Boolean objects instead of just boolean values ? Can't we avoid it ? 😅

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeActivity.java
```diff
@@ -184,19 +155,24 @@ private boolean openTab(HomeTab tab) {
...
+  }
+
   private boolean openConnectTab() {
     // Check if the wallet is setup
-    if (!mViewModel.isWalletSetUp()) {
+    if (!viewModel.isWalletSetUp()) {
       showWalletWarning();
       return false;
     }
+// Check for the permission, if it is not granted, ask for it
```
pierluca, line 168: there seems to be missing whitespace before this comment

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeActivity.java
```diff
@@ -66,61 +58,58 @@ protected void onCreate(Bundle savedInstanceState) {
...
-            laos -> {
-              if (laos.isEmpty()) {
-                socialMediaItem.setIcon(R.drawable.ic_common_social_media_transparent_foreground);
-              } else {
-                socialMediaItem.setIcon(R.drawable.ic_common_social_media_opaque_foreground);
-              }
+            enabled -> {
+              // We set the button icon depending on the livedata value
+              socialMediaItem.setIcon(
+                  Boolean.TRUE.equals(enabled)
```
pierluca, line 107: same here and there, can't we use the boolean value ? 😢

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeActivity.java
GabrielFleischer: Because it is the generic of the livedata.
And for some reason, in java, a generic must be a class. Not a primitive.

That is why we cannot use boolean. And it comes with the cost that we have not warranty concerning the nullability of a Boolean.

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeActivity.java
GabrielFleischer: sadly, no 😞

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/LaunchFragment.java
GabrielFleischer: I don't think there is a way to do the inflation implicitly 😅 

What we do here, is the use of [View Binding](https://developer.android.com/topic/libraries/view-binding)
It is a nice feature that generate code based on the activity xml so that we can get rid of the `findViewById()` 

As for the `setLifecycleOwner` call, this might not be needed. I'm looking into that.

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/wallet/ContentWalletFragment.java
GabrielFleischer: I... don't know

Let's just return there

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/LaunchFragment.java
GabrielFleischer: I don't understand the use cases I found. I think we can safely remove it.

Warhedgehog (approved):
LGTM

PR #1189, opened 2022-08-04
LaoRepository refactoring

Warhedgehog:
This PR is the first step of the refactoring of the `LAORepository`. In this step we introduce the `LaoView` which is what is retrievable from the repository. It doesn't allow change of the wrapped Lao but offers getters for the handlers to have access to needed state. To change the state of the Lao, one have to get the wrapped Lao from the `LaoView` which provides a deep copy. This is to move closer to a functional way of programming to ease the reasoning about how lao state is changed.

There is one exception for consensus attributes which are not deep copied because the whole implementation relies on changing things by reference and refactoring it all with the assurance to not add bugs would take more time and would therefore be done in a future PR

All code smells were considered and should be ignored (at least for now) in my opinion.


fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/remote/LAONetworkManager.java
```diff
@@ -177,6 +178,8 @@ private void handleMessages(List<MessageGeneral> messages, Channel channel) {
         messageHandler.handleMessage(repository, this, channel, msg);
       } catch (DataHandlingException e) {
         Log.e(TAG, "Error while handling received catchup message", e);
+      } catch (UnknownLaoException e) {
+        Log.e(TAG, "Error while handling received message", e);
```
pierluca, line 182:
```suggestion
        Log.e(TAG, "Error while handling received catchup message", e);
```
?

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/security/PublicKey.java
```diff
@@ -26,6 +26,11 @@ public PublicKey(String data) {
     verifier = new Ed25519Verify(this.data);
   }
 
+  public PublicKey(PublicKey organizer) {
+    super(organizer.data); // Deep copy of byte array is done in parent constructor
+    verifier = new Ed25519Verify(organizer.data);
```
pierluca, line 31:
```suggestion
    verifier = new Ed25519Verify(this.data);
```
should we use this, then, if the deep-copy happened in the parent constructor ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/LAORepository.java
```diff
@@ -116,4 +117,29 @@ public void updateNodes(Channel channel) {
...
+      throw new IllegalArgumentException();
+    }
+
+    if (laoById.containsKey(lao.getId())) {
+      // If the lao already exists, we can push the next update
+      laoById.get(lao.getId()).publish(lao);
+    } else {
+      // Otherwise, create the state
+      laoById.put(lao.getId(), new LAOState(lao));
+    }
```
pierluca, line 143: is this call supposed to be thread-safe ? If so, this `laoById` access wouldn't work without some form of locking.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/view/LaoView.java
```diff
@@ -0,0 +1,93 @@
...
+    Optional<RollCall> optional = lao.getRollCall(id);
+    return optional.map(RollCall::new); // If empty returns empty optional, if not
+    // returns optional with copy of retrieved RollCall
+  }
+
+  public Optional<Chirp> getChirp(MessageID messageID) {
+    Optional<Chirp> optional = lao.getChirp(messageID);
+    return optional.map(Chirp::new); // If empty returns empty optional, if not
+    // returns optional with copy of retrieved Chirp
+  }
```
pierluca, line 76: is the idea that later on, `getChirp` and other feature-specific functions will be somewhere other than the `LaoView` ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/view/LaoView.java
```diff
@@ -0,0 +1,93 @@
...
+
+  public PublicKey getOrganizer() {
+    return new PublicKey(lao.getOrganizer());
+  }
+
+  public Optional<ElectInstance> getElectInstance(MessageID messageId) {
+    // TODO uncomment that when consensus does not rely on call by reference
+    //    Optional<ElectInstance> optional = lao.getElectInstance(messageId);
+    //    return optional.map(ElectInstance::new); // If empty returns empty optional, if not
+    // returns optional with copy of retrieved ElectInstance
```
pierluca, line 90: If we absolutely need to keep comment-out code, okay, but do make sure we don't break the consensus 😅

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Lao.java
```diff
@@ -79,6 +79,83 @@ public Lao(String name, PublicKey organizer, long creation) {
...
+                            .collect(Collectors.toList())));
+    this.transactionByUser =
+        lao.transactionByUser.entrySet().stream()
+            .collect(
+                Collectors.toMap(
+                    entry -> new PublicKey(entry.getKey()),
+                    entry ->
+                        entry.getValue().stream()
+                            .map(TransactionObject::new)
+                            .collect(Collectors.toList())));
```
pierluca, line 156: this code and similar ones elsewhere are kind of scary / ugly. I believe I understand the general idea, but is there no cleaner / more readable / more succinct way of doing this ? 🤔

pierluca (commented):
Partial review, I've only reviewed 64% of the files and not the biggest ones.
Just a first salvo of remarks for now 😄

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/view/LaoView.java
Warhedgehog: We envision down the road to remove LaoView as it would not be needed anymore. We would like to end up with a functional handling of the objects. Furthermore we would like to break down the Lao object, it's too big and does too many things.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/view/LaoView.java
Warhedgehog: The new code is commented because of this :) We need to talk consensus at some point though 😰

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Lao.java
Warhedgehog: Unfortunately there is no deep copier of `Map` in java (from my search on S.O). 
I thought about making a custom function for that but the problem is that the `Object` class does not have a copy constructor in the interface. So I imagined adding an abstract type that does but it seemed a bit overkill for a few copy constructors.
Furthermore, we plan on breaking down the Lao object later so it should be less appalling 😄 

Finally the stream syntax might look a bit scary / ugly but the alternative is extremly verbose so I settled on that, Alternative without stream for line 148 to 156 looks like this just in case

````
```
    HashMap<PublicKey, List<TransactionObject>> copy = new HashMap<>();
    for (Map.Entry<PublicKey, List<TransactionObject>> entry : lao.transactionByUser.entrySet()){
      List<TransactionObject> listCopy = new ArrayList<>();
      for (TransactionObject transactionObject : entry.getValue()){
        listCopy.add(new TransactionObject(transactionObject));
      }
      copy.put(new PublicKey(entry.getKey()), listCopy);
    }
    this.transactionByUser = copy;
```
````
It's the same amount of lines but it is much more packed.

Comment (Warhedgehog) [reply fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Lao.java:156]: This was how I saw things when implementing it, if someone has a better idea I'll gladly take it 😄

File: fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/data/ChirpHandler.java [hunk 7]
```diff
@@ -28,14 +29,19 @@ private ChirpHandler() {
...
     PublicKey senderPk = context.getSenderPk();
 
-    Lao lao = laoRepository.getLaoByChannel(channel);
-
+    Log.d(TAG, "handleChirpAdd: " + channel + " id " + addChirp.getParentId());
+    Optional<LaoView> laoViewOptional = laoRepository.getLaoViewByChannel(channel);
+    if (!laoViewOptional.isPresent()) {
+      throw new UnknownLaoException(channel.extractLaoId());
+    }
+    LaoView laoView = laoViewOptional.get();
```
GabrielFleischer, line 44: Would it make sense to create a function inside the repository that throws an exception when no Lao are found ?

It seems like a usual usage of the repository.
Btw, you can write it like this :
```suggestion
    LaoView laoView = laoViewOptional.orElseThrow(() -> new UnknownLaoException(channel.extractLaoId()));
```

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/view/LaoView.java
```diff
@@ -0,0 +1,93 @@
...
+   * @param lao the Lao to be wrapped
+   */
+  public LaoView(Lao lao) {
+    if (lao == null) {
+      throw new IllegalArgumentException();
+    }
+    this.lao = new Lao(lao);
+  }
+
+  public Lao getLao() {
```
GabrielFleischer, line 30: To me, this function would be more meaningful if it was named `createCopy()`

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/data/LaoHandler.java
```diff
@@ -31,22 +33,22 @@ private LaoHandler() {
   @SuppressLint("CheckResult") // for now concerns Consensus which is not a priority this semester
   public static void handleCreateLao(HandlerContext context, CreateLao createLao) {
     LAORepository laoRepository = context.getLaoRepository();
-    Channel channel = context.getChannel();
+    Channel channel = new Channel(context.getChannel());
```
GabrielFleischer, line 36: Do we really need to copy the channel ?

I think it is immutable

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/data/LaoHandler.java
```diff
@@ -143,31 +156,38 @@ public static void handleStateLao(HandlerContext context, StateLao stateLao)
...
     lao.setModificationId(stateLao.getModificationId());
 
     PublicKey publicKey = context.getKeyManager().getMainPublicKey();
-    if (lao.getOrganizer().equals(publicKey) || lao.getWitnesses().contains(publicKey)) {
-      context.getMessageSender().subscribe(lao.getChannel().subChannel("consensus")).subscribe();
+    if (laoView.isOrganizer(publicKey) || laoView.isWitness(publicKey)) {
+      context
+          .getMessageSender()
+          .subscribe(laoView.getChannel().subChannel("consensus"))
+          .subscribe();
```
GabrielFleischer, line 172: If the subscribtion fails, the app will crash

GabrielFleischer (changes requested):
Great work !

I proposed some slight adjustments that you might want to take a look at.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/view/LaoView.java
pierluca: makes sense !

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/view/LaoView.java
pierluca: yup. let me know when we need to have that conversation 😁

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Lao.java
pierluca: Okay, I understand better.
I understand Lao being too big, and this makes it pretty obvious.
But if I recall correctly, the election object is almost as ugly.

I'd also support a `Copiable` interface, as it seems a bit weird for some classes constructor to have to deal with how different objects need to be copied.

pierluca: Okay, I understand better.
I understand Lao being too big, and this makes it pretty obvious.
But if I recall correctly, the election object is almost as ugly.

I'd also support a `Copiable` interface, as it seems a bit weird for some classes constructor to have to deal with how different objects need to be copied.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/DataRegistry.java
```diff
@@ -45,7 +44,7 @@ public Optional<Class<? extends Data>> getType(Objects obj, Action action) {
    * @throws DataHandlingException if an error occurs or if there was no handler
    */
   public void handle(HandlerContext context, Data data, Objects obj, Action action)
-      throws DataHandlingException {
+      throws DataHandlingException, UnknownLaoException {
```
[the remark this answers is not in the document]
pierluca: I quite like specific exceptions, to be honest 😅

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/consensus/ConsensusElect.java
```diff
@@ -23,6 +25,13 @@ public ConsensusElect(long creation, String objId, String type, String property,
     this.value = value;
   }
 
+  public ConsensusElect(ConsensusElect consensusElect) {
+    this.instanceId = consensusElect.instanceId;
+    this.key = new ConsensusKey(consensusElect.key);
+    this.creation = consensusElect.creation;
+    this.value = consensusElect.value; // This may break the deep copy principle
```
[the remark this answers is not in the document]
pierluca: when you add "at the moment",  I'd strongly suggest adding an author / date to the comment, as well as whatever context is needed for someone else to reevaluate this statement later.

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/data/LaoHandler.java
pierluca: unhelpful comment: that doesn't sound good 😅

pierluca (commented):
Finished the 2nd part of the review !

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/ElectInstance.java
```diff
@@ -107,6 +124,29 @@ public String toString() {
...
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    ElectInstance that = (ElectInstance) o;
+    return messageId.equals(that.messageId)
+        && channel.equals(that.channel)
+        && proposer.equals(that.proposer)
+        && elect.equals(that.elect)
+        && nodes.equals(that.nodes)
+        && acceptorToMessageId.equals(that.acceptorToMessageId)
```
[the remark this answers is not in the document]
GabrielFleischer: They use the equals of the objects inside the set/map.

So in our case, as we implemented (I think) all the equals, it will do a deep comparison.

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/data/ChirpHandler.java
```diff
@@ -54,25 +62,32 @@ public static void handleChirpAdd(HandlerContext context, AddChirp addChirp) {
...
+    Optional<Chirp> chirpOptional = laoView.getChirp(deleteChirp.getChirpId());
 
     if (!chirpOptional.isPresent()) {
       throw new InvalidMessageIdException(deleteChirp, deleteChirp.getChirpId());
     }
-    chirp = chirpOptional.get();
+    Chirp chirp = chirpOptional.get();
```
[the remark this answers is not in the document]
GabrielFleischer: Currently, any exception puts the message back into the processing queue with a 5 seconds delay.

Ideally we would want to filter that behavior based on the exception.

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/data/LaoHandler.java
jbsv: Can we catch that and log it ?

jbsv (approved):

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Election.java
```diff
@@ -51,6 +53,54 @@ public Election(String laoId, long creation, String name, ElectionVersion electi
...
+    this.end = election.end;
+    this.electionKey = election.electionKey;
+    this.electionQuestions =
+        election.electionQuestions.stream().map(ElectionQuestion::new).collect(Collectors.toList());
+    this.electionVersion = election.electionVersion;
+    this.openVoteByPublicKey =
+        election.openVoteByPublicKey.entrySet().stream()
+            .collect(
+                Collectors.toMap(
+                    entry -> new PublicKey(entry.getKey()),
```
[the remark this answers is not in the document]
Warhedgehog: True I hadn't taken taken that into consideration. Thinking about it now is the problem that someone might change the immutability of the class later, not knowing that it breaks this.

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/data/LaoHandler.java
Warhedgehog: Yep, good catch !

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/data/ChirpHandler.java
```diff
@@ -54,25 +62,32 @@ public static void handleChirpAdd(HandlerContext context, AddChirp addChirp) {
...
     if (chirp.getIsDeleted()) {
       Log.d(TAG, "The chirp is already deleted");
     } else {
       chirp.setIsDeleted(true);
       chirp.setText("");
     }
+
+    Lao lao = laoView.getLao();
+    lao.updateAllChirps(chirp.getId(), chirp);
+    laoRepository.updateLao(lao);
```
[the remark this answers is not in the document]
Warhedgehog: It is a naming problem. I think the person meant update the list of all chirps and shortened it. The result is bad unfortunately

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/data/ChirpHandler.java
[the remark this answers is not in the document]
Warhedgehog: True but I think that it should be done in an upcoming PR. We could also differentiate between temporary and permanent error from the server (with the error code)

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/ElectInstance.java
```diff
@@ -41,6 +42,21 @@ public ElectInstance(
     this.state = State.STARTING;
   }
 
+  public ElectInstance(ElectInstance electInstance) {
+    this.messageId = new MessageID(electInstance.messageId);
+    this.channel = new Channel(electInstance.channel);
+    this.proposer = new PublicKey(electInstance.proposer);
```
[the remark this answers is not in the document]
Warhedgehog: As long as they stay immutable it is indeed useless

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/election/ElectionEncryptedVote.java
```diff
@@ -24,8 +24,8 @@ public class ElectionEncryptedVote {
    * @param questionId id of the question
    * @param encryptedVote encrypted uniaue indice of the chosen vote
    * @param writeInEnabled indicates if write in is enabled
-   * @param encryptedWriteIn
-   * @param electionId
+   * @param encryptedWriteIn whether the election allows write-in
```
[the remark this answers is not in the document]
Warhedgehog: It would not be that big of change tough it would be a couple lines across multiple files (those involved with UI too). Therefore since it is also not the purpose of this PR, we could do a small separate PR with that.

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/data/ChirpHandler.java
Warhedgehog: For now it is the only behaviour we have when getting an LAO from the repo so it seems like a reasonable suggestion to me.

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/data/ChirpHandler.java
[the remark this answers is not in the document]
pierluca: I guess the current behaviour is "good enough" for now, as long as it's documented somewhere :-)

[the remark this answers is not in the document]
pierluca: I guess the current behaviour is "good enough" for now, as long as it's documented somewhere :-)

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/view/LaoView.java
```diff
@@ -0,0 +1,91 @@
...
+   */
+  public LaoView(Lao lao) {
+    if (lao == null) {
+      throw new IllegalArgumentException();
+    }
+    this.lao = new Lao(lao);
+  }
+
+  public Lao createLaoCopy() {
+    return new Lao(lao);
```
GabrielFleischer, line 29: If we already create a copy at construction, is it really necessary to create another one here ?

(I would suggest creating the copy only in the `createLaoCopy()` function, not at creation)

GabrielFleischer (approved):
I just have a slight proposition, otherwise it looks good to me !

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/view/LaoView.java
Warhedgehog: I think it should stay that way so that later change to the instance of Lao passed to the constructor does not affect the attribute of LaoView.
Same with `createLaoCopy`

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/data/LaoHandler.java
```diff
@@ -125,6 +138,7 @@ public static void handleUpdateLao(HandlerContext context, UpdateLao updateLao)
    * @param context the HandlerContext of the message
    * @param stateLao the message that was received
    */
+  @SuppressLint("CheckResult")
```
pierluca, line 124: why do we need this?

pierluca (commented):
Looks pretty good to me overalls, though I have a tiny question. 
Also, check out Sonar : it's pointing out a few good suggestions / issues.

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/data/LaoHandler.java
Warhedgehog: It's because we do not use the result of the subscribe (the javaRx one) which returns a disposable. For lifecycle aware parts of the codebase, those disposables are stored and then disposed of at the end of the lifecycle. 
In all our handlers those disposables are ignored. The suppressed lint is to make the IDE less aggressive about it.

I think we should provide the disposable container of the `LaoNetworkManager` to the handler so that they can add their disposables to it. What do you think @GabrielFleischer ?

I think it should be done in a separate PR. I can do it fairly quickly :)

Warhedgehog:
@pierluca  I've fixed the bug found by SC. The code smells are because of TODOs and FIXMEs mostly

PR #1191, opened 2022-08-11
Add the Copyable interface and use it to simplify copies

GabrielFleischer:


fe2-android/app/src/main/java/com/github/dedis/popstellar/model/Copyable.java
```diff
@@ -0,0 +1,47 @@
...
+   * Create a copy of a map containing lists
+   *
+   * @param source map
+   * @param <E> type of key
+   * @param <T> type of value inside the list
+   * @return the copy
+   */
+  static <E, T> Map<E, List<T>> copyMapOfList(Map<E, List<T>> source) {
+    return source.entrySet().stream()
+        .collect(Collectors.toMap(Map.Entry::getKey, entry -> new ArrayList<>(entry.getValue())));
```
pierluca, line 45: don't we need `T` to be `T extends Copyable<T>` here? Don't we need to deep-copy the list contents ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/ElectInstance.java
```diff
@@ -43,17 +43,12 @@ public ElectInstance(
...
-        electInstance.acceptorToMessageId.entrySet().stream()
-            .collect(
-                Collectors.toMap(
-                    entry -> new PublicKey(entry.getKey()),
-                    entry -> new MessageID(entry.getValue())));
+    this.messageId = electInstance.messageId;
+    this.channel = electInstance.channel;
+    this.proposer = electInstance.proposer;
+    this.elect = electInstance.elect;
+    this.nodes = electInstance.nodes;
```
pierluca, line 50: don't we need to copy the `ConsensusElect` ? 
and if not, can we annotate it so that future changes would have to explicitly break our assumptions ?

pierluca (commented):
I haven't checked everything (lack of time), but it seems a lot cleaner.
If possible, do make sure that we annotate the relevant data classes so that they don't suddenly become modifiable and break the assumptions everywhere in the code base. (e.g., PublicKey getting a conveniently named `setKey` function 😅). If annotation is impossible, big fat comments would be helpful 😁

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/Copyable.java
Warhedgehog: That's a good idea especially for later update by students unfamiliar with the issue

Warhedgehog (approved):
LGTM. Great work ! It is indeed much cleaner so

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/ElectInstance.java
GabrielFleischer: ConsensusElect is part of the network data classes.

Our current assumption (that I documented a second ago 😄) is that classes representing messages are immutable.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/Copyable.java
GabrielFleischer: Actually, I thought about it and I think creating a dedicated type is a bit overkill for the future of the app.

Our goal is to move to a mostly functional app state, and most of the copy will be done on immutable objects.

PR #1195, opened 2022-08-16
Re-add the add witness feature

GabrielFleischer:
During an earlier pull request, this feature was removed.

The PR therefore re-adds it in a ... clunky way.
With the current scanning system, I cannot write it in a way that I find clean enough.

But changing that would be a massive change to the codebase.


pierluca (approved):
It's ugly, but as discussed, we should be able to refactor this later along with the QR scanner code.

jbsv (approved):
LGTM

Warhedgehog (approved):
LGTM

PR #1197, opened 2022-08-19
Add disposables in handlers to message sender container

Warhedgehog:
This adds the disposables of the handlers to the disposables container of the `LaoNetworkManager`. Previously those were unused which can lead to some unpleasantness (an example [here](https://stackoverflow.com/a/49523600/18580049)).

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/remote/MessageSender.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/remote/MessageSender.java
@@ -67,4 +67,11 @@ public interface MessageSender extends Disposable {
    * @return an Observable of WebSocket events of the underlying connection
    */
   Observable<WebSocket.Event> getConnectEvents();
+
+  /**
+   * Adds the disposable to the container of disposables of the message sender
+   *
+   * @param disposable the disposable to add
+   */
+  void addToDisposableContainer(Disposable disposable);
 }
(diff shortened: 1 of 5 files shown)


fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/remote/MessageSender.java
```diff
@@ -67,4 +67,11 @@ public interface MessageSender extends Disposable {
    * @return an Observable of WebSocket events of the underlying connection
    */
   Observable<WebSocket.Event> getConnectEvents();
+
+  /**
+   * Adds the disposable to the container of disposables of the message sender
+   *
+   * @param disposable the disposable to add
+   */
+  void addToDisposableContainer(Disposable disposable);
```
pierluca, line 76: This seems odd as part of the API of `MessageSender`. It almost feels like it's in the wrong place.
If you look at the other functions, they sort of match the `MessageSender` abstraction. This doesn't.

Would there be a more appropriate place to put it ? 
The method also seems to be doing very little, maybe a bit of duplication would be cleaner than breaking the abstraction? 🤔

GabrielFleischer: I see the problem this is trying to fix and I don't think this is the way to do it.

We should not use the message sender's disposables for the handlers. I think the handlers should have their owns, but own exactly is another question.

Maybe we can provide the Application lifecycle to the handlers and observe it to know when the application goes to the pause/destroy state.

Warhedgehog:
Because of bad design and a codebase that has much evolved since then, this is given up

PR #1204, opened 2022-09-07
Transition to quasi immutability for Lao objects

Warhedgehog:
This PR moves the handling of Laos close to immutability. The reasoning is that only incoming messages from the backend may change the state of an LAO. Therefore the only part of the codebase where the lao object is directly used and modified is in the handlers. This PR completes the transtion for the rest of the codebase to use an immutable view of an Lao rather that the direct object


fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/event/election/fragments/CastVoteFragment.java
```diff
@@ -108,9 +108,9 @@ public View onCreateView(
     // setUp the cast Vote button
     voteButton = mCastVoteFragBinding.castVoteButton;
 
-    // Getting lao
-    Lao lao = mLaoDetailViewModel.getCurrentLao().getValue();
-    if (lao == null) {
+    // Getting laoView
+    LaoView laoView = mLaoDetailViewModel.getCurrentLao().getValue();
```
GabrielFleischer, line 112: Could we use the `getCurrentLaoValue()` function here ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/event/election/fragments/ElectionFragment.java
```diff
@@ -149,14 +150,27 @@ public View onCreateView(
                   "User should not be able to use the action button in this state :" + state);
           }
         });
-
-    election.getState().observe(getViewLifecycleOwner(), eventState -> setupElectionContent());
+    laoDetailViewModel
+        .getCurrentLao()
+        .observe(
+            getActivity(),
```
GabrielFleischer, line 156: I think it is better to use `getViewLifecycleOwner()`. It is non-null and more suited to what we want.

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/event/election/fragments/ElectionFragment.java
```diff
@@ -149,14 +150,27 @@ public View onCreateView(
...
+            });
 
     return view;
   }
 
   private void setupElectionContent() {
-    EventState electionState = election.getState().getValue();
+    if (getContext() == null) {
+      return;
+    }
```
GabrielFleischer, line 168: I am not sure I understand why this is needed

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/LaoDetailViewModel.java
```diff
@@ -536,17 +555,21 @@ public LiveData<List<RollCall>> getLaoAttendedRollCalls() {
...
+
+  public LaoView getCurrentLaoValue() throws UnknownLaoException {
+    return laoRepository.getLaoView(laoId);
   }
 
   @VisibleForTesting
-  public void setCurrentLao(Lao lao) {
-    mCurrentLao.setValue(lao);
+  public void setCurrentLao(LaoView laoView) {
+    mCurrentLao.postValue(laoView);
```
GabrielFleischer, line 572: Shouldn't we also set the `laoId` value here ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/socialmedia/SocialMediaViewModel.java
```diff
@@ -226,13 +239,7 @@ public LaoView getLaoView(String laoId) throws UnknownLaoException {
...
-  @Nullable
-  public Lao getCurrentLao() throws UnknownLaoException {
-    return getLao(getLaoId().getValue());
-  }
-
-  @Nullable
-  private Lao getLao(String laoId) throws UnknownLaoException {
-    return laoRepository.getLaoView(laoId).createLaoCopy();
+  public LaoView getCurrentLaoView() throws UnknownLaoException {
+    return getLaoView(getLaoId().getValue());
```
GabrielFleischer, line 243: I think it would be good to do that same thing as in LaoDetailViewModel: Store the laoId and use it to retrieve the LaoView.

What do you think ?

GabrielFleischer (approved):
Great work ! 👍 

I raised some small thingy here and there, nothing too important.

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/LaoDetailViewModel.java
```diff
@@ -203,16 +207,18 @@ public Completable openElection(Election e) {
 
   public Completable endElection(Election election) {
     Log.d(TAG, "ending election with name : " + election.getName());
-    Lao lao = getCurrentLaoValue();
-    if (lao == null) {
+    LaoView laoView;
+    try {
+      laoView = getCurrentLaoValue();
```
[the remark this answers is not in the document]
Warhedgehog: Very true. I hadn't realised this was now very counterintuitive

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/event/election/fragments/CastVoteFragment.java
Warhedgehog: Indeed, my mistake

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/event/election/fragments/ElectionFragment.java
Warhedgehog: Because for now this will get called everytime the LAO is updated which may occur when the fragment is not displayed and the context is null. 
I had no better idea how to handle this before the Lao object is split

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/socialmedia/SocialMediaViewModel.java
Warhedgehog: True. The mutable liveData is useless for how it is used

GabrielFleischer (dismissed):
LGTM

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/event/election/fragments/ElectionFragment.java
jbsv: Please add a comment so we still understand the reasoning in many months...

jbsv (approved):

PR #1208, opened 2022-09-14
Fix non updating rc state displayed

Warhedgehog:
This fixes a bug where attendees would not see their roll call state displayed on screen updated


GabrielFleischer (approved):
LGTM

PR #1210, opened 2022-09-14
Migrate handlers to object rather than static methods

GabrielFleischer:
This migration provides a way to inject dependencies inside handlers.

But this dependency injection makes tests more difficult to handle, as it needs to be done manually.

The result shown in this PR is the best I could come up with, the solution is simply to move the problem in one specific place (`DataRegistryModuleHelper`)

This also made me realize that the current handler tests are more instrumented tests that put the whole process from the receiving of a message to its handling into tests. This is a good idea, but it creates a lot of hard to read setup, and I think it would be beneficial to test the handlers as units.

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/di/DataRegistryModule.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/di/DataRegistryModule.java
@@ -30,15 +30,21 @@ public abstract class DataRegistryModule {
 
   @Provides
   @Singleton
-  public static DataRegistry provideDataRegistry() {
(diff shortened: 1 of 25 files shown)


pierluca (approved):
I see why you don't like the test setup, it doesn't feel great ... but it's also only 1-2 lines more than before. Good enough as far as I'm concerned 😁 

Overalls, this seems like a pretty solid improvement, good job !

Warhedgehog (approved):
Looks great to me ! Impressive work, it is much cleaner that way. Also concerning the tests, the supplementary boilerplate code seems quite reasonable imo.

PR #1212, opened 2022-09-21
Replace navigator clipboard by expo-clipboard

Xelowak:
Since we want the web app to work on mobile devices, the Navigator was something we needed to remove. The only use case of this was to copy to clipboard, so this has been replaced by expo's solution.


fe1-web/src/features/digital-cash/screens/SendReceive/SendReceive.tsx
```diff
@@ -256,6 +256,10 @@ export const SendReceiveHeaderRight = () => {
 
   const publicKey = useMemo(() => rollCallToken?.token.publicKey.valueOf() || '', [rollCallToken]);
 
+  const copyPkToClipBoard = async () => {
+    await Clipboard.setStringAsync(publicKey);
+  };
```
pierluca, line 261: seeing here (and elsewhere) that we're creating an async function (hence returning a promise) to await a call that returns a promise.... can't we just call `Clipboard.setStringAsync()` in the `onPress`. removing entirely this function ? It might be simpler 😄

fe1-web/src/core/components/CopyButton.tsx
```diff
@@ -1,19 +1,23 @@
+import * as Clipboard from 'expo-clipboard';
```
pierluca, line 1: I assume we're going to need a mock for this as well, no?

pierluca (commented):
Nice ! I just have a couple of suggestions to make this even simpler 👍

pierluca (approved):

PR #1231, opened 2022-09-22
Make sure the scanning fragment is opened when the permission is granted

### LINKED ISSUE #1115: [BUG] When accepting camera permission, the camera is not opened
When we use the app for the 1st time and must grant camera permission, when the permission is granted the app goes to home and connect is still selected on navbar

### Expected behavior
We expect the app to go to camera screen when camera permission is granted

### How to reproduce
1. Use the app without having already granted camera permission
2. Go to connect tab
3. Allow camera
4. See it brings you to home

### Version & Environment
This bug was reproduced on:
- Commit: [e6f7dc7](https://github.com/dedis/popstellar/commit/e6f7dc75ec530dc8fcb56ee8be52838011c7564f)

##### Front-ends:
- [ ] Not applicable

##### Back-ends:
- [ ] Be1-Go
- [ ] Be2-Scala
- [x] Not applicable

 - Device: [e.g. Pixel 4 XL emulator on Windows 10]

### Workaround
Reclicking on connect brings the user back to a normal state with no known problem caused further

### Impact
Very little. It would provide better UX to fix it

GabrielFleischer:
The PR fix #1115 which was caused by a failure when refactoring the tab system.

While reverting the deleted code, I discovered that the old code was not working in some cases because the listener of the permission grant result could be destroyed before receiving the response.

So, I decided to refactor the `CameraPermissionFragment` to merge it directly into the `QRCodeScanningFragment`.

It is a lot simpler to use, but add a new functionality to the pile that the `QRCodeScanningFragment` is.


Warhedgehog (approved):
Codewise LGTM 👍 I think it would be better to hide the rest of the fragment when we as for permission

GabrielFleischer:
> Codewise LGTM 👍 I think it would be better to hide the rest of the fragment when we as for permission 

I do not agree, I think a user might want to use the app without giving permission to use the camera.

With the current implemention, they can use the manual add to achieve it.

Warhedgehog:
Yeah you're right on second thought

PR #1232, opened 2022-09-22
Update dependencies

Tyratox:
First I just wanted to upgrade the test dependencies as of the discussion this morning but then I realized several other dependencies could be upgraded as well.

The new peer dependency system of npm together with expo made things a little complicated but it seems to be working now. All react-related libraries are fixed to version 18.0.0 which is what react-native 0.69.5 / [expo 46](https://docs.expo.dev/versions/latest/) support.

Similarly all jest dependencies are now on version 26 which is what `expo install jest-expo jest` installs.

Also I noticed a lot of unused dependencies which I removed :)

EDIT:

I fixed the installation of 'react-native-reanimated' by following https://docs.expo.dev/versions/latest/sdk/reanimated/#installation. 'react-native-reanimated' is not used by us directly but the console shows a warning if removed.


pierluca (commented):
The `tv*` properties everywhere sort of gave me a heart attack.
Would it make sense to upgrade to the rc6 of react native elements?
https://reactnativeelements.com/blog/2022/05/15/rneui-migration-guide

Happy to discuss this offline too.

.github/workflows/ci.yaml
```diff
@@ -58,7 +58,7 @@ jobs:
           npm install --legacy-peer-deps
 
       - name: Check npm security issues
-        run: npm audit fix
+        run: npm audit fix --legacy-peer-deps
```
pierluca, line 61: Is there any way we can get rid of the `--legacy-peer-deps` flags here ? 😢

Tyratox: I tried.. The problem is this: `react-native@0.69.5` ([which is required by the current expo version](https://docs.expo.dev/versions/latest/)) requires `react@18.0.0` as a peer. `@react-native-picker@2.4.2` requires `react@16` or `react@17`. The newer version, `@react-native-picker@2.4.4`, works works with react 18 but expo complains:

```
Some dependencies are incompatible with the installed expo package version:
 - @react-native-picker/picker - expected version: 2.4.2 - actual version installed: 2.4.4
Your project may not work correctly until you install the correct versions of the packages.
To install the correct versions of these packages, please run: expo install [package-name ...]
```

We can just ignore this warning, it seems to work but it doesn't feel like an ideal solution. I will quickly commit a version with this changed (to see if there is another issue, unfortunately I cannot reproduce all of the CI peer dependency issues locally) and then we can decide :)

Tyratox:
I am very happy to upgrade to the rc6 (I don't like these properties either), just wasn't sure whether this is even an options since we want the system to be more stable and I am not sure whether relying on a beta version helps in this regard. But if you think this is worth a try I am happy to try! According to the link you sent it should not take a lot of effort upgrading it but we should be able to get rid of the tv props :)

.github/workflows/ci.yaml
pierluca: Ehm, it seems there are a whole bunch of other issues ... that's what I was worried about :-/

pierluca:
I agree `rc6` is probably not the cleanest solution, but considering:
1. we don't need to release this software tomorrow (at least, not to 3rd parties)
2. if we get it working with rc6, moving to `4.x` should be somewhat painless
3. dealing with all these `tv*` properties is going to be unnatural for a while
I'd be willing to take the risk.

.github/workflows/ci.yaml
Tyratox: I think I just managed to get it working :) Not with all the latest versions but with more latest versions 😂

Tyratox:
Yes that sound reasonable! I'll try then! :)

Tyratox:
Uses https://github.com/react-native-elements/react-native-elements/issues/3607#issuecomment-1205035640 to make the new version run.

Tyratox:
Lastly there is the issue of

```
Warning: ReactDOM.render is no longer supported in React 18. Use createRoot instead. Until you switch to the new API, your app will behave as if it's running React 17. Learn more: https://reactjs.org/link/switch-to-createroot```

described [here](https://github.com/expo/expo/issues/18485).

I would suggest waiting until this is fixed by expo since it does not seem to cause issues for now?

File: fe1-web/src/features/social/navigation/SocialMediaNavigation.tsx [hunk 2]
```diff
@@ -72,11 +72,15 @@ const SocialMediaNavigation = () => {
       /* noop */
     });
 
+  const contextValue = useMemo(
+    () => ({
+      currentUserPopTokenPublicKey,
+    }),
+    [currentUserPopTokenPublicKey],
+  );
```
pierluca, line 80: isn't this a bit of a weird change in this PR ? What is it for?

fe1-web/webpack.config.js
```diff
@@ -7,7 +7,16 @@ const aliases = {
...
 
 async function configAsync(env, argv) {
-  const config = await createExpoWebpackConfigAsync(env, argv);
+  const config = await createExpoWebpackConfigAsync({
+      ...env,
+      babel: {
+          // This is a workaround for https://github.com/react-native-elements/react-native-elements/issues/3607
+          // to get the new version of react-native-elements working
+          // Revert this commit as soon as the new version is properly released and this
+          // issue is resolved
```
pierluca, line 16: You might also want to add a GitHub issue in our own repository to track this, documenting the rc6 + hack😄

fe1-web/package.json
```diff
@@ -28,23 +28,23 @@
...
+    "@rneui/base": "^4.0.0-rc.6",
+    "@rneui/themed": "^4.0.0-rc.6",
     "@sinonjs/fake-timers": "^9.1.2",
     "ajv": "^8.10.0",
     "ajv-formats": "^2.1.1",
     "base64url": "^3.0.1",
     "bip39": "^3.0.3",
-    "color": "^3.2.1",
     "ed25519-hd-key": "^1.2.0",
+    "eslint-config-airbnb": "^19.0.4",
```
pierluca, line 47: this should probably be a dev dependency

fe1-web/package.json
```diff
@@ -28,23 +28,23 @@
...
     "expect": "^26.6.2",
```
pierluca, line 48: shouldn't this be a dev dependency ?

fe1-web/package.json
```diff
@@ -28,23 +28,23 @@
...
+    "@react-native-picker/picker": "^2.4.4",
+    "@react-navigation/bottom-tabs": "^6.4.0",
+    "@react-navigation/core": "^6.4.0",
+    "@react-navigation/native": "^6.0.13",
+    "@react-navigation/routers": "^6.1.3",
+    "@react-navigation/stack": "^6.3.1",
+    "@reduxjs/toolkit": "^1.8.5",
+    "@rneui/base": "^4.0.0-rc.6",
+    "@rneui/themed": "^4.0.0-rc.6",
     "@sinonjs/fake-timers": "^9.1.2",
```
pierluca, line 41: also a dev dependency

fe1-web/src/features/home/screens/__tests__/ConnectOpenScan.test.tsx
```diff
@@ -34,7 +34,7 @@ jest.mock('@react-navigation/core', () => {
 });
 
 // Is mocked
-// eslint-disable-next-line react-hooks/rules-of-hooks
+
```
pierluca, line 37: do we need the empty line here? 😛

pierluca (changes requested):
This looks much better. Minor remarks, but very good overalls 👍

fe1-web/src/features/social/navigation/SocialMediaNavigation.tsx
Tyratox: True! I think we should set the linter to be more aggressive in these cases, for some reason it was reported to me once but now not anymore. I assumed it was because of a eslint update but it seems it was something else..

Anyway, the purpose is to prevent unnecessary rerenders, react only performs shallow equality checks and before the value passed to the context always was a different object which would trigger a re-render in all components making use of the context. I will add a comment to make this clear.

Tyratox: I guess it might be a good idea to find the corresponding linter rule and enable it

fe1-web/src/features/home/screens/__tests__/ConnectOpenScan.test.tsx
Tyratox: Nope 😁👍

pierluca:
Everything's been addressed.... except `expo doctor`. What shall we do with it?

PR #1233, opened 2022-09-23
Add Scan button in the rollcall fragment to go back to the scanning fragment

GabrielFleischer:
This PR adds a Scan button to the RollCall Fragment, it allows the organizer to go back to the scanning fragment when the roll call is opened.

I also refactored the RollCall Fragment to fix some codesmells, use data binding and have better naming.

### DIFF
--- a/fe2-android/app/src/test/ui/robolectric/com/github/dedis/popstellar/ui/detail/event/rollcall/RollCallFragmentTest.java
+++ b/fe2-android/app/src/test/ui/robolectric/com/github/dedis/popstellar/ui/detail/event/rollcall/RollCallFragmentTest.java
@@ -41,8 +41,7 @@ import io.reactivex.subjects.BehaviorSubject;
 
 import static androidx.test.espresso.action.ViewActions.click;
 import static androidx.test.espresso.assertion.ViewAssertions.matches;
-import static androidx.test.espresso.matcher.ViewMatchers.isDisplayed;
-import static androidx.test.espresso.matcher.ViewMatchers.withText;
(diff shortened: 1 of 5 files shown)


Warhedgehog (changes requested):
Very good improvement ! I have to admit that I'm not the fan of the UI. Maybe a round image button with a camera in the empty space ?

GabrielFleischer:
> Very good improvement ! I have to admit that I'm not a fan of the UI. Maybe a round image button with a camera in the empty space ?

I changed it to this :
image

Do you think it is better ?

Warhedgehog (approved):
Sorry for the delay I thought I had reviewed it. LGTM so !

fe2-android/app/src/test/ui/robolectric/com/github/dedis/popstellar/ui/detail/event/rollcall/RollCallFragmentTest.java
```diff
@@ -193,12 +205,28 @@ public void managementButtonCloseRollCallWhenOpened() {
     messageSenderHelper.assertSubscriptions();
   }
 
+  @Test
+  public void scanButtonOPenScanningTest() {
```
pierluca, line 209:
```suggestion
  public void scanButtonOpenScanningTest() {
```

pierluca (approved):
LGTM, but I'm not sure I agree with Johann with respect to icon vs. text.
What do you think @nkcr ?

Anyhow, I believe we can move forward for now and we'll revisit this as needed.

pierluca (approved):

Issue #1226, opened 2022-09-22
[BUG] Navigation with arrows does not work in Lao


Warhedgehog:
### Description (Actual behavior)
When in the lao screen, the back arrow and back buttons do not work

### Expected behavior
We expect the back arrows/buttons to take us back to the event list

### How to reproduce
1. Go to an lao
2. Create an event
3. Click on the event
4. Click on back arrow
5. See that it does nothing

### Version & Environment
This bug was reproduced on:
- Commit: [005c834](https://github.com/dedis/popstellar/commit/005c834e4f0b5145989c6b7f1a37ed5feff1957a)

##### Front-ends:
- [ ] Fe1-Web (please include browser's names & version)
- [x] Fe2-Android (please specify if phone or emulation, and Android version)
- [ ] Not applicable

##### Back-ends:
- [ ] Be1-Go
- [ ] Be2-Scala
- [x] Not applicable

### Workaround
Navigate with bottom tab

### Impact
It's an annoying bug but do not break any feature

### Possible root cause
The navigation refactoring overlooked this I think

PR #1234, opened 2022-09-23
Fix navigation bug

Warhedgehog:
This fixes issue #1226. I also removed the back buttons of election and roll call setup pages since the back arrow offers the same functionality

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/event/election/fragments/ElectionSetupFragment.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/event/election/fragments/ElectionSetupFragment.java
@@ -91,39 +91,39 @@ public class ElectionSetupFragment extends AbstractEventCreationFragment {
       @Nullable ViewGroup container,
       @Nullable Bundle savedInstanceState) {
 
-    ElectionSetupFragmentBinding setupElectionFragBinding =
+    ElectionSetupFragmentBinding binding =
         ElectionSetupFragmentBinding.inflate(inflater, container, false);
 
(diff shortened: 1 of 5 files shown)


fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/event/election/fragments/ElectionSetupFragment.java
```diff
@@ -94,39 +91,39 @@ public View onCreateView(
       @Nullable ViewGroup container,
       @Nullable Bundle savedInstanceState) {
 
-    mSetupElectionFragBinding = ElectionSetupFragmentBinding.inflate(inflater, container, false);
+    ElectionSetupFragmentBinding setupElectionFragBinding =
```
GabrielFleischer, line 94: In the rest of the app (where I renamed it too)I named the Binding `binding`. I feel like it is self-explanatory in the context, and it is enough.

What do you think ?

GabrielFleischer (approved):
Good job

We might need to do this fix at other places, but nothing comes to my mind right now. We'll create issues when necessary,

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/detail/event/election/fragments/ElectionSetupFragment.java
Warhedgehog: Much better I'll change it

Warhedgehog:
I don't think it will problem because the conditions for it to happen are quite specific i.e. it has to be a different fragment than the Default tab but opened from the Default tab. Both `HomeActivity` and `DigitalCashActivity` offer no access to a different fragment.

GabrielFleischer (approved):
LGTM

PR #1235, opened 2022-09-24
Fix disappearing qr code icon if lao title is too long

### LINKED ISSUE #1229: [BUG] Too long LAO titles hide the qr icon
When the lao title is quite long the qr icon (which let you display the qr code of the lao) is pushed out of the screen.

### Expected behavior
We expect it to be displayed regardless of the length of the title

### How to reproduce
1. Launch an LAO with a long title
2. See the qr code icon is not next to the lao title

### Version & Environment
This bug was reproduced on:
- Commit: 4e0323d020fbcfb97612de1fa685a4cddb220ab2

##### Front-ends:
- [ ] Not applicable

##### Back-ends:
- [ ] Be1-Go
- [ ] Be2-Scala
- [x] Not applicable

### Impact
If only android app participates it would be practically unusable

Warhedgehog:
This fixes #1229. The fix is quite ugly it defines a max width for the title. Value was chosen by trial and error on different screen sizes. For tiny phones, it might still be a problem though. 
I've looked for a better alternative online and found some people complaining about the same kind of problem. None of the other proposed solution worked/were satisfactory.

### DIFF
--- a/fe2-android/app/src/main/res/layout/lao_detail_fragment.xml
+++ b/fe2-android/app/src/main/res/layout/lao_detail_fragment.xml
@@ -33,10 +33,10 @@
         android:layout_width="@dimen/channel_qr_code_size"
         android:layout_height="@dimen/channel_qr_code_size"
         android:contentDescription="@string/qr_code_image_description"
+        app:layout_constraintTop_toTopOf="parent"
         app:layout_constraintBottom_toBottomOf="parent"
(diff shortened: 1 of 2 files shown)


GabrielFleischer (approved):
As you said, probably not the most beautiful fix.

But it's fine by me 😁

pierluca (approved):
Not great, but good enough for now, by far not our biggest issue 😁 👍

PR #1238, opened 2022-09-26
[Bug Fix] Fix multiple issues in the social media feature

GabrielFleischer:
The shuld be the final Pull Request from the social media decoupling, it aims to fix all issue found while testing the feature before merging it to master.

Issues found :
 - Modifications of the view was not done on the main thread
 - Subscriptions to the social media channels were not done correctly
 - The delete button was not refreshed, which caused multiple issues

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/socialmedia/SocialMediaViewModel.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/socialmedia/SocialMediaViewModel.java
@@ -193,16 +193,12 @@ public class SocialMediaViewModel extends NavigationViewModel<SocialMediaTab> {
   public Observable<List<Chirp>> getChirps() {
     return socialMediaRepository
         .getChirpsOfLao(laoId)
-        .observeOn(schedulerProvider.mainThread())
         // Retrieve chirp subjects per id
         .map(
             ids -> {
               List<Observable<Chirp>> chirps = new ArrayList<>(ids.size());
               for (MessageID id : ids) {
-                chirps.add(
-                    socialMediaRepository
-                        .getChirp(laoId, id)
-                        .observeOn(schedulerProvider.mainThread()));
+                chirps.add(socialMediaRepository.getChirp(laoId, id));
               }
               return chirps;
             })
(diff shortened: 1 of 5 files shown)


fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/data/RollCallHandler.java
```diff
@@ -138,23 +132,16 @@ public void handleCloseRollCall(HandlerContext context, CloseRollCall closeRollC
...
+    rollCall
+        .getAttendees()
+        .forEach(
+            token ->
+                context
+                    .getMessageSender()
+                    .subscribe(channel.subChannel("social").subChannel(token.getEncoded()))
+                    .subscribe(
+                        () -> Log.d(TAG, "subscription a success"),
+                        error -> Log.d(TAG, "subscription error")));
```
pierluca, line 144: The logic here changes rather brutally.

It's my understanding in the past (at least on Fe1) that you'd have to choose who you subscribe to.
So by default, you'd only subscribe to the main social channel, but then you could choose to subscribe to some, all or no PoP token.

This logic auto-subscribes to all the PoP tokens. Is it correct / what you intended?

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/socialmedia/SocialMediaViewModel.java
```diff
@@ -189,12 +193,16 @@ public Single<MessageGeneral> deleteChirp(MessageID chirpId, long timestamp) {
   public Observable<List<Chirp>> getChirps() {
     return socialMediaRepository
         .getChirpsOfLao(laoId)
+        .observeOn(schedulerProvider.mainThread())
```
Warhedgehog, line 196: What problem did this cause ?

fe2-android/app/src/test/unit/java/com/github/dedis/popstellar/utility/handler/RollCallHandlerTest.java
```diff
@@ -68,7 +67,7 @@ public void setup() throws GeneralSecurityException, IOException, KeyException {
     lenient().when(keyManager.getMainPublicKey()).thenReturn(SENDER);
     lenient().when(keyManager.getValidPoPToken(any(), any())).thenReturn(POP_TOKEN);
 
-    when(messageSender.subscribe(any())).then(args -> Completable.complete());
+    lenient().when(messageSender.subscribe(any())).then(args -> Completable.complete());
```
Warhedgehog, line 70: Why the use of lenient ?

Warhedgehog (approved):
Provided that the issue Pierluca raised is fixed, it's fine by me. Looking online, the use of `lenient()` seems a bit frowned upon though.

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/socialmedia/SocialMediaViewModel.java
GabrielFleischer: If this is not present, the whole computation will be done in the same thread as the one that published a new version, which was not the main thread.

This caused issues when trying to modify the views as this can only be done from the main thread.

fe2-android/app/src/test/unit/java/com/github/dedis/popstellar/utility/handler/RollCallHandlerTest.java
GabrielFleischer: By default, Mockito makes sure that mocks are actually used in tests. Using lenient disables this check.

This caused issues because every test do not trigger the subscribe function.

PR #1240, opened 2022-09-27
Add json schemas for pop token and identity qr codes

Warhedgehog:
In order to be able to verify the content of a scanned qr code, we need to add json schemas to protocol. This will mitigate error coming from user showing the wrong qr code to scan


Tyratox (approved):

protocol/qrcode/identity.json
```diff
@@ -0,0 +1,16 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "resource:/protocol/qrcode/identity.json",
+    "title": "QRCode data for the user's main public key",
+    "description": "Match a data contained in a qrcode encoding the main public key",
+    "type": "object",
+    "additionalProperties": false,
+    "properties": {
+        "identity": {
```
pierluca, line 9: I feel we might regret later on calling this merely `identity` 😅
do you have any idea for a more unique name ?

Warhedgehog: It's purpose is to display the main public key and I don't think it should change so we can call it `main_public_key` ?

jbsv: or if may play the 🐫 , I would name it: `mainPublicKey`

Warhedgehog: I think we have more 🐍 than 🐫 in schemas especially in messageData ones. But I don't know what's preferred for the pop project 🙂

jbsv: Well, the standard http://json-schema.org/draft-07/schema# tells me that it should be 🐫. Mixing up the 2 is of course a bad idea. Maybe @pierluca has some historical stake on this ?

jbsv: OK, talked to @pierluca and it looks like the pop project has taken the 🐍 path tic this regard. So we should continue with it as a pop "standard".

jbsv (approved):

protocol/qrcode/identity.json
Warhedgehog: Alright, thanks for the clarification !

Issue #1222, opened 2022-09-22
[BUG] Elections cannot be closed


Tyratox:
### Description (Actual behavior)
During the meeting on the 22nd of September, elections could not be closed.

```
Could not complete the election
[An error response was replied by the server
Error -4 - invalid message field: failed to publish: failed to handle a publish message: failed to process message: failed to process action 'election#end': invalid election#end message: failed to compute registered votes : registered votes is sO3qhBGeVx-9N-x2Z78U59ldSfaQ4KFVRWxkvqK7dAY=, should be sorted and equal to 2-PF0al4LnupnOzDgJx96glZu5q8-OrIwbMpn_2wA7k=]
com.github.dedis.popstellar.utility.error.JsonRPCErrorException: Error -4 - invalid message field: failed to publish: failed to handle a publish message: failed to process message: failed to process action 'election#end': invalid election#end message: failed to compute registered votes : registered votes is sO3qhBGeVx-9N-x2Z78U59ldSfaQ4KFVRWxkvqK7dAY=, should be sorted and equal to 2-PF0al4LnupnOzDgJx96glZu5q8-OrIwbMpn_2wA7k=
```

### How to reproduce
1. Create lao, roll call and election
2. Open election
3. Vote from multiple devices
4. Try to close the election

The sent message:
```
{
[...]

PR #1251, opened 2022-09-29
Extend test set

Tyratox:
Should help fixing #1222


pierluca (approved):
I can only trust you blindly on this one... 😅

PR #1262, opened 2022-09-30
Do not select tab when switching opens an activity

GabrielFleischer:
This tiny PR tries to fix the issue of the navigation bar that desynchronizes from the actual opened tab.

It removes the fact that the social media and digital cash activities are selected when pressing their respective button.

That way, going back to the Lao activity keeps the last selected tab, which is the one opened.


Warhedgehog (approved):
LGTM ! Tested it and seems to be working just fine

jbsv (approved):
So digital cash and social media are handled differently ? Is it temporary ?

PR #1267, opened 2022-10-04
Add firefox polyfill to fix #1258

### LINKED ISSUE #1258: [BUG] Requesting camera permissions does not work on firefox
Requesting camera permissions does not work on firefox. A javascript error is logged.

```
Uncaught (in promise) TypeError: 'camera' (value of 'name' member of PermissionDescriptor) is not a valid value for enumeration PermissionName.
    Z ExponentCameraManager.web.ts:89
    Babel 4
    Z 2.53ec6a87.chunk.js:8
    X 2.53ec6a87.chunk.js:8
    e ExponentCameraManager.web.ts:203
    Babel 4
    getCameraPermissionsAsync ExponentCameraManager.web.ts:115
    e Camera.tsx:163
    Babel 4
    value 2.53ec6a87.chunk.js:8
    d PermissionsHook.ts:39
    Babel 4
    Y PermissionsHook.ts:57
    React 7
    _animatedPropsCallback createAnimatedComponent.js:86
    value AnimatedProps.js:99
    value AnimatedValue.js:57
    value AnimatedValue.js:57
    value AnimatedValue.js:270
    value AnimatedValue.js:230
    value TimingAnimation.js:112
```

### How to reproduce
1. Go to the scanner tab
2. Open the js console

### Version & Environment
This bug was reproduced on:
- Commit: a0c8d16c74b8f64d4c710aa9b9acea2610bc4f8b

##### Front-ends:
- [x] Fe1-Web (please include browser's names & version)
- [ ] Not applicable

##### Back-ends:
- [ ] Be1-Go
- [ ] Be2-Scala
- [x] Not applicable

### Impact
Lao ID and URL must be entered manually, bad UX

Tyratox:
This is very hacky so I am happy to discuss whether there are better solutions

### DIFF
--- a/fe1-web/src/App.tsx
+++ b/fe1-web/src/App.tsx
@@ -18,10 +18,10 @@ import { persist, store } from 'core/redux';
 import { Color } from 'core/styles';
 import { configureFeatures } from 'features';
 
-import firefoxCameraPolyfill from './polyfill/firefox/expo-camera';
+import cameraPolyfill from './core/platform/camera/web-polyfill';
 
 // load polyfill when the app loads
-firefoxCameraPolyfill();
+cameraPolyfill();
 
 const { messageRegistry, keyPairRegistry, navigationOpts, context } = configureFeatures();
 configureKeyPair();
--- a/fe1-web/src/App.tsx
+++ b/fe1-web/src/App.tsx
@@ -18,6 +18,11 @@ import { persist, store } from 'core/redux';
 import { Color } from 'core/styles';
 import { configureFeatures } from 'features';
 
+import firefoxCameraPolyfill from './polyfill/firefox/expo-camera';
+
+// load polyfill when the app loads
+firefoxCameraPolyfill();
+
 const { messageRegistry, keyPairRegistry, navigationOpts, context } = configureFeatures();
 configureKeyPair();
 configureNetwork(messageRegistry, keyPairRegistry);
--- a/fe1-web/src/core/platform/camera/web-polyfill.native.ts
+++ b/fe1-web/src/core/platform/camera/web-polyfill.native.ts
@@ -1,0 +1,2 @@
+// replaces the polyfill
+export default () => {};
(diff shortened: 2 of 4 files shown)


pierluca:
In principle, I don't have issues with polyfills when they're needed. Though I'd definitely put this in `src/core/platform` and probably put a `Platform.OS === 'web'` guard around it, or even better create an empty file (alternatively) with `.native.ts` so that it's never compiled into a native build.

However, if you're telling me in #1258 that this is supposed to work, it might warrant a bit of investigation. I'd also consider raising an issue with the expo people to see if there's something we've obviously missed.

Tyratox:
Good points, thank you! Did not know about the `.native.ts` thing :)

The `TypeError: t.getCapabilities is not a function` error also happens in the online snack but it doesn't seem to be required for the qr code scanner functionality. 

Thus https://github.com/dedis/popstellar/pull/1267/files#diff-71c35ac79062b4526ec9a585bbce8a37d5b8b354e43ec45036a828b2410c1333R17 is mostly there to silence this error, especially during development it is quite annoying having the big error popup. (Which apparently cannot be prevented using error boundaries)

Regarding #1258 I am slightly confused why it works in the online setting. I tried looking at the compiled javascript code that is served online but then quickly deemed it to be not worth the time reverse engineering. Googling for this issue does not yield any `expo` related entries but for instance https://stackoverflow.com/a/53155894/2897827 which mentions that the permissions API is experimental. Not sure why `expo-camera` uses it honestly, especially since there are alternatives. Given the above I assumed they are using a polyfill and thus I looked for firefox polyfills and stumbled upon ttps://github.com/chromium/permissions.request/blob/6302dbd2fa5cbe7488b2534747ac0e51c4df3822/permissions.request.ts from which I copied the relevant parts and slightly adapted them.

An alternative to the polyfill would be to replace the respective code in the `expo-camera` dependency using https://www.npmjs.com/package/patch-package until it is fixed / supported by firefox. (e.g. replace the permissions.query with getUserMedia, similar to the polyfill but [should work more reliably cross-browser, except IE of course](https://caniuse.com/stream)).

pierluca (approved):
You have a point. Sold.

Issue #1217, opened 2022-09-22
[BUG] UI is not always updated


Tyratox:
### Description (Actual behavior)
The UI does not always update if there was a change on the back-end.

### Expected behavior
If necessary, the UI should update if the organizer or other attendees interacted with the LAO.

### How to reproduce
1. Create LAO and connect to it with a 2nd device (I only tested it with an iPhone)
2. Create an election, do *not* start it yet
3. Close the browser on the iPhone and do something else on the phone
4. Start the election
5. Re-open the browser and check if you see that the election has started

### Version & Environment
This bug was reproduced on:
- Commit: f78099a4c7c7c3a58acd71f4fddc67251b35ba53

##### Front-ends:
- [x] Fe1-Web (please include browser's names & version)
- [ ] Fe2-Android (please specify if phone or emulation, and Android version)
- [ ] Not applicable

##### Back-ends:
- [ ] Be1-Go
- [ ] Be2-Scala
- [x] Not applicable

### Workaround
Refresh the page in the browser and re-connect.

### Impact
The user is not informed about changes in the LAO

### Possible root cause
Probably the network connection is dropped. I thought I fixed this issue last semester but maybe it has now re-surfaced during the upgrades of the dependencies.

Issue #1224, opened 2022-09-22
[BUG] If the network stack in fe-1 crashes, the application becomes unusable


Tyratox:
### Description (Actual behavior)
On the 22nd of September we noticed that it was no longer possible to interact with the application (namely issuing money) after there was an error in one part of the application.

### How to reproduce
To be specified

### Version & Environment
This bug was reproduced on:
- Commit: f78099a4c7c7c3a58acd71f4fddc67251b35ba53

##### Front-ends:
- [x] Fe1-Web (please include browser's names & version)
- [ ] Fe2-Android (please specify if phone or emulation, and Android version)
- [ ] Not applicable

##### Back-ends:
- [ ] Be1-Go
- [ ] Be2-Scala
- [x] Not applicable

### Workaround
Refresh the website and re-connect to the lao

### Impact
Application becomes unusable without telling the user.

Tyratox:
@pierluca Do you remember how we got the network stack to crash?

PR #1268, opened 2022-10-06
Fix networking issues

Tyratox:
This PR should fix issues #1217 and #1224.

It does so by making the whole network stack more stable and improves the handling of requests that cannot be sent. (Re-try sending them even if the connection has been re-established)

What is left to do is writing tests (for NetworkConnection.ts, no tests exist so far 🙃) As this will probably take some time, it might be worth discussing to what extend I should try to model the network :)

As a side effect of this PR, both the navigation and the toasts can now be accessed globally, i.e. also outside of react components. This allows to trigger a toast and navigate to home when a connection breaks.

### DIFF
--- a/fe1-web/src/core/network/NetworkConnection.ts
+++ b/fe1-web/src/core/network/NetworkConnection.ts
@@ -67,9 +67,14 @@ export class NetworkConnection {
   private onInitialOpenCallback?: () => void;
 
   /**
+   * The timeout set when initially opening a connection
+   */
+  private readonly initalOpenTimeout: ReturnType<typeof setTimeout>;
+
+  /**
    * Function called when the connection closes for good
    */
-  private readonly onConnectionDeadCallback?: () => void;
+  private readonly onConnectionDeathCallback?: () => void;
 
   /**
    * Boolean indicating whether this connection is open or there is still hope
@@ -90,14 +95,27 @@ export class NetworkConnection {
     address: string,
     handler?: JsonRpcHandler,
     onInitialOpenCallback?: () => void,
-    onConnectionDeadCallback?: () => void,
+    onInitialOpenTimeout?: () => void,
+    onConnectionDeathCallback?: () => void,
   ) {
     this.address = address;
     this.onRpcHandler = handler !== undefined ? handler : defaultRpcHandler;
     this.onInitialOpenCallback = onInitialOpenCallback;
-    this.onConnectionDeadCallback = onConnectionDeadCallback;
+    this.onConnectionDeathCallback = onConnectionDeathCallback;
     this.alive = true;
     this.closeIntent = false;
+
+    this.initalOpenTimeout = setTimeout(() => {
+      // if the initial timeout expires, we deem this connection
+      // dead. this signals the rest of the class to ignore
+      // any further data that is received
+      this.alive = false;
+      this.disconnect();
+
+      if (onInitialOpenTimeout) {
+        onInitialOpenTimeout();
+      }
+    }, WEBSOCKET_CONNECTION_FAILURE_TIMEOUT_MS);
 
     this.ws = this.establishConnection(address);
   }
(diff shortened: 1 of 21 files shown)


Xelowak (approved):
LGTM! Very nice changes :)

fe1-web/src/core/network/NetworkConnection.ts
```diff
@@ -96,24 +218,46 @@ export class NetworkConnection {
...
+
+        // check if callback is set
+      } else if (this.onConnectionDeadCallback) {
+        // do not try to re-establish the connection, this connection seems
+        // to be broken for now and the near future
+        console.error(`Connection with ${this.address} broke for good`);
+        this.alive = false;
+        this.onConnectionDeadCallback();
+      }
+    }
```
pierluca, line 246: save an indentation, return early 😁 

```suggestion
    if (this.closeIntent) {
      return;
    }
    
    // it was not our intention to close the connection. try to re-connect
    this.failedConnectionAttempts += 1;
    console.error(`Trying to re-establish a connection at address : ${this.address}`);

    // only retry a certain number of times and add a wait before retrying
    if (this.failedConnectionAttempts <= WEBSOCKET_CONNECTION_MAX_ATTEMPTS) {
      setTimeout(() => {
        this.reconnectIfNecessary();
      }, WEBSOCKET_CONNECTION_FAILURE_TIMEOUT_MS);
      return;
    }
    
    if (this.onConnectionDeadCallback) {
      // do not try to re-establish the connection, this connection seems
      // to be broken for now and the near future
      console.error(`Connection with ${this.address} broke for good`);
      this.alive = false;
      this.onConnectionDeadCallback();
    }
    
```


which sort of brings about another question, shouldn't `console.error` and `this.alive = false` happen if `onConnectionDeadCallback` is not defined ? 

i.e. 
```typescript
    // do not try to re-establish the connection, this connection seems
    // to be broken for now and the near future
    console.error(`Connection with ${this.address} broke for good`);
    this.alive = false;

    if (this.onConnectionDeadCallback) {
      this.onConnectionDeadCallback();
    } 
```

fe1-web/src/core/network/NetworkConnection.ts
```diff
@@ -59,25 +113,93 @@ export class NetworkConnection {
...
+
+    // if there are pending requests, re-send them
+    // this can happen after the connection was re-established
+    for (const entry of this.pendingRequests.entries()) {
+      const request: PendingRequest = entry[1];
+
+      // clear timeout, connection was just re-established
+      clearTimeout(request.timeoutId);
+      this.resendRequest(request);
+    }
```
pierluca, line 196: 👍

fe1-web/src/features/home/screens/ConnectOpenScan.tsx
```diff
@@ -70,7 +70,7 @@ const ConnectOpenScan = () => {
     });
   }, [navigation]);
 
-  const handleScan = (data: string | null) => {
+  const handleScan = async (data: string | null) => {
```
pierluca, line 73: since this function is now `async`, do we need to check that the component is still mounted before initiating the navigation ? I know we do to change state, I'm not sure about navigation.

Also, this function is rather huge, we might want to break it up a bit 😅

fe1-web/src/features/lao/network/LaoGreetWatcher.ts
```diff
@@ -48,16 +48,24 @@ export const storeBackendAndConnectToPeers = (
...
+    );
+
+    // mark the lao#greet message as handled
+    dispatch(
+      handleGreetLaoMessage({
+        messageId: messageId.valueOf(),
+      }),
+    );
+  } catch {
+    // here we can actually ignore errors for now since
```
pierluca, line 64: since this is a "for now" thing, I'd suggest adding a console.warn indicating that this is currently an unsupported / unimplemented path. This way it's easier to find than a subtly hidden unimplemented feature 😉

pierluca (changes requested):
Pretty good, just a bug (if I understood correctly) and minor stylistic remarks 👍

fe1-web/src/core/network/NetworkConnection.ts
Tyratox: Oh yes, defintiely! Thank you for catching that! 😁

fe1-web/src/features/home/screens/ConnectOpenScan.tsx
Tyratox: True, otherwise we'll have interesting navigation problems again 😅 (when closing the scanner right after a qr code was visible)

pierluca (approved):
Can we improve the test coverage a bit, before merging? 🙏

Tyratox:
Yes of course! It is just that this might take some time as I assume writing tests for the network interface will be a bit cumbersome with respect to mocking etc :) But I will try to do so tomorrow / over the weekend!

pierluca (approved):
👍

Issue #1219, opened 2022-09-22
[BUG] Election checkbox label names cannot be clicked


Tyratox:
### Description (Actual behavior)
In elections it is not possible to select an option by clicking on the text label

### How to reproduce
1. Create LAO, roll call and an election
2. Open the election
3. Try to select one of the options by clicking on the label text

### Version & Environment
This bug was reproduced on:
- Commit: f78099a4c7c7c3a58acd71f4fddc67251b35ba53

##### Front-ends:
- [x] Fe1-Web (please include browser's names & version)
- [ ] Fe2-Android (please specify if phone or emulation, and Android version)
- [ ] Not applicable

##### Back-ends:
- [ ] Be1-Go
- [ ] Be2-Scala
- [x] Not applicable

PR #1269, opened 2022-10-08
Make election answers labels clickable

Xelowak:
Fixes issue #1219. Label of answers are now clickable instead of having to click on the checkbox specifically.

### DIFF
--- a/fe1-web/src/features/evoting/components/ElectionOpened.tsx
+++ b/fe1-web/src/features/evoting/components/ElectionOpened.tsx
@@ -146,7 +146,10 @@ const ElectionOpened = ({ election }: IPropTypes) => {
               </ListItem.Content>
             }
             onPress={() =>
-              setIsQuestionOpen({ ...isQuestionOpen, [question.id]: !isQuestionOpen[question.id] })
+              setIsQuestionOpen({
+                ...isQuestionOpen,
+                [question.id]: !isQuestionOpen[question.id],
+              })
             }
             isExpanded={!!isQuestionOpen[question.id]}>
             {question.ballot_options.map((ballotOption, ballotOptionIndex) => {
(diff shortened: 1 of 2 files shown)


fe1-web/src/features/evoting/components/ElectionOpened.tsx
```diff
@@ -159,19 +162,25 @@ const ElectionOpened = ({ election }: IPropTypes) => {
...
                       testID={`questions_${questionIndex}_ballots_option_${ballotOptionIndex}_checkbox`}
                       size={Icon.size}
                       checked={selectedBallots[questionIndex] === ballotOptionIndex}
-                      onPress={() =>
-                        setSelectedBallots({
-                          ...selectedBallots,
-                          [questionIndex]: ballotOptionIndex,
-                        })
-                      }
+                      onPress={onPress}
```
Tyratox, line 183: Is this still required if we already have the `onPress` on the wrapping `ListItem`?

Tyratox (dismissed):

Tyratox:
A very nice change! 😊 Finally we do not have to hit the checkboxes anymore 😂

fe1-web/src/features/evoting/components/ElectionOpened.tsx
Xelowak: I thought the same but if I delete it, clicking on the checkbox wouldn't do anything. At least now you have the choice to do both!

Tyratox: Huh, weird 😂👍

Tyratox (approved):

fe1-web/src/features/evoting/components/ElectionOpened.tsx
```diff
@@ -146,7 +146,10 @@ const ElectionOpened = ({ election }: IPropTypes) => {
               </ListItem.Content>
             }
             onPress={() =>
-              setIsQuestionOpen({ ...isQuestionOpen, [question.id]: !isQuestionOpen[question.id] })
+              setIsQuestionOpen({
+                ...isQuestionOpen,
+                [question.id]: !isQuestionOpen[question.id],
+              })
```
pierluca, line 152: Sonar says that lines 138 to 154 are now an exact duplicate of code in ElectionQuestions.
Does it make sense to factor this out into its own component ? Just asking, not suggesting :)

pierluca (commented):
👍

fe1-web/src/features/evoting/components/ElectionOpened.tsx
pierluca: @Tyratox I'll let you fix this in a different PR ... if it's worth fixing. Be the judge of that 😁

PR #1270, opened 2022-10-11
Changing UI of the Home screen

Tyratox:
I tried to implement the changes we discussed last week.

Screenshot 2022-10-11 at 11 32 48

Screenshot 2022-10-11 at 11 33 00

Screenshot 2022-10-11 at 20 45 07

(OLD:
Screenshot 2022-10-11 at 11 33 13
)

Thoughts, comments?
Should I add an additional text like "Camera not working?" above "Enter manually"?
Should I add additional instructions above the transparent qr code overlay?

--

Technical detail:

The reason why many snapshot files changed is because of a issue with react-navigation. It was surprisingly hard to get the bottom toolbar to be sticky. Usually it should be as simple as https://stackoverflow.com/a/31249011/2897827 but some reason react-navigation by default adds a scroll view which leads to very weird behavior if one actually wants to put scroll views in the proper places. This was disabled as described here: https://github.com/react-navigation/react-navigation/issues/9344#issuecomment-782520454

Because of this change many otherwise unrelated snapshot files now changed. (If they contained some navigation components)


fe1-web/src/features/home/navigation/HomeNavigation.tsx
```diff
@@ -40,67 +31,36 @@ const HomeNavigation = () => {
...
-          e.preventDefault();
-          navigation.navigate(STRINGS.navigation_app_connect);
-        },
-        tabBarIcon: makeIcon('scan'),
-        order: -10000,
-      } as HomeFeature.HomeScreen,
+        headerShown: false,
+      },
       // sort screens by order before rendering them
-    ].sort((a, b) => a.order - b.order);
```
Xelowak, line 58: How is it possible that we can remove orders of screens? Is it handled automatically?

fe1-web/src/features/home/navigation/HomeNavigation.tsx
```diff
@@ -40,67 +31,36 @@ const HomeNavigation = () => {
...
             name={id}
             component={Component}
-            listeners={{ tabPress }}
             options={{
               title: title || id,
               headerTitle: headerTitle || title || id,
               headerLeft,
               headerRight,
-              tabBarIcon: tabBarIcon || undefined,
               // hide the item if tabBarIcon is set to null
```
Xelowak, line 62: This comment can be removed.

fe1-web/src/features/home/screens/Launch.tsx
```diff
@@ -106,10 +104,6 @@ const Launch = () => {
           <PoPTextButton onPress={connectToTestLao}>
             [TEST] Connect to LocalMockServer.ts (use &apos;npm run startServer&apos;)
           </PoPTextButton>
```
Xelowak, line 106: Is it still useful? If it is the case, it might be nice to add it to `strings.ts`. I honestly never used this button, so I don't know if it's still relevant.

Xelowak (commented):
Nice addition! I added some minor comments.

Having generic screenOptions like this is so much cleaner, that's a very nice change!

I wasn't there during the UX meeting but I feel like it is very similar to the mockup you made together. However, I don't really like the QR code icon that shows on screen when you try to scan a LAO connection code. It is quite big and thick and it could be slimmer in my opinion.

fe1-web/src/features/home/navigation/HomeNavigation.tsx
Tyratox: Since it now is a stack navigator (i.e. no bottom tab navigation bar), an order would not matter :)

fe1-web/src/features/home/navigation/HomeNavigation.tsx
Tyratox: Thank you 👍

fe1-web/src/features/home/screens/Launch.tsx
Tyratox: Haha me neither! I guess we can remove it then:) (Together with LocalMockServer.ts) But I think I'll do that in a separate PR?

Tyratox:
I agree that the qr code is not ideal. At the moment it is simply an icon scaled up, hence the thick lines. I will try to replace it with something better

Tyratox:
@Xelowak 
This is now the icon used in the mockup. What do you think?
Screenshot 2022-10-11 at 20 45 07

Warhedgehog:
Very nice work ! Sorry for the delay.
A couple things. 

- I know it removes all possibility for confusion but I don't like having "create lao" and "join lao" with Lao in both. Since the title of the screen is LAOs, I wonder if we could assume "create" and "join" would be enough ?
- Also I wonder whether it would be clear to the user that the "logout" concerns the wallet.
- Finally, I like having "join" as a contained button and "create" as an outlined one, as joining should be much more common than creating. Do you have any strong feeling against it ?

Tyratox:
@Warhedgehog Do you prefer something like https://github.com/dedis/popstellar/pull/1275?

fe1-web/src/core/navigation/ScreenOptions.ts
```diff
@@ -0,0 +1,44 @@
+import { BottomTabNavigationOptions } from '@react-navigation/bottom-tabs';
+import { StackNavigationOptions } from '@react-navigation/stack';
+
+import { Color, Spacing, Typography } from 'core/styles';
+
+export const stackScreenOptionsWithoutHeader: StackNavigationOptions = {
+  headerShown: false,
+  // https://github.com/react-navigation/react-navigation/issues/9344#issuecomment-782520454
```
pierluca, line 8: it might be helpful to summarize briefly what this is all about :-)

pierluca (commented):
Pretty good, I'd suggest adding/updating the main fe1 documentation to reflect the updated navigation system.

pierluca (approved):
LGTM 👍

PR #1271, opened 2022-10-11
propose first version of implementation for polymorphic message and data output

tuomas-p:
Implement a polymorphic function in order to check the parameters, as well as output the message and the data from the message


be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/MessageHandler.scala
```diff
@@ -20,10 +20,13 @@ trait MessageHandler extends AskPatternConstants {
     */
   def dbActor: AskableActorRef = DbActor.getInstance
 
-  def checkParameters(rpcRequest: JsonRpcRequest, errorMsg: String): Future[GraphMessage] = {
+  def checkParameters[T](rpcRequest: JsonRpcRequest, errorMsg: String): Future[(GraphMessage, Message, Option[T])] = {
```
Ajkunas, line 23: As I said on slack, for me the function looks good. 
I suggest changing the name, since now the function does not only check, but also extract the message and data. Something like `extractParameters` ?

Ajkunas (changes requested):

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
```diff
@@ -44,9 +44,10 @@ class RollCallHandler(dbRef: => AskableActorRef) extends MessageHandler {
   def handleCreateRollCall(rpcRequest: JsonRpcRequest): GraphMessage = {
     val ask =
       for {
-        _ <- checkParameters(rpcRequest, serverUnexpectedAnswer)
-        message: Message = rpcRequest.getParamsMessage.get
-        data: CreateRollCall = message.decodedData.get.asInstanceOf[CreateRollCall]
+        (_, message, somedata) <- checkParameters[CreateRollCall](rpcRequest, serverUnexpectedAnswer)
+        data: CreateRollCall = somedata.get
```
Ajkunas, line 48: Very good, the code becomes even more concise

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/MessageHandler.scala
tuomas-p: sounds good to me

Ajkunas: Also one more comment, we can do the same for the function `checkLao` in term of extraction of channel

tuomas-p: addressed thanks!

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/MessageHandler.scala
```diff
@@ -20,14 +20,17 @@ trait MessageHandler extends AskPatternConstants {
...
+      case Some(_) => 
+        val message: Message = rpcRequest.getParamsMessage.get
+        val data: T = message.decodedData.get.asInstanceOf[T]
+        Future((Left(rpcRequest), message, Some(data)))
+      case _       => Future((Right(PipelineError(ErrorCodes.SERVER_ERROR.id, errorMsg, rpcRequest.id)), null, None))
     }
   }
 
-  def checkLaoChannel(rpcRequest: JsonRpcRequest, errorMsg: String): Future[GraphMessage] = {
+  def extractLaoChannel(rpcRequest: JsonRpcRequest, errorMsg: String): Future[GraphMessage] = {
```
Ajkunas, line 33: Thank you for the change of name! Also what I was thinking was to also do something similiar as `extractParameters` by extracting the laoChannel and returning it in case there is no error. (sorry it was not clear in the previous comments)
So the signature of the function would be something like this (ofc it is up to you what is better for the implementation): 
`def extractLaoChannel(rpc: JsonRpcRequest, errorMsg: String): Future[(GraphMessage, Option[Hash])]`

Ajkunas (changes requested):
Just a little change, otherwise everything is fine. Let me know when the change is done, I will then approve the PR :)

Ajkunas (dismissed):
Everything looks fine to me :)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
```diff
@@ -43,9 +43,10 @@ class RollCallHandler(dbRef: => AskableActorRef) extends MessageHandler {
   def handleCreateRollCall(rpcRequest: JsonRpcRequest): GraphMessage = {
     val ask =
       for {
-        _ <- checkParameters(rpcRequest, serverUnexpectedAnswer)
-        message: Message = rpcRequest.getParamsMessage.get
-        data: CreateRollCall = message.decodedData.get.asInstanceOf[CreateRollCall]
+        (_, message, somedata) <- extractParameters[CreateRollCall](rpcRequest, serverUnexpectedAnswer)
+        data: CreateRollCall = somedata.get
+        // message: Message = rpcRequest.getParamsMessage.get
+        // data: CreateRollCall = message.decodedData.get.asInstanceOf[CreateRollCall]
```
pierluca, line 49: if you want me to merge this, remove the old code, don't just comment it 😁 
(this applies throughout the file)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
```diff
@@ -67,8 +68,8 @@ class RollCallHandler(dbRef: => AskableActorRef) extends MessageHandler {
   def handleOpenRollCall(rpcRequest: JsonRpcRequest): GraphMessage = {
```
pierluca, line 68: Now that the function is a lot more readable, the logic seems a bit off to me.

Are you sure we're respecting the protocol ?

I may be missing something, but it seems that there's no check that the open message is actually targeting the right id. We're only checking that it's on the right channel. Same thing for the other roll call handlers.

pierluca (commented):
Much cleaner ! Nice !

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
tuomas-p: I believe that the checks are done within the validator, not the handler

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
tuomas-p: this is done in @Ajkunas ' refactoring PR but very good point thank you!

PR #1274, opened 2022-10-18
Restrict asset optimization to resource folder, disable optimization for now

Tyratox:
This PR disables automatic image optimization (as recommended here https://github.com/expo/expo-cli/issues/4459#issuecomment-1227860924) because of an issue that occurs on windows.

For now the script is still available as `npm run optimize-assets` which can be run from linux or mac os. Hence either the images have to be optimized manually (using some external optimizer) or a person running linux or mac os does it once when adding it to the repo. No need to run this script every time we are building the app as it currently prevents builds on windows.


pierluca (approved):
👍

PR #1275, opened 2022-10-18
Change home toolbar based on feedback

Tyratox:
Screenshot 2022-10-18 at 16 00 18

Compare to #1270


fe1-web/src/core/components/PoPButton.tsx
```diff
@@ -20,30 +20,36 @@ const styles = StyleSheet.create({
...
 
   const viewStyles = [styles.button];
-  if (negative && disabled) {
-    viewStyles.push(styles.disabledNegative);
-  } else if (negative) {
-    viewStyles.push(styles.negative);
+  if (negativeBorder && disabled) {
+    viewStyles.push(styles.disabledNegativeBorder);
+  } else if (negativeBorder) {
+    viewStyles.push(styles.negativeBorder);
```
pierluca, line 43: I'm not sure about the use of "negative", is it "negative" as in "negative space" ? 
What's the connotation here ?

fe1-web/src/core/components/PoPButton.tsx
```diff
@@ -54,15 +60,19 @@ const PoPButton = (props: IPropTypes) => {
 const propTypes = {
   onPress: PropTypes.func.isRequired,
   disabled: PropTypes.bool,
-  negative: PropTypes.bool,
+  outline: PropTypes.bool,
+  negativeBorder: PropTypes.bool,
+  margin: PropTypes.bool,
```
pierluca, line 65: This to me is a (growing) code smell. I understand the temptation, but I am concerned by the fact that the APIs of PoPButtons and PoPTexts (or equivalent) tend to grow until they encompass a larger and larger subset of the React button API + stylesheet. It's a constant tendency over this project's lifetime and I'll keep fighting it 😅 

Basically, we need to allow API users to pass in a stylesheet and/or define a set of stylesheets that should carry a semantic meaning in the UI. Look at Bootstrap as an example:
https://getbootstrap.com/docs/4.0/components/buttons/#examples
There's a clear style (whether we like it or not) that defines what "success" should look like, what a primary and secondary button should look like, enabled or disabled, etc.

I'd argue that we need to go in that direction (one small step at a time) rather than increasing the API surface to capture stylesheet concerns. This will make the code and intention clearer, because we won't be thinking about "big button with negative borders" but "primary button", "dialog button", etc.  Do note that if we design the stylesheets properly, they could be combined, e.g. "dialog + primary button"

pierluca (commented):
I quite like the new design, though I'd make sure both buttons are the same size.
However, I've some concerns regarding the changes to the components API.
Let me know what you think 😄

fe1-web/src/core/components/PoPButton.tsx
Tyratox: Negative as in colored background / white foreground rather then black/colored text on white background. I think this is the term that refers to this?

fe1-web/src/core/components/PoPButton.tsx
Tyratox: I 100% agree. Since I wasn't sure whether we will actually use these buttons I added them in a quick and dirty way but in the end we should have something along the lines of primary / secondary button, definitely!

What makes this more difficult in the case if buttons is that it seems we keep adding more different buttons at the moment. I think we now have three buttons: a normal, "primary" button with blue outline, a "secondary" button with just a blue border (the first two are the ones on the screenshot) and then one in case the button is used on colored background (in the wizard/setup phase) where it is almost equal to the primary button except that it also has a white border around it. (Previously we just had two, the primary on white background and the white outlined to be used on colored background)

In addition the standard "primary" button usually always has a margin below it but on the case of the toolbar this messes up the placement.

I am not yet sure what the best react API would look like. I could see us defining separate buttons for the toolbar but they should be the same ones as the others except for the margin. Thus it is also viable to have two buttons, one "primary" / colored one and an outlined one that each have the properties "margin" and "negative" that allows using them in the toolbar / on colored background. Regarding semantic naming it might make sense to rename the margin property to toolbar or something similar.

What are your thoughts? What API do you have in mind?

Tyratox:
Also a quick question regarding the size of the buttons 😅 Do you know of a way to ensure this apart from measuring both, taking the max and setting the width for both? The only other thing I can think of is setting a fixed width for both but I fear that this will break on certain devices with different fonts and different font rendering as text can't really be assumed to have a fixed pixel width 😅

fe1-web/src/core/components/PoPButton.tsx
pierluca: As discussed orally, we'll make a pass on this after the UI has been harmonized.

fe1-web/src/core/components/PoPButton.tsx
pierluca: As discussed, a HTML <button>-like API might be good enough (disabled, style), maybe even exposing a set of predefined styles which could be extended based on the specific needs of the UI (width, height, etc.)

pierluca:
Based on https://reactnative.dev/docs/height-and-width , it seems that fixed size isn't such a bad thing, because it's not pixel-related and stuff should look somewhat consistent across devices. I just wonder if screen's aspect ratio / total size could be an issue... we might need more testing tools for that, but we're not there yet :)

pierluca (approved):

pierluca (approved):
LGTM !

PR #1277, opened 2022-10-19
First draft at karate's CI

GabrielFleischer:
First draft at karate's CI

For now, it will only support backends


jbsv (dismissed):
LGTM, it really does :-D

Warhedgehog (dismissed):
LGTM

.github/workflows/karate.yml
```diff
@@ -0,0 +1,114 @@
+name: Integration Tests
+
+on:
+  schedule:
+    # Run the ci every day at 19:00
+    - cron:  '00 19 * * *'
```
pierluca, line 6: how long does it take ? 
By default I'd strongly suggest running it in the early morning, say 4 am, as it's typical for students to work in the evening and 7pm may run "while" they're merging stuff.

.github/workflows/karate.yml
```diff
@@ -0,0 +1,114 @@
...
+      - name: Set up JDK 11
+        uses: actions/setup-java@v3
+        with:
+          distribution: "zulu"
+          java-version: "11"
+
+      - name: Setup repo
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0 # Shallow clones should be disabled for a better relevancy of Sonar analysis
```
pierluca, line 23: we're not running Sonar, though, right? The comment might be a bit outdated 😁

pierluca (approved):
Minor remarks, but LGTM ! 👍

.github/workflows/karate.yml
```diff
@@ -0,0 +1,114 @@
+name: Integration Tests
+
+on:
+  schedule:
```
pierluca, line 4: also, is this running on the main branch by default ?

.github/workflows/karate.yml
GabrielFleischer: This might be an artifact of a copy-paste 😅

.github/workflows/karate.yml
GabrielFleischer: I have no idea

.github/workflows/karate.yml
GabrielFleischer: It takes around 15 minutes.

I can definitely move it to 4 am.

GabrielFleischer: It seems to be the default behavior, yes :

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule

.github/workflows/karate.yml
GabrielFleischer: It seems to be the default behavior, yes :

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule

pierluca (approved):

PR #1279, opened 2022-10-19
Refactor WitnessHandler file

Ajkunas:
Brief refactoring of the witness handler file


pierluca (commented):
I will let others review it, but it seems pretty good to me.

tuomas-p (approved):
Again clean code, not much to add!

pierluca (approved):
Much better as well, just be mindful of the max line length.

PR #1284, opened 2022-10-24
Improve karate's tests reliability and project's README

GabrielFleischer:
Some tests defined in karate are failing because they are buggy.

This PR tries to fix those issues

It also adds badges and links to the test results in the project's README. Currently, the badge is blank as the action does not exist yet.


.github/workflows/karate/be1-go.yaml
```diff
@@ -1,11 +1,11 @@
-name: Integration Tests
+name: Integration - Go
```
pierluca, line 1: Are you sure you can move this file in a subfolder?

tests/karate/src/test/java/be/constants.feature
```diff
@@ -236,9 +229,9 @@ Feature: Constants
     * def getValidRegisteredVotes = call createValidRegisteredVotes
     * def getInvalidRegisteredVotes = call createInvalidRegisteredVotes
 
-    * def getCoinIssuancePubKey = call createIssuancePubKey
-    * def getCreateSignatureForCoinIssuance = call createSignatureForCoinIssuance
-    * def getValidOutputs =   [{ "value": 32, "script": { "type": "P2PKH", "pubkey_hash": "2jmj7l5rSw0yVb-vlWAYkK-YBwk=" } }]
+    * def getCoinIssuancePubKey = call organizerPk
```
pierluca, line 232: Are you sure about this change? It seems contrary to the spec referenced in the original test implementation 🤔

pierluca (commented):
Thanks for looking into this.
I do have some concerns about the digital cash changes. Can we maybe have a quick chat about those?

tests/karate/src/test/java/be/constants.feature
GabrielFleischer: As discussed, yes

.github/workflows/karate/be1-go.yaml
GabrielFleischer: Well, you're right.

It does not seem to work

pierluca (approved):

Issue #1214, opened 2022-09-22
[BUG] Spacing between buttons in lao info is too big


Tyratox:
### Description (Actual behavior)
Spacing between buttons in lao info seems too big
<img width="406" alt="Screenshot 2022-09-22 at 09 28 29" src="https://user-images.githubusercontent.com/2634739/191685038-233e616c-dd74-4571-9f3e-3219164d94e9.png">

Also the "Add connection" button does not yet have a use case since this is currently not supported by the backends. Probably it is worth removing it. Also see #1118.

### How to reproduce
1. Connect to a lao
2. Click on the circled i in the top navigation bar

### Version & Environment
This bug was reproduced on:
- Commit: f78099a4c7c7c3a58acd71f4fddc67251b35ba53

##### Front-ends:
- [x] Fe1-Web (please include browser's names & version)
- [ ] Fe2-Android (please specify if phone or emulation, and Android version)
- [ ] Not applicable

##### Back-ends:
- [ ] Be1-Go
- [ ] Be2-Scala
- [x] Not applicable

### Impact
- Looks weird
- "Add connection" button is confusing, #1118

Tyratox:
This issue no longer exists as both buttons have been removed

Issue #1213, opened 2022-09-22
[BUG] The "Done" button and the title in modal headers are not aligned


Tyratox:
### Description (Actual behavior)
The "Done" button and the title in modal headers are not aligned

<img width="400" alt="Screenshot 2022-09-22 at 09 14 52" src="https://user-images.githubusercontent.com/2634739/191682655-e75d3883-50d5-49c5-97a1-9ef4fbb65f31.png">

### Expected behavior
The two should be on the same line. Add ["hyphens: auto"](https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens) or something similar (see [compatibility](https://caniuse.com/css-hyphens)) to cover cases of small device widths.

### How to reproduce
1. Connect to a lao
2. Press on the circled i in the top navgiation bar

### Version & Environment
This bug was reproduced on:
- Commit: f78099a4c7c7c3a58acd71f4fddc67251b35ba53

##### Front-ends:
- [x] Fe1-Web (please include browser's names & version)
- [ ] Fe2-Android (please specify if phone or emulation, and Android version)
- [ ] Not applicable

##### Back-ends:
- [ ] Be1-Go
- [ ] Be2-Scala
- [x] Not applicable

### Impact
Looks bad

### Possible root cause
Styling issue

Issue #1215, opened 2022-09-22
[BUG] Overlapping icons


Tyratox:
### Description (Actual behavior)
The qr code icon and circled i icon overlap in the title bar if either the device width is too small / the lao title is too long.

![signal-2022-09-22-093631_002](https://user-images.githubusercontent.com/2634739/191698754-ea2b1f5b-717b-40cc-95bf-b0f8b049e084.jpeg)

### Expected behavior
No overlapping

### How to reproduce
1. Create lao with a long name
2. Reduce the browser window to a minimum

### Version & Environment
This bug was reproduced on:
- Commit: f78099a4c7c7c3a58acd71f4fddc67251b35ba53

##### Front-ends:
- [x] Fe1-Web (please include browser's names & version)
- [ ] Fe2-Android (please specify if phone or emulation, and Android version)
- [ ] Not applicable

##### Back-ends:
- [ ] Be1-Go
- [ ] Be2-Scala
- [x] Not applicable

### Impact
Bad UI and UX, makes it very hard to press the correct button

PR #1285, opened 2022-10-27
Rework LAO Event screen part 1

Tyratox:
This PR is a first step in the direction we want to move with the lao (events) UI. This PR removes / hides the identity screen, moves the button showing the lao connection code to the events screen and puts the information that previously was accessible by clicking on the laos name with the circled i icon (the lao properties) into the same screen where the connection qr code can be found. At the moment I do not like the look of this screen. Partially because there is not enough spacing between the text and qr code but even with that it feels a bit off. The qr code is centered which makes it look better in a browser but on a narrow screen the non-alignment with the rest seems a bit weird to me. Does anybody have an idea on how to improve this?

Moreover the X button for disconnecting from the lao was replaced by a standard back button. Lastly in this PR, the way to access the notification screen was changed. Previously there was a bell button on the identity screen next to the lao connection qr code, now it is placed in the tab navigation bar since there now is space there given the removal of the identity screen.

In a next PR I will work on the past / current / future sections.

Any thoughts? :)

Screenshot 2022-10-27 at 10 51 16

Screenshot 2022-10-30 at 11 07 09


Tyratox:
Now also fixes #1213 by replacing the "Done" with an X and properly aligning it :)

Tyratox:
Also fixes #1214 by removing the two buttons as the first one is not needed yet and the functionality of the second was replicated (pressing the back button in the top navbar does the same)

Tyratox:
Now also takes care of https://github.com/dedis/popstellar/issues/1215 by removing the circled i, making the left and right headers take up as much space as they need, shrink the title if needed and restrict the title to one line (and all ... if it exceeds the available space)

pierluca (approved):
No particular remarks, this looks pretty good already 👍

PR #1288, opened 2022-11-02
Rework LAO Event screen part 2

Tyratox:
This PR separates the upcoming events out to a new screen and fixes various issues with the events screen.

### DIFF
--- a/fe1-web/src/features/events/components/EventList.tsx
+++ b/fe1-web/src/features/events/components/EventList.tsx
@@ -37,7 +37,7 @@ const EventList = () => {
   const eventListSelector = useMemo(() => makeEventListSelector(laoId.valueOf()), [laoId]);
   const events = useSelector(eventListSelector);
 
-  const [{ pastEvents, currentEvents }, setEvents] = useState<{
+  const [{ pastEvents, currentEvents, upcomingEvents }, setEvents] = useState<{
     pastEvents: EventState[];
     currentEvents: EventState[];
     upcomingEvents: EventState[];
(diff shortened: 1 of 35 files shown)


fe1-web/src/features/events/screens/UpcomingEvents.tsx
```diff
@@ -0,0 +1,75 @@
...
+          style={List.accordionItem}
+          content={
+            <ListItem.Content>
+              <ListItem.Title style={[Typography.base, Typography.important]}>
+                {STRINGS.events_list_upcoming}
+              </ListItem.Title>
+            </ListItem.Content>
+          }
+          isExpanded={showUpcoming}
+          onPress={() => setShowUpcoming(!showUpcoming)}>
```
pierluca, line 53: does it make sense to allow the user to hide this ?

fe1-web/src/features/rollCall/objects/RollCall.ts
```diff
@@ -58,7 +58,7 @@ export class RollCall {
   }
 
   public get end() {
-    return this.closedAt || this.proposedEnd;
```
pierluca, line 61: are you sure this won't have side effects? 🤔

fe1-web/src/features/events/screens/UpcomingEvents.tsx
Tyratox: Not it does not, you're right 😂

fe1-web/src/features/rollCall/objects/RollCall.ts
Tyratox: No, I am also a little scared what could break but so far it seems to work pretty well. There was only a few places where I saw this getter being used and there it should be fine..

Tyratox:
Just saw that I need to work on the coverage a bit 😅

fe1-web/src/features/events/components/EventList.tsx
```diff
@@ -1,159 +1,137 @@
...
-            isLastItem={idx === currentEvents.length - 1}
-            testID={`current_event_selector_${idx}`}
-          />
-        ))}
-      </ListItem.Accordion>
-      <ListItem.Accordion
-        containerStyle={List.accordionItem}
-        style={List.accordionItem}
-        content={
+    <View>
```
pierluca, line 60: later on we might want to address the cases when: 
1. all the events are "upcoming"
or
2. there are no events at all

pierluca (approved):

fe1-web/src/features/events/components/EventList.tsx
```diff
@@ -1,159 +1,137 @@
...
-      );
-
-      setEvents({
-        pastEvents: newPastEvents,
-        currentEvents: newCurrentEvents,
-        futureEvents: newFutureEvents,
-      });
-    }, 1000);
+    const interval = setInterval(
+      () => setEvents(categorizeEventsByTime(Timestamp.EpochNow(), events)),
```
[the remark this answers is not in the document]
Tyratox: Yes that is true but I guess for this we would need to compute the minimum time delta of a categorization change of any event? Or do you see a better approach? I do not think we can leverage redux for this since time passes without the state changing?

fe1-web/src/features/events/components/EventList.tsx
```diff
@@ -1,159 +1,137 @@
...
+                key={event.id}
+                eventId={event.id}
+                eventType={event.eventType}
+                isFirstItem={idx === 0}
+                isLastItem={idx === currentEvents.length - 1}
+                testID={`current_event_selector_${idx}`}
+              />
+            ))}
+          </ListItem.Accordion>
+        )}
```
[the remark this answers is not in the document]
Tyratox: Yes I think that would make sense 👍 :)

fe1-web/src/features/events/components/EventList.tsx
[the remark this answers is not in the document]
pierluca: I don't see a better approach. Indeed, we'd have to compute a delta. As long as we don't have 100+ events, we should be fine 😅 I'm happy to make this a priority when we start deploying (soon, hopefully).

pierluca (approved):
LGTM !

PR #1291, opened 2022-11-06
Refactor SocialMediaHandler file

Ajkunas:
Refactoring of the socialMediaHandler: 
- `parametersToBroadcast` function: returns the several inputs needed to broadcast the chirp 
- I could not insert the `dbBroadcast` function inside the `for loop`, I always had an incomplete Future, and therefore some of the tests failed. I tried several things, and could not find any other way to refactor the function by inserting the `dbBroadcast` inside the `for loop`. For me, the actual refactoring seems good. I am open to any suggestions ofc.


be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/SocialMediaHandler.scala
```diff
@@ -37,49 +41,56 @@ class SocialMediaHandler(dbRef: => AskableActorRef) extends MessageHandler {
...
-                val broadcastData: Base64Data = Base64Data.encode(notifyAddChirp.toJson.toString)
-                Await.result(dbBroadcast(rpcMessage, rpcMessage.getParamsChannel, broadcastData, broadcastChannel), duration)
-              case None => Right(PipelineError(ErrorCodes.SERVER_ERROR.id, "Server failed to extract chirp id for the broadcast", rpcMessage.id))
-            }
-          case None => Right(PipelineError(ErrorCodes.SERVER_ERROR.id, "Server failed to extract LAO id for the broadcast", rpcMessage.id))
-        }
-      case error @ Right(_) => error
-      case _                => Right(PipelineError(ErrorCodes.SERVER_ERROR.id, unknownAnswerDatabase, rpcMessage.id))
+        val lao_id: Hash = channelChirp.decodeChannelLaoId.get
+        val broadcastChannel: Channel = generateSocialChannel(lao_id)
```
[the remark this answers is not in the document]
Ajkunas: True!
I created a new function `parametersToBroadcast` that generates the several parameters that need to be broadcasted (or used to broadcast the chirp).

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
```diff
@@ -43,12 +43,9 @@ class RollCallHandler(dbRef: => AskableActorRef) extends MessageHandler {
...
     val ask =
       for {
-        (_, message, somedata) <- extractParameters[CreateRollCall](rpcRequest, serverUnexpectedAnswer)
-        data: CreateRollCall = somedata.get
-        // message: Message = rpcRequest.getParamsMessage.get
-        // data: CreateRollCall = message.decodedData.get.asInstanceOf[CreateRollCall]
+        (_, message, data) <- extractParameters[CreateRollCall](rpcRequest, serverUnexpectedAnswer)
         // we are using the rollcall id instead of the message_id at rollcall creation
-        rollCallChannel: Channel = Channel(s"${Channel.ROOT_CHANNEL_PREFIX}${data.id}")
+        rollCallChannel: Channel = Channel(s"${Channel.ROOT_CHANNEL_PREFIX}${data.get.id}")
```
tuomas-p, line 48: cleaner indeed!

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/SocialMediaHandler.scala
```diff
@@ -93,16 +86,24 @@ class SocialMediaHandler(dbRef: => AskableActorRef) extends MessageHandler {
   }
 
   def handleAddReaction(rpcMessage: JsonRpcRequest): GraphMessage = {
-    writeAndPropagate(rpcMessage)
+    val ask: Future[GraphMessage] = dbAskWritePropagate(rpcMessage)
```
tuomas-p, line 89: Here, I'm not sure whether it is more relevant to keep the Awaits explicit or not tbh

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/SocialMediaHandler.scala
```diff
@@ -37,49 +41,38 @@ class SocialMediaHandler(dbRef: => AskableActorRef) extends MessageHandler {
...
-      case error @ Right(_) => error
-      case _                => Right(PipelineError(ErrorCodes.SERVER_ERROR.id, unknownAnswerDatabase, rpcMessage.id))
+    val ask =
+      for {
+        _ <- extractLaoChannel(rpcMessage, "Server failed to extract LAO id for the broadcast")
+        _ <- extractParameters(rpcMessage, "Server failed to extract chirp id for the broadcast")
+        _ <- dbActor ? DbActor.WriteAndPropagate(rpcMessage.getParamsChannel, rpcMessage.getParamsMessage.get)
+      } yield ()
+
+    Await.ready(ask, duration).value match {
```
tuomas-p, line 69: Even though we still have one chained Await, it's way better and easily readable than before!

tuomas-p (commented):
I'm happy to approve this if we quickly go over the writeAndPropagate removal

pierluca (approved):
Much better than before.

The lines of code seem very long though, to the pointer of hindering a bit the readability (ex. lines 54-56, 72-74). Is there anything we can do to improve that?

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/SocialMediaHandler.scala
```diff
@@ -36,50 +40,46 @@ class SocialMediaHandler(dbRef: => AskableActorRef) extends MessageHandler {
...
+        _ <- extractLaoChannel(rpcMessage, "Server failed to extract LAO id for the broadcast")
+        _ <- extractParameters[AddChirp](rpcMessage, "Server failed to extract chirp id for the broadcast")
+        _ <- dbActor ? DbActor.WriteAndPropagate(rpcMessage.getParamsChannel, rpcMessage.getParamsMessage.get)
+      } yield ()
+
+    Await.ready(ask, duration).value match {
+      case Some(Success(_)) =>
+        val (chirp_id, channelChirp, data, broadcastChannel) = parametersToBroadcast[AddChirp](rpcMessage)
+        val notifyAddChirp: NotifyAddChirp = NotifyAddChirp(chirp_id, channelChirp, data.timestamp)
+        broadcastData[NotifyAddChirp](rpcMessage, notifyAddChirp, channelChirp, broadcastChannel)
```
Ajkunas, line 62: I just implemented this function `broadcastData`, which encodes and broadcast the data. It should be more readable. What do you think?

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/SocialMediaHandler.scala
```diff
@@ -40,6 +40,13 @@ class SocialMediaHandler(dbRef: => AskableActorRef) extends MessageHandler {
 
   private def generateSocialChannel(lao_id: Hash): Channel = Channel(Channel.ROOT_CHANNEL_PREFIX + lao_id + Channel.SOCIAL_MEDIA_CHIRPS_PREFIX)
 
+  // the following function encodes the data and broadcast it
+  private def broadcastData[T](rpcMessage: JsonRpcRequest, data: T, channel: Channel, broadcastChannel: Channel): GraphMessage = {
```
pierluca, line 44: this still seems beyond the 120 chars limit.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/SocialMediaHandler.scala
```diff
@@ -40,6 +40,13 @@ class SocialMediaHandler(dbRef: => AskableActorRef) extends MessageHandler {
...
+    val encodedData: Base64Data = Base64Data.encode(data.toString)
```
pierluca, line 45: in doing this refactoring, we went from `data.toJson.toString` to `data.toString`, was that forgotten or is it intentional ?

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/SocialMediaHandler.scala
pierluca: It's becoming more concise and the intention remains pretty clear, so that's good.
However, it does add a level of indirection and hides where the `Await.result` happens.

I'm wondering: the new `broadcastData` function doesn't seem to be doing much. Would it make sense to change `dbBroadcast` to become a generic (i.e., `dbBroadcast[T]`) and move the `Base64Data.encode` there ? 
Then we'd only have to keep the `Await.result`.

Ajkunas: Yes that is true that it only moves the `Await.result`, thus it is not really useful. I removed it. 

I took in account your remark and I found out that the function does not need to be generic. To move the `Base64Data.encode` into the `dbBroadcast` function, we simply need to take a `String` instead of a `Hash` as a parameter for the `broadcastData`. I think that it is the maximum generalization that we can have for this function (for now on).

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/MessageHandler.scala
```diff
@@ -92,7 +91,7 @@ trait MessageHandler extends AskPatternConstants {
     * @return
     *   the database answer wrapped in a [[scala.concurrent.Future]]
     */
-  def dbBroadcast(rpcMessage: JsonRpcRequest, channel: Channel, broadcastData: Base64Data, broadcastChannel: Channel): Future[GraphMessage] = {
+  def dbBroadcast(rpcMessage: JsonRpcRequest, channel: Channel, broadcastData: String, broadcastChannel: Channel): Future[GraphMessage] = {
```
pierluca, line 94: Since the `String` type here allows to pass garbage in this function much more easily (even if only by mistake), and that every caller pretty much does `.toJson.toString` , would it make sense to take a JSON as parameter instead ?

pierluca (commented):
It's very nice to see the progress. I'm feeling a bit ambivalent about how `dbBroadcast` could be misused. Do you have any opinion @RaulinN ?

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/MessageHandler.scala
Ajkunas: I guess we can also do that. Indeed, we always call `.toString` after the JSON encoding of the data. 
It is a way to generalize the `dbBroadcast` function. 
We can also still stick with the initial implementation where the `Base64.encode` is outside of the function.
We have two options. I do not really see a more favourable one, except the generalization argument.

Ajkunas: I see your point about passing something wrong to the function. The `Base64.encode` did not ensure anything more reliable as well (with the previous implementation of the function, where the function took a Base64 Hash instead of a String or JSON value), we could have also passed anything wrong that was simply encoded I think.

pierluca (approved):
LGTM !

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/MessageHandler.scala
pierluca: I agree with you, we didn't ensure anything more coherent. However, the intention of what was supposed to go in there was a little bit more obvious (and it is again with the `JsValue`). Basically, type safety wasn't so much the key point as much as giving a good `tip` to the developer as to what was expected 😁

Issue #1221, opened 2022-09-22
[BUG] social media feature crashes


Tyratox:
### Description (Actual behavior)
After having posted several chirps (not yet sure whether this is actually required, this is just how we observed it during the meeting on the 22nd of September), reloading the website in the browser, re-connecting to the lao it was no longer possible to go to the social media tab. The website simply crashed.

### How to reproduce
To be specified

### Version & Environment
This bug was reproduced on:
- Commit: f78099a4c7c7c3a58acd71f4fddc67251b35ba53

##### Front-ends:
- [x] Fe1-Web (please include browser's names & version)
- [ ] Fe2-Android (please specify if phone or emulation, and Android version)
- [ ] Not applicable

##### Back-ends:
- [ ] Be1-Go
- [ ] Be2-Scala
- [x] Not applicable

### Impact
The social media feature can no longer be used

Tyratox:
Was fixed in an earlier PR, probably #1293

PR #1293, opened 2022-11-14
Offline mode

Tyratox:
This PR adds an offline mode which allows user to access previously downloaded LAO content read-only.

When connecting after tapping on a lao item on the home screen fails, the application automatically enters this new offline mode, shows a toast to the user indicating this and shows a orange "offline" banner at the bottom of the screen while in offline mode. All buttons triggering network requests are disabled.

This new feature allows accessing past election results, roll call members or digital cash balance without requiring network connection or while the lao server is down. It is a more general form of what we had before, just the wallet being accessible without connecting to a lao.

Any thoughts or comments on this?

Screenshot 2022-11-14 at 09 46 06

Screenshot 2022-11-14 at 09 46 14

Screenshot 2022-11-14 at 09 46 19

Screenshot 2022-11-14 at 09 46 25


fe1-web/src/features/digital-cash/components/TransactionHistory.tsx
```diff
@@ -191,7 +192,8 @@ const TransactionHistory = ({ laoId }: IPropTypes) => {
 };
 
 const propTypes = {
-  laoId: PropTypes.string.isRequired,
+  laoId: PropTypes.oneOfType([PropTypes.instanceOf(Hash).isRequired, PropTypes.string.isRequired])
```
pierluca, line 195: It makes me slightly uncomfortable to relax the typing on the properties, but it's good that we're moving everything to tolerate `Hash` as an input.

fe1-web/src/features/digital-cash/hooks/DigitalCashHooks.ts
```diff
@@ -23,26 +23,32 @@ export namespace DigitalCashHooks {
   };
 
   /**
-   * Gets the current lao id
+   * Gets the current lao id, throws an error if there is none
    */
-  export const useCurrentLaoId = () => useDigitalCashContext().useCurrentLaoId();
+  export const useAssertCurrentLaoId = () => useDigitalCashContext().useAssertCurrentLaoId();
```
pierluca, line 28: I seem to recall we discussed a lot about the naming of `useAssertCurrentLaoId`, I'm still not convinced, but I'm very frustrated that I can't think of a better name ... the only alternative I have in mind is `useCurrentLaoIdUnsafe`😅

fe1-web/src/features/rollCall/screens/ViewSingleRollCall.tsx
```diff
@@ -39,22 +39,38 @@ const ViewSingleRollCall = () => {
...
+          isOrganizer={isOrganizer}
+        />
+      );
     case RollCallStatus.REOPENED:
     case RollCallStatus.OPENED:
       return (
         <RollCallOpen
           rollCall={rollCall}
           laoId={laoId}
+          isConnected={isConnected}
```
pierluca, line 60: should this be a parameter? shouldn't we just use the hook within those components? 🤔

fe1-web/src/features/social/screens/SocialHome.tsx
```diff
@@ -79,8 +82,7 @@ const SocialHome = () => {
             value={inputChirp}
             onChangeText={setInputChirp}
             onPress={publishChirp}
-            // The publish button is disabled when the user public key is not defined
-            publishIsDisabledCond={!currentUserPopTokenPublicKey}
+            publishIsDisabledCond={publishDisabled}
```
pierluca, line 85: that's a weird attribute name, which causes an even weirder logic.
It might make sense to flip it to `publishEnabled` or at least rename it as `publishDisabled`

fe1-web/src/features/digital-cash/components/TransactionHistory.tsx
Tyratox: Hmm yes. But I think Hash is probably the way to go for the laoId so I guess we we can also do a refactor that drops support for strings? Should be easier now :)

fe1-web/src/features/digital-cash/hooks/DigitalCashHooks.ts
Tyratox: Haha yeah me neither 😂 Honestly I think `useCurrentLaoIdUnsafe` is not much better as `unsafe` usually implies that you should not use it whereas I would argue in this case we should actually use it (almost) everywhere. Maybe it is even possible to actually use it everywhere now given that we have a clear separation of screens there we want to assert the existence of a current lao and screens where it will always be undefined. The error thrown by the function is caught be the error boundary that then makes the application navigate to the home screen.

fe1-web/src/features/rollCall/screens/ViewSingleRollCall.tsx
Tyratox: That is a good question but also applies to the other props here, especially laoId. All of them can be retrieved from the global application state but we can also pass them using props. At the time I am writing this I do not see a big advantage for either way. What do you think?

fe1-web/src/features/social/components/ChirpCard.tsx
```diff
@@ -103,6 +105,10 @@ const ChirpCard = (props: IPropTypes) => {
   const [deleteModalIsVisible, setDeleteModalIsVisible] = useState(false);
 
   const addReaction = (reaction_codepoint: string) => {
+    if (!isConnected) {
+      return;
```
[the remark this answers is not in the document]
Tyratox: True!

Oh interesting idea with the message queue :)

fe1-web/src/features/social/screens/SocialHome.tsx
Tyratox: Yes, I thought that too! Most definitely was on my todo list for refactoring haha :)

fe1-web/src/resources/strings.ts
```diff
@@ -95,9 +96,6 @@ namespace STRINGS {
 
   export const navigation_wallet_digital_cash_wallet_scanner = 'Digital Cash Wallet Scanner';
 
-  export const navigation_wallet_digital_cash_transaction_history =
-    'Digital Cash Transaction History';
```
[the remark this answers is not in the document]
Tyratox: Yes it was unused. I assume we wanted to have a separate screen for this at some point but this did not end up happening.

The strings.ts file is prone to containing unused strings. We should write a check for that at some point

fe1-web/src/features/digital-cash/hooks/DigitalCashHooks.ts
Tyratox: If we use it everywhere we could even rename it to just `useCurrentLaoId` and remove the old function named like this?

fe1-web/src/features/social/components/ChirpCard.tsx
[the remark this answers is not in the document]
Tyratox: While doing this change I noticed a bug in the social media feature: The currentUserPublicKey could be undefined but wasn't typed like this which most likely lead to https://github.com/dedis/popstellar/issues/1221. I fixed this in this PR as well so it got a little bit bigger, sorry for that!

fe1-web/src/features/digital-cash/hooks/DigitalCashHooks.ts
pierluca: Yes, since my comment above, my thinking has evolved exactly along those lines. 😁

fe1-web/src/features/social/components/ChirpCard.tsx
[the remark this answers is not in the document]
pierluca: Fair enough. Although it would have been nice for #1221 to be fixed in a separate PR 😅

fe1-web/src/features/digital-cash/components/TransactionHistory.tsx
pierluca: Yes, in principle moving everything to Hash would be nice.

However, I also remember students struggled in the past with Hash's though, due to a couple of things:
1. they're objects, so equality comparisons do not work as expected
2. when rehydrating data from redux, it's easy to forget rehydrating a string into a Hash 

What do you think ?

fe1-web/src/features/social/navigation/SocialMediaNavigation.tsx
```diff
@@ -47,7 +47,7 @@ const iconSelector =
  * This class manages the social media navigation and creates the corresponding navigation bar.
  */
 const SocialMediaNavigation = () => {
-  const [currentUserPopTokenPublicKey, setCurrentUserPopTokenPublicKey] = useState(
+  const [currentUserPublicKey, setCurrentUserPublicKey] = useState(
```
pierluca, line 50: could we rename this to clarify that we're talking about PoP tokens and not the (constant) public key ?
maybe `currentPopToken` would be enough ?

I seem to recall the long name, at the time, was because @Xelowak felt that `PopToken` would imply having the private key there as well. That could still be a valid point. Just giving you the history 😄

pierluca (commented):
Minor remarks, looks pretty good

fe1-web/src/features/digital-cash/components/TransactionHistory.tsx
Tyratox: Hmm.. I can definitely see how 1) causes an issue but (unfortunately) this applies to many objects and should probably be learned at some point? I mean it already starts with `({x: 1}) == ({x: 1})`

And for 2) I see the problem too but I guess it should be noticeable if a lot of `new Hash(` appear in the code (assuming we changed the interfaces to only return Hashes) as the other components / functions expect it to be of this type.

I think doing a refactor of this should be doable but I'll do it in a separate PR :)

fe1-web/src/features/digital-cash/hooks/DigitalCashHooks.ts
Tyratox: 👍😁

Tyratox: Also separate PR I assume 😂

fe1-web/src/features/social/components/ChirpCard.tsx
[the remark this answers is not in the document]
Tyratox: Yeah I agree.. I actually wanted to but realized I could not test these changes without it being fixed.. 😬

fe1-web/src/features/rollCall/screens/ViewSingleRollCall.tsx
pierluca: In terms of examples in the codebase, it might make sense to use the hook for `isConnected`. While the other params make sense for a roll call (conceptually), `isConnected` feels wrong in the API, it shouldn't be an external concern for a roll call.

Actually, I was even thinking, should we have the `isConnected` hook (almost) everywhere or could we put it in a global context ?

fe1-web/src/features/social/navigation/SocialMediaNavigation.tsx
pierluca: Either of those works for me, whichever you think is better 😅

fe1-web/src/features/digital-cash/components/TransactionHistory.tsx
Tyratox: Yes, that it should for sure!

fe1-web/src/features/rollCall/screens/ViewSingleRollCall.tsx
Tyratox: It kind of already is, it is in the feature context provided by the lao feature. Putting it in a separate context comes with the question where. Only the lao feature knows how to retrieve this information from redux. Hence we would have to put it in the LaoNavigator component wrapping all lao screens and export the context from the lao feature and all other features would then need to import the context. If we only put the `isConnected` hook there this seems to be a bit of an overkill but now when thinking about it, I think we could put other things in there too such as the laoId. For instance we could get rid of the `getCurrentLaoId` completely and would reduce the number of feature in- and exports :)

Tyratox:
A few more changes are needed to fix the issues discovered this morning

fe1-web/src/features/rollCall/screens/ViewSingleRollCall.tsx
pierluca: Oh right, I forgot about that implementation aspect. Thanks for the reminder. I see the point you're making.

It does make sense, but at this stage, where the system as a whole is becoming pretty stable, I'm not sure we'd want to do yet another far-reaching change as opposed to stabilizing what we already have. What do you think?

pierluca (approved):

pierluca:
@Tyratox are you okay to merge as it is, or is the PR too small for your standards? 😜

Tyratox:
Haha 😂😜

PR #1296, opened 2022-11-17
Add Logs to the connection object

GabrielFleischer:
In the last meeting, we had issues receiving messages from the server.

The problem is that there are not enough logs related to the connection object.

This tiny PR adds **a lot** of logs in the connection object. That way, we will be able to pinpoint the cause with ease.


Warhedgehog (dismissed):
Good addition 👍
Just wondering if it might make more sense to log those in info rather than debug, because it will add quite some logs I imagine

fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/remote/Connection.java
```diff
@@ -1,31 +1,65 @@
...
     // subject
-    this.disposable =
-        laoService.observeMessage().subscribe(messagesSubject::onNext, messagesSubject::onError);
+    this.disposables.add(
+        laoService
+            .observeMessage()
+            .doOnNext(msg -> Log.d(TAG, "Received a new message from remote: " + msg))
+            .subscribe(messagesSubject::onNext, messagesSubject::onError));
+
+    // Add logs on connection stated
```
jbsv, line 36: typo? 
```suggestion
    // Add logs on started connection
```

fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/remote/Connection.java
```diff
@@ -39,7 +73,8 @@ public Observable<WebSocket.Event> observeConnectionEvents() {
   }
 
   public void close() {
-    disposable.dispose();
+    disposables.dispose();
+    messagesSubject.onComplete();
```
jbsv, line 77: please add a short comment for its purpose

jbsv (changes requested):

fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/remote/Connection.java
GabrielFleischer: added

fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/remote/Connection.java
GabrielFleischer: fixed

jbsv (approved):

pierluca:
I've requested a re-review as the CI seems to fail to unrelated changes and the issues seem to have been addressed.

Warhedgehog (approved):
LGTM !

PR #1299, opened 2022-11-20
Refactor Roll Call Creation

Warhedgehog:
This PR fixes the behaviour  of the dates. New behaviour is:

- On opening rc creation screen, suggested date and time is today from now to now + 1 hour
- If the user clicks on start date, both start and end date get deleted
- If the user clicks on end date, only the end date gets deleted
- If the user clicks on start time, both start and end time get deleted
- If the user clicks on end time, only the end time gets deleted

This PR also does minor rework of the UI of the roll call creation screen and adds the location field, which was statically set to Lausanne for some reason

rc_creation


pierluca:
Just a tiny question based on the screenshot: is it normal that the time fields are not aligned with the date fields ? 
I'm also wondering if we shouldn't add a bit of (horizontal) spacing between the start and end fields in general.

Warhedgehog:
> Just a tiny question based on the screenshot: is it normal that the time fields are not aligned with the date fields ? I'm also wondering if we shouldn't add a bit of (horizontal) spacing between the start and end fields in general.

With the current implementation it would require to redo the date and time selection components (and probably merge them in a single one with both). It would not be a much work, but it's not a one line tweak either. 
Moreover I'm not sure it would look good since they have different size. We could make the time fields as big as the date fields but I don't know any elegant way to do that (One can always set a minimum width). 

I'll try several spacing so that we can compare :)

Warhedgehog:
8dp :


12dp


16dp


20dp

Warhedgehog:
I think 16dp, what do you think @pierluca ?

GabrielFleischer (dismissed):
LGTM

jbsv (approved):
LGTM

PR #1301, opened 2022-11-22
More consistent typing

Tyratox:
Yes this is one of these big PRs sorry..

With this PR I changed Ids from `string` to `Hash` in all places I found. Moreover `Hash`, `PublicKey` etc. all now have a new method `toState()` similar to `RollCall`, `Election` etc. which allows `toState()` to be called recursively. In combination with the new `fromState()` static method this would also allow the serialization type to be changed in the future. `toString()` should be used for some string representation of the value, `valueOf()` to access the underlying primitive and `toState()` for serialization.

While at it, I added some new classes such as `Question`, `Vote`, `EncryptedVote` all with `toState()` and `fromState()` methods to make sure this pattern is consistently used.

In the case of notifications this (unfortunately) required further changes as before the feature operated exclusively on the notification states and not on class instances.


fe1-web/src/__tests__/utils/TestUtils.ts
```diff
@@ -35,23 +35,25 @@ export const org = new PublicKey(mockPublicKey);
 // MOCK LAO
 export const mockLaoName = 'MyLao';
 export const mockLaoCreationTime = new Timestamp(1600000000);
-export const mockLaoIdHash: Hash = Hash.fromStringArray(
+export const mockLaoId: Hash = Hash.fromStringArray(
```
Tyratox, line 38: Given that now "ids" should be of type `Hash`, I updated the name of the variables to reflect this

fe1-web/src/core/network/ingestion/ExtendedMessage.ts
```diff
@@ -1,28 +1,34 @@
...
+  laoId: HashState;
 
-  data: string;
-  sender: string;
-  signature: string;
-  message_id: string;
+  data: Base64UrlDataState;
+  sender: PublicKeyState;
+  signature: SignatureState;
+  message_id: HashState;
```
Tyratox, line 31: I added type aliases such as `type TimestampState = number` to make it easier to change and also to make the respective "unserialized" types more visible in these type definitions.

fe1-web/src/core/network/ingestion/ExtendedMessage.ts
```diff
@@ -113,6 +119,22 @@ export class ExtendedMessage extends Message implements ProcessableMessage {
   }
 
   public toState(): ExtendedMessageState {
-    return JSON.parse(JSON.stringify(this));
```
Tyratox, line 116: Explicit return values are most likely easier to understand and are also needed given the new `toState()` methods

fe1-web/src/core/network/ingestion/MessageReducer.ts
```diff
@@ -33,7 +33,7 @@ const messagesSlice = createSlice({
   reducers: {
     // Add a Message to the list of known Messages
     addMessages: {
-      prepare(messages: ExtendedMessageState | ExtendedMessageState[]): any {
```
Tyratox, line 36: I tried to add `prepare` methods in most places that take care of the conversion of class instances to the respective state values

fe1-web/src/features/lao/objects/__tests__/Server.test.ts
```diff
@@ -1,96 +0,0 @@
-import {
-  mockAddress,
-  mockKeyPair,
-  mockLaoId,
-  mockLaoIdHash,
-  mockPopToken,
-  mockPublicKey,
-  mockPublicKey2,
-} from '__tests__/utils';
```
Tyratox, line 9: Somehow this file was still in the codebase, was replaced by `LaoServer` a while ago.

fe1-web/src/core/objects/Hash.ts
```diff
@@ -78,4 +80,37 @@ export class Hash extends Base64UrlData {
...
+    return super.valueOf();
+  }
+
+  /**
+   * Returns the serialized version of the hash that can for instance be stored
+   * in redux stores
+   */
+  public toState(): HashState {
+    return super.valueOf();
+  }
```
pierluca, line 108: aren't you redefining / hiding the inherited methods here, whilst providing an identical implementation ?

fe1-web/src/core/objects/Hash.ts
```diff
@@ -78,4 +80,37 @@ export class Hash extends Base64UrlData {
...
+   */
+  public valueOf(): string {
+    return super.valueOf();
+  }
+
+  /**
+   * Returns the serialized version of the hash that can for instance be stored
+   * in redux stores
+   */
+  public toState(): HashState {
```
pierluca, line 106: I do agree that it gives flexibility, but I don't see the state fundamentally changing, given the `Hash` itself is a fancy encapsulation of a string. 
It seems that `string` here would make it inheritable (one to one) from Base64UrlData

fe1-web/src/core/objects/PrivateKey.ts
```diff
@@ -13,4 +13,13 @@ export class PrivateKey extends Base64UrlData {
...
   }
+
+  /**
+   * Returns the serialized version of the private key that can for instance be stored
+   * in redux stores
+   * @returns The serialized private key
+   */
+  toState(): string {
+    return this.toString();
+  }
```
pierluca, line 24: maybe dumb question, no `fromState` here?

fe1-web/src/core/objects/Base64UrlData.ts
```diff
@@ -43,4 +45,37 @@ export class Base64UrlData extends String {
...
+  }
+
+  /**
+   * Returns the primitive value used for representing the Base64UrlData,
+   * a string
+   * If you want to serialize an instance use .toState() instead
+   */
+  public valueOf(): string {
+    return super.valueOf();
+  }
```
pierluca, line 65: this is already defined in `String`, that we inherit. What's the advantage of re-implementing it?

fe1-web/src/core/objects/Base64UrlData.ts
```diff
@@ -43,4 +45,37 @@ export class Base64UrlData extends String {
...
   }
+
+  /**
+   * Returns *some* string representation of this object.
+   * If you need access to the unterlying data type use .valueOf() and
+   * if you want to serialize an instance use .toState() instead
+   */
+  public toString(): string {
+    return super.toString();
+  }
```
pierluca, line 56: same question

fe1-web/src/core/objects/Timestamp.ts
```diff
@@ -88,4 +90,37 @@ export class Timestamp extends Number implements Number {
...
+    return super.valueOf();
+  }
+
+  /**
+   * Returns the serialized version of the base64url that can for instance be stored
+   * in redux stores
+   */
+  public toState(): TimestampState {
+    return super.valueOf();
+  }
```
pierluca, line 118: In this particular case, though, I fully agree with implementing the `TimestampState`.
The number is indeed an arbitrary representation of a timestamp, and we could instead use ISO-8601.

fe1-web/src/core/objects/Timestamp.ts
```diff
@@ -88,4 +90,37 @@ export class Timestamp extends Number implements Number {
...
+
+  /**
+   * Returns the primitive value used for representing the Timestamp,
+   * the number of seconds that passed since 00:00:00 UTC
+   * If you want to serialize an instance use .toState() instead
+   */
+  public valueOf(): number {
+    return super.valueOf();
+  }
```
pierluca, line 110: precisely because the underlying representation is arbitrary, and we could instead use a string formatted according to ISO 8601, I would avoid exposing it... unless there's a strong need to? 

`toState` seems like a good way to hide the actual implementation details.

fe1-web/src/core/objects/Timestamp.ts
```diff
@@ -88,4 +90,37 @@ export class Timestamp extends Number implements Number {
...
   }
+
+  /**
+   * Returns *some* string representation of this object.
+   * If you need access to the unterlying data type use .valueOf() and
+   * if you want to serialize an instance use .toState() instead
+   */
+  public toString(): string {
+    return super.toString();
+  }
```
pierluca, line 101: if we want a `toString` representation, I'd be almost tempted to make it an ISO-8601 representation 😅

fe1-web/src/core/objects/Hash.ts
```diff
@@ -1,8 +1,10 @@
 import { sha256 } from 'js-sha256';
 
-import { Base64UrlData } from './Base64Url';
+import { Base64UrlData } from './Base64UrlData';
 import { PublicKey } from './PublicKey';
 
+export type HashState = string;
+
```
pierluca, line 7: comment for lower in this file: 
if we change the implementation of `fromStringArray` to take a `String[]` input instead of a `string[]` input, we could now do a `.valueOf()` inside that function and avoid doing it everywhere else in the application :)

pierluca (commented):
The review is so big that I can't go through everything manually in a short time frame.
As such, I've done more of a random sampling and a few targeted file reviews.

I'm not convinced by some of the changes, while others seem very relevant.
Let me know what you think.

fe1-web/src/core/network/ingestion/ExtendedMessage.ts
Tyratox: I agree that it is not clear what the underlying type is but this is what partially motivated the change as it is no longer necessary to know how the different objects are actually stored. The only 2 functions that you need to know that are `fromState` and `toState`. As long as the underlying storage type is not changed it is possible to abstract and completely ignore the actual value of `XXState`. When writing code it should suffice that there is some storage type and I have access to a function that creates an instance from this type (`fromState`) and a function that transforms an instance to this type (`toState`).

My hope is that this also reduces the likelihood of people using the `XXState` anywhere in the application again (as it was the case for `Hash / string`). To me this more clearly indicates that it should only be used for storage, i.e. it is not the "proper" type.

But this of course is a very subjective interpretation and if this is not what other people think when looking at it we should think of a better solution / revert this change!

fe1-web/src/core/network/ingestion/MessageReducer.ts
```diff
@@ -56,10 +56,10 @@ const messagesSlice = createSlice({
 
     // Remove a Message to the list of unprocessed Messages
     processMessages: {
-      prepare(messageIds: String | String[]): any {
+      prepare(messageIds: Hash | Hash[]): any {
```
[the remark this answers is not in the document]
Tyratox: Yes, definitely!

fe1-web/src/core/network/ingestion/__tests__/MessageReducer.test.ts
```diff
@@ -70,18 +70,19 @@ describe('MessageReducer', () => {
 
   it('should process the message', async () => {
     const extMsg = createExtendedMessage();
-    const msgId = extMsg.message_id.toString();
+    const msgId = extMsg.message_id;
+    const serializedMsgId = msgId.toString();
```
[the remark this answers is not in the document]
Tyratox: Just missed this. There were just too many of these 😂

fe1-web/src/core/objects/Hash.ts
Tyratox: Mostly in order to add the comments that describe the purpose of the different functions. But to make this more explicit and hopefully avoid misuse I could also see `toString` to be changed for most of these to something like ```
```
return `Hash(${super.toString()})`;
```

or as you commented for timstamps the ISO date

fe1-web/src/core/objects/Base64UrlData.ts
Tyratox: Only the minor advantage of having a comment indicating the usage and it would allow us to change it to "Base64Url(...)" which might underline the different use cases for the functions?

Tyratox: While for Hash and Base64Url it does not make a difference I think it reduces the mental load when reading and writing code when we try to consistently use the appropriate variants as for other data types there are real differences between the functions.

fe1-web/src/core/objects/Timestamp.ts
Tyratox: Yes, I will do that actually :)

fe1-web/src/core/objects/Timestamp.ts
Tyratox: Yes but how should we do that? Have a `valueOf()` that returns null?

Tyratox: (Otherwise the `valueOf()` function is still inherited)

fe1-web/src/core/objects/Timestamp.ts
Tyratox: 👍👍👍😁

fe1-web/src/core/objects/Hash.ts
Tyratox: Oh wow this is great, love it! 🥳😁

fe1-web/src/core/objects/Timestamp.ts
Tyratox: Okay just realized that this results in a problem with a lot of hash computations which would have to be changed to something else 😂 Hence we still need access to `valueOf()` in one way or another 🤔

Tyratox: fromStringArray() could for instance also allow Timestamps and convert them using `valueOf()` ..? Probably the cleanest solution?

fe1-web/src/core/network/ingestion/ExtendedMessage.ts
pierluca: Actually, I find your argument convincing 👍

fe1-web/src/core/objects/Hash.ts
pierluca: I like your suggestion, but I'm concerned that there are still cases where we use `toString` instead of `valueOf` 🤔 
Don't you think this is typical one of those changes that risks breaking stuff everywhere? 😅

fe1-web/src/core/objects/Base64UrlData.ts
pierluca: Fair enough, I can see the argument.
I'm however slightly concerned by the ease with which `toString` is called implicitly by javascript and might fail in subtle ways. If we're comparing a Hash and a Base64Url data , and they contain the same thing, both `toString` should return the same string... otherwise some comparisons might fail.
See https://stackoverflow.com/questions/50036448/when-is-tostring-get-called-implicitly-in-javascript for an example (agreed: strict equalities should be preferred, but this codebase is worked on by many students...).

fe1-web/src/core/objects/Timestamp.ts
pierluca: Yup, bad idea of mine. Nevermind ^^

pierluca (commented):
I've answered your remarks. The changes so far LGTM.

fe1-web/src/core/objects/Hash.ts
Tyratox: Yeah... Usually I would like argue it would only break in places that would require fixing since it will just fail at a later point but given the current state I think I agree that we might not want to do that right now 😅😅

fe1-web/src/core/objects/Base64UrlData.ts
Tyratox: Uhhhhh.. I did not know about `==` calling `toString()`, that is terrible and definitely good to know, thank you 😂 I mean for the specific case of comparisons we fortunately already have  a linter rule forbidding it (https://eslint.org/docs/latest/rules/eqeqeq) but I can definitely see your point! I'll remove the `toString()` methods for the classes inheriting from String 👍

pierluca: Agreed. JavaScript is one backstabbing son of a bad programming language 😅 

If you want some pretty good (and hilarious) examples of JavaScript's beautiful counter-intuitiveness ... https://www.destroyallsoftware.com/talks/wat

pierluca (approved):

fe1-web/src/core/objects/Base64UrlData.ts
Tyratox: Haha hilarious indeed, thank you 🤣

PR #1303, opened 2022-11-29
Workaround for react-camera issue

Tyratox:
This PR fixes the issue discovered today where no more than 2 attendees can be added to a roll call. The problem is actually not directly our code but rather a bug in `react-camera`.

It seems as if the barcodescanner handle function is only set initially by expo-camera and if the passed prop changes,  this is ignored. Hence variables such as the current list of attendees referenced by the function won‘t update. You can see the problem in this snack: https://snack.expo.dev/0MS_5nPEp. Observer the counter value after scanning a few qr codes. Previously this problem did not occur since `setState(oldState => …)` was used where `oldState` received in the callback is always the latest and the reference to `setState` did not change. This PR works around this issue by additionally using a mutable reference that allows for proper duplicate checking which most likely was broken since the migration to `react-camera`.

### DIFF
--- a/fe1-web/src/core/components/QrCodeScanner.tsx
+++ b/fe1-web/src/core/components/QrCodeScanner.tsx
@@ -50,6 +50,11 @@ const styles = StyleSheet.create({
   },
 });
 
+/**
+ * Attention: react-camera currently has the problem that the onBarCodeScanned property is only set initially, i.e
+ * the function cannot be changed later meaning if we pass a function that references other variables, they will
+ * only have their proper values if they are mutated which in general should be avoided.
+ */
 const QrCodeScanner = ({ showCamera, children, handleScan }: IPropTypes) => {
   const [permission, requestPermission] = Camera.useCameraPermissions();
   const [cameraType, setCameraType] = useState<CameraType>(CameraType.back);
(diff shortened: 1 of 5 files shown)


fe1-web/src/core/components/QrCodeScanner.tsx
```diff
@@ -50,6 +50,11 @@ const styles = StyleSheet.create({
   },
 });
 
+/**
+ * Attention: react-camera currently has the problem that the onBarCodeScanned property is only set initially, i.e
```
pierluca, line 54: I'd suggest adding a date on this comment and a link to the GitHub issue describing this problem

fe1-web/src/features/rollCall/screens/RollCallScanner.tsx
```diff
@@ -63,6 +63,10 @@ const RollCallOpened = () => {
 
   // this is needed as otherwise the camera may stay turned on
   const [showScanner, setShowScanner] = useState(false);
+  // use a mutable ref to avoid the problem that QrCodeScanner's handle scan function
+  // is never changed, i.e. the bound variables stay the same during the scanner's lifetime
+  // this is due to a react-camera bug and is hopefully fixed at some point in the future
+  const attendeePopTokens = useRef(navigationAttendeePopTokens);
```
pierluca, line 69: does this behave correctly when switching between different roll calls, assuming multiple roll calls exist in the LAO?

Tyratox: As far as I can tell it does, yes. Tested having 2 open roll calls, switching between them and closing the one created 2nd and this should be the only reachable scenario. Closing a non-latest roll call is impossible and you can only get to this screen when a roll call is open.

pierluca (approved):

PR #1307, opened 2022-12-06
Patch dependency to allow black drawer overlay to be pressed to close the drawer

Tyratox:
Not sure if we want to take this route but I thought I'd put it up for discussion


pierluca:
As per our discussion on Slack, this is an interesting approach. I'd just make sure that it's suitably "noisy" and that it doesn't subtly break things when we upgrade. I'm okay with a hack failing with a (massive) bang.
On the other hand, I'd document the hell out of this so that if/when it goes out with a bang, we can find a suitable replacement if a hack is still needed.

pierluca (approved):

pierluca (approved):

PR #1308, opened 2022-12-06
Update dependencies

Tyratox:
My hope was that updating the dependencies would fix the issue we have with reanimated / react-navigation (i.e. the drawer cannot be closed by clicking on the black overlay) and/or the issue with react-camera (https://github.com/dedis/popstellar/pull/1303) but unfortunately neither was fixed. Still, upgrading our dependencies does not hurt and will only become harder the longer we don't do it.


pierluca (commented):
Overalls this looks very good and I'd love to be able to merge this.
Given there are blocking upstream issues, can you link them here so that we can monitor them and update this PR when they get solved ?

Also, do you know if this would (finally) work with node 18.x ?

fe1-web/package.json
```diff
@@ -130,7 +130,7 @@
     "sharp-cli": "^2.1.0",
     "ts-jest": "^26.5.6",
     "typedoc": "^0.23.14",
-    "typescript": "^4.8.3",
```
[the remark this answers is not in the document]
Tyratox: Can't remember how or why that happened, fixed it in a new commit.

Tyratox:
Oh that is actually a very good point, it is worth checking whether the old node version is the culprit especially given this comment: https://github.com/facebook/jest/issues/9324#issuecomment-1304780623

I wanted to link this issue https://github.com/facebook/jest/issues/9324 just now and saw this comment which I apparently skipped up to now. There are so many solution suggestions in the linked issue 😂 But trying node 18 is definitely worth a shot!

Tyratox:
Unfortunately also fails on node 18 and 19 :(

But yes, without the jest issue it now also runs on node 18 and 19 so I guess there is that 🙃

Tyratox:
I hoped the issue was fixed in the meantime but unfortunately that is not the case :(

Tyratox:
After more obnoxious debugging I finally found the culprit: Somehow the updated dependencies badly interacted with the sinon fake timers. Fortunately jest 28 fixed the issue why we needed them and hence additionally updating the jest dependencies allowed me to use the standard jest timers and now the test pass again 🥳🥳

Hence we could merge this and then have the CI run more modern versions of node (also multiple versions again!), and hopefully make the project easier to maintain :)

Tyratox:
So for node v18, the expo developers suggest to add prefix the commands with

`npx cross-env NODE_OPTIONS=--openssl-legacy-provider`

https://github.com/expo/expo-cli/issues/4575#issuecomment-1295006089

until they updated the webpack dependency or the metro bundler (https://docs.expo.dev/guides/customizing-metro/) will be the new default. I tried manually updating webpack to v5 and it kind of worked but it would have required tons of additional hacks to make all dependencies load properly since there were major changes between v4 and v5 (for instance node modules are no longer automatically polyfilled and we rely on this in quite a few places for things such as "Buffer"s)

Tyratox:
`NODE_OPTIONS=--openssl-legacy-provider` is not allowed for node16, great 😂 (it exits with an error)

fe1-web/docs/README.md
```diff
@@ -631,7 +631,7 @@ If you write E2E tests, please be mindful of making them maintainable through th
 Please reach out to the DEDIS Engineering team members to deploy a build to an
 internet accessible host.
 
-The web application can be packaged for deploynent by executing `npm run build-web`.
+The web application can be packaged for deploynent by executing `npm run build-web`. In case you experience the error `--openssl-legacy-provider is not allowed in NODE_OPTIONS` then it is due to a workaround introduced here: https://github.com/dedis/popstellar/pull/1308#issuecomment-1426683039. The problem is that expo only works with node version > 16 with the `--openssl-legacy-provider` option but this option makes builds fail on v16. You can still create a build using node v16 by manually running `npx expo export:web` rather than `npm run build-web`.
```
pierluca, line 634: This is one long line :-) 

Would it make sense to split over the next few lines and maybe put a warning block of some kind ?

fe1-web/docs/README.md
```diff
@@ -631,7 +631,7 @@ If you write E2E tests, please be mindful of making them maintainable through th
 Please reach out to the DEDIS Engineering team members to deploy a build to an
 internet accessible host.
```
pierluca, line 632: actually, we could mention that a GitHub release now automatically deploys to fe1.personhood.online

pierluca (commented):
I see you use `cross-env` , but have you been able to test this on Windows + Mac + Linux ?

fe1-web/docs/README.md
Tyratox: Ohh I always forget to change these kind of things if the linter does not do it automatically 🙈😂

yes for sure, something like this or what did you have in mind with a "warning block"?

fe1-web/docs/README.md
Tyratox: True 😁

Tyratox:
> I see you use cross-env , but have you been able to test this on Windows + Mac + Linux ?

Mac yes, linux yes by running it on the CI, windows not so far but I windows is the only reason for using `cross-env`, Mac & Linux use the same format for environment variables. I'll try to start it on a windos machine 👍

Tyratox:
I was able to start development mode & create a build on windows as expected using node v18 using this branch (after installing git) :) Btw, should we maybe add a windows machine on the CI for fe-1? i.e. have it being built on linux on v16 & v18 and on macOS / windows just v18. We can of course also do all combinations but it would for probably further increase the time it takes to let the CI run 🤔

pierluca:
My bad, I think I merged something that caused a conflict, but I'd be happy to merge this 👍

pierluca (approved):
LGTM !

PR #1314, opened 2022-12-14
Seperate screen for chirp creation, disable reactions after usage

Tyratox:


fe1-web/src/features/social/reducer/SocialReducer.ts
```diff
@@ -277,15 +277,15 @@ export const makeChirpsListOfUser = (laoId?: Hash) => (user?: PublicKey) => {
 };
 
 const createReactionsEntry = (reactionByUser: Record<string, string[]>) => ({
-  '👍': reactionByUser['👍'] ? reactionByUser['👍'].length : 0,
-  '👎': reactionByUser['👎'] ? reactionByUser['👎'].length : 0,
-  '❤️': reactionByUser['❤️'] ? reactionByUser['❤️'].length : 0,
+  '👍': reactionByUser['👍'] ? reactionByUser['👍'].map(PublicKey.fromState) : [],
```
pierluca, line 280: isn't this format (and its usage) going to severely degrade the runtime characteristics (RAM usage, responsiveness) of the application ?

pierluca (commented):
Seems pretty good to me, only one question.

fe1-web/src/features/social/reducer/SocialReducer.ts
Tyratox: Most likely, yes.. Not doing the `.map(PublicKey.fromState)` would already be better performance-wise but it contradicts the pattern we want to use in general.

A most likely better solution would be to revert the changes to this selector and add a new one that is just there to check whether the user already reacted :)

Tyratox:
And here are the screenshots :)

Screenshot 2022-12-15 at 18 32 05
Screenshot 2022-12-15 at 18 32 17

pierluca (approved):
LGTM !

PR #1316, opened 2022-12-15
fix Scala tests redundant error

tuomas-p:
remove non-unique Actor names that could conflict due to tests running in parallel

### DIFF
--- a/be2-scala/src/test/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandlerTest.scala
+++ b/be2-scala/src/test/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandlerTest.scala
@@ -57,7 +57,7 @@ class ElectionHandlerTest extends TestKit(ActorSystem("Election-DB-System")) wit
           system.log.info(s"Received - error $x")
       }
     })
-    system.actorOf(dbActorMock, "MockedDB-NACK")
+    system.actorOf(dbActorMock)
   }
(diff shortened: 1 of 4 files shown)


pierluca (commented):
In principle, this works for me. I'm just wondering if it reduces the understandability of the tests during their debugging ? If so, we could just append some random ASCII characters after the actors' names.

Outside the scope of this PR, it seems to me the tests don't seem very well formatted/clean. Is it just my impression as a non-Scala-native or do you agree ?

tuomas-p:
My other alternative would have been to indeed add a "name" argument to the mock actor creation function, which would allow to have unique names across every test, but this also comes "with a price", as we want to make sure that all names are unique and this solution scales a bit less well. From what I have seen, the names may be useful in some circumstances, to decrypt the logs for example, however, this is a very tedious task and since the tests are pretty simple, it is often way easier to go look at a single test and run it again separately if necessary, which removes the need for names.

For the test format, I agree that it could maybe be a tad cleaner, however, the current testing method seems to be the easiest for me (mock a few actors for the different cases and then each test follows the same simple structure: create actor, create message, send message, observe effects, stop actor. So at least in my opinion, the structure of the individual tests is fine, the actor creation could be a bit cleaner, but defining a receive function is always a bit messy with multiple cases and creating one separate actor per tests does not really make sense scalability-wise.

pierluca (approved):

PR #1317, opened 2022-12-15
Make the user profile in ChirpCard clickable

Tyratox:
I caught myself multiple times wanting to do that and it did not work yet so I added this functionality too. Unfortunately adding this was a little more complex that I had hoped for. The problem is that in order to properly support this feature, all tabs of the social media feature had to become stack navigators that include separate instances of SocialUserProfile. The reason being that if the instance is shared, the user experience is extremely confusing (why did pressing on the profile icon suddenly take me to a different tab? why do I get to see the profile screen that I viewed before when I switch tabs?) and make certain screens even inaccessible. (Before there was only one instance in the user list / search tab. When I added the code to directly navigate there, the back button (as it should) took me back to the previous screen, e.g. the feed. But now when tapping on the user list / search tab, the profile screen is shown instead of the user list since that navigator is still stuck in that state and there is now way to recover from that since the back button will just take me back to where I just came from, e.g. again the Feed)


pierluca (approved):
I've approved the code changes but I find it hard to visualize what this change to the navigation actually does. Your description helps me a little but I guess I'll need to see it.
In principle, I'm inclined to merge if you think this is a more intuitive navigation.

Tyratox:
Yeah I understand this is hard to visualize but I don't think it would help if I added screenshots..

This adds the ability to press on the user profile to navigate to that users profile. Now chirps that display a profile image (and hence from which this navigation action can be taken) are on multiple screens (Feed, Search, Profile).

Now if we only kept the Profile instance in the search and always navigated there, this would set the state of the StackNavigator of the Search / User list screen to have this as its current screen. When the user presses the back button, they land back on the feed. So far so good. But now the StackNavigator of the search / user list is stuck in that state because no matter how you navigate there, the back button will never refer to taking the user back to the search / user list page. I kind of repeated myself but I hope it might have helped a little?

pierluca (approved):
Makes sense. Thanks !

PR #1319, opened 2022-12-15
Move toasts to the bottom of the screen

Tyratox:
I spent quite some time trying to figure out how we can solve the problem that toasts cover the whole width and make the content below unclickable.

Obviously the fix from https://github.com/dedis/popstellar/pull/1110 was not enough even though I would have assumed https://github.com/arnnis/react-native-toast-notifications/blob/66a275e03da7b82896c8116740cfd24b5ede3c96/src/toast-container.tsx#L147 is exactly what we wanted. Unfortunately there is also this `width:100%` https://github.com/arnnis/react-native-toast-notifications/blob/master/src/toast.tsx#L347 applied to a container that each toast is wrapped in. Hence all toasts always use the full width and since "box-none" only makes the element itself unclickable but the children (in this case the full width toast containers) still catch pointer events. Thus the current behavior.

Also this `width:100%` cannot be changed, see https://github.com/arnnis/react-native-toast-notifications/blob/master/src/toast.tsx#L312-L316.

Hence as a workaround, this PR simply moves the toasts to the bottom of the screen circumventing the problem in most of the cases. It also places the toast nearer the finger of the user on mobile allowing them to be swiped away more quickly. To compensate for the loss of readability when combining bottom warning toasts while in offline mode, this PR also changes the color used for the offline mode banner. Feel free to suggest a better color than yellow :)


pierluca (approved):
Not super satisfying, but it makes sense :)

PR #1320, opened 2022-12-20
UI changes to wallet initialization

Tyratox:
This PR changes the UI of the wallet initialization to make it more consistent with the rest of the application. (No negative text, no blue background, buttons in the toolbar)

Screenshot 2022-12-20 at 14 42 21
Screenshot 2022-12-20 at 14 44 04


fe1-web/src/resources/strings.ts
```diff
@@ -397,15 +397,23 @@ namespace STRINGS {
     ', you can continue and start using the POPStellar application. ' +
     'The purpose of this application is to demonstrate use cases for proofs of personhood (PoP).';
 
-  export const wallet_welcome_start_exploring = 'Start exploring POPStellar 🤩';
+  export const wallet_welcome_already_know_seed_1 =
+    'If you have already used this application before and remember your previous';
+  export const wallet_welcome_already_know_seed_2 =
+    ', you can restore your wallet by pressing the respective button.';
```
pierluca, line 403: does it make sense that we start with a comma ?

fe1-web/src/features/wallet/screens/WalletCreateSeed.tsx
```diff
@@ -83,75 +77,72 @@ const WalletCreateSeed = () => {
...
-        <View style={containerStyles.centeredY}>
-          <Text style={[Typography.heading, Typography.negative]}>
-            {STRINGS.wallet_welcome_heading}
+    <ScreenWrapper toolbarItems={toolbarItems}>
+      <View style={containerStyles.centeredY}>
+        <Text style={Typography.heading}>{STRINGS.wallet_welcome_heading}</Text>
+        <Text style={Typography.paragraph}>{STRINGS.wallet_welcome_text_first_time}</Text>
+        <Text style={Typography.paragraph}>
+          <Text>{STRINGS.wallet_welcome_text_wallet_explanation_1}</Text>
+          <Text> </Text>
```
pierluca, line 100: does it make sense to have all these empty `<Text>` fields ? 
Is there no better way ?

pierluca (commented):
Minor comments, but LGTM overalls

fe1-web/src/resources/strings.ts
Tyratox: This sentence is split into two parts with "seed 🔑" in between (colored in blue). If we would have it the same color or put the comma inside a <Text/> as with the spaces, we would could combine the two / not start the sentence with a comma. What do you think?

Screenshot 2022-12-20 at 14 42 21

fe1-web/src/features/wallet/screens/WalletCreateSeed.tsx
Tyratox: Well either that or have sentences start or end with a space which is bad too 😅

Tyratox: Or, of course, rewrite the sentences / combine them as suggested above. The reason for these artifacts are the differently colored text sections. Or do you see a another option?

pierluca (approved):
Okay, I understand better. We could do better, but there's no point in investing more effort in this now. Good enough ! 👍

PR #1323, opened 2022-12-26
Refactor CoinHandler and MeetingHandler files

Ajkunas:
Brief refactoring of `handleCreateMeeting` and `handlePostTransaction` functions.


be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/MeetingHandler.scala
```diff
@@ -1,22 +1,27 @@
...
 
   def handleCreateMeeting(rpcMessage: JsonRpcRequest): GraphMessage = {
-    rpcMessage.getParamsChannel.decodeChannelLaoId match {
-      case Some(_) =>
-        val ask: Future[GraphMessage] = dbAskWritePropagate(rpcMessage)
-        Await.result(ask, duration)
-      case _ => Right(PipelineError(
-          ErrorCodes.INVALID_DATA.id,
-          s"Unable to create meeting: invalid encoded laoId '${rpcMessage.getParamsChannel}'",
-          rpcMessage.id
```
pierluca, line 18: here as well, having this split on multiple lines was helpful to read it quickly.

pierluca (approved):
Looks good to me, just be mindful that removing too many newlines might hurt readability a bit.
In general it's preferable to keep line length below some limit, usually 80 to 120, depending on the organization you work in 😄 
According to the Scala editor config, this should be 120: 
https://github.com/dedis/popstellar/blob/master/be2-scala/.editorconfig#L7

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/MeetingHandler.scala
Ajkunas: Understood, I agree that it is easily readable.

pierluca (approved):
Good enough for now 😁

PR #1333, opened 2023-01-05
Make social media navigation more intuitive

Tyratox:
In the last meeting the most intuitive behavior was deemed to be that tapping on "User List" will always go back to the user list, even if a user previously navigated to a specific profile within that screen.

Hence it was also possible to remove the duplicate profile screens as the issue described [here](https://github.com/dedis/popstellar/pull/1317#issuecomment-1353480210) no longer persists.


pierluca (approved):
This comes across as much more intuitive in terms of code too !

PR #1335, opened 2023-01-05
Make reaction buttons less attention grabbing

Tyratox:
This PR changes the button style of the chirp reaction buttons to be less attention grabbing as depicted bellow. (Based on discussion in the last meeting). Moreover it contains a small fix for the icon color of disabled PopIconButtons.

Screenshot 2023-01-05 at 15 28 17

Screenshot 2023-01-05 at 15 30 30

### DIFF
--- a/fe1-web/src/core/components/PoPIconButton.tsx
+++ b/fe1-web/src/core/components/PoPIconButton.tsx
@@ -15,6 +15,18 @@ const SIZE_MAP = {
 const PoPIconButton = (props: IPropTypes) => {
   const { onPress, buttonStyle, disabled, toolbar, size, testID, name } = props;
 
+  let iconColor = Color.contrast;
+
+  if (buttonStyle === 'secondary') {
+    // in case of an outlined button, the icon color
+    // should be the same as the border's
+    if (disabled) {
+      iconColor = Color.inactive;
+    } else {
+      iconColor = Color.accent;
+    }
+  }
+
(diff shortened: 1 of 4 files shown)


fe1-web/src/core/components/PoPIconButton.tsx
```diff
@@ -15,18 +15,26 @@ const SIZE_MAP = {
 const PoPIconButton = (props: IPropTypes) => {
   const { onPress, buttonStyle, disabled, toolbar, size, testID, name } = props;
 
+  let iconColor = Color.contrast;
+
+  if (buttonStyle === 'secondary') {
+    // in case of an outlined button, the icon color
+    // should be the same as the border's
+    if (disabled) {
+      iconColor = Color.inactive;
```
pierluca, line 24: is the `inactive` color only useful for the `secondary` button style, or should it be used for all button tyles ?

pierluca (approved):
LGTM, barring a tiny question

fe1-web/src/core/components/PoPIconButton.tsx
Tyratox: The primary button (colored background, white icon) also has a disabled state but there only the background color changes which is done in `PoPButton`. But for the secondary button style (colored icon, outlined button) it is necessary to change the color of the icon to gray. Does this explanation make sense?

PR #1336, opened 2023-01-05
Change how sender is displayed in chirps

Tyratox:
As discussed in the last meeting, this PR changes the Chirp view so that no longer the full public key of the sender is show with every chirp but only the first four characters right below the profile picture.

Screenshot 2023-01-05 at 15 40 34


pierluca (approved):
Very nice !

PR #1337, opened 2023-01-05
Refactor type architecture of the Vote objects

GabrielFleischer:
Currently, the encrypted and plain vote objects are completely distinct, and this leads to a good amount of duplicated code.

This PR creates a common parent in the form of the `Vote` interface.

This allows a good amount of refactor and simplifications throughout the election process.


Warhedgehog (approved):
Great work ! LGTM

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Election.java
```diff
@@ -219,40 +238,32 @@ public static String generateEncryptedElectionVoteId(
...
-            ? getListOfVoteIds(openVoteByPublicKey, ElectionVote::getId)
-            : getListOfVoteIds(encryptedVoteByPublicKey, ElectionEncryptedVote::getId);
-
-    if (listOfVoteIds.isEmpty()) {
+    String[] ids =
+        messageMap.values().stream()
+            .map(votesBySender::get)
+            // Merge lists and drop nulls
+            .flatMap(
+                electionVotes -> electionVotes != null ? electionVotes.stream() : Stream.empty())
```
pierluca, line 246: Do you have null-coalescing operators that would allow this ?

```suggestion
                electionVotes -> electionVotes?.stream() ?? Stream.empty())
```

GabrielFleischer: In Kotlin, yes
In java, I don't think so. At least not in java 8 to 11

PR #1339, opened 2023-01-09
Fix small bugs found on the android front-end

GabrielFleischer:
This PR fixes two issues :
 1. In the `QRScanningFragment`, when a dialog is dismissed, it used the wrong function and this could lead to an application crash if the camera permission was not granted
 2. The `MessageHandler` was not checking that a message had already been handled before. This could lead to infinite loops and performance drops


jbsv (approved):
LGTM, thanks for this fix

PR #1341, opened 2023-01-09
Refactor the event list data usage

GabrielFleischer:
This PR merge the roll call and the elections into a single collection to use them in the event list.

This allows a simpler data pipeline.


Warhedgehog (changes requested):
It's a bit of a tough call. Codewise, nothing to say it's very clean. 
Yet I think it is a bit of an overkill to implement a generic repo interface. In my mind this could be achieved in a more simple way by merging the observables of the roll call and election repos (like here https://www.learnrxjs.io/learn-rxjs/operators/combination/merge). Is there a reason you went with this heavy-ish architecture ?

Also since the work is already done, I don't know if it would make sense to throw it away in favor of the easier solution (provided that the simple one even works)

GabrielFleischer:
> It's a bit of a tough call. Codewise, nothing to say it's very clean. Yet I think it is a bit of an overkill to implement a generic repo interface. In my mind this could be achieved in a more simple way by merging the observables of the roll call and election repos (like here https://www.learnrxjs.io/learn-rxjs/operators/combination/merge). Is there a reason you went with this heavy-ish architecture ?
> 
> Also since the work is already done, I don't know if it would make sense to throw it away in favor of the easier solution (provided that the simple one even works)

It's mostly to be able to reuse the code that observe the ids, and extract/merge the observable per id.

Otherwise, the whole function was mostly duplicated.
And to me, it is not a big change, only a simple interface with functions.

Also, in a far future, there is a goal to have an independent module per feature, and this is a first step in that direction.

Warhedgehog (approved):
Given the subsequent PR #1344 , LGTM !

PR #1342, opened 2023-01-10
Fix CI go

Anghul:
Fix some of the errors happening sometimes with the go tests.

5 different errors have been identified:

1.  When testing the chirp deletion, the deletechirp message was sometimes processed before the addchirp message, failing the test.
For now, a small delay has been added in the test to avoid the issue, but ultimately a way for the program to retry when receiving out of order messages should be implemented. It wasn't done for now as it would probably need to be added in multiple places of the program.

2. When catching up on a LAO channel, the first two messages were sometimes out of order when they had the same timestamp.
It was fixed in two steps : first the sort function used previously didn't guarantee the order to stay the same when the timestamp is equal. The sort function has therefore been changed. Second a map in go doesn't guarantee the order when transformed to an array. An array has therefore been added, containing pointers to the messages to keep this order.

3. Sometimes when testing if the greeting sending worked, the test failed because some arguments couldn't be converted to string for the error message.
The arguments were probably missused in the test, and have been removed as they aren't necessary, even if they help when fixing bugs.

4. Sometimes when testing the transmission of the election key, the message couldn't be unmarshaled.
The unmarshall in itself was done weirdly, and the error didn't appear again after it was cleaned up, but the bug may not be fixed.

5. When testing the timeout of the consensus, sometimes the timout doesn't happen. Locally, the failure happens after multiple thousands of test run, and takes too much time to gather more information. This bug isn't fixed yet because of this.

### DIFF
--- a/be1-go/channel/election/mod_test.go
+++ b/be1-go/channel/election/mod_test.go
@@ -476,13 +476,10 @@ func Test_Sending_Election_Key(t *testing.T) {
 
 	electionKeyMsg := catchupAnswer[1]
 
-	data := messagedata.ElectionKey{}.NewEmpty()
+	var dataKey messagedata.ElectionKey
 
-	err := electionKeyMsg.UnmarshalData(data)
+	err := electionKeyMsg.UnmarshalData(&dataKey)
 	require.NoError(t, err, electionKeyMsg)
-
-	dataKey, ok := data.(*messagedata.ElectionKey)
-	require.True(t, ok)
 
 	key, err := base64.URLEncoding.DecodeString(dataKey.Key)
 	require.NoError(t, err)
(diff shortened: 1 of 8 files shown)


be1-go/channel/election/mod_test.go
```diff
@@ -476,13 +476,10 @@ func Test_Sending_Election_Key(t *testing.T) {
...
 	electionKeyMsg := catchupAnswer[1]
 
-	data := messagedata.ElectionKey{}.NewEmpty()
+	var dataKey messagedata.ElectionKey
 
-	err := electionKeyMsg.UnmarshalData(data)
-	require.NoError(t, err)
-
-	dataKey, ok := data.(*messagedata.ElectionKey)
-	require.True(t, ok)
```
pierluca, line 485: why dropping the check on the data ?

Anghul: The check on the data is simply a cast, with the check itself verifying if the cast worked or not. It is in my opinion unnecssary as dataKey already has the correct type.
For the failure in itself, it could be this, even if the error message didn't indicate it (it was only "unexpected end of JSON message"), and if so it is fixed now.

pierluca (approved):

be1-go/channel/election/mod_test.go
pierluca: Thanks !

PR #1343, opened 2023-01-10
Drawer menu

Warhedgehog:
This adds a drawer menu to the application. A lot of code common to the three lao activities - laodetail, digitalcash and social media - is now done in the abstract LaoActivity (which used to be `NavigationActivity`). A further step would be to merge the 3 activities into 1.
The QR icon in `LaoDetailActivity` has been moved to the drawer (as Invite like in fe1) and the view it opens is now a separate fragment
The drawer header now displays the name of the lao as well as the role
Here are screenshots. The current version is with the main blue but I've tried also the dark blue

I'm very sorry for the size of the PR, I never thought it would be that high.

The code coverage is not great unfortunately but for some reason I cannot open the drawer in tests. The only way (afaik) does not work and I've given up trying to fix it after some time. The consequence is that it is not possible to test the drawer navigation (which is quite a few LoCs because of all the switches). I don't think it's too bad since a breakdown would be immediately noticed upon using the app even a little

I tested the app manually and did a 2 phones pop party


fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/navigation/LaoViewModel.java
```diff
@@ -0,0 +1,174 @@
...
+
+import io.reactivex.disposables.CompositeDisposable;
+import io.reactivex.disposables.Disposable;
+
+/**
+ * Abstract view model that provides the components of the navigation bar
+ *
+ * <p>Any activity that uses a navigation bar should have its view model extends from the this *
+ * class
+ */
```
GabrielFleischer, line 26: I think this comment is outdated too

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/navigation/LaoViewModel.java
```diff
@@ -0,0 +1,174 @@
...
+   * messages flows
+   *
+   * <p>They will be disposed of when the view model is cleaned which ensures that the subscription
+   * stays relevant throughout the whole lifecycle of the activity and it is not bound to a fragment
+   *
+   * @param disposable to add
+   */
+  public void addDisposable(Disposable disposable) {
+    this.disposables.add(disposable);
+  }
```
GabrielFleischer, line 154: Oh, it went here.

Makes sense

fe2-android/app/src/test/ui/robolectric/com/github/dedis/popstellar/ui/socialmedia/ChirpListAdapterTest.java
```diff
@@ -27,10 +28,20 @@
...
+
+  private static final PublicKey SENDER_1 = SENDER_KEY_1.getPublicKey();
+  private static final PublicKey SENDER_2 = SENDER_KEY_2.getPublicKey();
+
+  private static final String LAO_ID = Lao.generateLaoId(SENDER_1, CREATION_TIME, LAO_NAME);
 
   private final Intent intent =
       new Intent(ApplicationProvider.getApplicationContext(), SocialMediaActivity.class)
-          .putExtra("OPENED_FROM", "");
+          .putExtra(Constants.LAO_ID_EXTRA, "");
```
GabrielFleischer, line 44: Just for the sake of consistency, could you set the LAO_ID ?
```suggestion
          .putExtra(Constants.LAO_ID_EXTRA, LAO_ID);
```

GabrielFleischer (changes requested):
Very great !

It's good to see things take form.

I found some small issues, happy to approve once they're addressed 😄

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/navigation/LaoActivity.java
```diff
@@ -0,0 +1,227 @@
...
+                                  return rc.equals(rollCallRepo.getLastClosedRollCall(laoId));
+                                } catch (NoRollCallException e) {
+                                  return false;
+                                }
+                              });
+                  laoViewModel.setIsAttendee(isLastRollCallAttended);
+                },
+                error ->
+                    ErrorUtils.logAndShow(this, TAG, error, R.string.unknown_roll_call_exception)));
+  }
```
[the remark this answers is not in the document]
Warhedgehog: Is the injection not done in the constructor in viewmodels ? Because if yes I'd rather leave it here for the sake of simplicity.

GabrielFleischer (approved):
LGTM

PR #1344, opened 2023-01-11
Attempt to simplify the event repositories

GabrielFleischer:
This is a test PR to show an idea I got that could simplify the events systems (at least for now)

What do you think about it ?


Warhedgehog (approved):
LGTM !

PR #1346, opened 2023-01-13
Remove unused resources

GabrielFleischer:
This PR remove the huge amount of unused resources that were present in the application.


Warhedgehog (approved):
LGTM. Just to be sure did you try to build it on Android Studio ? Because sometimes there are false positive in SC and their absence then make the linking fail. I see that it builds on GH but I'd rather be sure :)

GabrielFleischer:
> LGTM. Just to be sure did you try to build it on Android Studio ? Because sometimes there are false positive in SC and their absence then make the linking fail. I see that it builds on GH but I'd rather be sure :)

Yes, it was working fine.

PR #1348, opened 2023-01-18
Adds a deploy action and some build info

nkcr:
I wrote a CI action to automatically deploy fe1-web. In that process I added some build info and made an attempt to include them in the layout (bottom left). Though this is probable not the cleanest way to do it. I did my best but I have a limited understanding of the codebase.

Screenshot 2023-01-18 at 10 49 41

It gets deployed to https://fe1.personhood.online/

### DIFF
--- a/fe1-web/src/App.tsx
+++ b/fe1-web/src/App.tsx
@@ -4,7 +4,7 @@ import { ActionSheetProvider } from '@expo/react-native-action-sheet';
 import { NavigationContainer } from '@react-navigation/native';
 import { registerRootComponent } from 'expo';
 import React from 'react';
-import { Platform, StatusBar } from 'react-native';
+import { StyleSheet, Platform, StatusBar } from 'react-native';
 import { SafeAreaProvider } from 'react-native-safe-area-context';
 import Toast, { ToastProvider } from 'react-native-toast-notifications';
 import { Provider } from 'react-redux';
@@ -31,32 +31,33 @@ configureNetwork(messageRegistry, keyPairRegistry);
 persist.persist();
 
 const BuildInfo = () => {
-  const containerStyle = {
-    position: 'fixed',
-    bottom: '4px',
-    left: '3px',
-    zIndex: 100,
-    color: '#757575',
-    fontSize: '8px',
-    fontFamily: 'monospace',
-    textTransform: 'uppercase',
-  };
-
-  const linkStyle = {
-    textDecoration: 'none',
-    color: '#757575',
-  };
+  const styles = StyleSheet.create({
+    container: {
+      position: 'absolute',
+      bottom: '4px',
+      left: '3px',
+      zIndex: 100,
+      color: '#757575',
+      fontSize: '8px',
+      fontFamily: 'monospace',
+      textTransform: 'uppercase',
+    },
+    link: {
+      textDecorationLine: 'none',
+      color: '#757575',
+    },
+  });
 
   return (
-    <div style={containerStyle}>
+    <div style={styles.container}>
       <a
-        style={linkStyle}
+        style={styles.link}
         href={`https://github.com/dedis/popstellar/releases/tag/${Constants?.expoConfig?.extra?.appVersion}`}
         target="_blank">
         {Constants?.expoConfig?.extra?.appVersion}
       </a>
       {' | '}
-      <a style={linkStyle} href={Constants?.expoConfig?.extra?.buildURL} target="_blank">
+      <a style={styles.link} href={Constants?.expoConfig?.extra?.buildURL} target="_blank">
         {Constants?.expoConfig?.extra?.shortSHA}
       </a>
       {' | '}
--- a/fe1-web/src/App.tsx
+++ b/fe1-web/src/App.tsx
@@ -10,6 +10,7 @@ import Toast, { ToastProvider } from 'react-native-toast-notifications';
 import { Provider } from 'react-redux';
 import { PersistGate } from 'redux-persist/integration/react';
 
+import { BuildInfo } from 'core/components';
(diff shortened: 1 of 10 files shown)


jbsv (approved):
LGTM

pierluca:
@Tyratox can I let you review this and apply the necessary changes for code consistency ? Thank you ! 😄

fe1-web/src/App.tsx
```diff
@@ -29,6 +30,42 @@ configureNetwork(messageRegistry, keyPairRegistry);
 // start persisting the redux state after all reducers have been registered
 persist.persist();
 
+const BuildInfo = () => {
+  const styles = StyleSheet.create({
+    container: {
+      position: 'absolute',
+      bottom: '4px',
+      left: '3px',
```
Tyratox, line 38: If possible it would be great if the "standard" spacing units (https://github.com/dedis/popstellar/blob/master/fe1-web/src/core/styles/spacing.ts) are used for these kind of placements as it in general results in a more consistent view.

What is more important is that we cannot use pixel values in react native, this will most likely break if we want to build the application for iOS / Android and will only work in the web build.
> All dimensions in React Native are unitless, and represent density-independent pixels.
https://reactnative.dev/docs/height-and-width

Tyratox, line 38: If possible it would be great if the "standard" spacing units (https://github.com/dedis/popstellar/blob/master/fe1-web/src/core/styles/spacing.ts) are used for these kind of placements as it in general results in a more consistent view.

What is more important is that we cannot use pixel values in react native, this will most likely break if we want to build the application for iOS / Android and will only work in the web build.
> All dimensions in React Native are unitless, and represent density-independent pixels.
https://reactnative.dev/docs/height-and-width

fe1-web/src/App.tsx
```diff
@@ -29,6 +30,42 @@ configureNetwork(messageRegistry, keyPairRegistry);
...
+      zIndex: 100,
+      color: '#757575',
```
Tyratox, line 40: If possible colors from the color scheme (https://github.com/dedis/popstellar/blob/master/fe1-web/src/core/styles/color.ts) are preferred so that they can be changed all at once. `inactive` (not the best name for sure) / #8E8E8E should be similar to #757575.

fe1-web/src/App.tsx
```diff
@@ -29,6 +30,42 @@ configureNetwork(messageRegistry, keyPairRegistry);
...
+      fontSize: '8px',
+      fontFamily: 'monospace',
+      textTransform: 'uppercase',
+    },
+    link: {
+      textDecorationLine: 'none',
+      color: '#757575',
```
Tyratox, line 47: Same comment regarding color scheme

fe1-web/src/App.tsx
```diff
@@ -29,6 +30,42 @@ configureNetwork(messageRegistry, keyPairRegistry);
...
+        target="_blank">
+        {Constants?.expoConfig?.extra?.appVersion}
+      </a>
+      {' | '}
+      <a style={styles.link} href={Constants?.expoConfig?.extra?.buildURL} target="_blank">
+        {Constants?.expoConfig?.extra?.shortSHA}
+      </a>
+      {' | '}
+      {Constants?.expoConfig?.extra?.buildDate}
+    </div>
```
Tyratox, line 65: If we want to maintain compatibility with native builds we cannot use standard HTML elements such as `div` or `a` or even plain unwrapped text.

react-native's `View` more or less corresponds to the HTML `div` and all texts have to be wrapped in `<Text>xyz</Text>` elements.

> The most fundamental component for building a UI, View is a container that supports layout with [flexbox](https://reactnative.dev/docs/flexbox), [style](https://reactnative.dev/docs/style), [some touch handling](https://reactnative.dev/docs/handling-touches), and [accessibility](https://reactnative.dev/docs/accessibility) controls. View maps directly to the native view equivalent on whatever platform React Native is running on, whether that is a UIView, \<div>, android.view, etc.

(https://reactnative.dev/docs/view)

For displaying a link it becomes even more cumbersome:

```jsx
<Text onPress={() => Linking.openURL('http://google.com')}>Google</Text>
```

(From https://stackoverflow.com/a/30540502/2897827)

fe1-web/src/App.tsx
```diff
@@ -29,6 +30,42 @@ configureNetwork(messageRegistry, keyPairRegistry);
 // start persisting the redux state after all reducers have been registered
 persist.persist();
 
+const BuildInfo = () => {
```
Tyratox, line 33: Since this is a pretty much self-contained, it might be worth putting the component in its own file?

Tyratox (commented):
Nice, very cool! 😁 I added a few comments, especially the ones related to the pixel values and HTML elements are important as this will break native builds. (They are not fully functional yet but this will move us further away from having them working)

Regarding placement I think it should be fine as it is only debug information anyway but given that the hash is already displayed in the drawer

Screenshot 2023-01-18 at 13 27 20

it might be worth considering either moving it there or removing the hash in the drawer?

Tyratox:
If you like me to I am also happy to apply the mentioned changes of course, whatever you prefer :)

nkcr:
@Tyratox thank you for the review. As I was curious I made the changes myself, but I let you correct anything else if you judge necessary :)

Tyratox (approved):
Cool 😄 No, now it looks very good to me! 😊
(except the minor issues indicated by the linter but I assume you know how to fix them :))

(To update the snapshots you can run `npm run test -- --u`, just in case this message was lost in the huge log output)

Tyratox (approved):

pierluca:
@Tyratox @nkcr thank you both for the great work 😄

PR #1350, opened 2023-01-19
Make fe-1 wallet consistent with fe-2

Tyratox:
Make the fe-1 wallet UI more consistent with fe-2 (#1328)

One intended difference is the naming of Current Token / Current instead of Active Token / Active as I thought it is more consistent with the "previous" label. Feel free to correct me on that though :)

Screenshot 2023-01-19 at 20 02 20


pierluca (approved):
LGTM.
Quick question: is there any way from the token display to see information about the associated roll call ?

Tyratox:
Not yet, I forgot that we were talking about this! It should be rather straight forward to add a button to navigate to the roll call event :) What do you think, should I do a PR for this or create a "Good First Issue" as this might be a nice task to get to know the navigation?

PR #1351, opened 2023-01-20
Adds build info and automatic deployment

nkcr:
Gets deployed to https://be1.personhood.online/. 
Info can be fetched with https://be1.personhood.online/client/infos

### DIFF
--- a/be1-go/Makefile
+++ b/be1-go/Makefile
@@ -1,7 +1,17 @@
+version=$(shell git describe --abbrev=0 --tags || echo '0.0.0')
+versionFlag="popstellar.Version=$(version)"
+versionFile=$(shell echo $(version) | tr . _)
+timeFlag="popstellar.BuildTime=$(shell date +'%d/%m/%y_%H:%M')"
+shortsha=$(shell git rev-parse --short HEAD)
+shaFlag="popstellar.ShortSHA=$(shortsha)"
+
 .PHONY: test
 
 build: protocol
-	go build -o pop ./cli
+	go build -ldflags="-X $(versionFlag) -X $(timeFlag) -X $(shaFlag)" -o pop ./cli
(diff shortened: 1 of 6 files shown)


be1-go/Makefile
```diff
@@ -1,7 +1,17 @@
...
+shortsha=$(shell git rev-parse --short HEAD)
+shaFlag="popstellar.ShortSHA=$(shortsha)"
+
 .PHONY: test
 
 build: protocol
-	go build -o pop ./cli
+	go build -ldflags="-X $(versionFlag) -X $(timeFlag) -X $(shaFlag)" -o pop ./cli
+	GOOS=linux GOARCH=amd64 go build -ldflags="-X $(versionFlag) -X $(timeFlag) -X $(shaFlag)" -o pop-linux-amd64-$(versionFile) ./cli
+	GOOS=darwin GOARCH=amd64 go build -ldflags="-X $(versionFlag) -X $(timeFlag) -X $(shaFlag)" -o pop-darwin-amd64-$(versionFile) ./cli
```
pierluca, line 13: it might be convenient to have a darwin arm64 release too.

be1-go/network/server.go
```diff
@@ -63,6 +66,8 @@ func NewServer(hub hub.Hub, addr string, port int, st socket.SocketType, log zer
 	mux := http.NewServeMux()
 	mux.HandleFunc(path, server.ServeHTTP)
 
+	mux.HandleFunc(infoPath, server.infoHandler)
```
pierluca, line 69: In principle, this is not ideal as it helps the fingerprinting of the binary for attacks.
On the other hand, at this stage, this is going to be super helpful. Thanks !

pierluca (approved):
Nice !

Is be1 getting deployed on the popdemo server ?
And if so, would it make sense for fe1 to be deployed on the same server ?

be1-go/Makefile
nkcr: added

nkcr:
> And if so, would it make sense for fe1 to be deployed on the same server ?
Yes, we could.

pierluca (approved):

PR #1352, opened 2023-01-23
Adds automatic deployment for scala

nkcr:
It gets deployed to https://be2.personhood.online/.

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -113,3 +113,51 @@ jobs:
             rsync -chav --delete \
               -e 'ssh -p ${{secrets.POPDEMO_DEPLOY_PORT}} -i ./deploy_key -o StrictHostKeyChecking=no' \
               ./pop ${{env.dest}}
+
+  deploy_be2:
+    name: Release and deploy be2-scala
+
+    if:  ${{ startsWith(github.event.release.tag_name, 'be2-') }}
+    runs-on: ubuntu-latest
+
+    env:
+      base_folder: be2-scala
+      scala_version: 2.13
+
+    defaults:
+      run:
+        working-directory: ./${{ env.base_folder }}
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Set up JDK 11
+        uses: actions/setup-java@v2
+        with:
+          distribution: "zulu"
+          java-version: "11"
+
+      - name: build
+        run: |
+          sbt assembly
+
+      - name: Upload release
+        uses: softprops/action-gh-release@v1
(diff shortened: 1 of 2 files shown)


be2-scala/README.md
```diff
@@ -68,8 +70,38 @@ The project relies on several sbt dependencies (external libraries) :
...
+# install java zulu 11
+sdk install java 11.0.18-zulu
+# install sbt
+sdk install sbt
+# check that everything went well
+java -version
+> openjdk version "11.0.18" 2023-01-17 LTS
+```
+
+## Build and run
```
pierluca, line 94:
```suggestion
## Package and run
```

I reckon for day-to-day development, we should probably instead point to `sbt run`

be2-scala/README.md
```diff
@@ -68,8 +70,38 @@ The project relies on several sbt dependencies (external libraries) :
...
+## Install java and sbt on linux and Mac
+
+[skdman](https://sdkman.io/) does a pretty good job for installing the required
+environment:
+
+```sh
+# install sdkman
+curl -s "https://get.sdkman.io" | bash 
+# install java zulu 11
+sdk install java 11.0.18-zulu
```
pierluca, line 86: are you sure? 
On popdemo we install OpenJDK 17.

pierluca (commented):
@Ajkunas @tuomas-p you might want to have a look at this PR as well as this one: https://github.com/dedis/popstellar/pull/1351

Do you think it's possible to do something similar, where an endpoint on the Scala backend would give us the version ? It would be super convenient.

be2-scala/README.md
nkcr: The only example I had at that point was the CI, I blindly followed what is done there 🙈

nkcr: But this is fine, it will run. We update this one and the CI in the future.

.github/workflows/deploy.yaml
```diff
@@ -112,4 +112,52 @@ jobs:
...
+        working-directory: ./${{ env.base_folder }}
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Set up JDK 11
+        uses: actions/setup-java@v2
+        with:
+          distribution: "zulu"
+          java-version: "11"
```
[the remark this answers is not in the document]
nkcr: To be consistent with the CI. This won't prevent the jar from running.

nkcr:
> Do you think it's possible to do something similar, where an endpoint on the Scala backend would give us the version ? It would be super convenient.

I won't give it a try because I'm not familiar enough with the codebase, but that would be very helpful.

pierluca (approved):

PR #1356, opened 2023-01-31
Fix vote encryption in android

GabrielFleischer:
The order of ids was wrongly computed

I took some time to simplify the way data is stored and the flow of the tests


jbsv (approved):
Looks good to me. Worth a try during a pop party indeed.

Warhedgehog (approved):
LGTM

PR #1362, opened 2023-02-01
Prevent text overflow on lao item

Tyratox:
In case of very long lao titles, the text overflowed over the screen. With this PR the name is truncated at the end and adds ellipsis (...). On web this will be `text-overflow: ellipsis` (https://www.w3schools.com/cssref/playdemo.php?filename=playcss_text-overflow&preval=ellipsis)


pierluca (approved):
LGTM

PR #1364, opened 2023-02-01
Display election status on election detail screens

Tyratox:
As discussed in the last meeting, for non-organizer it was previously not clear what the current state of an election is once they are on the detail screen.

Screenshot 2023-02-01 at 21 29 04


pierluca (approved):
LGTM !

PR #1365, opened 2023-02-01
Make election notice less attention grabbing

Tyratox:
Before:

Screenshot 2023-02-01 at 21 34 00

After:

(Option 1, current)
Screenshot 2023-02-10 at 11 37 17

(Option 2, previous)
Screenshot 2023-02-01 at 21 33 31


Tyratox:
Another option would be to have a white background, I think this might be a good tradeoff and is more consistent with the rest of the UI:
Screenshot 2023-02-10 at 11 37 17

pierluca (approved):
Yup, smooth, me likey !

PR #1367, opened 2023-02-05
Fix bug in transaction validation

Tyratox:
This PR fixes a bug where fe-1 crashed when trying to process a transaction referring to an invalid input and adds corresponding tests.


pierluca (approved):
The cognitive complexity of this stuff is insane, but it'd require a much bigger refactoring than makes sense for this bugfix. 

LGTM !

PR #1368, opened 2023-02-05
Add more try/catch clauses to network handlers

Tyratox:
To prevent issues such as the one fixed in #1367, this PR adds try/catch clauses around constructors in the message handlers that might throw exceptions. For most of the errors it should not be possible to trigger them thanks to the json schemas.


pierluca:
Seems okay to me but I haven't got the time to review this much code reasonably fast.
@Xelowak can you review in detail?

Xelowak (approved):
LGTM!
Thanks for these changes, I believe they will make the system more robust to issues you mentioned in the description.
Adding tests to ChirpHandler and ReactionHandler is also nice.

PR #1369, opened 2023-02-05
Ensure attendee list is sorted

Tyratox:
As discussed in the last meeting, we should ensure that the order of the list of pop tokens should not correspond to the scanning order as this would facilitate de-anonymization.

**This is a breaking change and should only be merged when all subsystems implemented this change**


fe1-web/src/features/rollCall/network/__tests__/RollCallMessageApi.test.ts
```diff
@@ -203,7 +203,9 @@ describe('MessageApi', () => {
     const attendeePks: PublicKey[] = [
       'NdCY6Ga7yVK5rZay3S4xGjlz2jHZQrg2gw7fyI6tfuo=',
       'BEW-uVz_NG_prXFuaKrI9Ae0EbBLGWehLQ8aLZFWY4w=',
-    ].map((a) => new PublicKey(a));
+    ]
+      .map((a) => new PublicKey(a))
+      .sort();
```
pierluca, line 208: are `PublicKey` objects sortable by their string content ?
I'd have believed `sort()` needed to happen before the `map()`.
This might be true in other places as well.

fe1-web/src/features/rollCall/objects/RollCall.ts
```diff
@@ -84,56 +84,79 @@ export class RollCall {
...
+      if (!isAttendeeListSorted[0]) {
+        throw new Error(
+          'Attendee list is not sorted alphabetically, rejecting due to the risk of de-anonymization',
+        );
+      }
+    }
     this.attendees = obj.attendees;
 
-    this.openedAt = obj.openedAt;
-    this.closedAt = obj.closedAt;
```
pierluca, line 136: If we can't merge this quickly, I'm a bit concerned this PR might grow stale because of this change that might bring in conflicts later on.

fe1-web/src/features/rollCall/network/__tests__/RollCallMessageApi.test.ts
Tyratox: Since `PublicKey` is a `String` it works, yes :) But I agree that in this particular case the other way around is easier to understand

PR #1370, opened 2023-02-07
Add form feedback when creating events

Xelowak:
I added clear feedback for when the confirm button is disabled during event creation.
When fields are left empty, the user is aware of it. The same happens when the user is offline.

Meeting
Capture d’écran 2023-02-07 à 16 57 39

Capture d’écran 2023-02-07 à 16 57 53

Roll Call
Capture d’écran 2023-02-07 à 16 58 24

Election
Capture d’écran 2023-02-07 à 16 58 41

Capture d’écran 2023-02-13 à 16 42 19

Not connected to a LAO (Meeting)
Capture d’écran 2023-02-07 à 17 33 16

### DIFF
--- a/fe1-web/src/features/evoting/screens/CreateElection.tsx
+++ b/fe1-web/src/features/evoting/screens/CreateElection.tsx
@@ -309,9 +309,7 @@ const CreateElection = () => {
       )}
       {questions.some(isQuestionInvalid) && (
         <Text style={[Typography.paragraph, Typography.error]}>
-          {STRINGS.election_create_invalid_questions_1 +
-            MIN_BALLOT_OPTIONS +
-            STRINGS.election_create_invalid_questions_2}
+          {STRINGS.election_create_invalid_questions.replace('{}', MIN_BALLOT_OPTIONS.toString())}
         </Text>
       )}
       {!haveQuestionsSameTitle(questions) && (
--- a/fe1-web/src/features/evoting/screens/CreateElection.tsx
+++ b/fe1-web/src/features/evoting/screens/CreateElection.tsx
@@ -70,6 +70,15 @@ const isQuestionInvalid = (question: NewQuestion): boolean =>
   question.question === '' || question.ballot_options.length < MIN_BALLOT_OPTIONS;
 
 /**
+ * Checks whether a question title is not unique within a list of questions
+ * @param questions The list of questions
+ */
+const haveQuestionsSameTitle = (questions: NewQuestion[]): boolean => {
+  const questionTitles = questions.map((q: NewQuestion) => q.question);
+  return questionTitles.length === new Set(questionTitles).size;
+};
+
+/**
  * Creates a new election based on the given values and returns the related request promise
  * @param laoId The id of the lao in which the new election should be created
  * @param version The version of the lection that should be created
@@ -141,7 +150,10 @@ const CreateElection = () => {
 
   // Confirm button only clickable when the Name, Question and 2 Ballot options have values
   const confirmButtonEnabled: boolean =
-    isConnected === true && electionName !== '' && !questions.some(isQuestionInvalid);
+    isConnected === true &&
+    electionName !== '' &&
+    !questions.some(isQuestionInvalid) &&
+    haveQuestionsSameTitle(questions);
 
   const onCreateElection = () => {
     createElection(currentLao.id, version, electionName, questions, startTime, endTime)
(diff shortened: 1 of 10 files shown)


fe1-web/src/features/evoting/screens/CreateElection.tsx
```diff
@@ -285,6 +285,24 @@ const CreateElection = () => {
...
+          {STRINGS.event_creation_name_not_empty}
+        </Text>
+      )}
+      {questions.some(isQuestionInvalid) && (
+        <Text style={[Typography.paragraph, Typography.error]}>
+          {STRINGS.election_create_invalid_questions_1 +
+            MIN_BALLOT_OPTIONS +
+            STRINGS.election_create_invalid_questions_2}
+        </Text>
+      )}
```
Tyratox, line 304: We might also want to catch the case where there are multiple questions with the same name as this will cause question id collisions..

> "Question ID: Hash : HashLen('Question', election_id, question)"

(https://github.com/dedis/popstellar/blob/master/docs/messageData.md#setting-up-an-election-electionsetup)

fe1-web/src/resources/strings.ts
```diff
@@ -264,13 +267,16 @@ namespace STRINGS {
...
   export const modal_button_start_now = 'Start it now';
   export const modal_button_go_back = 'Cancel';
+  export const event_creation_must_be_connected = 'You must be connected to a LAO.';
+  export const event_creation_name_not_empty = 'Name must not be empty.';
+  export const event_creation_location_not_empty = 'Location must not be empty.';
 
   /* --- Cast Vote Strings --- */
   export const cast_vote = 'Cast Vote';
-  export const cast_vote_success = 'Sucessfully casted vote';
+  export const cast_vote_success = 'Successfully casted vote';
```
Tyratox, line 279: whoops 😳😂

Tyratox (approved):
LGTM! 😊 Finally some feedback on the forms 🥳

At first I was not sure whether we might want some additional form validation logic handled by a dependency (https://formik.org/, https://react-hook-form.com/) but given the simplicity of the validation and the fact that this is way way easier to understand I think it's way better like this :)

Tyratox:
And I guess the snapshots have to be fixed for the tests to pass :)

fe1-web/src/features/evoting/screens/CreateElection.tsx
Xelowak: Okay, I'm ont it!

Xelowak:
> And I guess the snapshots have to be fixed for the tests to pass :)

That is weird though because everything passes locally

fe1-web/src/features/evoting/screens/CreateElection.tsx
```diff
@@ -285,6 +297,29 @@ const CreateElection = () => {
...
+      {electionName === '' && (
+        <Text style={[Typography.paragraph, Typography.error]}>
+          {STRINGS.event_creation_name_not_empty}
+        </Text>
+      )}
+      {questions.some(isQuestionInvalid) && (
+        <Text style={[Typography.paragraph, Typography.error]}>
+          {STRINGS.election_create_invalid_questions_1 +
+            MIN_BALLOT_OPTIONS +
+            STRINGS.election_create_invalid_questions_2}
```
pierluca, line 314: This seems like a slightly fragile approach to handling a variable in a translation.
It's generally safer to have the full text in one string, with a variable that gets replaced here.
e.g. `myStr = "there are {} elephants"` , and `myStr.replace("{}",3)`, because then it becomes easy to change the text to say `"{} elephants were seen"` , or translate to a different language where `{}` would be in a different place in the string.

pierluca:
I see from the screenshots that the warning messages were all put at the bottom. 
Would it make sense to show them next to / above / below the field to which they relate ?

Tyratox:
Hmm I see arguments for both.. Preventing the UI from changing when errors appear / disappear is doable by having empty padding there if there is no error but I am unsure about the first point. I think we could have the best of both them by showing an error message directly below the field only if the field is dirty or the user tried pressing the submit button. We can then show an additional error at the bottom that is displayed if there are any errors in the whole form which tells the user to check the form inputs. But this will result in much more complex code and if we want to take this approach we might want to use a form library as they usually offer all / most of these functionalities?

Tyratox (approved):

pierluca:
Okay, I hadn't realized that the errors show up from the very beginning, without waiting for the fields to be dirty or the button pressed. That's a bit of an unusual choice, but I understand how it might make things more expedient.
I agree with @Tyratox that probably a library would simplify this, but this is already a useful increment.

PR #1382, opened 2023-02-13
Fix persistence bug

Warhedgehog:
This fixes the persistence bug where channels would not be saved for an unkown reason. Changing the lifecycle method where the saving is done is the workaround I've found.


pierluca (approved):
Ugly, but I can work with ugly & documented 😁

PR #1385, opened 2023-02-14
Adapt architecture diagram

Warhedgehog:


GabrielFleischer (approved):
As discussed, I completely agree with these changes

pierluca (approved):
Makes sense ! 👍

PR #1386, opened 2023-02-14
Update documentation of fe1

Xelowak:


fe1-web/docs/README.md
```diff
@@ -47,13 +47,20 @@ Here's the annotated directory tree:
 │
 │   ├── components               # library of simple, reusable UI component
 │   │
+│   ├── contexts                 # feature context type definition
+│   │
+│   ├── functions                # module containing functions that are useful in the whole application
+│   │
+│   ├── hooks                    # contains a wrapper for a useful react hook
```
Tyratox, line 54: Not sure whether we should be this specific in the documentation as maybe there will be multiple ones in this directory later?

fe1-web/docs/README.md
```diff
@@ -47,13 +47,20 @@ Here's the annotated directory tree:
...
+│   │
 │   ├── keypair                  # module dealing with the storage of a global, unique keypair
 │   │
 │   ├── navigation               # module dealing with the top-level React navigation
 │   │
 │   ├── network                  # module to network with the backend
 │   │   ├── ingestion              # implementation and configuration of the processing of incoming messages
 │   │   ├── jsonrpc                # network & protocol objects
+│   │   ├── strategies             # sending strategies when there are multiple server
```
Tyratox, line 63: Maybe my english is just bad but I think there should be an `s`?
```suggestion
│   │   ├── strategies             # sending strategies when there are multiple servers
```

fe1-web/docs/README.md
```diff
@@ -62,14 +69,23 @@ Here's the annotated directory tree:
...
 ├── features                 # independent features in the system
 │
-│   ├── connect                  # feature dealing with LAO/server connection
+│   ├── digital cash             # feature dealing with digital cash
+│   │
+│   ├── events                   # feature dealing with events happening in a LAO
+│   │
+│   ├── evoting                  # feature dealing with E-Voting and Elections
+│   │
+│   ├── home                     # feature dealing with the app's home screen
```
pierluca, line 84: should we put them after the `lao`, which provides the general structure of a feature ?

fe1-web/docs/README.md
pierluca: not sure I understand your comment @Tyratox . Do you mean the following ?
```suggestion
│   ├── hooks                    # contains utility react hooks
```

fe1-web/docs/README.md
Xelowak: Sure we can! It's just that I opted for the same order as the one in the IDE, which is alphabetical. But I agree it would be clearer this way, so I'm going to change it.

fe1-web/docs/README.md
Xelowak: True!

Tyratox (approved):

PR #1394, opened 2023-02-16
Add back navigation for selected fragments

Warhedgehog:
This PR adds back navigation for non tabs fragments in `LaoActivity`. Back navigation for tabs fragment (i.e. those that you can select from the navigation drawer) is still to be done.


pierluca (approved):
LGTM. Just wondering if there is any way to make this less repetitive 🤔

Warhedgehog:
Is it a bit better like that ?

pierluca (approved):
LGTM !

PR #1411, opened 2023-02-23
Resolved IntelliJ missing protocol folder copy

RaulinN:
+ Added IntelliJ troubleshooting tips

### DIFF
--- a/be2-scala/build.sbt
+++ b/be2-scala/build.sbt
@@ -42,13 +42,13 @@ Scapegoat/ scalacOptions -= "-Xfatal-warnings"
 Scapegoat/ scalacOptions += "-P:scapegoat:overrideLevels:all=Warning"
 
 // Reload changes automatically
-Global / onChangedBuildSource := ReloadOnSourceChanges
-Global / cancelable := true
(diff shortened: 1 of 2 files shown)


be2-scala/build.sbt
```diff
@@ -76,7 +76,7 @@ copyProtocolTask := {
 }
 
 // Add the copyProtocolTask to compile and test scopes
-(Compile/ compile) := ((Compile/ compile) dependsOn copyProtocolTask).value
+Compile / compile := (Compile / compile).dependsOn(copyProtocolTask).value
```
Ajkunas, line 79: The syntax of the line changed a bit, and it is not the same as the line 80. Was it intentional? If not, maybe just sticking with the previous one would be better in terms of code uniformity. (minor comment btw)

Ajkunas (commented):
Everything is fine for me, just a minor comment on the build.sbt file.

be2-scala/build.sbt
RaulinN: You are right! Added consistency throughout the whole file =)

Issue #1387, opened 2023-02-14
Prevent opening any roll call when one is already opened 


Warhedgehog:
While it is (or at least should be) handled by the backend, the organizer should not be able to click on open of a roll call if any other roll call is open.

PR #1414, opened 2023-02-26
Fix the opening of a roll call when another one is opened

matteosz:
This PR addresses issue #1387. 
Specifically, it checks whether there are any other roll calls currently opened before allowing the user to open a new one. This removes the need for the backend server to perform this check.

Changes made include:

- Added a new function `canOpenRollCall()` that verifies if there are any open roll calls in the RollCallRepository for the given LAO

- Added a new exception `DoubleOpenedRollCallException` to handle the error

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/event/rollcall/RollCallViewModel.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/event/rollcall/RollCallViewModel.java
@@ -144,7 +144,12 @@ public class RollCallViewModel extends AndroidViewModel implements QRCodeScannin
     }
 
     if (!rollCallRepo.canOpenRollCall(laoId)) {
-      Log.d(TAG, "failed to open roll call with id " + id + ", laoID: " + laoView.getId());
+      Log.d(
+          TAG,
+          "failed to open roll call with id "
+              + id
+              + " because another roll call was already opened, laoID: "
+              + laoView.getId());
       return Completable.error(new DoubleOpenedRollCallException(id));
     }
 
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/event/rollcall/RollCallViewModel.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/event/rollcall/RollCallViewModel.java
@@ -142,6 +142,12 @@ public class RollCallViewModel extends AndroidViewModel implements QRCodeScannin
       ErrorUtils.logAndShow(getApplication(), TAG, e, R.string.unknown_lao_exception);
       return Completable.error(new UnknownLaoException());
     }
+
+    if (!rollCallRepo.canOpenRollCall(laoId)) {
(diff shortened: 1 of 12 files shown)


pierluca:
Don't forget to associate the e-mail you use in your commits with your GitHub account, sign the [CLA](https://github.com/dedis/popstellar/pull/1414#issuecomment-1445380670) and ask a team mate for a review :)

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/event/rollcall/RollCallViewModel.java
```diff
@@ -142,11 +142,17 @@ public Completable openRollCall(String id) {
       ErrorUtils.logAndShow(getApplication(), TAG, e, R.string.unknown_lao_exception);
       return Completable.error(new UnknownLaoException());
     }
+
+    if (!rollCallRepo.canOpenRollCall(laoId)) {
+      Log.d(TAG, "failed to open roll call with id " + id + ", laoID: " + laoView.getId());
```
simone-kalbermatter, line 147: The log message could be more precise (failed because another roll call was already open)

fe2-android/app/src/test/unit/java/com/github/dedis/popstellar/utility/handler/RollCallHandlerTest.java
```diff
@@ -148,6 +147,9 @@ public void testHandleCreateRollCall()
   public void testHandleOpenRollCall()
       throws DataHandlingException, UnknownLaoException, UnknownRollCallException,
           UnknownElectionException, NoRollCallException {
+    // Check if the Roll Call can be opened
+    assertTrue(rollCallRepo.canOpenRollCall(LAO.getId()));
```
simone-kalbermatter, line 151: It might be better to create a dedicated test for the blocking functionality instead of inserting it into both testHandleOpenRollCall() and testHandleCloseRollCall()

simone-kalbermatter (dismissed):

fe2-android/app/src/test/unit/java/com/github/dedis/popstellar/utility/handler/RollCallHandlerTest.java
matteosz: Yes, you're right, it's definetely cleaner like that

simone-kalbermatter (approved):

Warhedgehog (approved):
Codewise this is perfect ! Impressive work, well documented and tested.

I wonder on a spec level about rejecting network messages that open a roll call when one is already open. Missing a close rc message would put the app in an undesired state from which there would be no easy escape.
What do you think @jbsv ?

jbsv (approved):
LGTM (looks good to me)

jbsv:
> I wonder on a spec level about rejecting network messages that open a roll call when one is already open. Missing a close rc message would put the app in an undesired state from which there would be no easy escape. What do you think @jbsv ?

Good point. Now if I understand correctly, this PR fixes the issue at the user interaction level. @matteosz can you please open a new issue for the network message filtering, and I would recommend to discuss this with the fe1 group or in the PoP party meeting.

Issue #1380, opened 2023-02-13
The election protocol should not allow multiple questions with the same name


GabrielFleischer:
As the question id depends on the election id and the question name, two questions with the same name yields to the same clashing id.

This issue should be addressed in the protocol and the front-ends should disallow it in the UI.

MeKHell:
This is already fixed in fe1.

PR #1418, opened 2023-02-28
Addresses #1380 by disallowing multiple questions with the same ID in the election protocol 

MariemBaccari:
Addresses #1380 : Previously, the election protocol allows the creation of a channel with multiple questions with the same ID. The suggested fix follows the fail-fast concept and does not allow the creation of a channel in that case.


be1-go/channel/election/mod.go
```diff
@@ -819,12 +823,16 @@ func updateVote(msgID string, sender string, castVote messagedata.VoteCastVote,
...
+// Creates the questions for the election channel or returns nil if they are not valid
 func getAllQuestionsForElectionChannel(questions []messagedata.ElectionSetupQuestion) map[string]*question {
 	qs := make(map[string]*question)
 	for _, q := range questions {
 		ballotOpts := make([]string, len(q.BallotOptions))
 		copy(ballotOpts, q.BallotOptions)
-
+		_, exists := qs[q.ID]
+		if exists {
+			return nil
```
Nazianzenov, line 834: maybe log an error here?

MariemBaccari: Thanks, I just added that in a new commit.

be1-go/go.mod
```diff
@@ -25,7 +25,7 @@ require (
 	github.com/shurcooL/sanitized_anchor_name v1.0.0 // indirect
 	go.dedis.ch/fixbuf v1.0.3 // indirect
 	golang.org/x/crypto v0.0.0-20210921155107-089bfa567519 // indirect
-	golang.org/x/sys v0.0.0-20211019181941-9d821ace8654 // indirect
+	golang.org/x/sys v0.3.0 // indirect
```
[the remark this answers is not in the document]
MariemBaccari: From what I read online go.mod and go.sum automatically keep track of the dependencies so could using a new json file (in the test method) have caused those modifications ?

be1-go/channel/election/mod.go
Anghul: Why not directly returning an error ? And checking the error value when the function is called ?

Anghul (changes requested):

be1-go/channel/election/mod.go
MariemBaccari: The function is called when creating an election channel using a constructor. I tried implementing the function by having it return two values (the map and an error) but it seems that constructors in go don't allow you to initialize two fields in the same line (cf attached image).
getAllQuestions

Anghul: You would need to initialize `questions` outside the constructor, the same way `registry` is done. You would get something like:

```
newChannel := &Channel{
...
}

questions, err := getAllQuestionsForElectionChannel()
if err != nil {
    return
}
newChannel.questions = questions
```

Anghul (approved):
Looking good to me now

protocol/examples/messageData/election_setup/bad_election_setup_question_identical_id.json
```diff
@@ -0,0 +1,27 @@
...
+    },
+    {
+      "id": "2PLwVvqxMqW5hQJXkFpNCvBI9MZwuN8rf66V1hS-iZU=",
+      "question": "Is this project fun?",
+      "voting_method": "Plurality",
+      "ballot_options": ["Yes", "Yes"],
+      "write_in": true
+    }
+  ]
+}
```
Tyratox, line 27: Great that you saw the protocol folder! 😁

Usually in addition to the examples there is a test file that makes sure that bad examples do not pass the json schema validation and good examples do. You can find it at https://github.com/dedis/popstellar/blob/master/protocol/test/main.test.js. But to be fair I don't think this is something that can be checked with JSON schemas by default. Although it seems the validator we're using (at least in protocol tests and also on fe-1) has a keyword https://github.com/ajv-validator/ajv-keywords#uniqueitemproperties that can be used for this :)

Moreover it would be good to document the change in https://github.com/dedis/popstellar/blob/master/docs/messageData.md#setting-up-an-election-electionsetup.

MariemBaccari: Thank you for pointing that out ! I added the example in main.test.js. It looks like this case can be checked since https://github.com/dedis/popstellar/blob/master/docs/messageData.md#setting-up-an-election-electionsetup already specifies "uniqueItems" as true. Considering this, I think the docs is already complete (probably thanks to fe1 because I noticed that this bug was fixed from their end). How does it look now ?

protocol/examples/messageData/election_setup/bad_election_setup_question_identical_id.json
```diff
@@ -0,0 +1,27 @@
...
+      "question": "Is this project fun?",
+      "voting_method": "Plurality",
+      "ballot_options": ["Yes", "No"],
+      "write_in": false
+    },
+    {
+      "id": "2PLwVvqxMqW5hQJXkFpNCvBI9MZwuN8rf66V1hS-iZU=",
+      "question": "Is this project fun?",
+      "voting_method": "Plurality",
+      "ballot_options": ["Yes", "Yes"],
```
Tyratox, line 23: This is not allowed by the scheme so while it is correct that validation of this will always fail it is for the wrong reason 😅

```suggestion
      "ballot_options": ["Yes", "No", "Prefer not to answer"],

```

protocol/examples/messageData/election_setup/bad_election_setup_question_identical_id.json
Tyratox: Thank you for adding it! :)

I just double checked since `uniqueItems` should not check the individual properties of the array elements but just the whole objects and it turns out that the test passed (i.e. validation failed) because of other issues with the file that I missed before. I added comments in the respective places.

Hence I'd still update the docs a bit to make sure that the "question" property has to be unique. You can use the "$comment" field for this.

Also I quickly checked out `uniqueitemproperties` that I linked in my previous comment and it seems there is only partial support for it and since it might lead to other systems having issues with a non-standard key (genuinely don't know but I think if we want to enforce it, it should be enforced on all subsystems). I would suggest to either forget about this and make sure the documentation & implementation make clear that the question property has to be unique OR look for some other solution?

Also I just reviewed the changes regarding the protocol, not the changes in the go code. You will need to ask @Anghul for a review of it 😄

MariemBaccari: Thank you for the feedback ! I just pushed the changes you suggested: updating the docs and just commenting the test as one that cannot be checked by json schemas like some other ones above it.

Tyratox (commented):
Looks good to me now 👍 Minor detail: The documentation (`/docs`) and the json schemas (`/protocol/query/method/message/data`) should always have a 1:1 correspondence but given I forgot to mention this to you I quickly pushed this change myself. Also you added a trailing comma to the JSON which is not allowed but I quickly removed it 😅

Just pointing it out for future protocol changes always update 1) the actual file used for validation in `/protocol/query/`, 2) the examples in `/protocol/examples`, 3) the tests in `/protocol/test` and 4) the documentation :) 

(I know this was just a change in a comment which does not affect validation but still)

@Anghul Can this PR be approved? Just not sure since you did not explicitly approve the PR on github before :)

Tyratox:
Also since you started out on fixing this issue on your subsystem you should ensure all subsystems are consistent and do not break because of each other.

I know fe-1 checks to prevent duplicate question names but merging this PR without having fe-2 also add such a check might lead to fe-2 crashing if it allowed for two questions with the same name (depending on the implementation since the server then replies with an error).

MariemBaccari:
@Tyratox Thank you for clarifying the procedure for protocol changes ! I believe that @Anghul approved the go changes two days ago when he left a comment but it might have been "cancelled" when you requested more changes. I will also discuss this issue with fe-2 before merging.

Tyratox (approved):
Here you go then :)

Anghul (approved):
I confirm, my approval was removed when the changes were pushed (to avoid bypassing the need for approval). It still looks good to me.

PR #1419, opened 2023-02-28
Fixing the out-of-order processing for Deleting reactions

Nazianzenov:
As for the deleting chirp issue, introduced a small delay before reprocessing a delete reaction in case an add reaction was published out-of-order.


MariemBaccari (approved):
Similarly to #1415, this looks good to me !

Anghul (approved):
LGTM

PR #1423, opened 2023-03-02
Fixing the out-of-order processing for Closing Roll Calls.

Nazianzenov:
When a roll call opens and closes quickly, it can happen that the open message is received after the close message. Fixing that issue is done by retrying to process the close message after some delay. Also replaced some method argument's names by _ when unused in the lao package.

### DIFF
--- a/be1-go/channel/lao/mod.go
+++ b/be1-go/channel/lao/mod.go
@@ -26,6 +26,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


Nazianzenov:
After reviewing myself the code, it seems like it is a bad idea to tackle out-of-order roll calls creations and deletions, especially because it is unlikely such a situation will happen ( roll calls are usually opened  during a few minutes).

PR #1425, opened 2023-03-03
Prevent the user to submit an election with duplicated questions names

matteosz:
This addresses the issue #1380 in response to the PR #1418.

The new version checks whether the user types a question with the same name as a previous one; in that case an error box notifies the problem and the submit button becomes unavailabe to create the election.

Changes made include:

- Added a new function `hasDuplicate()`, which verifies if in a list of strings there is any duplicate (case-insensitive comparison).

- Added the check perfomed by that function to the ViewHolder of the question title, within the function `checkIfAnInputIsValid()`. If a duplicate in the list of questions is found, a flag is set to False to prevent the user pressing on submit.

- Added the relative tests to the code written.

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/event/election/adapters/ElectionSetupViewPagerAdapter.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/event/election/adapters/ElectionSetupViewPagerAdapter.java
@@ -85,12 +85,17 @@ public class ElectionSetupViewPagerAdapter
             // On each text change we edit the list of questions
             // and we add or remove the question from the list of filled question
             String questionText = s.toString();
+            // Prevents the user from creating two different questions with the same name
+            if (questions.contains(questionText)) {
+              electionQuestionText.setError("Two different questions can't have the same name");
+            }
             if (!electionQuestionText.getText().toString().trim().isEmpty()) {
               questions.set(holder.getAdapterPosition(), questionText);
               listOfValidQuestions.add(holder.getAdapterPosition());
             } else {
               listOfValidQuestions.remove(holder.getAdapterPosition());
             }
+
             checkIfAnInputIsValid();
           }
         });
@@ -162,6 +167,12 @@ public class ElectionSetupViewPagerAdapter
       }
     }
 
+    // Check if there's a duplicate question, if true invalidate the input
+    if (hasDuplicate(questions)) {
+      isAnInputValid.setValue(false);
+      return;
+    }
+
     for (Integer i : listOfValidQuestions) {
       if (listOfValidBallots.contains(i)) {
         isAnInputValid.setValue(true);
@@ -182,6 +193,11 @@ public class ElectionSetupViewPagerAdapter
       }
     }
     return listOfValidInputs;
+  }
+
+  private boolean hasDuplicate(List<String> strings) {
+    Set<String> uniqueStrings = new HashSet<>();
+    return strings.stream().map(String::toLowerCase).anyMatch(s -> !uniqueStrings.add(s));
   }
 
   /** Adds a view of ballot option to the layout when user clicks the button */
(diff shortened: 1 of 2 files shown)


fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/event/election/adapters/ElectionSetupViewPagerAdapter.java
```diff
@@ -85,12 +85,18 @@ public void afterTextChanged(Editable s) {
             // On each text change we edit the list of questions
             // and we add or remove the question from the list of filled question
             String questionText = s.toString();
+
+            // Prevents the user from creating two different questions with the same name
+            if (questions.contains(questionText)) {
```
simone-kalbermatter, line 90: Just a question: In line 239, why do we check `!text.isEmpty()` but here we don't?

simone-kalbermatter (approved):
Looks good to me!

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/event/election/adapters/ElectionSetupViewPagerAdapter.java
```diff
@@ -184,6 +196,18 @@ public List<Integer> getValidInputs() {
...
+  /**
+   * This function checks if a given list of strings has at least a duplicate. The comparison is
+   * performed case-insensitive.
+   *
+   * @param strings List of Strings
+   * @return true if the list of strings has at least a duplicate, false otherwise
+   */
+  public static boolean hasDuplicate(List<String> strings) {
+    Set<String> uniqueStrings = new HashSet<>();
+    return strings.stream().map(String::toLowerCase).anyMatch(s -> !uniqueStrings.add(s));
```
pierluca, line 208: A case insensitive comparison makes sense from a usability perspective, but you might want to sync up with fe1 to make sure that you have the same definition of a valid input :)

matteosz: Right, going to change that!

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/event/election/adapters/ElectionSetupViewPagerAdapter.java
matteosz: You've spotted a very subtle bug indeed! I didn't insert that check, as the next piece of code adds the string to the `questions` list only when it's not empty (lines 93-98). Also, that method is called when the text is changed, so even if at the starting point it's empty, it's not a problem. However, it could happen that a user writes the first question name, creates a new question, and goes back to the first one, clearing the text.

Unfortunately, even if we add the isEmpty() check as you suggested, there's still another potential issue. A user could write a certain title (e.g. 'A'), create another question, go back to the first one and clear its text. Then the application doesn't allow the user to write the same title ('A') for the second question, as it has already been inserted in the `questions` list and not removed (due to the lines 93-98, which leave the question list unchanged if the text is empty).

So, I need to modify those lines to address this scenario.

matteosz:
- Solved the issue discussed in the comments above. The solution follows the FE1 current implementation methodology: the string comparison is case-sensitive and the content is not trimmed (e.g. “A “ is different from “a “ and different from “A”).
- Additionally, when testing the FE1 in the election creation, I discovered a bug that here in FE2 is already addressed. The issue is presented here: #1428

simone-kalbermatter (approved):
Looks good to me!

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/event/election/adapters/ElectionSetupViewPagerAdapter.java
```diff
@@ -85,12 +85,18 @@ public void afterTextChanged(Editable s) {
...
+            if (!questionText.isEmpty() && questions.contains(questionText)) {
+              electionQuestionText.setError("Two different questions can't have the same name");
+            }
             if (!electionQuestionText.getText().toString().trim().isEmpty()) {
-              questions.set(holder.getAdapterPosition(), questionText);
               listOfValidQuestions.add(holder.getAdapterPosition());
             } else {
               listOfValidQuestions.remove(holder.getAdapterPosition());
             }
+            questions.set(holder.getAdapterPosition(), questionText);
```
Warhedgehog, line 98: Why add the question text when it's empty ?

Warhedgehog (changes requested):
Looks very good, just the one question :)

Warhedgehog (approved):
LGTM

PR #1426, opened 2023-03-03
Refactored the MeetingValidator class

BelghmiAmine:


be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/MeetingValidator.scala
```diff
@@ -39,23 +39,39 @@ sealed class MeetingValidator(dbActorRef: => AskableActorRef) extends MessageDat
...
+          ),
+          checkTimestampStaleness(
+            rpcMessage,
+            data.start,
+            validationError(s"stale 'start' timestamp (${data.start})")
+          ),
+          checkTimestampOrder(
+            rpcMessage,
+            data.creation,
+            data.end.get,
```
K1li4nL, line 56: I think that line could be problematic, if data.end is undefined an exception will be thrown instead of using the error pipeline.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/MeetingValidator.scala
```diff
@@ -39,23 +39,39 @@ sealed class MeetingValidator(dbActorRef: => AskableActorRef) extends MessageDat
...
+            validationError(s"'end' (${data.end.get}) timestamp is smaller than 'creation' (${data.creation})")
+          ),
+          checkTimestampOrder(
+            rpcMessage,
+            data.start,
+            data.end.get,
```
K1li4nL, line 62: Same as above, I think that line could be problematic, if data.end is undefined an exception will be thrown instead of using the error pipeline.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/MeetingValidator.scala
```diff
@@ -70,24 +86,52 @@ sealed class MeetingValidator(dbActorRef: => AskableActorRef) extends MessageDat
...
+          ),
+          checkTimestampStaleness(
+            rpcMessage,
+            data.start,
+            validationError(s"stale 'start' timestamp (${data.start})")
+          ),
+          checkTimestampOrder(
+            rpcMessage,
+            data.creation,
+            data.end.get,
```
K1li4nL, line 109: Same as above, I think that line 109 could be problematic, if data.end is undefined an exception will be thrown instead of using the error pipeline.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/MeetingValidator.scala
```diff
@@ -70,24 +86,52 @@ sealed class MeetingValidator(dbActorRef: => AskableActorRef) extends MessageDat
...
+            validationError(s"'end' (${data.end.get}) timestamp is smaller than 'creation' (${data.creation})")
+          ),
+          checkTimestampOrder(
+            rpcMessage,
+            data.start,
+            data.end.get,
```
K1li4nL, line 115: Same as above, I think that line 115 could be problematic, if data.end is undefined an exception will be thrown instead of using the error pipeline.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/MeetingValidator.scala
Ajkunas: Absolutely! As you can see in the initial code, we first check if the `data.end` is defined (i.e. not `None`) and only after that we perform the actual check we want. 
Try to find a nice and clear way to do the same thing :) And think about something that can be used in similar cases (try to generalize as much as possible)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/MeetingValidator.scala
```diff
@@ -70,24 +86,52 @@ sealed class MeetingValidator(dbActorRef: => AskableActorRef) extends MessageDat
         val laoId: Hash = rpcMessage.extractLaoId
         val expectedHash: Hash = Hash.fromStrings(EVENT_HASH_PREFIX, laoId.toString, data.creation.toString, data.name)
```
Ajkunas, line 87: You can split up this line, so that the code is a bit more readable.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/MeetingValidator.scala
```diff
@@ -70,24 +86,52 @@ sealed class MeetingValidator(dbActorRef: => AskableActorRef) extends MessageDat
         val laoId: Hash = rpcMessage.extractLaoId
```
Ajkunas, line 86: You can use the function `extractData` from the `MessageValidator.scala` file. Thus, you can make your code more concise.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/MeetingValidator.scala
```diff
@@ -70,24 +86,52 @@ sealed class MeetingValidator(dbActorRef: => AskableActorRef) extends MessageDat
...
+            data.modification_signatures,
+            data.modification_id,
+            validationError("witness key-signature pairs are not valid for the given modification_id")
+          )
+        )
       case _ => Right(validationErrorNoMessage(rpcMessage.id))
     }
   }
+
+  private def checkWitnessSignatures(rpcMessage: JsonRpcRequest, witnessesKeyPairs: List[WitnessSignaturePair], data: Hash, error: PipelineError): GraphMessage = {
```
Ajkunas, line 130: This overlaps with the PR made by Kilian on the Lao refactoring. Same comment as him, this function could be moved in the same place as `validateWitnessSignatures`. 
Also, when defining new functions, do not forget to comment them for more clarity :)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/MeetingValidator.scala
```diff
@@ -4,8 +4,8 @@ import ch.epfl.pop.model.network.JsonRpcRequest
 import ch.epfl.pop.model.network.method.message.Message
 import ch.epfl.pop.model.network.method.message.data.ObjectType
 import ch.epfl.pop.model.network.method.message.data.meeting.{CreateMeeting, StateMeeting}
-import ch.epfl.pop.model.objects.{Channel, Hash, PublicKey}
-import ch.epfl.pop.pubsub.graph.validators.MessageValidator._
+import ch.epfl.pop.model.objects.{Channel, Hash, PublicKey, WitnessSignaturePair}
+import ch.epfl.pop.pubsub.graph.validators.MessageValidator.{checkOwner, _}
```
Ajkunas, line 8: I guess it is better to simply keep the underscore, since `checkOwner` is already included into it.

Ajkunas (changes requested):
I will let you take into account the comments we gave you. There is minor changes, otherwise the refactoring is good. 

But be careful with the naming of your branch. Here you have refactored the MeetingHandler, but your branch suggests that it is the ElectionHandler. It can be confusing. 
Same comment for your commits messages. I suggest you to change the message by using the commant `git rebase`.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/MeetingValidator.scala
```diff
@@ -38,10 +37,9 @@ sealed class MeetingValidator(dbActorRef: => AskableActorRef) extends MessageDat
 
         val sender: PublicKey = message.sender
         val channel: Channel = rpcMessage.getParamsChannel
-        if (!data.end.isDefined)
-          Right(validationError("data.end is not defined"))
 
         runChecks(
+          checkDataEndIsDefined(rpcMessage, data.end, validationError("data.end is not defined")),
```
K1li4nL, line 42: I don't think this behavior is correct, from the protocol spec for [Meeting#Create](https://github.com/dedis/popstellar/blob/master/docs/messageData.md#creating-a-meeting-meetingcreate)
the end of the meeting is optional so an error should only be thrown if an end is defined but invalid (e.g the end is before the beginning)

Ajkunas: Nice comment. Thank you for reminding us the protocol (forgot it as well oops).

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/MessageDataContentValidator.scala
```diff
@@ -161,4 +161,41 @@ trait MessageDataContentValidator extends ContentValidator with AskPatternConsta
...
+    * @param witnessesKeyPairs
+    *   the witness key signature pairs
+    * @param data
+    *
+    * @param error
+    *   the error to forward in case of invalid modifications
+    * @return
+    *   GraphMessage: passes the rpcMessages to Left if successful, else, right with pipeline error.
+    */
+  final def checkWitnessSignatures(rpcMessage: JsonRpcRequest, witnessesKeyPairs: List[WitnessSignaturePair], dataEnd: Hash, error: PipelineError): GraphMessage = {
```
Ajkunas, line 178: Kilian has merged his previous PR where he has already defined this function. Try to merge the master branch into your branch.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/MeetingValidator.scala
Ajkunas: I reiterate with this comment. Try to use `extractData`.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/MeetingValidator.scala
```diff
@@ -39,23 +38,40 @@ sealed class MeetingValidator(dbActorRef: => AskableActorRef) extends MessageDat
         val sender: PublicKey = message.sender
```
Ajkunas, line 38: You can also use `extractData` here to make your code more concise.

Ajkunas (changes requested):
Still some little things to change. 
Take a particular attention to Kilians comment who reminded us the protocol. So, the idea of defining a function that checks that the value is defined will not be really useful here.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/MeetingValidator.scala
BelghmiAmine: Okey, I get it, so the idea would be to run the checks concerning data.end only in the case where data.end is defined, if it is not, then no need to run these checks since they will throw an exception ?

Ajkunas: Exactly :)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/MessageDataContentValidator.scala
```diff
@@ -58,6 +58,15 @@ trait MessageDataContentValidator extends ContentValidator with AskPatternConsta
       Right(error)
   }
 
+  /** This method behaves the same as checkTimestampOrder, except that the param <second> is not necessarily defined. (The protocol states that <data.en> is not necessarily defined for certain type of messages as CreateMeeting)
```
Ajkunas, line 61: `(Wrt to the protocol, some fields are not necessarily defined for certain type of messages, such as CreateMeeting)`

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/MessageDataContentValidator.scala
```diff
@@ -58,6 +58,15 @@ trait MessageDataContentValidator extends ContentValidator with AskPatternConsta
...
+    */
+  def checkOptionalTimestampOrder(rpcMessage: JsonRpcRequest, first: Timestamp, second: Option[Timestamp], error: PipelineError): GraphMessage = {
```
Ajkunas, line 63: Good idea! But you have not used the function everywhere it is needed.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/MeetingValidator.scala
Ajkunas: Here you can use your function `CheckOptionalTimestampOrder`.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/MeetingValidator.scala
Ajkunas: Here you can use your function CheckOptionalTimestampOrder.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/MeetingValidator.scala
Ajkunas: See above.

Ajkunas (changes requested):
You forgot to use your new function `checkOptionalTimestampOrder` in some places, and try to use `extractData`. :)

Ajkunas (approved):
LGTM

PR #1429, opened 2023-03-04
CoinValidator Refactoring

BelghmiAmine:


K1li4nL (dismissed):
All good to me

Ajkunas:
You need to change back your base branch to `work-be2-amine-ElectionHandlerRefactoring` because there is still the changes of the refactoring of MeetingValidator that appear.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/CoinValidator.scala
```diff
@@ -17,18 +17,36 @@ case object CoinValidator extends MessageDataContentValidator {
...
     }
   }
+  private def checkTransactionId(rpcMessage: JsonRpcRequest, transactionId: Hash, expectedTransactionId: Hash, error: PipelineError): GraphMessage = {
+    if (transactionId == expectedTransactionId)
+      Left(rpcMessage)
+    else {
+      Right(error)
+    }
+  }
+  private def checkTransactionSignatures(rpcMessage: JsonRpcRequest, transaction: Transaction, error: PipelineError): GraphMessage = {
```
Ajkunas, line 37: Maybe some comments for these two new functions. :)

Ajkunas (changes requested):
See my comments, there are some little changes to be done.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/CoinValidator.scala
```diff
@@ -17,18 +17,36 @@ case object CoinValidator extends MessageDataContentValidator {
...
+    if (transaction.checkSignatures())
+      Left(rpcMessage)
+    else {
+      Right(error)
+    }
+  }
+  private def checkSumOutputs(rpcMessage: JsonRpcRequest, transaction: Transaction): GraphMessage = {
```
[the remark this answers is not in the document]
Ajkunas: I think you can also give as a parameter the `error: PipelinError` to this function. Thus, the functions are consistant between themself.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/CoinValidator.scala
```diff
@@ -17,18 +17,51 @@ case object CoinValidator extends MessageDataContentValidator {
...
-          Right(validationError("bad signature"))
-        } else {
-          data.transaction.sumOutputs() match {
-            case Left(err) => Right(validationError(err.getMessage()))
-            case Right(_)  => Left(rpcMessage)
-          }
-        }
+
+        runChecks(
+          checkId(rpcMessage, data.transactionId, data.transaction.transactionId, validationError("incorrect transaction id")),
```
Ajkunas, line 22: Also try to split up those functions for readability.

Ajkunas (changes requested):
Last little comments. I am happy to approve your PR after you take them in consideration. :)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/CoinValidator.scala
```diff
@@ -17,18 +17,51 @@ case object CoinValidator extends MessageDataContentValidator {
...
+    *   rpc message to validate
+    * @param transaction
+    *   The transaction to be checked
+    * @param error
+    *   the error to forward in case the transaction's sum is not valid.
+    * @return
+    *   GraphMessage passes the rpcMessages to Left if successful, else right with pipeline error
+    */
+  private def checkSumOutputs(rpcMessage: JsonRpcRequest, transaction: Transaction, error: PipelineError): GraphMessage = {
+    def validationError(reason: String): PipelineError = super.validationError(reason, "PostTransaction", rpcMessage.id)
```
Ajkunas, line 58: You do not need to define this function inside. It is already defined in the class (see line 15). We simply redefine the validationError which concerns the message, thus we do it inside the class, because it is used in every validators.

Ajkunas: Mmh after those changes, I believe that the optimal implementation was to keep the same structure, aka: 
`transaction match {
            case Left(err) => Right(validationError(err.getMessage()))
            case Right(_)  => Left(rpcMessage)`

Sorry, I was not realizing that then you would need to redefine a variable :(
I think keeping it as a match case is also more clear. And then I guess you also do not need to put the `error: PipelineError` in the parameters (Sorry again :( )

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/CoinValidator.scala
Ajkunas: I reiterate with this comment.

Ajkunas (changes requested):
Okeey I made a bad suggestion previously, my bad ! 
I will let you see my comments, but the initial implementation of `checkSumOutputs` was way more optimal, just realizing it now. 

Do not forget to split up the functions with too many lines. 

After this, I am happy to approve your PR :)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/CoinValidator.scala
BelghmiAmine: The code doesn't compile when I delete line 58 (def validationError()...
)

Ajkunas: Mmmh okeey I see now why it does not compile.. 
Indeed you need to access the `err` to use it inside the `validationError` function. 

For now, the parameter `error: PipelineError` is not used inside your function. Can't you try to give as a parameter the function `validationError`, and then use it inside. So, that you do not need to redefine a local one inside.

Ajkunas: So, you delete the parameter `error: PipelineError` and give as a parameter `validationError: String => PipelineError` instead.

K1li4nL (approved):

PR #1430, opened 2023-03-05
Adresses 1399 - Returning an error when processing a message received twice 

MariemBaccari:
This PR adresses the issue #1399. Before, when a server receives a "publish" message for the second time, it used to simply log an error. This fix makes it return an error instead. It also splits the Test_Handle_Publish test in two parts: one to test receiving a publish from a client and another one from a server since the previous test used to test both at the same time and would fail since the fix doesn't allow receiving a message twice. A new test for handling receiving a publish message twice is also added.

### DIFF
--- a/be1-go/hub/standard_hub/message_handling.go
+++ b/be1-go/hub/standard_hub/message_handling.go
@@ -3,7 +3,6 @@ package standard_hub
 import (
 	"encoding/base64"
 	"encoding/json"
-	"github.com/rs/zerolog/log"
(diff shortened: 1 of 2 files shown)


Nazianzenov (approved):
Looks good to me, tests are really clean!

Anghul (approved):
Good job, really clean and well commented tests. Splitting the tests for the handling message from clients and servers is also cleaner than keeping them in the same test.

Issue #1409, opened 2023-02-22
Replace three dots by trash can in socal media feature


Tyratox:
Be sure to add a confirmation modal (there is a ConfirmModal component) to make sure chirps are not deleted accidentally.

<img width="562" alt="Screenshot 2023-02-22 at 12 32 44" src="https://user-images.githubusercontent.com/2634739/220608501-92d89eca-ca29-41fc-ad02-2adb888b9570.png">

<img width="562" alt="Screenshot 2023-02-22 at 12 32 45" src="https://user-images.githubusercontent.com/2634739/220608510-963a4076-d058-4446-bb2f-c6243a6e4aac.png">

Yasir761:
I want to work on this but I didn't get what to add in replace of three dots icon.

Tyratox:
I thought a trash can would for instance be more user friendly as currently the only action hidden behind the three dots is the chirp's deletion which means we can also just put that one action there instead of the three dots.

Most likely a confirmation modal is still desired to prevent accidental deletions of chirps.

Does this make sense?

PR #1432, opened 2023-03-06
 Replace three dots by trash can in socal media feature #1409

dayan9265:
This pull request addresses the feature #1409: to replace the option list on the chirp with a simple delete button (with a confirmation window). This was done since currently the option list only had the delete option.


fe1-web/src/features/social/components/ChirpCard.tsx
```diff
@@ -255,16 +245,23 @@ const ChirpCard = ({ chirp, isFirstItem, isLastItem }: IPropTypes) => {
...
+                    onPress={() => setShowDeleteConfirmation(true)}
                     size="small"
                     buttonStyle="secondary"
                     toolbar
                   />
+                  <ConfirmModal
+                    onConfirmPress={deleteChirp}
+                    visibility={showDeleteConfirmation}
+                    description="Are you sure you want to delete this chirp?"
+                    title="Confirm Deletion"
```
Tyratox, line 262: We try to put all strings in `src/resources/strings.ts`. While this solution is not perfect it allows for easy translation of the app and without having to search the whole application for static strings :)

fe1-web/src/features/social/components/ChirpCard.tsx
```diff
@@ -157,13 +153,7 @@ const ChirpCard = ({ chirp, isFirstItem, isLastItem }: IPropTypes) => {
 
   const listStyle = List.getListItemStyles(isFirstItem, isLastItem);
 
-  const actionSheetOptions: ActionSheetOption[] = [];
-  if (isSender && !chirp.isDeleted && isConnected) {
-    actionSheetOptions.push({
-      displayName: STRINGS.social_media_delete_chirp,
```
Tyratox, line 163: Also regarding the strings file: If you delete a `STRINGS.x` statement, check if it is still used somewhere else and delete it from `strings.ts` if it is not. Maybe we can also add some linter rule to check for this in the future 🤔

Tyratox (changes requested):
Great work!! 🥳 It seems you already figured out how to use react components, react hooks and even how to re-generate the snapshot tests 😁 And it is great to see that you updated the corresponding test, you even have 100% coverage of the PR 🥳😁

I just added two minor comments regarding the strings file, otherwise the PR looks good to me and I'm happy to approve it when the static strings are gone :)

For the future it would be nice to post a screenshot with a PR but it is fine for this one now 😄

dayan9265:
Thank you for the feedback!  Indeed I did not see that we had a file with all the strings, so now normally the static strings should have been replaced and I will keep in mind that for future changes in the UI it is not a bad idea to add a screenshot to have an idea on how it looks like ^^

fe1-web/src/features/social/components/ChirpCard.tsx
pierluca: I like linter rules. Linters are our friends 😁 

image

PR #1434, opened 2023-03-06
Ignore Sonar for Dependabot PRs

pierluca:
Dependabot PRs currently fail due to the Sonar token being a secret, unavailable in the CI at that point.
This disables Sonar in Dependabot PRs, allowing the rest of the CI to be executed and thereby verifying if the upgrade is safe.


Tyratox (approved):
Looks good to me :) I assume the slash at the end of dependabot is intended?

PR #1436, opened 2023-03-06
Add overlay texts to all QRCodes

### LINKED ISSUE #1332: Add text to scannable roll call qr codes like the "Scan to Connect"
For the qr code that is used to connect to a LAO there is an overlay text saying "Scan to Connect".

Screenshot 2023-01-05 at 15 12 52

We should add such a text to the pop token qr codes that are currently without text:

Screenshot 2023-01-05 at 15 14 21

The `QRCode` component has a `overlayText` that just needs to be set to the right value in the proper places.

MeKHell:
This PR adds overlays to all the QR codes that are actually shown (and that did not have already an overlay). 
This includes : 
- The QR code in the digital cash feature
image
- The QR code that displays user tokens 
image
- The QR code that should be scanned in order to enter
image

This resolves #1332

### DIFF
--- a/fe1-web/src/core/components/QRCode.tsx
+++ b/fe1-web/src/core/components/QRCode.tsx
@@ -21,7 +21,7 @@ const styles = StyleSheet.create({
       and a height of 50%. centering this gives us 25% margin on every side
     */
     position: 'absolute',
-    padding: '4px',
+    padding: '4',
     top: '25%',
     left: '25%',
     right: '25%',
--- a/fe1-web/src/core/components/QRCode.tsx
+++ b/fe1-web/src/core/components/QRCode.tsx
@@ -21,6 +21,7 @@ const styles = StyleSheet.create({
       and a height of 50%. centering this gives us 25% margin on every side
     */
     position: 'absolute',
+    padding: '1%',
     top: '25%',
     left: '25%',
     right: '25%',
--- a/fe1-web/src/core/components/QRCode.tsx
+++ b/fe1-web/src/core/components/QRCode.tsx
@@ -21,7 +21,7 @@ const styles = StyleSheet.create({
       and a height of 50%. centering this gives us 25% margin on every side
     */
     position: 'absolute',
-    padding: '1%',
+    padding: '4px',
     top: '25%',
     left: '25%',
     right: '25%',
--- a/fe1-web/src/features/wallet/screens/__tests__/__snapshots__/WalletSingleRollCall.test.tsx.snap
+++ b/fe1-web/src/features/wallet/screens/__tests__/__snapshots__/WalletSingleRollCall.test.tsx.snap
@@ -356,8 +356,7 @@ exports[`ViewSingleRollCallScreenHeader renders correctly 1`] = `
                           <View>
                             <View>
                               <qrcode
-                                overlayText="Scan to get
-user token"
+                                overlayText="Scan to get user token"
                                 value="{"pop_token":"oKHk3AivbpNXk_SfFcHDaVHcCcY8IBfHE7auXJ7h4ms="}"
                               />
                             </View>
--- a/fe1-web/src/features/wallet/screens/__tests__/__snapshots__/WalletSingleRollCall.test.tsx.snap
+++ b/fe1-web/src/features/wallet/screens/__tests__/__snapshots__/WalletSingleRollCall.test.tsx.snap
@@ -356,7 +356,8 @@ exports[`ViewSingleRollCallScreenHeader renders correctly 1`] = `
                           <View>
                             <View>
                               <qrcode
-                                overlayText="To be scanned"
+                                overlayText="Scan to get
+user token"
                                 value="{"pop_token":"oKHk3AivbpNXk_SfFcHDaVHcCcY8IBfHE7auXJ7h4ms="}"
                               />
                             </View>
@@ -829,7 +830,8 @@ exports[`WalletSingleRollCall renders correctly 1`] = `
                           <View>
                             <View>
                               <qrcode
-                                overlayText="To be scanned"
+                                overlayText="Scan to get
+user token"
                                 value="{"pop_token":"oKHk3AivbpNXk_SfFcHDaVHcCcY8IBfHE7auXJ7h4ms="}"
                               />
                             </View>
--- a/fe1-web/src/features/wallet/screens/__tests__/__snapshots__/WalletSingleRollCall.test.tsx.snap
+++ b/fe1-web/src/features/wallet/screens/__tests__/__snapshots__/WalletSingleRollCall.test.tsx.snap
@@ -830,8 +830,7 @@ exports[`WalletSingleRollCall renders correctly 1`] = `
                           <View>
                             <View>
                               <qrcode
-                                overlayText="Scan to get
-user token"
+                                overlayText="Scan to get user token"
                                 value="{"pop_token":"oKHk3AivbpNXk_SfFcHDaVHcCcY8IBfHE7auXJ7h4ms="}"
                               />
                             </View>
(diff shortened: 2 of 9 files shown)


fe1-web/src/resources/strings.ts
```diff
@@ -291,6 +291,7 @@ namespace STRINGS {
     'The Roll Call is currently open and you as the organizer should start adding attendees by scanning their PoP tokens.';
   export const roll_call_open_attendee =
     'The Roll Call is currently open and you as an attendee should let the organizer scan your PoP token encoded in the QR Code below.';
+  export const roll_call_QRcode_text = 'To be scanned';
```
Tyratox, line 294: I think this text is a bit too general, it would be nice to have something that indicates the purpose of the qrcode. It is clear that it should be scanned but what for?

Also it would be nice to stick to the naming conventions, i.e. camel case except in this file because of legacy reasons. At some point we can make it all camel case here too though😁

fe1-web/src/resources/strings.ts
```diff
@@ -427,6 +428,7 @@ namespace STRINGS {
   export const wallet_home_rollcall_current_pop_tokens = 'Current tokens';
   export const wallet_home_rollcall_previous_pop_tokens = 'Previous tokens';
   export const wallet_home_rollcall_pop_token_valid = 'Current';
+  export const wallet_home_rollcall_QRcode_text = 'Scan to get\nuser token';
```
Tyratox, line 431: Same here regarding the variable name.

fe1-web/src/resources/strings.ts
```diff
@@ -448,6 +450,7 @@ namespace STRINGS {
     'Your balance for this lao is the sum of the balances of each roll call token you own in this lao. ' +
     'The distribution of your total balance is visible in the list below. ' +
     'To send cash from one of the accounts, simply tap on it.';
+  export const digital_cash_wallet_QRcode_text = 'Scan to set\nas recipient';
```
Tyratox, line 453: Same here regarding the variable name.

Tyratox (changes requested):
Looks very good in general 😄 A few minor comments:)

fe1-web/src/resources/strings.ts
```diff
@@ -291,7 +291,7 @@ namespace STRINGS {
     'The Roll Call is currently open and you as the organizer should start adding attendees by scanning their PoP tokens.';
   export const roll_call_open_attendee =
     'The Roll Call is currently open and you as an attendee should let the organizer scan your PoP token encoded in the QR Code below.';
-  export const roll_call_QRcode_text = 'To be scanned';
+  export const roll_call_qrcode_text = 'To be scanned to connect';
```
MeKHell, line 294: I am not sure that this string is better than the precedent one, but I do not how to change it to be more explicit.

Tyratox: I would try to make it either represent the intention of the attendee or the organizer. Since we have so far phrased most overlays from the perspective of the person scanning it, the intention of the organizer might be more appropriate so maybe something along the lines of "Scan to add attendee"?

Tyratox:
You should also review each others code so feel free to ping me again when @dayan9265 has left a review :)

dayan9265:
The code looks good to me, small detail: I don't know if it is easy to make the text automatically wrap in the blue circle instead of manually add '\n' (because maybe on different devices it might render in a weird way). But if that is not the case, I don't think it is worth spending time for this detail ^^

MeKHell:
>  I don't know if it is easy to make the text automatically wrap in the blue circle instead of manually add '\n' (because maybe on different devices it might render in a weird way).

There is already an automatic wrapping. I added the '\n' since I think it renders way better than the original wrapping. In addition, if I wanted to add more styling rule to use a custom "automatic" wrapping, it would imply to change the QRCode properties. 

This would be pretty dirty compared to this single '\n'.

fe1-web/src/features/wallet/screens/__tests__/__snapshots__/WalletSingleRollCall.test.tsx.snap
```diff
@@ -356,6 +356,8 @@ exports[`ViewSingleRollCallScreenHeader renders correctly 1`] = `
                           <View>
                             <View>
                               <qrcode
+                                overlayText="Scan to get
+user token"
                                 value="{"pop_token":"oKHk3AivbpNXk_SfFcHDaVHcCcY8IBfHE7auXJ7h4ms="}"
```
Xelowak, line 361: Are you sure it is normal that pop token values disappeared from snapshots?

Xelowak (commented):
Nice job! 
It just feels weird to me that PoP Token values disappeared from snapshots, is it expected?

Xelowak:
> The code looks good to me, small detail: I don't know if it is easy to make the text automatically wrap in the blue circle instead of manually add '\n' (because maybe on different devices it might render in a weird way). But if that is not the case, I don't think it is worth spending time for this detail ^^

You're right! I think it might be possible by adding a `width` attribute to the overlay text in `QRCode`. I don't think it is necessary though, so if it does not work you can just let it as it is.

MeKHell:
With padding: 
image

Without padding: 
image

fe1-web/src/features/wallet/screens/__tests__/__snapshots__/WalletSingleRollCall.test.tsx.snap
MeKHell: They did not disappear. 

GitHub identified a potential syntax error and render it in dark red. [More info,](https://stackoverflow.com/questions/30667675/what-does-a-red-text-background-mean-in-github-comparison-view)

dayan9265 (approved):
The code looks good to me 👍

fe1-web/src/core/components/QRCode.tsx
```diff
@@ -21,6 +21,7 @@ const styles = StyleSheet.create({
       and a height of 50%. centering this gives us 25% margin on every side
     */
     position: 'absolute',
+    padding: '1%',
```
Tyratox, line 24: Is there a reason you use percentage padding? The overlay is made in percentages to ensure decoding will still work but I think you should be able to use non-relative padding here (i.e. Spacing.x1) since the font size is also not relative to the size of the qr code.

fe1-web/src/features/wallet/screens/__tests__/__snapshots__/WalletSingleRollCall.test.tsx.snap
Xelowak: Okay, sorry!

Xelowak (commented):
This looks good! I'll let you handle the padding issue Nico came up with.

fe1-web/src/core/components/QRCode.tsx
MeKHell: Using `Spacing.x` is less elegant for me, since the padding is related to the size of the blue circle, which is directly related to the hard-coded size of the QRCode in the QRCode.txs file. In addition, the text styles used inside the QRCode are not defined using `Spacing.x`.

Using a constant padding instead of a relative is indeed better, and this change will be implemented in the next commit. 

Btw, using the `centeredBase` style instead of `base` and `centered` gives a too large padding for potential larger texts later.

Tyratox: Hmm I see your point 🤔

But you should not add the px suffix, react native uses unitless values :) (a bit weird, yes)
https://reactnative.dev/docs/height-and-width

And you‘re right, somehow (probably it was even me) there is an additional margin for centeredBase 😅

fe1-web/src/core/components/QRCode.tsx
```diff
@@ -21,6 +21,7 @@ const styles = StyleSheet.create({
       and a height of 50%. centering this gives us 25% margin on every side
     */
     position: 'absolute',
+    padding: '4px',
```
Tyratox, line 24: Remove the px (i.e. just write 4) and we‘re good to go 🚀😁 
  
https://reactnative.dev/docs/height-and-width

Tyratox (changes requested):

fe1-web/src/core/components/QRCode.tsx
MeKHell: Done

Tyratox (approved):

dayan9265 (approved):
Looks good to me 👍

Issue #1295, opened 2022-11-17
[BUG] Cannot create election with emoji-only question


Tyratox:
### Description (Actual behavior)
Cannot create an election with a pure emoji question text

### Expected behavior
It should be possible to create an election with any question text

### How to reproduce
1. Create election with question text only containing emojis
2. See error

### Version & Environment
This bug was reproduced on:
- Commit (fe-1): dae78823068b5cab25bf20fd44eae1d03b83c4ca
- Commit (be-2): a24500837

##### Front-ends:
- [x] Fe1-Web (please include browser's names & version)
- [ ] Fe2-Android (please specify if phone or emulation, and Android version)
- [ ] Not applicable

##### Back-ends:
- [ ] Be1-Go
- [x] Be2-Scala
- [ ] Not applicable

Tyratox:
Confirmed again on current master (a01eae4d5e74bce693bd72766d4c6607f196fc41)

Tyratox:
Confirmed fe-1 to work with go, so this is a scala issue

Ajkunas:
Problem solved by the PR #1440 ! I guess we can close this issue :)

PR #1440, opened 2023-03-08
Fix issue #1295

K1li4nL:
Fix issue #1295,
The problem came from our calculation of lengths for strings, instead of having our length in bytes we were counting the characters, for pure text 1byte = 1char but not for emojis. 

Two tests were added to confirm the fix.


Ajkunas (approved):
LGTM ! Good job :)

PR #1441, opened 2023-03-08
Integrate the build of the APK in the release

matteosz:
The following PR is intended to integrate the automation of the APK build process into the workflows pipeline so that we can deliver a new up-to-date release before each PoP party. 

> This is still a work-in-progress, as the configuration file currently only generates the APK file and uploads it as an artifact. We still need to upload it in the release and make it available online on the server.

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -161,3 +161,80 @@ jobs:
             rsync -chav --delete \
               -e 'ssh -p ${{secrets.POPDEMO_DEPLOY_PORT}} -i ./deploy_key -o StrictHostKeyChecking=no' \
               ./pop-scala.jar ${{env.dest}}
+
+  deploy_fe2:
+    name: Release and deploy fe2-android
+
+    if:  ${{ startsWith(github.event.release.tag_name, 'fe2-') || startsWith(github.event.release.tag_name, 'all-') }}
+    runs-on: ubuntu-latest
+
+    env:
+      base_folder: fe2-android
+      main_project_module: app
+      playstore_name: ${{ github.repository }}
+
+    defaults:
+      run:
+        working-directory: ./${{ env.base_folder }}
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Set up JDK
+        uses: actions/setup-java@v3
+        with:
+          java-version: 11
+
+      - name: Change wrapper permissions
+        run: chmod +x ./gradlew
+
+      # Save some build information to inject into the APP
+      - name: Get the version
+        id: get_version
+        run: |
+          echo ::set-output name=version::$(echo ${GITHUB_REF/refs\/tags\//})
+          echo ::set-output name=version_file::$(echo ${GITHUB_REF/refs\/tags\//} | tr . _)
+          echo "::set-output name=shortsha::$(git rev-parse --short ${GITHUB_SHA})"
+          echo "::set-output name=buildurl::${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}/commit/${GITHUB_SHA}"
+          echo "::set-output name=date::$(date +'%d/%m/%y %H:%M')"
+
+      # Run Build Project
+      - name: Build gradle project
+        run: ./gradlew build
+
+      # Create APK Debug
+      - name: Build apk debug project (APK) - ${{ env.main_project_module }} module
+        run: ./gradlew assembleDebug
+
+      # Create APK Release
+      - name: Build apk release project (APK) - ${{ env.main_project_module }} module
+        run: ./gradlew assemble
+
+      # Create Bundle AAB Release
+      - name: Build app bundle release (AAB) - ${{ env.main_project_module }} module
+        run: ./gradlew ${{ env.main_project_module }}:bundleRelease
+
+      # Upload Artifact Build
+      - name: Upload APK Debug - ${{ steps.get_version.outputs.version }}
+        uses: actions/upload-artifact@v2
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yaml
```diff
@@ -161,3 +161,80 @@ jobs:
...
+      - name: Build apk release project (APK) - ${{ env.main_project_module }} module
+        run: ./gradlew assemble
+
+      # Create Bundle AAB Release
+      - name: Build app bundle release (AAB) - ${{ env.main_project_module }} module
+        run: ./gradlew ${{ env.main_project_module }}:bundleRelease
+
+      # Upload Artifact Build
+      - name: Upload APK Debug - ${{ steps.get_version.outputs.version }}
+        uses: actions/upload-artifact@v2
```
pierluca, line 219: Why not directly use the [softprops/action-gh-release](https://github.com/softprops/action-gh-release) to upload the artifacts ? It'll directly integrate with the other systems in a GitHub release.

.github/workflows/deploy.yaml
```diff
@@ -161,3 +161,80 @@ jobs:
...
+          name: ${{ steps.get_version.outputs.version }} - ${{ env.playstore_name }} - ${{ steps.get_version.outputs.date }} - APK(s) release generated
+          path: ${{ env.main_project_module }}/build/outputs/apk/release/
+
+      - name: Upload AAB (App Bundle) Release - ${{ steps.get_version.outputs.version }}
+        uses: actions/upload-artifact@v2
+        with:
+          name: ${{ steps.get_version.outputs.version }} - ${{ env.playstore_name }} - ${{ steps.get_version.outputs.date }} - App bundle(s) AAB release generated
+          path: ${{ env.main_project_module }}/build/outputs/bundle/release/
+
+      # Upload release
```
pierluca, line 236: Uploading to the Play Store is unfortunately going to take a while on my side, as we might formally need to get access from EPFL, create a "development" version, and so on. I can look into it, but it won't be available before a few weeks - that's how long it took in the past for the iOS App Store to get all the EPFL-internal formalities done.

.github/workflows/deploy.yaml
```diff
@@ -161,3 +161,80 @@ jobs:
...
+      # TODO
+
+      # Use rsync to deploy the apk on the website
+      # TODO
```
pierluca, line 240: Feel free to do something like this:

```suggestion
      # Use rsync to deploy the new website
      - name: Sync
        env:
            dest: "popdemo@${{secrets.POPDEMO_DEPLOY_ADDR}}:fe2-android/"
        run: |
            echo "${{secrets.POPDEMO_DEPLOY_KEY}}" > deploy_key
            chmod 600 ./deploy_key
            rsync -chav --delete \
              -e 'ssh -p ${{secrets.POPDEMO_DEPLOY_PORT}} -i ./deploy_key -o StrictHostKeyChecking=no' \
              ./pop-latest.apk ${{env.dest}}
```
(assuming your APK is going to be named `pop-latest.apk` 😁 )

I'll make sure that works :)

pierluca (commented):
Thank you for looking into this, a couple of comments and we should be good to go 😄

.github/workflows/deploy.yaml
matteosz: Ok, so I'll remove the bundle for now!

.github/workflows/deploy.yaml
matteosz: Sure! I'll use that action so we have all the systems in sync.

.github/workflows/deploy.yaml
```diff
@@ -161,3 +161,82 @@ jobs:
...
+      # Use rsync to deploy the release apk on the website
+      - name: Sync
+        env:
+            dest: "popdemo@${{secrets.POPDEMO_DEPLOY_ADDR}}:fe2-android/"
+        run: |
+            echo "${{secrets.POPDEMO_DEPLOY_KEY}}" > deploy_key
+            chmod 600 ./deploy_key
+            rsync -chav --delete \
+              -e 'ssh -p ${{secrets.POPDEMO_DEPLOY_PORT}} -i ./deploy_key -o StrictHostKeyChecking=no' \
+              ./${{ env.main_project_module }}/build/outputs/apk/release/${{ env.app_name }}-R-${{ steps.get_version.outputs.version }}.apk ${{env.dest}}
```
pierluca, line 242: Given the filenames are going to be different every time (unlike fe1/be1/be2, because those "live" on the server), do you need me to enable the webserver's autoindexing features for fe2.personhood.online ? (i.e. something looking like this https://www.linickx.com/files/2015/02/df_nginx_autoindex.png )

.github/workflows/deploy.yaml
```diff
@@ -161,3 +161,82 @@ jobs:
...
+
+      # Create APK Release
+      - name: Build apk release project (APK) - ${{ env.main_project_module }} module
+        run: ./gradlew assemble
+
+      # Rename the Release APK
+      - name: Rename APK release
+        run: mv ${{ env.main_project_module }}/build/outputs/apk/release/app-release-unsigned.apk ${{ env.main_project_module }}/build/outputs/apk/release/${{ env.app_name }}-R-${{ steps.get_version.outputs.version }}.apk
+
+
```
pierluca, line 218: that's probably a newline too many 😁

pierluca (approved):
Minor remarks, but pretty good already !

.github/workflows/deploy.yaml
matteosz: Yeah, I think it'd be a good feature to have a sort of release history (if it's not a big deal to enable)👍🏼

pierluca (approved):
LGTM ! 

FYI, we'll have a release history in GitHub anyways, but I've also enabled nginx's auto-index.

Issue #1401, opened 2023-02-20
Change the RollCallData and ElectionData channel path 


Ajkunas:
You can see in the documentation of the Scala back-end the following path : 

- for RollCallData: `root/rollcall/lao_id`
- for ElectionData: `root/private/election_id`

They do not respect the initial Scala protocol: the channel should first be separated by lao_id since the rollcall and election are a part of a specific LAO. 
A brief refactoring is needed to change it so that they are conform to the initial protocol. 

**HINT**: take a look at the `DbActor` file ;).

PR #1445, opened 2023-03-11
Change of RollCallData and ElectionData channel paths.

BelghmiAmine:
See #1401

### DIFF
--- a/be2-scala/docs/README.md
+++ b/be2-scala/docs/README.md
@@ -206,8 +206,8 @@ Summary of the keys used to retrieve data:
 - for a message: `channel#message_id`
 - for ChannelData: `channel`
 - for LaoData: `root/lao_id#laodata`
-- for RollCallData: `root/rollcall/lao_id`
-- for ElectionData: `root/private/election_id`
+- for RollCallData: `root/lao_id/rollcall`
+- for ElectionData: `root/lao_id/private/election_id`
 
 We use `/` as a separator for parts of a channel and `#` as a separator for data objects when needed.
 
--- a/be2-scala/docs/README.md
+++ b/be2-scala/docs/README.md
@@ -306,7 +306,7 @@ final case class DbActorReadLaoDataAck(laoData: LaoData) extends DbActorMessage
 For the Social Media functionality, each user has their own channel with the identifier `root/lao_id/own_pop_token` and each broadcast containing the message_id of a post will be written to `root/lao_id/posts`.
 
 For the Election functionality, we need to have a key pair stored safely somewhere so that we can encrypt/decrypt messages. That is why we use a `ElectionData` object to store the key pairs for the corresponding election.
-The path root is `root/private/election_id` as stated above.
+The path root is `root/lao_id/private/election_id` as stated above.
 The key pair can be stored and retrieved by the following functions.
 
 ```scala
--- a/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandler.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandler.scala
@@ -54,7 +54,7 @@ class ElectionHandler(dbRef: => AskableActorRef) extends MessageHandler {
       _ <- dbActor ? DbActor.WriteAndPropagate(rpcMessage.getParamsChannel, message)
       _ <- dbActor ? DbActor.CreateChannel(electionChannel, ObjectType.ELECTION)
       _ <- dbActor ? DbActor.WriteAndPropagate(electionChannel, message)
-      _ <- dbActor ? DbActor.CreateElectionData(electionId, keyPair)
+      _ <- dbActor ? DbActor.CreateElectionData(electionId, keyPair, rpcMessage.extractLaoId)
     } yield (data, electionId, keyPair, electionChannel)
 
     Await.ready(combined, duration).value match {
--- a/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandler.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandler.scala
@@ -107,16 +107,10 @@ class ElectionHandler(dbRef: => AskableActorRef) extends MessageHandler {
       case _        => Nil
     }
     val electionChannel: Channel = rpcMessage.getParamsChannel
-    val laoId = electionChannel.decodeChannelLaoId match {
-      case Some(id) => id
-      case _        => Nil
-    }
-    if (laoId == Nil) {
-      Right(PipelineError(ErrorCodes.INVALID_DATA.id, "could not get laoId", rpcMessage.id))
-    }
     val combined = for {
+      (_, Some(laoId)) <- extractLaoChannel(rpcMessage, "could not get laoId")
 
-      electionQuestionResults <- createElectionQuestionResults(electionChannel, electionChannel.decodeChannelLaoId.get)
+      electionQuestionResults <- createElectionQuestionResults(electionChannel, laoId)
       // propagate the endElection message
       _ <- dbAskWritePropagate(rpcMessage)
       // data to be broadcast
--- a/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandler.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandler.scala
@@ -107,7 +107,15 @@ class ElectionHandler(dbRef: => AskableActorRef) extends MessageHandler {
       case _        => Nil
     }
     val electionChannel: Channel = rpcMessage.getParamsChannel
+    val laoId = electionChannel.decodeChannelLaoId match {
+      case Some(id) => id
+      case _ => Nil
+    }
+    if(laoId == Nil){
+      Right(PipelineError(ErrorCodes.INVALID_DATA.id, "laoId does not exist", rpcMessage.id))
+    }
     val combined = for {
+
       electionQuestionResults <- createElectionQuestionResults(electionChannel, electionChannel.decodeChannelLaoId.get)
       // propagate the endElection message
       _ <- dbAskWritePropagate(rpcMessage)
--- a/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandler.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandler.scala
@@ -139,7 +139,7 @@ class ElectionHandler(dbRef: => AskableActorRef) extends MessageHandler {
       setupMessage <- electionChannel.getSetupMessage(dbActor)
       // associate the questions ids to their ballots
       questionToBallots = setupMessage.questions.map(question => question.id -> question.ballot_options).toMap
-      DbActorReadElectionDataAck(electionData) <- dbActor ? DbActor.ReadElectionData(setupMessage.id, laoId)
+      DbActorReadElectionDataAck(electionData) <- dbActor ? DbActor.ReadElectionData(laoId, setupMessage.id)
     } yield {
       // set up the table of results
       val resultsTable = mutable.HashMap.from(for {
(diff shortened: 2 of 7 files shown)


be2-scala/src/main/scala/ch/epfl/pop/storage/DbActor.scala
```diff
@@ -139,8 +139,8 @@ final case class DbActor(
   }
 
   @throws[DbActorNAckException]
-  private def createElectionData(electionId: Hash, keyPair: KeyPair): Unit = {
-    val channel = Channel(s"${ROOT_CHANNEL_PREFIX}private/${electionId.toString}")
+  private def createElectionData(electionId: Hash, keyPair: KeyPair, laoId: Hash): Unit = {
```
pierluca, line 142: Here and elsewhere, try to define function signatures where the parameter order is obvious to someone calling. Because there's a relationship between `LAO` and `election`, they should at least be next to each other, and it might even make sense to ask `laoId` first considering their hierarchical relationship.

be2-scala/src/main/scala/ch/epfl/pop/storage/DbActor.scala
```diff
@@ -139,8 +139,8 @@ final case class DbActor(
...
+    val channel = Channel(s"${ROOT_CHANNEL_PREFIX}${laoId.toString}/private/${electionId.toString}")
```
pierluca, line 143: do you still need the `/private` segment in the path ? 
Conversely, how are you going to ensure that this key is always interpreted as an election in the database ? 

This question applied before your changes too, I'm just raising a question that could've been addressed by students in past iterations 😄

pierluca (commented):
Since I was looking at your PR for the CI issue, here's a couple of comments.
It's just a quick look, not an exhaustive review 😁

be2-scala/src/main/scala/ch/epfl/pop/storage/DbActor.scala
K1li4nL: I suppose using `/election` would be better ? At least it would be clearer and more coherent with the other keywords used so far (rollcall, social, reaction ...etc)

Ajkunas: The purpose of this object `ElectionData` is simply to keep the private key stored somewhere where we cannot access it. 
There is already the channel `root/laoId/electionId` which is defined for the election channel. The problem of storing the `ElectionData` inside this channel, was when we wanted to retrieve messages from the election channel, which is done in the validators or when computing the results of election, anyone could access to the private key. 
To dodge this problem, we decided to create another channel defined with a `private` just to differentiate it from the main election channel. This channel is simply used to store the `ElectionData` where we can find the private key used for the election in question.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandler.scala
```diff
@@ -108,7 +108,7 @@ class ElectionHandler(dbRef: => AskableActorRef) extends MessageHandler {
     }
     val electionChannel: Channel = rpcMessage.getParamsChannel
     val combined = for {
-      electionQuestionResults <- createElectionQuestionResults(electionChannel)
+      electionQuestionResults <- createElectionQuestionResults(electionChannel, rpcMessage.extractLaoId)
```
Ajkunas, line 111: Be careful here. What you are extracting is the `electionId` and not the `laoId`. Indeed, the function `extractLaoId` extracts the last id in the channel defined, i.e. for a channel `root/laoId/electionId` the last substring extracted is `electionId`. If you want to access the `laoId`, you need to use the function `decodeChannelLaoId`. 
We need to be careful with `EndElection` messages because those ones are sent on the **election** channel. 
For the `SetupElection` messages, they are sent on the main channel, i.e the lao channel, which means that the channel is as `root/laoId`. So, there is no problem using the `extractLaoId`, because it will take the last substring, meanin the `laoId` ;)
Also, see protocol (and slack) for more informations.

be2-scala/src/main/scala/ch/epfl/pop/storage/DbActor.scala
```diff
@@ -205,7 +205,7 @@ final case class DbActor(
 
   // generates the key of the RollCallData to store in the database
   private def generateRollCallDataKey(laoId: Hash): String = {
-    s"${ROLL_CALL_DATA_PREFIX}${laoId.toString}"
```
Ajkunas, line 208: You can also remove completely this defined `String`, found in the file `Channel.scala` I think.

Ajkunas (changes requested):
Be careful with the use of `extractLaoId`, I think we need to redefine those function, because they can be confusing.. 

Also, do not hesitate to update the scala documentation where we define the `ElectionData` and `RollCallData` paths. (this should take you 1 min)

K1li4nL:
Now that I think about it, the scala documentation should reflect the changes, I suppose that could be done in this PR ?

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandler.scala
```diff
@@ -108,7 +108,7 @@ class ElectionHandler(dbRef: => AskableActorRef) extends MessageHandler {
     }
     val electionChannel: Channel = rpcMessage.getParamsChannel
     val combined = for {
-      electionQuestionResults <- createElectionQuestionResults(electionChannel, rpcMessage.extractLaoId)
+      electionQuestionResults <- createElectionQuestionResults(electionChannel, electionChannel.decodeChannelLaoId.get)
```
Ajkunas, line 111: Be careful here. In case you have a empty LAO id, it will throw an exception. 
How can you deal with that ?

Ajkunas: Also, maybe it would be more clear if you define a value for the laoId

Ajkunas (changes requested):
Always be careful when you do a `get` on an optional on an `Option`.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandler.scala
Ajkunas: Okeey I realize now that we forgot to speak about this during the meeting.. 

So, your implementation is actually good ! But you can do something more concise, and actually you have a already-implemented function that you can use.. 
Remember the function `extractParameters` in the `MessageHandler.scala` ? There is a similar function there that can really help you for this. :)

be2-scala/docs/README.md
```diff
@@ -206,8 +206,8 @@ Summary of the keys used to retrieve data:
 - for a message: `channel#message_id`
 - for ChannelData: `channel`
 - for LaoData: `root/lao_id#laodata`
-- for RollCallData: `root/rollcall/lao_id`
-- for ElectionData: `root/private/election_id`
+- for RollCallData: `root/lao_id/rollcall`
+- for ElectionData: `root/lao_id/private/election_id`
```
Ajkunas, line 210: You also need to change the line 309 for consistency.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandler.scala
```diff
@@ -54,7 +54,7 @@ class ElectionHandler(dbRef: => AskableActorRef) extends MessageHandler {
       _ <- dbActor ? DbActor.WriteAndPropagate(rpcMessage.getParamsChannel, message)
       _ <- dbActor ? DbActor.CreateChannel(electionChannel, ObjectType.ELECTION)
       _ <- dbActor ? DbActor.WriteAndPropagate(electionChannel, message)
-      _ <- dbActor ? DbActor.CreateElectionData(electionId, keyPair)
+      _ <- dbActor ? DbActor.CreateElectionData(rpcMessage.extractLaoId, electionId, keyPair)
```
K1li4nL, line 57: `extractLaoId` above may or may not need to be changed depending on pr #1459, I am not sure if here the actual laoId is really retrieved, gotta be careful

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandler.scala
```diff
@@ -107,16 +107,10 @@ class ElectionHandler(dbRef: => AskableActorRef) extends MessageHandler {
...
     val electionChannel: Channel = rpcMessage.getParamsChannel
-    val laoId = electionChannel.decodeChannelLaoId match {
-      case Some(id) => id
-      case _        => Nil
-    }
-    if (laoId == Nil) {
-      Right(PipelineError(ErrorCodes.INVALID_DATA.id, "could not get laoId", rpcMessage.id))
-    }
     val combined = for {
+      (_, Some(laoId)) <- extractLaoChannel(rpcMessage, "could not get laoId")
```
Ajkunas, line 111: I would suggest as error message : `There is an issue with the id of the LAO`

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandler.scala
Ajkunas: A Setup Election message is sent on the main channel `root/laoId`, so here by applying `extractLaoId` we actually retrieve the laoId. 
It is the only exception among the election messages. Otherwise, we should be careful with the other messages.

be2-scala/src/main/scala/ch/epfl/pop/storage/DbActor.scala
```diff
@@ -251,9 +251,9 @@ final case class DbActor(
         case failure              => sender() ! failure.recover(Status.Failure(_))
       }
 
-    case ReadElectionData(electionId) =>
+    case ReadElectionData(electionId, laoId) =>
       log.info(s"Actor $self (db) received a ReadElectionData request for election '$electionId'")
-      Try(readElectionData(electionId)) match {
+      Try(readElectionData(electionId, laoId)) match {
```
K1li4nL, line 256: I believe you forgot to switch laoId and electionId to match the function definition

be2-scala/src/main/scala/ch/epfl/pop/storage/DbActor.scala
```diff
@@ -74,8 +74,8 @@ final case class DbActor(
   }
 
   @throws[DbActorNAckException]
-  private def readElectionData(electionId: Hash): ElectionData = {
-    Try(storage.read(s"${ROOT_CHANNEL_PREFIX}private/${electionId.toString}")) match {
+  private def readElectionData(electionId: Hash, laoId: Hash): ElectionData = {
```
Ajkunas, line 77: For more clarity, maybe you should also switch the paramters here in the signature. There will be less confusion when you use the `ReadElectionData`  function

Ajkunas (changes requested):
Little changes with the signatures of the functions.

be2-scala/src/main/scala/ch/epfl/pop/storage/DbActor.scala
```diff
@@ -251,9 +251,9 @@ final case class DbActor(
         case failure              => sender() ! failure.recover(Status.Failure(_))
       }
 
-    case ReadElectionData(electionId) =>
+    case ReadElectionData(electionId, laoId) =>
```
Ajkunas, line 254: Here your ReadElectionData paramters order does not match the definition (See line 394). You need to switch the order of laoId and electionId

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandler.scala
```diff
@@ -129,15 +131,15 @@ class ElectionHandler(dbRef: => AskableActorRef) extends MessageHandler {
...
+  private def createElectionQuestionResults(electionChannel: Channel, laoId: Hash): Future[List[ElectionQuestionResult]] = {
     for {
       // get the last votes of the CastVotes messages
       castsVotesElections <- electionChannel.getLastVotes(dbActor)
       // get the setupElection message of the channel
       setupMessage <- electionChannel.getSetupMessage(dbActor)
       // associate the questions ids to their ballots
       questionToBallots = setupMessage.questions.map(question => question.id -> question.ballot_options).toMap
-      DbActorReadElectionDataAck(electionData) <- dbActor ? DbActor.ReadElectionData(setupMessage.id)
+      DbActorReadElectionDataAck(electionData) <- dbActor ? DbActor.ReadElectionData(setupMessage.id, laoId)
```
Ajkunas, line 142: Here too, you need to switch laoId and the electionId (setupMessage.id)

Ajkunas (approved):

PR #1447, opened 2023-03-12
Implement dark mode

simone-kalbermatter:
### Changes made in this PR include:
- Dark mode or light mode is automatically enabled based on the system theme
- Defined a default button style that is automatically applied everyhwere (no need to set a style in xml)
- Defined other custom styles for buttons, toolbars, text views and card views to remove hard coded colors in layout files
- QR codes are now generated in white if dark mode is enabled (this was manually tested)
- The scanning of a white QR code was manually tested


pierluca:
Looks awesome !

Tiny suggestion based on the screenshot below:
Would it make sense to keep the text white on the highlighted menu (i.e. "Events") ?
I'm thinking it might improved readability when the screen is dimmed 😄

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/digitalcash/DigitalCashReceiveFragment.java
```diff
@@ -110,4 +115,18 @@ public void handleOnBackPressed() {
...
+   * @return the color of the QR code (either Color.WHITE or Color.BLACK)
+   */
+  public int getQRCodeColor() {
+    Configuration configuration = getResources().getConfiguration();
+    int nightModeFlags = configuration.uiMode & Configuration.UI_MODE_NIGHT_MASK;
+    if (nightModeFlags == Configuration.UI_MODE_NIGHT_YES) {
+      return Color.WHITE;
+    }
+    return Color.BLACK;
+  }
```
matteosz, line 131: To avoid code replication I was wondering if it would be better to have such function written only once as static, perhaps taking as parameter a `Configuration`, so that it has not to be replicated multiple times across the different fragments.

matteosz (commented):
Very good work, also well documented and tested! 

Appearance-wise, I like the general palette, but as Pierluca suggested one change could be to ensure that every UI has the proper constrast, as it is particularly important in dark mode. This can be translated into a white selection on the side menu and some brighter text (for small text the current grey tone is totally fine for me, for headings/bigger text I'd suggest a whiter color e.g. wallet description 'This is only...' and 'LAO Coins' in the screenshots).

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/digitalcash/DigitalCashReceiveFragment.java
simone-kalbermatter: Yes you are right, I was thinking about this but couldn't really find a spot where it fits... Do you have a suggestion?

simone-kalbermatter: And about the text contrast, good point, I will change that!

matteosz: Probably it could fit as utility function in ActivityUtils, but there might be other better places

simone-kalbermatter:
Increased the contrasts of some elements to address Pierluca's and Matteo's comments:

matteosz (approved):
LGTM now!

Warhedgehog (approved):
LGTM ! Great work, very clean

PR #1452, opened 2023-03-14
Improved the back navigation

### LINKED ISSUE #1406: Improve back navigation
Right now, back navigation works well (is defined and implemented) for non tabs screen. By that it is meant screens that are not directly reachable from the drawer menu, and for wich there's a back arrow as the left icon of the top app bar. Here is a screenshot of a roll call screen as an example 

What must be improved is for other pages, like tab paged (reachable from the drawer menu). Below is an example screenshot

For such pages, you should define a coherent back pressed policy for the whole app, if possible in accordance with fe1, and implement it.

matteosz:
- This PR aims to improve the back navigation in the application for a smoother usability.
- Resolves #1406 
  
Modifies include:
-------------------
- Pressing back button on the list of LAOs exits the application (coherent with the default behaviour of all the other apps).
- Pressing back on a "secondary" tab (Invite, Social Media, Digital Cash, Witnessing, Tokens) returns to the Events List tab.
- Pressing back on a Social Media sub-tab (Search, Following, Profile) returns to Social Media Home.
- Pressing on the transaction history icon a second time works as going back.
- Refactor of previous code to reduce repeatability.

Here's a video highlighting these new functionalities:

https://user-images.githubusercontent.com/83547952/224983416-ac281bbd-11fa-4c51-a43d-545e68f9663c.mp4


fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/LaoActivity.java
```diff
@@ -145,11 +145,18 @@ private void observeToolBar() {
...
+            // already displayed, then consider it as a back button pressed
+            if (!historyFlag) {
+              setCurrentFragment(
+                  getSupportFragmentManager(),
+                  R.id.fragment_digital_cash_history,
+                  DigitalCashHistoryFragment::newInstance);
+            } else {
+              openDigitalCashTab();
+            }
+            historyFlag = !historyFlag;
```
simone-kalbermatter, line 159: Using a boolean is a good solution for this but it was not directly obvious to me what it does. Maybe there is some way to check that the transaction screen is already displayed instead of using the boolean? Or probably changing the name of the boolean to `isHistorySelectedFlag` or something similar would help.

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/ActivityUtils.java
```diff
@@ -128,4 +134,24 @@ public static boolean activitySavingRoutine(
             + subscriptions);
     return storePersistentData(context, new PersistentData(seed, serverAddress, subscriptions));
   }
+
+  /**
+   * The following function creates an object of type OnBackPressedCallback given a specific
+   * callback function. This avoids code repetitions.
```
simone-kalbermatter, line 140: Very nice, cleans up the code a lot!

simone-kalbermatter (changes requested):
Very clean code and a lot of changes!

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/ActivityUtils.java
matteosz: Thank you 😄

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/LaoActivity.java
matteosz: Yeah, you're right this boolean it's not exactly the cleanest way to do it. I'll remove it and use the fragment manager to detect whether it's already opened.

simone-kalbermatter (approved):
LGTM!

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/LaoActivity.java
```diff
@@ -69,9 +69,7 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
     observeDrawer();
     setupDrawerHeader();
 
-    // Open Event list on activity creation
```
Warhedgehog, line 72: I would keep the comment as it is a bit clearer why this is done

Warhedgehog (changes requested):
Great work, very clean. Just the two details but this is really nitpicking

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/LaoActivity.java
```diff
@@ -265,6 +271,12 @@ private void openSocialMediaTab() {
         getSupportFragmentManager(), R.id.fragment_social_media_home, SocialMediaHomeFragment::new);
   }
 
+  /** Open Event list on activity creation */
```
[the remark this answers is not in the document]
matteosz: Yeah, you're totally right I forgot to update those comments 👍🏼

Warhedgehog (dismissed):
LGTM !

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeActivity.java
```diff
@@ -122,6 +122,10 @@ public void onBackPressed() {
     if (!(fragment instanceof SeedWalletFragment)) {
       setCurrentFragment(getSupportFragmentManager(), R.id.fragment_home, HomeFragment::new);
     }
+    // Exits the application if back is pressed on home
+    if (fragment instanceof HomeFragment) {
+      moveTaskToBack(true);
```
jbsv, line 127: The comment says that it exists the application, but the function says that it moves it. This sounds inconsistent. Isn't it a false exit, like the application is still opened but in the background ?

jbsv (commented):
overall very good. Just one question about exiting the application: is the application really closed or is it only pushed back on the applications' stack

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/home/HomeActivity.java
matteosz: Yeah it simply moves it in background, as it happens generally to every other mobile application, then you can close it through the task manager. With "exit" I was meaning just that, you exit the application but not close it

jbsv: Well, either properly close the application as in [](https://stackoverflow.com/questions/17719634/how-to-exit-an-android-app-programmatically) for example, and/or update the misleading comment. Exiting an application to me means that the application is closed and an exit code code is returned, but maybe I am too old fashion :-D

jbsv (approved):

matteosz:
- Added the PageObject classes as suggested by @GabrielFleischer

pierluca (approved):
Approving blindly based on past approves by @Warhedgehog and @jbsv

PR #1454, opened 2023-03-14
Fix Left/Right scala pipe convention for the GraphMessage type

Jharaxus:
The scala convention for the use of Either is using Right as the success pipe and Left as the error pipe. However, the inverse is done in this project regarding the GraphMessage type. This works as is but is a technical specificity that needs to be learnt by newcomers in the project, and as the scala standard library is based on a different convention, this prevents us from using parts of it. Therefore, conforming us to the standard convention could let us use these features which would help turn the code into a cleaner form (the first commit shows a usage example in the `validateWitnessMessage`, turned into `WitnessValidatorTest`, of the `WitnessValidator` class).

### DIFF
--- a/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/MessageDecoder.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/MessageDecoder.scala
@@ -17,8 +17,8 @@ object MessageDecoder {
 
   /** Graph component: takes a string as input and parses it into a JsonRpcMessage (stored into a GraphMessage)
     */
-  val jsonRpcParser: Flow[Either[JsonString, PipelineError], GraphMessage, NotUsed] = Flow[Either[JsonString, PipelineError]].map {
-    case Left(jsonString) => Try(jsonString.parseJson.asJsObject) match {
+  val jsonRpcParser: Flow[Either[PipelineError, JsonString], GraphMessage, NotUsed] = Flow[Either[PipelineError, JsonString]].map {
+    case Right(jsonString) => Try(jsonString.parseJson.asJsObject) match {
         case Success(obj) =>
           val fields: Set[String] = obj.fields.keySet
 
@@ -34,7 +34,7 @@ object MessageDecoder {
           ))
       }
 
-    case Right(pipelineError) => Left(pipelineError) // implicit typecasting
+    case Left(pipelineError) => Left(pipelineError) // implicit typecasting
   }
 
   /** Graph component: takes a GraphMessage and parses the 'data' field of any JsonRpcRequest query containing a message. The result is stored in the input JsonRpcRequest's Message's decodedData field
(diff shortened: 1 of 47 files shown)


be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/WitnessValidator.scala
```diff
@@ -17,25 +16,19 @@ object WitnessValidator {
 }
 
 sealed class WitnessValidator(dbActorRef: => AskableActorRef) extends MessageDataContentValidator {
+  def bindToPipe[T](rpcMessage: JsonRpcMessage, opt: Option[T], pipelineError: PipelineError): GraphMessage = {
```
Jharaxus, line 19: This will can be moved to an outer class (probably `MessageDataContentValidator`) in a later PR, but I didn't want to make that one heavier than it already is

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/WitnessValidator.scala
```diff
@@ -17,25 +16,19 @@ object WitnessValidator {
...
+    if (opt.isEmpty) Left(pipelineError) else Right(rpcMessage)
+  }
```
Jharaxus, line 21: This will can be moved to an outer class (probably `MessageDataContentValidator`) in a later PR, but I didn't want to make that one heavier than it already is

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/CoinValidator.scala
```diff
@@ -48,9 +46,9 @@ case object CoinValidator extends MessageDataContentValidator {
     */
```
K1li4nL, line 46: Line 45 comments should be corrected to match the new convention

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/CoinValidator.scala
```diff
@@ -65,8 +63,8 @@ case object CoinValidator extends MessageDataContentValidator {
...
```
K1li4nL, line 63: Same as above:  Line 62 comments should be corrected to match the new convention

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/MessageDecoder.scala
```diff
@@ -23,18 +23,18 @@ object MessageDecoder {
           val fields: Set[String] = obj.fields.keySet
```
K1li4nL, line 23: Line 21: Shouldn't we change that `Left()` into a `Right()` too ?

K1li4nL (changes requested):
I took a look at everything, it mostly look good, but there are two files bothering me still: 
- MessageDecoder.scala
- SchemaVerifier.scala

They seem to partially (or not at all for `SchemaVerifier`) follow the new convention. For example, `verifyRpcSchema` still returns `Either[JsonString, PipelineError]`

Otherwise I didn't spot anything else beside some forgotten comments for which I took the liberty to fix myself. Good job !

Ajkunas (commented):
I reviewed all files, except the handler test files. 
You did a good job, I haven't found anything to fix, except a little comment. 

I agree with @K1li4nL about the `SchemaVerifier.scala` and `MessageDecoder.scala` files. They are quite confusing.

pierluca:
As discussed, forcing merge to let the CI run Karate.

Issue #1402, opened 2023-02-21
Continuity of the refactoring of the validators 


Ajkunas:
The initial implementation of the validators consist of a cascade of `if else`. Each check was performed by a `if else`. 
After several checks, we see that the implementation is not **scalable** if we need to verify more and more fields of the message. 
The new suggested implementation is to avoid this `if else` cascade by simply running the checks as a sort of list.

The refactoring has already started: the validators of the RollCall in the file `RollCallValidator.scala` are entirely refactored. 
The idea of the validators is to "validate" or check the fields of the message. Each field is checked by a specific function:

For example, we have function such as `checkOwner` in the file `MessageValidator.scala`. We also have function such as `checkTimestampOrder` in the file `MessageDataContentValidator.scala`. 

If we look in more detail one of the validator of the `RollCallValidator.scala` file, here are the main steps: 

- We extract the information fields to check of the message. 
- We define the expected information, i.e. the `expectedRollCallId`. 
- We run the several specific checks of the message with the function `runChecks`.

The recursive function `runChecks` is found in the file `MessageValidator.scala` and it simply goes through the checks. If one of the check fails, i.e. the check return `Right` with a `PipelineError` then the function `runChecks` stops and returns the error found. In case there is no error, i.e. each check returns `Left(rpcMessage)`, then the function also returns a `Left(rpcMessage)` meaning that the message is conform and valid. 

The idea is to refactor in the same way every validator file of the Scala back-end. Of course, the suggestion can be improved. Any other ideas or suggestions is welcomed.

PR #1455, opened 2023-03-14
Refactor MessageValidator

Jharaxus:
Refactor `MessageValidator` according to issue [#1402](https://github.com/dedis/popstellar/issues/1402)

### DIFF
--- a/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/package.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/package.scala
@@ -7,7 +7,11 @@ package object graph {
 
   type GraphMessage = Either[PipelineError, JsonRpcMessage]
 
-  def bindToPipe(rpcMessage: JsonRpcMessage, success: Boolean, error: PipelineError) = {
-    if (success) Left(rpcMessage) else Right(error)
+  def bindToPipe(rpcMessage: JsonRpcMessage, success: Boolean, error: PipelineError): GraphMessage = {
+    if (success) Right(rpcMessage) else Left(error)
+  }
+
+  def bindToPipe[T](rpcMessage: JsonRpcMessage, opt: Option[T], pipelineError: PipelineError): GraphMessage = {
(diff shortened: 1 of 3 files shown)


be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/package.scala
```diff
@@ -5,4 +5,8 @@ import ch.epfl.pop.model.network.JsonRpcMessage
 package object graph {
   type JsonString = String
   type GraphMessage = Either[JsonRpcMessage, PipelineError] // FIXME convention states that Right(_) is the success value
+
+  def bindToPipe(rpcMessage: JsonRpcMessage, success: Boolean, error: PipelineError) = {
```
Jharaxus, line 9: If this function is accepted, it would probably be good to wait for the other validators refactoring happening with respect to issue [#1402 ](https://github.com/dedis/popstellar/issues/1402) to be merged to master so as to their validators use this function too before completing the pull request.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/package.scala
```diff
@@ -5,4 +5,8 @@ import ch.epfl.pop.model.network.JsonRpcMessage
...
+    if (success) Left(rpcMessage) else Right(error)
+  }
```
Jharaxus, line 11: If this function is accepted, it would probably be good to wait for the other validators refactoring happening with respect to issue [#1402 ](https://github.com/dedis/popstellar/issues/1402) to be merged to master so as to their validators use this function too before completing the pull request.

K1li4nL: I came up with such a function at the beginning of the refactoring process but I gave up the idea as it kinda defeat the purpose of replacing the if-else pattern. My opinion is that the point of using these `checkSomething()` function is to have abstraction and having this `bindToPipe()` kinda short circuit that.

Jharaxus:
The actual implementation of the refactoring became obsolete due to previous commits done and should be remade from scratch.

Issue #1428, opened 2023-03-04
Question name in the election can be made of only spaces


matteosz:
### Description (Actual behavior)
When creating a question, the check is only on the fact that the name is actually empty. However, it can be made of just space characters, resulting visually empty as well (i.e. the text is not trimmed).

![elec_screen](https://user-images.githubusercontent.com/83547952/222894979-85a0f70c-7099-4fd7-9a66-853cf6c628df.png)

### Expected behavior
The current implementation in FE2 doesn't allow the user to type a string of just spaces as the first question (otherwise it's not possible to create the election), while it gives the possibility to do so with next questions. However such questions are not considered valid during the election and, thus, not displayed.

![Screenshot (26)](https://user-images.githubusercontent.com/83547952/222895430-e23dd6c9-2890-49e0-809f-e3afddec41a3.png)
|:--:| 
| *First question of election with spaces* |

![Screenshot (27)](https://user-images.githubusercontent.com/83547952/222895494-130ce802-b593-4475-a0a8-0733438a4720.png)
|:--:| 
| *Second question with spaces after a valid first question* |

![Screenshot (28)](https://user-images.githubusercontent.com/83547952/222895521-0f2bb50f-2924-4a7d-8f9b-870938b34f10.png)
|:--:| 
| *Final election result with second question discarded* |

### How to reproduce
1. Go to 'Create Election'
2. Insert the name
3. Type in Question 1 box a space " "
4. Add random balot options
5. Click on Add question
6. Type in Question 2 box 2 spaces "  "
7. Add random balot options
8. Click confirm

### Version & Environment
This bug was reproduced on the current online version of the FE1 (https://fe1.personhood.online)

##### Front-ends:
- [X] Fe1-Web
- [ ] Fe2-Android
- [ ] Not applicable

##### Back-ends:
- [ ] Be1-Go
- [ ] Be2-Scala
- [X] Not applicable

### Impact
Not impacting in terms of functionality, but just in terms of usability and coherence.

pierluca:
Nice catch ! 

Just to clarify, "this bug was reproduced on the current online version of the FE1 (https://fe1.personhood.online/) "
This was FE1-V1.3 | A332107 | 21/02/23 22:36

Also, @matteosz, are you saying that fe2 silently removes the question ?
If so, shouldn't it block the creation of the question in the first place, instead of silently removing it ?

matteosz:
[...]

PR #1456, opened 2023-03-15
Election: avoid space question

MeKHell:
This PR solves #1428 . 
It prevents to have: 
- blank question name
- blank election name
- blank ballot answers

We achieve this by trimming every user input and preventing users to enter whitespace-only strings or strings starting with whitespace.

This does not impact the UI. So no pictures.


dayan9265:
It looks good to me, I just have a small question, what was the reason to only trim the beginning of the user input when shown and not also the end?

dayan9265:
> > It looks good to me, I just have a small question, what was the reason to only trim the beginning of the user input when shown and not also the end?
> 
> If we also trim the end on the reacting inputs, the user won't be able to type space while they write and would need to insert the space (after they have written what they want) in the middle of their text.
> 
> But when the written text is ready to be processed (when the organizer confirms the election), we usually trim both going further.

Oh okay it makes sense, clearly it would not be very practical if the user couldn't type white space at the end of the text ^^

dayan9265 (dismissed):
It looks good to me :)

fe1-web/src/features/evoting/screens/CreateElection.tsx
```diff
@@ -282,7 +284,7 @@ const CreateElection = () => {
                   id === idx
                     ? {
                         ...item,
-                        ballot_options: ballot_options,
+                        ballot_options: ballot_options.filter((option) => option.trim() !== ''),
```
Tyratox, line 287: Why omit empty ballot options here? I understand that we don't want empty ballot options in the final object but showing an error to the user is a better option than just omitting data without telling the user.

It works because the TextInputList is not a controlled component (https://stackoverflow.com/a/42522792/2897827) but a future refactor should probably change this and ensure the state is in sync with what is displayed by the `TextInputList` component.

fe1-web/src/features/evoting/screens/CreateElection.tsx
```diff
@@ -302,7 +304,7 @@ const CreateElection = () => {
           {STRINGS.event_creation_must_be_connected}
         </Text>
       )}
-      {electionName === '' && (
+      {electionName.trim() === '' && (
```
Tyratox, line 307: Yes, this is great! If something is off usually it is best to tell the user and let them fix it. Trimming whitespace is fine without prompting the user is fine as it does not change the meaning but having an empty election name is something they have to fix.

fe1-web/src/features/evoting/screens/CreateElection.tsx
```diff
@@ -74,7 +74,9 @@ const isQuestionInvalid = (question: NewQuestion): boolean =>
  * @param questions The list of questions
  */
 const haveQuestionsSameTitle = (questions: NewQuestion[]): boolean => {
-  const questionTitles = questions.map((q: NewQuestion) => q.question);
+  const questionTitles = questions
+    .map((q: NewQuestion) => q.question.trim())
```
Tyratox, line 78: This is great, in the final questions will be trimmed and thus we should check for equality where they are trimmed. But since we now make a transformation between the data displayed to the user and what will be used to create the election we should ensure we perform the checks on the same data that will then be used to create the election. Otherwise the two things might differ and we have code duplication.

One way to solve is is by separating the values displayed to the user and the ones used for processing later. For instance you can create an additional `electionName` and `questions` variables and rename the old variables to something like `electionNameInput` and `questionsInput`.

E.g.
```typescript
const [questionsInput, setQuestionsInput] = useState<NewQuestion[]>([EMPTY_QUESTION]);
...


const questions = useMemo(() => {
  return questionsInput.map((q: NewQuestion) => {...q, question: q.question.trim()});
}, [questionsInput]);
```

In this case `questionsInput` represents what is entered by the user and `questions` what we use for processing. All checks should be performed on `questions` and tell the user what they need to change in `questionsInput` to make the input valid. `useMemo` updates its value if one of it's dependencies (in the array) change.

This way we're sure that the data used for processing (`questions`, derived from `questionsInput` ) is valid.

Does this make sense?

fe1-web/src/features/evoting/screens/CreateElection.tsx
```diff
@@ -74,7 +74,9 @@ const isQuestionInvalid = (question: NewQuestion): boolean =>
...
+    .filter((string) => string !== '');
```
Tyratox, line 79: Again, this should be done by the user as this is not something one would expect to happen automatically.

fe1-web/src/features/evoting/screens/CreateElection.tsx
```diff
@@ -151,7 +153,7 @@ const CreateElection = () => {
   // Confirm button only clickable when the Name, Question and 2 Ballot options have values
   const confirmButtonEnabled: boolean =
     isConnected === true &&
-    electionName !== '' &&
+    electionName.trim() !== '' &&
```
Tyratox, line 156: Same comment as above. Try to separate the value displayed by the user and the one used for processing and perform the check on the one used for processing.

At the moment you perform the processing and the checks in the same place.

fe1-web/src/features/evoting/screens/CreateElection.tsx
```diff
@@ -105,9 +107,9 @@ const createElection = (
   // compute the id for all questions and add the write_in property
   const questionsWithId = questions.map((item) =>
     Question.fromState({
-      id: Hash.fromArray(EventTags.QUESTION, electionId, item.question).toString(),
-      question: item.question,
-      ballot_options: item.ballot_options,
+      id: Hash.fromArray(EventTags.QUESTION, electionId, item.question.trim()).toString(),
+      question: item.question.trim(),
+      ballot_options: item.ballot_options.map((s) => s.trim()),
```
Tyratox, line 112: Same comment as above. Try to create the "ready for processing" values using `useMemo`, perform the checks on these and then directly use them here.

If you perform processing here you cannot be sure whether the resulting data satisfies the desired properties.

Tyratox (changes requested):
Cool that you made it work but honestly it is difficult to tell whether all places were covered and whether there are not missed edge cases.

I would try to take a step back and separate the processing of the data from the checks. You can do so by having separate variables, one for the raw user input data and one for the processed (e.g. trimmed) data. Then you can perform the checks on the processed/trimmed data and if that data satisfies the desired properties, directly use that data without performing (in the best case the same) processing again.

I made a lot of comments so it might not everywhere be clear what I mean. I am happy to do a quick zoom call tomorrow or on Tuesday if anything is unclear.

fe1-web/src/core/components/TextInputList.tsx
```diff
@@ -20,7 +20,7 @@ const TextInputList = (props: IPropTypes) => {
   const updateParent = (options: { id: number; value: string }[]) => {
     // Gets the distinct options which are not empty ('')
     const distinctValues = [...new Set(options.map((option) => option.value))].filter(
-      (value) => value !== '',
+      (value) => value.trim() !== '',
     );
```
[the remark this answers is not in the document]
Xelowak: I agree. This component is from `core`, meaning that it should be generic to allow most uses in the app. When we need a list of text inputs, we should look into `TextInputLIst`. This kind of behaviour you are adding may not be expected at first, leading in something too specific for this kind of component.

Xelowak (commented):
@Tyratox 's comments are really relevant and I honestly don't have much to add.

I feel like the trimming is done in a lot of different locations, when it could probably be more centralized. As @Tyratox suggested, I think it is better to create values that are directly trimmed and perform checks on them, rather than trimming everything before processing. This may solve the problem of trimming strings everywhere.

fe1-web/src/features/evoting/screens/CreateElection.tsx
```diff
@@ -220,7 +222,7 @@ const CreateElection = () => {
       </Text>
       <Input
         value={electionName}
-        onChange={setElectionName}
+        onChange={(s) => setElectionName(s.trimStart())}
```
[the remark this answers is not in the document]
MeKHell: The idea is to give a live feedback to the user so that he understand that whitespace only strings are not allowed

fe1-web/src/features/evoting/screens/CreateElection.tsx
```diff
@@ -255,15 +257,15 @@ const CreateElection = () => {
...
             testID={`question_selector_${idx}`}
             onChange={(text: string) =>
               setQuestions((prev) =>
                 prev.map((item, id) =>
-                  id === idx
+                  id === idx && text.trimStart() !== ''
                     ? {
                         ...item,
-                        question: text,
+                        question: text.trimStart(),
```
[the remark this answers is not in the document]
MeKHell: See https://github.com/dedis/popstellar/pull/1456#discussion_r1143079249

fe1-web/src/features/evoting/screens/CreateElection.tsx
```diff
@@ -255,15 +257,15 @@ const CreateElection = () => {
...
           </Text>
           <Input
-            value={questions[idx].question}
+            value={value.question}
             testID={`question_selector_${idx}`}
             onChange={(text: string) =>
               setQuestions((prev) =>
                 prev.map((item, id) =>
-                  id === idx
+                  id === idx && text.trimStart() !== ''
```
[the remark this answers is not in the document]
MeKHell: Indeed. The idea was that the `InputList` will remove it. Which might be wrong.
In any case, with the refactoring of the `InputList`, all the logic has been centralized in `CreateElection` and this part will also change.

MeKHell:
Based on the comment above, I decided to refactor the following components: 
 - RemovableTextInput
 - TextInputList
 in order to correctly test the new CreateElectionScreen (and that the bug as disappeared), I added the ability to store default questions in the election reducer.

This is not the optimal solution. But it might be useful if we later add a feature to set default questions.

The other solution would have been to allow to pass the default questions as navigation parameters. But this would be even less logic and useful for later development.

fe1-web/src/features/evoting/screens/CreateElection.tsx
```diff
@@ -141,22 +153,46 @@ const CreateElection = () => {
...
+
   // Confirm button only clickable when the Name, Question and 2 Ballot options have values
   const confirmButtonEnabled: boolean =
     isConnected === true &&
-    electionName !== '' &&
-    !questions.some(isQuestionInvalid) &&
-    haveQuestionsSameTitle(questions);
+    electionName.trim() !== '' &&
+    areQuestionsValid(trimmedQuestions) && // Checks there are enough valid questions to open an election
+    !questions.some(isNonEmptyQuestionInvalid) && // Checks that there are no questions that will be silently removed
```
Tyratox, line 183: If this is the only place `isNonEmptyQuestionInvalid` is used you might even call it `isSilentlyRemoved`? Then the could would directly read like your comment :)

fe1-web/src/features/evoting/screens/CreateElection.tsx
```diff
@@ -141,22 +153,46 @@ const CreateElection = () => {
...
+  }, [questions]);
+
   // Confirm button only clickable when the Name, Question and 2 Ballot options have values
   const confirmButtonEnabled: boolean =
     isConnected === true &&
-    electionName !== '' &&
-    !questions.some(isQuestionInvalid) &&
-    haveQuestionsSameTitle(questions);
+    electionName.trim() !== '' &&
+    areQuestionsValid(trimmedQuestions) && // Checks there are enough valid questions to open an election
```
Tyratox, line 182: If this is the only check performed by this function, feel free to call it like that or in this particular case write it inline and keep the comment :)

fe1-web/src/features/evoting/screens/CreateElection.tsx
```diff
@@ -141,22 +153,46 @@ const CreateElection = () => {
...
+    !questions.some(isNonEmptyQuestionInvalid) && // Checks that there are no questions that will be silently removed
+    !trimmedQuestions.some(areBallotOptionsInvalid) &&
```
Tyratox, line 184: Picky detail but I think it could read even a bit nicer :)
```suggestion
    !trimmedQuestions.some(hasInvalidBallotOptions) &&
```

fe1-web/src/features/evoting/screens/CreateElection.tsx
```diff
@@ -50,38 +50,49 @@ type NavigationProps = CompositeScreenProps<
...
+const areBallotOptionsInvalid = (question: NewQuestion): boolean =>
+  question.question !== '' &&
+  (new Set(question.ballot_options).size !== question.ballot_options.length ||
+    question.ballot_options.length < MIN_BALLOT_OPTIONS);
+
+/**
+ * Checks whether a question title is not unique withcreateEin a list of questions
  * @param questions The list of questions
  */
 const haveQuestionsSameTitle = (questions: NewQuestion[]): boolean => {
```
Tyratox, line 85: You are checking whether they are unique, right?
```suggestion
const haveUniqueQuestionTitles = (questions: NewQuestion[]): boolean => {
```

fe1-web/src/features/evoting/screens/CreateElection.tsx
```diff
@@ -141,22 +153,46 @@ const CreateElection = () => {
...
+        return {
+          ...item,
+          question: item.question.trim(),
+          ballot_options: item.ballot_options
+            .map((val) => val.trim())
+            .filter((value) => value !== ''),
+        };
+      })
+      .filter((question) => question.question !== '');
+  }, [questions]);
```
Tyratox, line 176: 🥳👏

fe1-web/src/features/evoting/screens/CreateElection.tsx
```diff
@@ -141,22 +153,46 @@ const CreateElection = () => {
...
+    !questions.some(isNonEmptyQuestionInvalid) && // Checks that there are no questions that will be silently removed
+    !trimmedQuestions.some(areBallotOptionsInvalid) &&
+    haveQuestionsSameTitle(trimmedQuestions);
 
   const onCreateElection = () => {
-    createElection(currentLao.id, version, electionName, questions, startTime, endTime)
+    createElection(
+      currentLao.id,
+      version,
+      electionName.trim(),
```
Tyratox, line 191: I guess it's fine if it is just twice but it might also be nice to have a `trimmedElectionName` analogous  to `trimmedQuestions`?

fe1-web/src/features/evoting/screens/CreateElection.tsx
```diff
@@ -220,7 +256,7 @@ const CreateElection = () => {
       </Text>
       <Input
         value={electionName}
-        onChange={setElectionName}
+        onChange={(s) => setElectionName(s)}
```
Tyratox, line 259: Is there a reason for this change?

fe1-web/src/features/evoting/screens/CreateElection.tsx
```diff
@@ -250,12 +286,12 @@ const CreateElection = () => {
       {questions.map((value, idx) => (
         // FIXME: Do not use index in key
         // eslint-disable-next-line react/no-array-index-key
-        <View key={idx.toString()}>
+        <View key={idx.toString()} style={{ marginBottom: Spacing.x2 }}>
```
Tyratox, line 289: Inline styles are discouraged, I would create a small StyleSheet at the top of the file with a `questionWrapper` style :)

fe1-web/src/features/evoting/screens/CreateElection.tsx
```diff
@@ -250,12 +286,12 @@ const CreateElection = () => {
       {questions.map((value, idx) => (
```
Tyratox, line 286:
```suggestion
      {questions.map((question, idx) => (
```

?

fe1-web/src/features/evoting/screens/CreateElection.tsx
```diff
@@ -269,27 +305,43 @@ const CreateElection = () => {
...
+            values={value.ballot_options}
             placeholder={STRINGS.election_create_option_placeholder}
             onChange={(ballot_options: string[]) =>
               setQuestions((prev) =>
                 prev.map((item, id) =>
                   id === idx
                     ? {
                         ...item,
-                        ballot_options: ballot_options,
+                        ballot_options: ballot_options.filter((option) => option !== ''),
```
Tyratox, line 324: This likely causes a ballot option to disappear when it's text is deleted even if it is not the last one..? For `trimmedQuestions` you filter them a second time anyway 😄

fe1-web/src/features/evoting/screens/CreateElection.tsx
Tyratox: Okay you have to do trimming here again, I think having a `trimmedElectionName` would definitely be a good idea 😅

fe1-web/src/features/evoting/screens/__tests__/CreateElection.test.tsx
```diff
@@ -45,10 +72,96 @@ describe('CreateElection', () => {
...
+        <FeatureContext.Provider value={contextValue}>
+          <MockNavigator component={CreateElection} />
+        </FeatureContext.Provider>
+      </Provider>,
+    );
+    const electionName = getByTestId('election_name_selector');
+    const question = getByTestId('question_selector_0');
+    fireEvent.changeText(electionName, '     ');
+    fireEvent.changeText(question, '        ');
+    expect(toJSON()).toMatchSnapshot();
```
Tyratox, line 96: I think waiting here is necessary for the snapshot to properly reflect the changes? 😅

fe1-web/src/features/evoting/screens/__tests__/CreateElection.test.tsx
```diff
@@ -45,10 +72,96 @@ describe('CreateElection', () => {
...
+      },
+      {
+        question: '   Question     1        ',
+        ballot_options: ['    Answer 0    ', '  Answer 1  ', 'will be removed'],
+        voting_method: STRINGS.election_method_Plurality,
+        id: '',
+        write_in: false,
+      },
+    ];
+    jest.setSystemTime(Date.UTC(2023, 3, 21));
```
Tyratox, line 116: Why is this necessary? If you want to workaround specific time values you can always use https://jestjs.io/docs/expect#expectanything

Tyratox (commented):
Phew that was quite the refactor, great work!! 🥳🥳👏

Very happy to see you managed to get the test working!! While I agree that the workaround is not ideal I think it is fair especially given the use case you describe 😄

I added a few minor comments (a few of them are just some additional thoughts) but it looks very very good overall! 😄

Tyratox:
Also I just saw that sonar cloud complains about the component being to big. Maybe you can find a part that can be isolated into a separate function or component :)

fe1-web/src/core/components/__tests__/__snapshots__/TextInputList.test.tsx.snap
```diff
@@ -0,0 +1,309 @@
...
+                "borderRadius": 8,
+                "borderWidth": 1,
+                "flex": 1,
+                "padding": 8,
+                "width": 50,
+              },
+            ]
+          }
+          testID="x_option_1_input"
+          value="   test 2"
```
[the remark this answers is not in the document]
MeKHell: No since the input is trimmed only on submission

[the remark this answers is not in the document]
Tyratox: Ohhh you‘re right 👍👍

fe1-web/src/core/components/RemovableTextInput.tsx
```diff
@@ -25,33 +26,40 @@ const styles = StyleSheet.create({
...
-      <PoPTouchableOpacity
-        containerStyle={styles.icon}
-        onPress={() => onRemove(id)}
-        testID={testID ? `${testID}_remove` : undefined}>
-        <PoPIcon name="delete" color={Color.primary} size={Icon.size} />
-      </PoPTouchableOpacity>
+      {/* Added for UI coherence and alignment */}
+      {!isRemovable && (
+        <PoPTouchableOpacity containerStyle={styles.icon} onPress={() => {}}>
+          <NavigationPadding />
```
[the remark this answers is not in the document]
MeKHell: I think that `ButtonPadding` is a bit more specific and closer to the usage of this component.

fe1-web/src/features/evoting/screens/CreateElection.tsx
MeKHell: This was avoided to not have some `question.question` few lines below, I think. I am not sure if it would be clearer. Changed to a longer but better name.

fe1-web/src/features/evoting/screens/__tests__/CreateElection.test.tsx
MeKHell: It is not necessary.

fe1-web/src/features/evoting/screens/__tests__/CreateElection.test.tsx
MeKHell: It is required, because I want the ID to be checked too. And it depends on the current time.

Tyratox: I see 👍
I guess you did not come across this yet but the global system time should already be fixed for all tests here: https://github.com/dedis/popstellar/blob/master/fe1-web/jest/setupFakeTimers.js#L6 :)

fe1-web/src/features/evoting/screens/CreateElection.tsx
MeKHell: And I just noticed that this behavior was wrong. Now I need to filter them three times because we do not have a unique question id at this stage. :sweat_smile:

fe1-web/src/features/evoting/screens/CreateElection.tsx
```diff
@@ -63,25 +63,51 @@ const MIN_BALLOT_OPTIONS = 2;
...
+    .filter((ballot) => ballot !== '');
+  return (
+    question.ballot_options.length > 0 &&
+    (new Set(trimmedBallotOptions).size !== question.ballot_options.length ||
+      trimmedBallotOptions.length < MIN_BALLOT_OPTIONS)
+  );
+};
 
 /**
  * Checks whether a question title is not unique within a list of questions
```
Tyratox, line 99:
```suggestion
 * Checks whether a question title is unique within a list of questions
```

fe1-web/src/features/evoting/screens/CreateElection.tsx
```diff
@@ -125,6 +151,67 @@ const createElection = (
...
+  };
+};
+
+/**
+ * Display the error message that are not directly related to a question
+ * @param isConnected If the user isConnected to the LAO
+ * @param electionName the trimmed name of the election
+ * @param trimmedQuestions the trimmed questions of the election
+ */
+const globalErrorMessages = (
```
Tyratox, line 174: Is there a particular reason why you used a function instead of a component?

fe1-web/src/features/evoting/screens/__tests__/CreateElection.test.tsx
```diff
@@ -45,10 +72,97 @@ describe('CreateElection', () => {
...
+        </FeatureContext.Provider>
+      </Provider>,
+    );
+    const electionName = getByTestId('election_name_selector');
+    const question = getByTestId('question_selector_0_input');
+    const option = getByTestId('question_0_ballots_option_0_input');
+    fireEvent.changeText(electionName, '     ');
+    fireEvent.changeText(question, '        ');
+    fireEvent.changeText(option, '        ');
+    waitFor(() => expect(toJSON()).toMatchSnapshot());
```
Tyratox, line 98: Not sure this accomplished what you want 😅 Wait for executes the passed function multiple times and succeeds if the passed function succeeds once. (https://testing-library.com/docs/dom-testing-library/api-async/#waitfor)

Tyratox (approved):
LGTM! 😁

Tyratox:
Just make sure the sonar cloud issues are fine and mark them accordingly :)

fe1-web/src/features/evoting/screens/__tests__/CreateElection.test.tsx
MeKHell: Indeed, it is not needed, but I added it after https://github.com/dedis/popstellar/pull/1456#discussion_r1161118777, because I thought you wanted a more explicit test. My bad.

fe1-web/src/features/evoting/reducer/__tests__/ElectionReducer.test.ts
```diff
@@ -77,20 +72,14 @@ describe('ElectionReducer', () => {
...
-            allIds: [],
-          } as ElectionReducerState,
-          updateElection(mockElectionState),
-        ),
-      ).toThrow();
-    });
+    expect(() =>
+      electionReduce(defaultElectionReducerState, updateElection(mockElectionState)),
+    ).toThrow();
+    it('throws an error when trying to update an inexistent election', () => {});
```
dayan9265, line 82: I suppose these lines 3 lines of test should be in the test 'throws an error when trying to update an inexistent election'?

fe1-web/src/features/evoting/reducer/__tests__/ElectionReducer.test.ts
```diff
@@ -102,26 +91,17 @@ describe('ElectionReducer', () => {
...
-            allIds: [],
-          } as ElectionReducerState,
-          removeElection(mockElection.id),
-        ),
-      ).toThrow();
-    });
+    expect(() =>
+      electionReduce(defaultElectionReducerState, removeElection(mockElection.id)),
+    ).toThrow();
+    it('throws an error when trying to remove an inexistent election', () => {});
```
dayan9265, line 104: same here for the test

dayan9265 (changes requested):
That was quite a big change to avoid space in question ^^, but it looks good, I have two added two comments about some tests that looked weird, but otherwise good job 👍

fe1-web/src/features/evoting/reducer/__tests__/ElectionReducer.test.ts
MeKHell: Nice catch. This is indeed an error.

fe1-web/src/features/evoting/reducer/__tests__/ElectionReducer.test.ts
MeKHell: fixed

fe1-web/src/features/evoting/reducer/__tests__/ElectionReducer.test.ts
```diff
@@ -77,18 +72,13 @@ describe('ElectionReducer', () => {
...
     it('throws an error when trying to update an inexistent election', () => {
       expect(() =>
-        electionReduce(
-          {
-            byId: {},
-            allIds: [],
-          } as ElectionReducerState,
-          updateElection(mockElectionState),
-        ),
+        electionReduce(defaultElectionReducerState, removeElection(mockElection.id)),
```
dayan9265, line 81: is it wanted that the `updateElection` is now `removeElection`?

dayan9265 (approved):
Looks good to me, I just have a small question (see comment), but otherwise it seems perfect 😄

fe1-web/src/features/evoting/reducer/__tests__/ElectionReducer.test.ts
MeKHell: No, I do not know how I did not catch it before. Thanks

Tyratox (approved):
A few small comments but I don't want to block this PR further. If you think the filtering is correct like this, feel free to handle the other comments how you prefer :)

fe1-web/src/features/evoting/screens/CreateElection.tsx
MeKHell: If we do not remove the empty ones, we will have a new empty line for every character input, since our list is now a controlled component.

Tyratox: Ah I see you're not removing it, then it is necessary yes. I thought you changed it (https://github.com/dedis/popstellar/pull/1456/files/dc095b6cec3c47c3966620b3a5ff2a1baa70577a#r1161113914) because I am pretty sure that now there is a bug if we delete all text in the not last ballot option (it is removed). Personally I would deem this weird/bad UX but it seems it is what you want?

PR #1458, opened 2023-03-15
Added hash tests for emoji

### LINKED ISSUE #1442: Add emojis corner case to hash function unit tests
Just to follow up on issue #1295 and the tests added in pr #1440 in the scala backend:

All the backend and frontend should agree on these added inputs/outputs.

Test added : https://github.com/dedis/popstellar/blob/master/be2-scala/src/test/scala/ch/epfl/pop/model/objects/HashSuite.scala#L56-L75

- [ ] be1-go
- [x] fe1-web
- [x] fe2-android

This issue should be closed once we all agree/confirm the correctness of this updated truth table. 

The hash function: https://github.com/dedis/popstellar/blob/master/docs/protocol.md#concatenation-for-hashing

matteosz:
- This PR resolves #1442, aligning with the other sub-systems.
- It simply adds 3 new unit tests for the hash function in presence of special characters (emoji)


Warhedgehog (approved):
LGTM

PR #1459, opened 2023-03-16
Change extractLaoId behavior

K1li4nL:
This Pr  should solve the following problem: 

`JsonRpcRequest.extractLaoId()` actually calls `Channel.extractChildId()` hence on a channel such as `/root/lao_id/election_id` it will not extract `lao_id` but `election_id`

### DIFF
--- a/be2-scala/src/main/scala/ch/epfl/pop/model/network/JsonRpcRequest.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/model/network/JsonRpcRequest.scala
@@ -62,7 +62,7 @@ class JsonRpcRequest(
     case None => None
   }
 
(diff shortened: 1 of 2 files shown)


BelghmiAmine (changes requested):
Hey ! 


Previously the extractLaoId method was tested on a channel "root/laoId/" (see : JsonRpcRequestSuite.scala). I suggest that you change this file so that it is tested with a channel path that does not end with a laoId. Otherwise it is all good for me ;))

BelghmiAmine (approved):
All good for me !

Ajkunas (approved):
All good to me too !
Little rermark: I think it would be a good idea to define a `HASH_ERROR` variable, so that we can use it in any files. I will let you figure out where we should put it.

PR #1460, opened 2023-03-16
Social - make chirp like/dislike exclusive

dayan9265:
These changes makes that if a user likes a chirp that he previously had disliked, it will remove the dislike reaction (same thing for a liked chirp). This small change avoids the situation of a chirp being liked and disliked which in theory doesn't make a lot of sense.

### DIFF
--- a/fe1-web/src/features/social/components/ChirpCard.tsx
+++ b/fe1-web/src/features/social/components/ChirpCard.tsx
@@ -155,9 +155,8 @@ const ChirpCard = ({ chirp, isFirstItem, isLastItem }: IPropTypes) => {
 
   const [showDeleteConfirmation, setShowDeleteConfirmation] = useState(false);
 
-  const thumbsPressed = (isThumbsUp: boolean) => {
-    const reactionToAdd = isThumbsUp ? '👍' : '👎';
-    const reactionToRemove = isThumbsUp ? '👎' : '👍';
+  const thumbsPressed = (reactionToAdd: '👍' | '👎') => {
+    const reactionToRemove = reactionToAdd === '👍' ? '👎' : '👍';
 
     if (reacted[reactionToAdd]) {
       deleteReaction(reacted[reactionToAdd].id);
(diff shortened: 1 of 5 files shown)


MeKHell (changes requested):

MeKHell (changes requested):
Simple refactoring needed.

fe1-web/src/features/social/components/ChirpCard.tsx
```diff
@@ -155,6 +155,26 @@ const ChirpCard = ({ chirp, isFirstItem, isLastItem }: IPropTypes) => {
 
   const [showDeleteConfirmation, setShowDeleteConfirmation] = useState(false);
 
+  const thumbsPressed = (isThumbsUp: boolean) => {
+    const reactionToAdd = isThumbsUp ? '👍' : '👎';
+    const reactionToRemove = isThumbsUp ? '👎' : '👍';
```
Tyratox, line 160: What about

```suggestion
  const thumbsPressed = (reactionToAdd: '👍' | '👎') => {
    const reactionToRemove = isThumbsUp ? '👎' : '👍';
```

fe1-web/src/features/social/components/ChirpCard.tsx
```diff
@@ -155,6 +155,26 @@ const ChirpCard = ({ chirp, isFirstItem, isLastItem }: IPropTypes) => {
...
+    // If the user has already reacted with the other reaction, remove it
+    if (reacted[reactionToRemove]) {
+      deleteReaction(reacted[reactionToRemove].id);
+    }
+
+    addReaction(reactionToAdd);
+  };
+
+  const thumbsUpPressed = () => thumbsPressed(true);
+  const thumbsDownPressed = () => thumbsPressed(false);
```
Tyratox, line 176: If you need to pass an argument to a function feel free to write

`onPress={() => thumbsPressed("👍")}`

in the component directly. If there are no arguments, `onPress={someFunction}` is usually preferred.

MeKHell (approved):
This looks good to me

Tyratox (approved):
Looks good to me as well 😊

PR #1469, opened 2023-03-22
Fix #2 to apk deploy + caching

matteosz:
This PR should:
- fix the error in the deploy of fe2 
- implement caching of the gradle build command


pierluca (approved):
LGTM

PR #1474, opened 2023-03-27
First prototype for the Authorization server of PopCHA

Nazianzenov:
- major objects definitions (authorization server, storage)
- new Authorization server serving both WebSocket and HTTP
- in-memory storage
- html page for the QR code
- small prototype without authorization request validation
- not tested yet


be1-go/network/socket/socket.go
```diff
@@ -24,6 +24,8 @@ const (
 
 	// ServerSocketType denotes a server.
 	ServerSocketType SocketType = "server"
+
+	AuthServerSocketType SocketType = "authorize"
```
MariemBaccari, line 28: Maybe you can add a comment here for AuthServerSocketType like for ServerSocketType just for consistency

be1-go/network/socket/socket.go
```diff
@@ -288,3 +290,17 @@ func NewServerSocket(receiver chan<- IncomingMessage,
 			conn, wg, done, log),
 	}
 }
+
+type AuthServerSocket struct {
```
MariemBaccari, line 294: Same remark about comments here

be1-go/network/socket/socket.go
Nazianzenov: yep i'll do that, thanks!

PR #1476, opened 2023-03-28
Add heartbeat get_messages_by_id Json support

K1li4nL:
This Pr does a couple things: 

1) Refactor our Params type to ParamsWithChannel and convert Params to an abstract class

The above is done so that JsonRpcType can accept heartbeat and get_messages_by_id, since they don't have a `channel` type.

2) Add ParamsWithMap which fits the new methods
3) Add the support to read and write the new methods jsons

### DIFF
--- a/be2-scala/src/main/scala/ch/epfl/pop/model/network/method/Params.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/model/network/method/Params.scala
@@ -2,6 +2,9 @@ package ch.epfl.pop.model.network.method
 
 import ch.epfl.pop.model.objects.Channel
 
-class Params(val channel: Channel) {
-  def hasMessage: Boolean = false
+abstract class Params {
+  // Default channel
+  val channel: Channel = Channel.ROOT_CHANNEL
+  def hasMessage: Boolean
+
 }
--- a/be2-scala/src/test/scala/ch/epfl/pop/model/network/JsonRpcRequestSuite.scala
+++ b/be2-scala/src/test/scala/ch/epfl/pop/model/network/JsonRpcRequestSuite.scala
@@ -3,7 +3,7 @@ package ch.epfl.pop.model.network
 import ch.epfl.pop.model.network.method.message.Message
 import ch.epfl.pop.model.network.method.message.data.lao.CreateLao
 import ch.epfl.pop.model.network.method.message.data.{ActionType, MessageData, ObjectType}
-import ch.epfl.pop.model.network.method.{Params, ParamsWithMessage}
+import ch.epfl.pop.model.network.method.{ParamsWithChannel, ParamsWithMessage}
 import ch.epfl.pop.model.objects._
 import org.scalatest.funsuite.{AnyFunSuite => FunSuite}
 import org.scalatest.matchers.should.Matchers
@@ -16,7 +16,7 @@ class JsonRpcRequestSuite extends FunSuite with Matchers {
   private final val electionId: String = "defg"
   private final val channelEx: Channel = Channel(Channel.ROOT_CHANNEL_PREFIX + laoId)
   private final val channelElectionEx: Channel = Channel(Channel.ROOT_CHANNEL_PREFIX + laoId + Channel.CHANNEL_SEPARATOR + electionId)
-  private final val params: Params = new Params(channelEx)
+  private final val params: ParamsWithChannel = new ParamsWithChannel(channelEx)
   private final val paramsWithMessage: ParamsWithMessage = new ParamsWithMessage(channelEx, messageEx)
   private final val paramsWithMessage2: ParamsWithMessage = new ParamsWithMessage(channelElectionEx, messageEx)
   private final val rpc: String = "rpc"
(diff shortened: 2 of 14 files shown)


be2-scala/src/main/scala/ch/epfl/pop/model/network/method/Params.scala
BelghmiAmine, line 1: Quick suggestion to add a boolean field (def hasMap: Boolean) in order to support JsonRpc's with maps.

BelghmiAmine (changes requested):

be2-scala/src/main/scala/ch/epfl/pop/model/network/method/Params.scala
K1li4nL: I thought about that but it didn't seemed helpful because for now at least `hasMessage` and `hasChannel` are enough to differentiate the 3 types of `Params`. Since every function added in `Params` has to be implemented in the classes that extends it, it can become a bit heavy.

BelghmiAmine (dismissed):

be2-scala/src/test/scala/ch/epfl/pop/json/HighLevelProtocolSuite.scala
```diff
@@ -126,6 +130,48 @@ class HighLevelProtocolSuite extends FunSuite with Matchers {
         an[IllegalArgumentException] should be thrownBy Message.buildFromJson(source)
     }
   }
+
+  test("parse correctly heartbeat") {
+    val heartbeat = "{\n    \"jsonrpc\": \"2.0\",\n    \"method\": \"heartbeat\",\n    \"params\": {\n        \"/root/nLghr9_P406lfkMjaNWqyohLxOiGlQee8zad4qAfj18=/social/8qlv4aUT5-tBodKp4RszY284CFYVaoDZK6XKiw9isSw=\": [\n            \"DCBX48EuNO6q-Sr42ONqsj7opKiNeXyRzrjqTbZ_aMI=\"\n        ],\n        \"/root/nLghr9_P406lfkMjaNWqyohLxOiGlQee8zad4qAfj18=/HnXDyvSSron676Icmvcjk5zXvGLkPJ1fVOaWOxItzBE=\": [\n            \"z6SbjJ0Hw36k8L09-GVRq4PNmi06yQX4e8aZRSbUDwc=\",\n            \"txbTmVMwCDkZdoaAiEYfAKozVizZzkeMkeOlzq5qMlg=\"\n        ]\n    }\n}"
```
Ajkunas, line 135: A good idea would be to add a function which directly returns the pretty print version of the json methods found in the resources file.
You can find a similar function, `getExampleMessage`  in the `MessageDataProtocolSuite.scala`.

Ajkunas (commented):
All is good. I have a little suggestion for the generation of your tests.

be2-scala/src/test/scala/ch/epfl/pop/json/HighLevelProtocolSuite.scala
K1li4nL: I got my way out of writing a pretty print function. I can call `HighLevelProtocol.write()` to get a `spray.json` and use that library's pretty print function. Then I read from that string with `JsonRpcRequest.buildFromJson` and check if every field is what's expected. I added some more tests to get a better coverage but I don't think I can go much higher than that, most of the coverage missing is from the renaming of `ParamsHandler` to `ParamsWithChannelHandler` for which I added tests too.

BelghmiAmine (dismissed):

Ajkunas (changes requested):
Nice  tests ! 
I'll let you take in account the remark and I'll approve your PR just after !

be2-scala/src/test/scala/ch/epfl/pop/model/network/JsonRpcRequestSuite.scala
```diff
@@ -19,14 +21,17 @@ class JsonRpcRequestSuite extends FunSuite with Matchers {
...
   private final val methodType: MethodType.MethodType = MethodType.BROADCAST
+  private final val methodType2: MethodType.MethodType = MethodType.HEARTBEAT
   private final val rpcReq: JsonRpcRequest = JsonRpcRequest(rpc, methodType, params, id)
   private final val rpcReq2: JsonRpcRequest = JsonRpcRequest(rpc, methodType, paramsWithMessage, id)
   private final val paramsWithMessageAndDecoded: ParamsWithMessage = new ParamsWithMessage(channelEx, messageLao)
   private final val rpcReq3: JsonRpcRequest = JsonRpcRequest(rpc, methodType, paramsWithMessageAndDecoded, id)
   private final val rpcReq4: JsonRpcRequest = JsonRpcRequest(rpc, methodType, paramsWithMessage2, id)
+  private final val rpcReq5: JsonRpcRequest = JsonRpcRequest(rpc, methodType2, paramsWithMap, None)
 
   test("Constructor works for regular Params and ParamsWithMessage") {
```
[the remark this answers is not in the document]
K1li4nL: Done !

Ajkunas (approved):
All good !
To merge the PR, @pierluca has to force it :)

pierluca:
Merging blindly based on a request by @Ajkunas and @K1li4nL

PR #1479, opened 2023-03-29
Implementing Heartbeat and GetMessagesById methods 

MariemBaccari:
This PR implements the Heartbeat and GetMessagesById methods by:
- Defining the method structs in go 
- Adds a ticker that activates sending a heartbeat message every x seconds
- The server handles a heartbeat message by either not doing anything or sending a getMessagesById if there are any Ids missing
- The answers to getMessagesById are handled in the same way as catchups

However, sending the same message on two different channels might not be well supported for the moment.


be1-go/hub/standard_hub/mod_test.go
```diff
@@ -1793,3 +1952,38 @@ func (f *fakeSocket) ID() string {
 func (f *fakeSocket) Type() socket.SocketType {
 	return socket.ClientSocketType
 }
+
+func IntPointer(i int) *int {
```
Nazianzenov, line 1956: is this used anywhere? also is it public on purpose?

be1-go/hub/standard_hub/message_handling.go
```diff
@@ -193,15 +197,68 @@ func (h *Hub) handleAnswer(senderSocket socket.Socket, byteMessage []byte) error
...
+func (h *Hub) handleGetMessagesByIdAnswer(senderSocket socket.Socket, answerMsg answer.Answer) {
+	messages := answerMsg.Result.GetMessagesByChannel()
+	for channel, messageArray := range messages {
+		for _, msg := range messageArray {
+
+			var messageData message.Message
+
+			err := json.Unmarshal(msg, &messageData)
+			if err != nil {
+				h.log.Error().Msgf("failed to unmarshal message during getMessagesById answer handling: %v", err)
```
Nazianzenov, line 226: To confirm, in this case if a message throws an error when unmarshalled, you still continue and just log the error right?

Nazianzenov (changes requested):
Very good code overall, handlers are quite clean, i still had a few comments about some methods and the way you handle errors in some cases (see comments previously).

be1-go/hub/standard_hub/mod_test.go
MariemBaccari: Good catch, I was supposed to remove it.

be1-go/hub/standard_hub/message_handling.go
MariemBaccari: Yes, I handled it almost exactly the same way as the catchup messages but now that you made a comment I'm not sure. @Anghul could you clarify whether there's a reason catchup answers are handled that way ?

Anghul: They were handled that way because when receiving an array of message, maybe one of these messages would fail while unmarshalling, and I wanted the code to be able to still treat the following messages and not just fail in the middle of the array, but then you could want to do it another way dending on your protocol.

MariemBaccari: Thanks, it does make sense to do it that way. I think I'll keep it like this for now @Nazianzenov

Nazianzenov (approved):
LGTM now

Nazianzenov (dismissed):
The only change was the dependecies and the go.sum files it seems.
It's good for me.

Anghul (changes requested):
Good job overall, just a few comments.
And finaly, it may be better to do it in another PR, but think if you need to keep the code previously done for the broadcast and/or catchup or not.

be1-go/hub/standard_hub/mod.go
```diff
@@ -79,13 +85,21 @@ type Hub struct {
 	// rootInbox and queries are used to help servers catchup to each other
 	rootInbox inbox.Inbox
 	queries   queries
+
+	//globalInboy is used to remember all messages
+	globalInbox inbox.Inbox
+
+	// messageIdsByChannel stores all the message ids and the corresponding channel ids
+	// to help servers
```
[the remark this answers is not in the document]
MariemBaccari: To help servers determine which channel corresponds to the message ids since the channel isn't stored in the message. I'll clarify that in a new commit.

be1-go/hub/standard_hub/mod.go
```diff
@@ -580,3 +678,47 @@ func generateKeys() (kyber.Point, kyber.Scalar) {
...
+		for _, msg := range messages {
+			_, alreadyStored := h.globalInbox.GetMessage(msg.MessageID)
+			if !alreadyStored {
+				h.globalInbox.StoreMessage(msg)
+				h.addMessageId(channelId, msg.MessageID)
+			}
+		}
+	}
+
+}
```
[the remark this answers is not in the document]
MariemBaccari: I wasn't sure how necessary it was to do that but I was thinking about the messages that are created and sent by the server itself and are only stored at the channel level such as the lao greet. Is there another way to retrieve them ?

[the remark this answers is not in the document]
MariemBaccari: Update: Actually I'm checking with the scala team and if we only need the messages received by the clients I can just remove it entirely.

be1-go/hub/standard_hub/message_handling.go
```diff
@@ -454,3 +480,120 @@ func (h *Hub) handleCatchup(socket socket.Socket,
...
+	var heartbeat method.Heartbeat
+
+	err := json.Unmarshal(byteMessage, &heartbeat)
+	if err != nil {
+		return xerrors.Errorf("failed to unmarshal heartbeat message: %v", err)
+	}
+
+	receivedIds := heartbeat.Params
+
+	h.updateRecords()
```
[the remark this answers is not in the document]
MariemBaccari: After testing the servers without it, I think you're correct, it's redundant. I just removed it in the latest commit.

be1-go/hub/standard_hub/mod.go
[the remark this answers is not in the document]
MariemBaccari: Update 2: After discussions, it looks like we also need those messages for now but I think we can have this as a temporary solution since we will be optimizing the heartbeat to send ids by batches as oppposed to sending all the ids we know (for examples the latest N messages, to be decided in the protocol).

Anghul (approved):
LGTM. I won't ask you to change the way to get messages, as you told it will change in the future, but for the greet messages (and others) you could also simply store them when they are created/processed by this server.

PR #1480, opened 2023-03-29
Modifying the makefile of the go backend to disable CGO:

Nazianzenov:
* disables importing of C into go file
* binary doesn't have external C dependencies
* avoiding library incompatibilities between ubuntu and debian systems


pierluca (approved):
LGTM

PR #1484, opened 2023-04-03
PoPCHA HTTP auth server with in-memory storage

Nazianzenov:
- uncomplete storage
- no communication with the app yet
- unit testing
- property-based testing


be1-go/popcha/storage/mod.go
```diff
@@ -0,0 +1,73 @@
...
+	popTokenToIdentifier map[string]string //hash map enforcing uniqueness on (PopToken,identifier)
+	nonces               map[string]struct{}
+	log                  zerolog.Logger
+}
+
+// NewStorage initializes a Storage object given a logger
+func NewStorage(log zerolog.Logger) (*Storage, error) {
+	return &Storage{
+		lock:                 &sync.Mutex{},
+		popTokenToIdentifier: make(map[string]string, 0),
```
MariemBaccari, line 25: Is there a reason why you specify the 0 argument in `make()` for some usages and others not ? I think 0 is already a default value but I'm not 100% sure about that.

MariemBaccari (changes requested):
The code looks generally clean and well structured. The comments help users who don't have experience with the POPCHA subproject understand the big picture of what's happening. I made a few comments about some small things that can be fixed before I approve the PR.

be1-go/popcha/server.go
```diff
@@ -0,0 +1,410 @@
...
+	"sync"
+	"time"
+)
+
+const (
+	// formatting for auth request validation error
+	errValidAuthFormat = "Error while validating the auth request: %s"
+
+	//error message for invalid response type
+	errInvalidRepsonseType = "response type is invalid for the implicit flow"
```
MariemBaccari, line 31: I already made this comment in my first review but there's this small typo here :)

be1-go/popcha/server.go
```diff
@@ -0,0 +1,410 @@
...
+		as.handleBadRequest(w, err, "Error while generating PoPCHA QRCode")
+	}
+}
+
+func (as *AuthorizationServer) handleBadRequest(w http.ResponseWriter, err error, msg string) {
+	as.log.Err(err).Msg(msg)
+	w.WriteHeader(badRequestCode)
+	_, err = w.Write([]byte(strconv.Itoa(badRequestCode) + " - " + err.Error()))
+	if err != nil {
+		as.log.Err(err).Msg("Error while writing error message in the HTTP repsonse body")
```
MariemBaccari, line 274: Same "repsonse" typo here as well

be1-go/popcha/server.go
```diff
@@ -0,0 +1,410 @@
...
+	ResponseType = "response_type"
+	State        = "state"
+	ResTypeMulti = "id_token token"
+	OpenID       = "openid"
+	Profile      = "profile"
+	LoginHint    = "login_hint"
+)
+
+// clientParams implements op.Client
+// as PoPCHA don't support registered clients, clientParams is an alternative which defines  parameters
```
MariemBaccari, line 63: I might be nitpicking here but there are two spaces between "defines" and "parameters"

be1-go/popcha/server.go
```diff
@@ -0,0 +1,410 @@
...
+// for any client sending an auth request. As an example, any client requesting authorization has to use
+// the implicit flow.
+type clientParams struct {
+	clientID     string
+	redirectURIs []string
+	resType      oidc.ResponseType
+}
+
+/*
+Most of the methods of op.Client are implemented, however some of them
```
MariemBaccari, line 73: Small suggestion: I think that physically separating the functions that are implemented and the ones that are not by some comments (in two different "blocs") would make the code more readable. Future programmers might find it useful to only see the functions they can directly use and the ones that just panic separately.

MariemBaccari (changes requested):
Nice job ! Thanks for adding the comments. I wrote a few comments about typos and readability but the rest looks good to me.

be1-go/popcha/server.go
Nazianzenov: Oh my bad, i will correct that asap

be1-go/popcha/server.go
Nazianzenov: same, i will correct that!

be1-go/popcha/server.go
Nazianzenov: i think i use Goland IDE formatting, i will take a look at that !

be1-go/popcha/server.go
Nazianzenov: True! That's a very good suggestion definitely.

MariemBaccari (dismissed):
Looks good to me now !

Anghul (changes requested):
Good job overall. Good, well commented tests. Just a few comments before being able to approve it.

be1-go/popcha/server.go
```diff
@@ -0,0 +1,416 @@
...
+	"sync"
+	"time"
+)
+
+const (
+	// formatting for auth request validation error
+	errValidAuthFormat = "Error while validating the auth request: %s"
+
+	//error message for invalid response type
+	errInvalidResponseType = "response type is invalid for the implicit flow"
```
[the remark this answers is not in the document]
Nazianzenov: Unfortunately, the library doesn't cover the specific check for the response type we use. However, I can change the string and specify which response type is valid.

be1-go/popcha/server.go
```diff
@@ -0,0 +1,416 @@
...
+	ClientID     = "client_id"
+	Nonce        = "nonce"
+	Scope        = "scope"
+	RedirectURI  = "redirect_uri"
+	ResponseType = "response_type"
+	State        = "state"
+	ResTypeMulti = "id_token token"
+	OpenID       = "openid"
+	Profile      = "profile"
+	LoginHint    = "login_hint"
```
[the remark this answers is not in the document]
Nazianzenov: It is very likely i will need those outside, in order to avoid duplicating those variable.

Anghul (approved):
LGTM now

PR #1486, opened 2023-04-04
Back navigation on transactions history

matteosz:
This PR makes the following changes:
- going back from the history of transactions opens the last accesses fragment before entering there. Previously there was a non-smooth behaviour for which going back from the history was always taking to digital cash home.
- Fixed the UI of the roll call in landscape orientation: if the QR is displayed, then the rotation is locked (as it could be not displayed entirely). Also there's another fix on the list of attendees in landscape orientation

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/remote/GlobalNetworkManager.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/remote/GlobalNetworkManager.java
@@ -20,7 +20,7 @@ import io.reactivex.disposables.Disposable;
 @Singleton
 public class GlobalNetworkManager implements Disposable {
 
-  private static final String DEFAULT_URL = "ws://10.0.2.2:9000/organizer/client";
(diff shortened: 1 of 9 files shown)


fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/remote/GlobalNetworkManager.java
```diff
@@ -20,7 +20,7 @@
 @Singleton
 public class GlobalNetworkManager implements Disposable {
 
-  private static final String DEFAULT_URL = "ws://10.0.2.2:9000/organizer/client";
+  private static final String DEFAULT_URL = "ws://10.0.2.2:9000/client";
```
simone-kalbermatter, line 23: Good idea to change this because it was confusing.

matteosz: Yeah, maybe in the past it was in sync but now it's just a waste of time as to test you have everytime to modify it

Warhedgehog (approved):
LGTM

PR #1493, opened 2023-04-07
Connection to multiple servers

matteosz:
This PR aims to improve the reliability of the system by allowing the client to connect to multiple servers at the same time. In particular, this can be useful in the future to develop mechanisms to bypass censorship.

MultiConnection
----------------
Currently it implements a MultiConnection class that simply subscribes to the channel with each peer. In the future optimisations can be made to reduce traffic (e.g. exchanging ids with peers instead of whole messages).
The peers considered are only the ones to the "main" server, so successive GreetLao messages have no effect.

I tested it by connecting two times to the same server (Go) and to two local servers (Scala and Go).

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/remote/MultiConnection.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/remote/MultiConnection.java
@@ -10,23 +10,59 @@ import java.util.function.Function;
 
 import io.reactivex.Observable;
 
+/**
+ * This class extends Connection to represent a set of Connections to multiple servers.
+ *
+ * <p>The superclass works as main connection to the first server, the others are instantiated in
+ * the connectionMap.
+ */
 public class MultiConnection extends Connection {
-  private Function<String, Connection> connectionProvider;
+
+  /** Caller method for the connectionFactory to instantiate connections */
+  private final Function<String, Connection> connectionProvider;
+
+  /** Map a PeerAddress (url for now) to its connection */
   private final Map<PeerAddress, Connection> connectionMap = new HashMap<>();
 
-  public MultiConnection(Connection connection) {
-    super(connection);
-  }
-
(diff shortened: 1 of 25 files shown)


simone-kalbermatter (dismissed):
Impressive work! Clean, well documented and keeps the existing structure.

fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/remote/MultiConnection.java
Warhedgehog, line 1: I'm missing the point of having the architecture done so. Why treating the first connection different than the others ? By that I mean that I would imagine the result to be simpler with simply populating the connection with every connection including the first one

Warhedgehog (changes requested):
Impressive work and well documented 👍 A couple question on my end

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/LaoActivity.java
```diff
@@ -277,15 +273,6 @@ private void setEventsTab() {
...
 
-  /** Restore the fragment contained in the stack as container of the current lao */
-  public void resetLastFragment() {
-    Fragment fragment = fragmentStack.pop();
-    getSupportFragmentManager()
-        .beginTransaction()
-        .replace(R.id.fragment_container_lao, fragment)
-        .commit();
-  }
-
```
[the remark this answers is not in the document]
matteosz: Yeah I just realised that I've committed some conflicting and unwanted changes due to a previous merge I've done 😅 , I'll fix that

fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/remote/MultiConnection.java
Warhedgehog: Makes sense 👍

Warhedgehog (approved):
LGTM, great work !

PR #1494, opened 2023-04-07
Refactoring (general + hardcoded strings removed)

### LINKED ISSUE #1377: There exist string elements displayed to the user that are hard-coded
In Android, most displayed string should be retrieved from the resources, as it allows the application to be easily localized in another language.

Most string values are retrieved from the resources, but there are some elements that are still hard-coded.
They must be addressed as it is a bad practice.

matteosz:
This PR resolves #1377.

It removes all the hardcoded strings which are displayed to the user and does a general refactoring under the following aspects:
- removed unnecessary `this` references
- added square brackets for each conditional statement
- converted some for loops in their stream version
- removed deprecated methods

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/di/KeysetModule.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/di/KeysetModule.java
@@ -4,6 +4,7 @@ import android.content.Context;
 import android.content.SharedPreferences;
 import android.util.Log;
 
+import com.google.crypto.tink.KeyTemplates;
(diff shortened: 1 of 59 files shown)


simone-kalbermatter (dismissed):
LGTM! Cleaned up the code a lot 😄

Warhedgehog (changes requested):
Nice work ! Some of the classes cleaned up were quite old and dusty 😄 
I personally don't have anything against some use of `this`, but as we said it's your project now 😄 
Just one question

Warhedgehog (approved):
LGTM

PR #1495, opened 2023-04-08
Encrypted votes to non-UI thread

### LINKED ISSUE #1261: While the encrypted vote is computed, the UI is not disabled
While the vote is being encrypted, the UI does not signal the user it's computing (which can take a few seconds) and allows the user to click on stuff when they should not be allowed to.

Additionally, it's too long so there is probably an issue. In any case it could be done in a non UI thread.

### LINKED ISSUE #1331: Consistent behaviour after voting on both front-ends
Currently there is a discrepancy between the two front-ends with respect to UX:
fe-1 stays on the voting screen after having pressed the vote button while fe-2 navigates back to a screen.

[ ] Decide what the behavior should be
[ ] Change one or both front-ends to reflect this

matteosz:
This PR resolves #1261 and resolves #1331 .

Changes
---------

- Moved the computation of the encrypted vote to a non-UI thread, displayed a progress bar and blocked touch inputs during the loading screen
- Moved to another thread the keyset computation
- Solved the inconsistency with FE1, which is staying on the castVoteFragment after voting

*Disclaimer*: the code coverage is under threshold as KeySetModule cannot be directly tested with Roboeletric, but rather it's tested anyway without calling those methods

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/di/KeysetModule.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/di/KeysetModule.java
@@ -48,7 +48,6 @@ public class KeysetModule {
   @Singleton
   public static AndroidKeysetManager provideDeviceKeysetManager(
       @ApplicationContext Context applicationContext) {
-
     try {
       SharedPreferences.Editor editor =
           applicationContext
@@ -70,14 +69,14 @@ public class KeysetModule {
                       .withMasterKeyUri(DEVICE_MASTER_KEY_URI)
                       .build();
                 } catch (GeneralSecurityException | IOException e) {
-                  throw new SecurityException(
-                      "Could not retrieve the device keyset from the app", e);
+                  throw new SecurityException(e);
                 }
               },
               Executors.newSingleThreadExecutor());
 
       return future.join();
     } catch (GeneralSecurityException e) {
+      Log.e(TAG, "Could not retrieve the device keyset from the app", e);
       throw new SecurityException("Could not retrieve the device keyset from the app", e);
     }
   }
@@ -90,12 +89,25 @@ public class KeysetModule {
     try {
       AesGcmKeyManager.register(true);
       AeadConfig.register();
-      return new AndroidKeysetManager.Builder()
-          .withSharedPref(applicationContext, WALLET_KEYSET_NAME, WALLET_SHARED_PREF_FILE_NAME)
-          .withKeyTemplate(KeyTemplates.get("AES256_GCM_RAW"))
-          .withMasterKeyUri(WALLET_MASTER_KEY_URI)
-          .build();
-    } catch (GeneralSecurityException | IOException e) {
+
+      CompletableFuture<AndroidKeysetManager> future =
+          CompletableFuture.supplyAsync(
+              () -> {
+                try {
+                  return new AndroidKeysetManager.Builder()
+                      .withSharedPref(
+                          applicationContext, WALLET_KEYSET_NAME, WALLET_SHARED_PREF_FILE_NAME)
+                      .withKeyTemplate(KeyTemplates.get("AES256_GCM_RAW"))
+                      .withMasterKeyUri(WALLET_MASTER_KEY_URI)
+                      .build();
+                } catch (GeneralSecurityException | IOException e) {
+                  throw new SecurityException(e);
+                }
+              },
+              Executors.newSingleThreadExecutor());
+
+      return future.join();
+    } catch (GeneralSecurityException e) {
       Log.e(TAG, "Could not retrieve the wallet keyset from the app", e);
       throw new SecurityException("Could not retrieve the wallet keyset from the app", e);
     }
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/di/KeysetModule.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/di/KeysetModule.java
@@ -92,7 +92,7 @@ public class KeysetModule {
       AeadConfig.register();
       return new AndroidKeysetManager.Builder()
           .withSharedPref(applicationContext, WALLET_KEYSET_NAME, WALLET_SHARED_PREF_FILE_NAME)
-          .withKeyTemplate(AesGcmKeyManager.rawAes256GcmTemplate())
+          .withKeyTemplate(KeyTemplates.get("AES256_GCM_RAW"))
           .withMasterKeyUri(WALLET_MASTER_KEY_URI)
           .build();
     } catch (GeneralSecurityException | IOException e) {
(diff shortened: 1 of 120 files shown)


simone-kalbermatter (changes requested):
Nice work! Just one comment I'm not sure about.

fe2-android/app/src/main/java/com/github/dedis/popstellar/di/KeysetModule.java
```diff
@@ -56,14 +58,25 @@ public static AndroidKeysetManager provideDeviceKeysetManager(
...
+                      .withKeyTemplate(KeyTemplates.get("ED25519_RAW"))
+                      .withMasterKeyUri(DEVICE_MASTER_KEY_URI)
+                      .build();
+                } catch (GeneralSecurityException | IOException e) {
+                  throw new SecurityException(e);
+                }
+              },
+              Executors.newSingleThreadExecutor());
+
+      return future.join();
```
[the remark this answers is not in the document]
matteosz: I see the point but then I should change the signature to return a Completable rather than the `AndroidKeysetManager` and as it's injected I'm not very sure I'm allowed to

fe2-android/app/src/main/java/com/github/dedis/popstellar/di/KeysetModule.java
```diff
@@ -76,12 +89,25 @@ public static AndroidKeysetManager provideWalletKeysetManager(
...
+                      .withKeyTemplate(KeyTemplates.get("AES256_GCM_RAW"))
+                      .withMasterKeyUri(WALLET_MASTER_KEY_URI)
+                      .build();
+                } catch (GeneralSecurityException | IOException e) {
+                  throw new SecurityException(e);
+                }
+              },
+              Executors.newSingleThreadExecutor());
+
+      return future.join();
```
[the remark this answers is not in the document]
matteosz: same comment here as previous thread

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/event/election/ElectionViewModel.java
```diff
@@ -191,10 +202,22 @@ private CastVote createCastVote(List<PlainVote> votes, Election election, LaoVie
...
+                new Handler(Looper.getMainLooper())
+                    .post(
+                        () ->
+                            Toast.makeText(getApplication(), "Vote encrypted !", Toast.LENGTH_LONG)
+                                .show());
+                return new CastVote(encryptedVotes, election.getId(), laoView.getId());
+              },
+              Executors.newSingleThreadExecutor());
+
+      return future.join();
```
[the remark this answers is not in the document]
matteosz: Yes here I can modify the signature and return a completable

simone-kalbermatter (approved):
LGTM!

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/event/election/fragments/CastVoteFragment.java
```diff
@@ -165,15 +193,10 @@ private void castVote(View voteButton) {
...
-                        R.id.fragment_event_list,
-                        EventListFragment::newInstance);
-                    // Toast ? + send back to election screen or details screen ?
-                    Toast.makeText(requireContext(), "vote successfully sent !", Toast.LENGTH_LONG)
-                        .show();
-                  },
+                  () ->
+                      Toast.makeText(
+                              requireContext(), "vote successfully sent !", Toast.LENGTH_LONG)
+                          .show(),
```
Warhedgehog, line 199: Have you checked with fe1 for behaviour coherence ?

fe2-android/app/src/test/ui/robolectric/com/github/dedis/popstellar/ui/lao/event/election/CastVoteFragmentTest.java
```diff
@@ -200,4 +231,30 @@ public void castVoteSendsACastVoteMessage() {
...
+    fragmentRule
+        .getScenario()
+        .onActivity(
+            activity ->
+                activity
+                    .getSupportFragmentManager()
+                    .beginTransaction()
+                    .replace(containerId(), CastVoteFragment.newInstance(SECRET_ELECTION_ID))
+                    .commit());
+  }
```
Warhedgehog, line 259: This feels dirty. Though I don't have a better idea at the moment. Maybe you have one @GabrielFleischer ?

Warhedgehog (approved):
Nice work ! Please don't merge yet as I'd like an additional opinion

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/event/election/fragments/CastVoteFragment.java
matteosz: Yes, we're keeping updated the doc with the inconsistencies so we decided this common behaviour

fe2-android/app/src/test/ui/robolectric/com/github/dedis/popstellar/ui/lao/event/election/CastVoteFragmentTest.java
GabrielFleischer: That's ok, but I think the best would be to move this test to a new class where the scenario rule would create a CastVote with a Secret Election id.

fe2-android/app/src/test/framework/robolectric/java/com/github/dedis/popstellar/di/TestKeysetModule.java
```diff
@@ -39,7 +40,7 @@ public static AndroidKeysetManager provideDeviceKeysetManager() {
       Ed25519PrivateKeyManager.registerPair(true);
       PublicKeySignWrapper.register();
       when(manager.getKeysetHandle())
-          .thenReturn(KeysetHandle.generateNew(Ed25519PrivateKeyManager.rawEd25519Template()));
+          .thenReturn(KeysetHandle.generateNew(KeyTemplates.get("ED25519_RAW")));
```
GabrielFleischer, line 43: Was that necessary ?

matteosz: `rawEd25519Template()` is currently deprecated (see [this](https://google.github.io/tink/javadoc/tink-android/HEAD-SNAPSHOT/deprecated-list.html))

Warhedgehog (approved):
LGTM ! It is indeed much better with the separate test file

Warhedgehog (approved):

PR #1498, opened 2023-04-12
Create basic hello screen for PoPcha

dayan9265:
This PR contains the basic structure for the implementation of the popcha scanner, currently it just displays a hello message with the current laoID and can be accessed with the menu on the left. (The icon chosen is not a definitive choice, it was just to test that it correctly displays an icon)

### DIFF
--- a/fe1-web/src/core/navigation/typing/PoPchaParamList.ts
+++ b/fe1-web/src/core/navigation/typing/PoPchaParamList.ts
@@ -1,1 +1,3 @@
-export type PoPchaParamList = {};
+export type PoPchaParamList = {
+  ['simpleScreen']: undefined;
+};
--- a/fe1-web/src/core/navigation/typing/PoPchaParamList.ts
+++ b/fe1-web/src/core/navigation/typing/PoPchaParamList.ts
@@ -1,0 +1,1 @@
+export type PoPchaParamList = {};
--- a/fe1-web/src/features/popcha/README.md
+++ b/fe1-web/src/features/popcha/README.md
@@ -1,0 +1,6 @@
+# PoPCha
+
+Provides a scanner that allows the user to scan a QR code to authenticate to a 3rd party website that it belongs to a specific LAO. (Maybe later, will display all tokens the user has, to be able to revoked them).
+
+## Dependencies
+
--- a/fe1-web/src/features/popcha/hooks/PoPchaHooks.ts
+++ b/fe1-web/src/features/popcha/hooks/PoPchaHooks.ts
@@ -1,0 +1,17 @@
+import { useContext } from 'react';
+
+import FeatureContext from '../../../core/contexts/FeatureContext';
+import { POPCHA_FEATURE_IDENTIFIER, PoPchaReactContext } from '../interface';
+
+export namespace PoPchaHooks {
+  export const usePoPchaContext = (): PoPchaReactContext => {
+    const featureContext = useContext(FeatureContext);
+    // assert that the social context exists
+    if (!(POPCHA_FEATURE_IDENTIFIER in featureContext)) {
+      throw new Error('PoPcha context could not be found!');
+    }
+    return featureContext[POPCHA_FEATURE_IDENTIFIER] as PoPchaReactContext;
+  };
+
+  export const useCurrentLaoId = () => usePoPchaContext().useCurrentLaoId();
+}
--- a/fe1-web/src/features/popcha/hooks/PoPchaHooks.ts
+++ b/fe1-web/src/features/popcha/hooks/PoPchaHooks.ts
@@ -6,9 +6,8 @@ import { POPCHA_FEATURE_IDENTIFIER, PoPchaReactContext } from '../interface';
 export namespace PoPchaHooks {
   export const usePoPchaContext = (): PoPchaReactContext => {
     const featureContext = useContext(FeatureContext);
-    // assert that the social context exists
+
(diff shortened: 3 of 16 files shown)


fe1-web/src/features/popcha/interface/Features.ts
```diff
@@ -0,0 +1,8 @@
+import { LaoParamList } from '../../../core/navigation/typing/LaoParamList';
+import { NavigationDrawerScreen } from '../../../core/navigation/typing/Screen';
```
MeKHell, line 2: The imports could be simplified to `core/xxxxx`

fe1-web/src/features/popcha/interface/Configuration.ts
```diff
@@ -0,0 +1,16 @@
+import { Hash } from '../../../core/objects';
+import FeatureInterface from '../../../core/objects/FeatureInterface';
```
MeKHell, line 2: The imports could be simplified to `core/xxxxx`

fe1-web/src/features/popcha/hooks/PoPchaHooks.ts
```diff
@@ -0,0 +1,17 @@
+import { useContext } from 'react';
+
+import FeatureContext from '../../../core/contexts/FeatureContext';
```
MeKHell, line 3: The imports could be simplified to `core/xxxxx`

fe1-web/src/features/popcha/screens/__tests__/PoPchaScanner.test.tsx
```diff
@@ -0,0 +1,25 @@
+import { render } from '@testing-library/react-native';
+import React from 'react';
+
+import FeatureContext from 'core/contexts/FeatureContext';
+
+import { mockLaoId } from '../../../../__tests__/utils';
```
MeKHell, line 6: The imports could be simplified to `__tests__/xxxxx`

MeKHell (changes requested):
Few comments on imports. The rest is good

dayan9265:
Good catch for the import path, it should be good now :)

Tyratox:
Is this already ready for review? :)

fe1-web/src/features/popcha/screens/__tests__/PoPchaScanner.test.tsx
MeKHell: You have forgotten this one

MeKHell, line 6:
```suggestion
import { mockLaoId } from '__tests__/utils';
```

fe1-web/src/features/popcha/index.ts
```diff
@@ -0,0 +1,11 @@
+import { POPCHA_FEATURE_IDENTIFIER, PoPchaConfiguration, PoPchaInterface } from './interface';
+import { popchaScannerScreen } from './screens/PoPchaScanner';
+
+export const configure = (configuration: PoPchaConfiguration): PoPchaInterface => {
+  console.log('PoPcha configuration: ', configuration);
```
Tyratox, line 5: Can probably be removed?

fe1-web/src/core/navigation/typing/PoPchaParamList.ts
```diff
@@ -0,0 +1,3 @@
+export type PoPchaParamList = {
+  ['simpleScreen']: undefined;
+};
```
Tyratox, line 3: Do you need this?

Tyratox (commented):
Great work getting the whole dependency injection / inter-feature dependencies right in the first try 🥳🥳

A minor comment: In general we have so far written pop as one word, i.e. Pop(-Token) or pop(-token) but not PoP but if you prefer PoP, that works too :) The former might be easier in some cases, i.e. PopChaHooks etc.

fe1-web/src/features/popcha/README.md
```diff
@@ -0,0 +1,6 @@
+# PoPCha
+
+Provides a scanner that allows the user to scan a QR code to authenticate to a 3rd party website that it belongs to a specific LAO. (Maybe later, will display all tokens the user has, to be able to revoked them).
```
Xelowak, line 3:
```suggestion
Provides a scanner that allows the user to scan a QR code to authenticate to a 3rd party website that it belongs to a specific LAO. (Maybe later, will display all tokens the user has, to be able to revoke them).
```
It could be nice to add a date and a GitHub username when adding comments that use "later" or "for now" to know when you wrote it. You may add something like `(dayan9265, 05-2023)`.

fe1-web/src/features/popcha/interface/Configuration.ts
```diff
@@ -0,0 +1,17 @@
+import { Hash } from 'core/objects';
+import FeatureInterface from 'core/objects/FeatureInterface';
+
+import { PoPchaFeature } from './Features';
+
+export const POPCHA_FEATURE_IDENTIFIER = 'popcha';
+
+export interface PoPchaConfiguration {
+  useCurrentLaoId: () => Hash;
```
Xelowak, line 9: For now there is only one function that is defined in PoPchaConfiguration so this is not really an issue but I would recommend to get used to add complete documentation to each element of the configuration, as we did in other features.

Xelowak (commented):
I've made a couple of comments and I agree with everything Nico said.
I would also suggest to rename `PoPcha` into `PopCha`, at least for variable names in the code as it feels weird to me to write it this way compared to other features/variables that are already there.

PR #1500, opened 2023-04-12
Add Monitor actor

K1li4nL:
This pr add the new actor for the scala backend: Monitor

This actor is tasked with the scheduling of heartbeats.

### DIFF
--- a/be2-scala/src/main/scala/ch/epfl/pop/decentralized/Monitor.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/decentralized/Monitor.scala
@@ -17,10 +17,9 @@ import scala.concurrent.duration.{DurationInt, FiniteDuration}
 // Periodic heartbeats are sent with a period of PERIODIC_HEARTBEAT seconds.
 final case class Monitor(
     heartbeatGenRef: ActorRef,
-    PERIODIC_HEARTBEAT: FiniteDuration,
-    MESSAGE_DELAY: FiniteDuration
-)(implicit system: ActorSystem) extends Actor with ActorLogging with Timers {
-  import system.dispatcher
+    heartbeatRate: FiniteDuration,
+    messageDelay: FiniteDuration
+) extends Actor with ActorLogging with Timers {
 
   // These keys are used to keep track of the timers states
   private val periodicHbKey = 0
@@ -35,7 +34,7 @@ final case class Monitor(
     case Monitor.AtLeastOneServerConnected =>
       connectionMediatorRef = sender()
       if (!timers.isTimerActive(periodicHbKey))
-        timers.startTimerWithFixedDelay(periodicHbKey, TriggerHeartbeat, PERIODIC_HEARTBEAT)
+        timers.startTimerWithFixedDelay(periodicHbKey, TriggerHeartbeat, heartbeatRate)
 
     case Monitor.NoServerConnected =>
       timers.cancelAll()
@@ -45,16 +44,16 @@ final case class Monitor(
       timers.cancel(singleHbKey)
       heartbeatGenRef ! Monitor.GenerateAndSendHeartbeat(connectionMediatorRef)
 
-    case Right(jsonRpcRequest: JsonRpcRequest) =>
-      jsonRpcRequest.getParams match {
+    case Right(jsonRpcMessage: JsonRpcRequest) =>
+      jsonRpcMessage.getParams match {
 
         case _: Heartbeat       => /* Actively ignoring this specific message */
         case _: GetMessagesById => /* Actively ignoring this specific message */
         // For any other message, we schedule a single heartbeat to reduce messages propagation delay
         case _ =>
           if (!timers.isTimerActive(singleHbKey) && timers.isTimerActive(periodicHbKey)) {
-            log.info("Scheduling single heartbeat")
-            timers.startSingleTimer(singleHbKey, TriggerHeartbeat, MESSAGE_DELAY)
+            log.info(s"Scheduling single heartbeat")
+            timers.startSingleTimer(singleHbKey, TriggerHeartbeat, messageDelay)
           }
       }
 
@@ -63,8 +62,8 @@ final case class Monitor(
 }
 
 object Monitor {
-  def props(heartbeatGenRef: ActorRef, PERIODIC_HEARTBEAT: FiniteDuration = 30.seconds, MESSAGE_DELAY: FiniteDuration = 3.seconds)(implicit system: ActorSystem): Props =
-    Props(new Monitor(heartbeatGenRef, PERIODIC_HEARTBEAT, MESSAGE_DELAY)(system))
+  def props(heartbeatGenRef: ActorRef, heartbeatRate: FiniteDuration = 30.seconds, messageDelay: FiniteDuration = 3.seconds): Props =
+    Props(new Monitor(heartbeatGenRef, heartbeatRate, messageDelay))
 
   def sink(monitorRef: ActorRef): Sink[GraphMessage, NotUsed] = {
     Sink.actorRef(
--- a/be2-scala/src/main/scala/ch/epfl/pop/decentralized/Monitor.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/decentralized/Monitor.scala
@@ -25,6 +25,9 @@ final case class Monitor(
   private val periodicHbKey = 0
   private val singleHbKey = 1
 
+  // State of connected servers
+  private var someServerConnected = false
+
   // Monitor is self-contained,
   // To that end it doesn't know the ref of the connectionMediator
   private var connectionMediatorRef = ActorRef.noSender
@@ -32,11 +35,13 @@ final case class Monitor(
   override def receive: Receive = LoggingReceive {
 
     case Monitor.AtLeastOneServerConnected =>
+      someServerConnected = true
       connectionMediatorRef = sender()
       if (!timers.isTimerActive(periodicHbKey))
         timers.startTimerWithFixedDelay(periodicHbKey, TriggerHeartbeat, heartbeatRate)
 
     case Monitor.NoServerConnected =>
+      someServerConnected = false
       timers.cancelAll()
 
     case Monitor.TriggerHeartbeat =>
@@ -46,11 +51,10 @@ final case class Monitor(
 
     case Right(jsonRpcMessage: JsonRpcRequest) =>
       jsonRpcMessage.getParams match {
-
         case _: ParamsWithMap => /* Actively ignoring this specific message */
         // For any other message, we schedule a single heartbeat to reduce messages propagation delay
         case _ =>
-          if (!timers.isTimerActive(singleHbKey) && timers.isTimerActive(periodicHbKey)) {
+          if (someServerConnected && !timers.isTimerActive(singleHbKey)) {
             log.info(s"Scheduling single heartbeat")
             timers.startSingleTimer(singleHbKey, TriggerHeartbeat, messageDelay)
           }
--- a/be2-scala/src/main/scala/ch/epfl/pop/decentralized/Monitor.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/decentralized/Monitor.scala
@@ -1,0 +1,79 @@
+package ch.epfl.pop.decentralized
+
+import akka.NotUsed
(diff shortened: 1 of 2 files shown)


be2-scala/src/main/scala/ch/epfl/pop/decentralized/Monitor.scala
```diff
@@ -0,0 +1,83 @@
...
+      timers.cancelAll()
+
+    case Monitor.TriggerHeartbeat =>
+      log.info("triggering a heartbeat")
+      timers.cancel(singleHbKey)
+      heartbeatGenRef ! Monitor.GenerateAndSendHeartbeat(connectionMediatorRef)
+
+    case Right(jsonRpcMessage: JsonRpcRequest) =>
+      jsonRpcMessage.getParams match {
+
```
Jharaxus, line 49: Is this line return intended (even if it is a small pattern matching)?

be2-scala/src/test/scala/ch/epfl/pop/decentralized/MonitorSuite.scala
```diff
@@ -0,0 +1,80 @@
...
+  override def afterAll(): Unit = {
+    // Stops the test actor system
+    TestKit.shutdownActorSystem(system)
+  }
+
+  test("monitor should start scheduling heartbeats when told a server connected") {
+
+    val testProbe = TestProbe()
+    val monitorRef = system.actorOf(
+      Monitor.props(testProbe.ref, heartbeatRate = 2.seconds, messageDelay = 1.seconds)
```
Jharaxus, line 26: These test settings variables (n seconds) are repeated across the tests so they could gain consistency by becoming attributes of the class

be2-scala/src/test/scala/ch/epfl/pop/decentralized/MonitorSuite.scala
```diff
@@ -0,0 +1,80 @@
...
+    )
+
+    testProbe.send(monitorRef, Monitor.AtLeastOneServerConnected)
+    testProbe.expectMsgType[Monitor.GenerateAndSendHeartbeat]
+    testProbe.expectMsgType[Monitor.GenerateAndSendHeartbeat]
+    testProbe.expectMsgType[Monitor.GenerateAndSendHeartbeat]
```
Jharaxus, line 32: It may be more reliable to add our own timeout limit instead of relying on akka's default one

be2-scala/src/test/scala/ch/epfl/pop/decentralized/MonitorSuite.scala
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  test("monitor should schedule single heartbeat when receiving a Right GraphMessage") {
```
Jharaxus, line 35: This test also tests that the Right(message) only HB if message is a JsonRpcRequest without params with map, so this should be reflected in the test name or be put into a separate test (lines 49-50 and 55-56)

be2-scala/src/test/scala/ch/epfl/pop/decentralized/MonitorSuite.scala
```diff
@@ -0,0 +1,80 @@
...
+    testProbe.expectNoMessage(5.seconds)
+
+    Source.single(Right(JsonRpcRequestExample.subscribeRpcRequest)).to(sink).run()
+    testProbe.expectMsgType[Monitor.GenerateAndSendHeartbeat]
+
+    Source.single(Right(getMessagesById)).to(sink).run()
+    testProbe.expectNoMessage(5.seconds)
+  }
+
+  test("monitor should send heartbeats only when servers are connected") {
```
Jharaxus, line 59: This test could also check that no HB is sent upon receiving a Right(JsonRpcRequest) that would otherwise expect a single HB

Jharaxus (changes requested):
Minor changes, but the overall code is great!

be2-scala/src/main/scala/ch/epfl/pop/decentralized/Monitor.scala
```diff
@@ -0,0 +1,83 @@
...
+        case _: ParamsWithMap => /* Actively ignoring this specific message */
+        // For any other message, we schedule a single heartbeat to reduce messages propagation delay
+        case _ =>
+          if (!timers.isTimerActive(singleHbKey) && timers.isTimerActive(periodicHbKey)) {
```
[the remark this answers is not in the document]
K1li4nL: Definitely, fixed !

be2-scala/src/main/scala/ch/epfl/pop/decentralized/Monitor.scala
K1li4nL: Fixed

be2-scala/src/test/scala/ch/epfl/pop/decentralized/MonitorSuite.scala
K1li4nL: I decided to break it into two, couldn't find a good name that would express all that concisely.

be2-scala/src/test/scala/ch/epfl/pop/decentralized/MonitorSuite.scala
K1li4nL: Also done !

be2-scala/src/main/scala/ch/epfl/pop/decentralized/Monitor.scala
```diff
@@ -25,18 +25,23 @@ final case class Monitor(
...
   // Monitor is self-contained,
   // To that end it doesn't know the ref of the connectionMediator
   private var connectionMediatorRef = ActorRef.noSender
 
   override def receive: Receive = LoggingReceive {
 
     case Monitor.AtLeastOneServerConnected =>
+      someServerConnected = true
       connectionMediatorRef = sender()
       if (!timers.isTimerActive(periodicHbKey))
```
Jharaxus, line 40: You could use `someServerConnected` for this condition too

be2-scala/src/test/scala/ch/epfl/pop/decentralized/MonitorSuite.scala
```diff
@@ -23,58 +27,76 @@ class MonitorSuite extends TestKit(ActorSystem("MonitorSuiteActorSystem")) with
...
-    testProbe.expectNoMessage(5.seconds)
+    testProbe.expectNoMessage(timeout)
   }
 
   test("monitor should send heartbeats only when servers are connected") {
 
     val testProbe = TestProbe()
     val monitorRef = system.actorOf(
-      Monitor.props(testProbe.ref, heartbeatRate = 1.seconds, messageDelay = 60.seconds)
+      Monitor.props(testProbe.ref, heartbeatRate = fastRate, messageDelay = slowRate)
```
Jharaxus, line 83: `messageDelay` should be `fastRate` otherwise even if an unexpected single HB is scheduled it would not be sent before the test timeout

Jharaxus (changes requested):

be2-scala/src/main/scala/ch/epfl/pop/decentralized/Monitor.scala
K1li4nL: Not sure, the logic here is that I want the timer to be set again only when it wasn't, regardless of server being connected or not. I just want to avoid the very narrow corner case where for some reason `Monitor` receives repetitively `AtLeastOneServerConnected` and have the timer reset every time.

be2-scala/src/test/scala/ch/epfl/pop/decentralized/MonitorSuite.scala
K1li4nL: Good point ! The test wasn't very insightful that way...

be2-scala/src/main/scala/ch/epfl/pop/decentralized/Monitor.scala
Jharaxus: You can do the following to use `someServerConnected`:
```scala
connectionMediatorRef = sender()
if (!someServerConnected) {
  timers.startTimerWithFixedDelay(periodicHbKey, TriggerHeartbeat, heartbeatRate)
  someServerConnected = true
}
```

Jharaxus (dismissed):

be2-scala/src/test/scala/ch/epfl/pop/decentralized/MonitorSuite.scala
```diff
@@ -0,0 +1,102 @@
...
+    // Wait for the first heartbeat then "disconnect" servers
+    testProbe.expectMsgType[Monitor.GenerateAndSendHeartbeat](timeout)
+    testProbe.send(monitorRef, Monitor.NoServerConnected)
+    testProbe.expectNoMessage(timeout)
+
+    // No single heartbeat should be scheduled either
+    Source.single(Right(JsonRpcRequestExample.subscribeRpcRequest)).to(Monitor.sink(monitorRef)).run()
+    testProbe.expectNoMessage(timeout)
+
+    // Connect a server and check for heartbeats again
```
Ajkunas, line 98: I do not think that those last lines do really belong to the corresponding test. 
You already tested it in previous tests, and it makes the test a bit confusing. I understand that the purpose was to test corner cases, but maybe writing another test by specifying in the description that you want to test if by connecting and disconnecting the servers, the Monitor still behaves correctly.

Ajkunas (changes requested):
Overall really good ! 
I only have some minor comments that can be discussed. :)

be2-scala/src/main/scala/ch/epfl/pop/decentralized/Monitor.scala
```diff
@@ -0,0 +1,88 @@
...
+        case _: ParamsWithMap => /* Actively ignoring this specific message */
+        // For any other message, we schedule a single heartbeat to reduce messages propagation delay
+        case _ =>
+          if (someServerConnected && !timers.isTimerActive(singleHbKey)) {
+            log.info(s"Scheduling single heartbeat")
+            timers.startSingleTimer(singleHbKey, TriggerHeartbeat, messageDelay)
+          }
+      }
+
+    case _ => /* DO NOTHING */
```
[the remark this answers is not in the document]
K1li4nL: I am not sure I see the point, this last case is mostly filtering out Left(...) messages for which we don't want to trigger heartbeats. Messages received are already pretty well logged, I feel like this won't add much debugging value and will cluter the logs.

be2-scala/src/main/scala/ch/epfl/pop/decentralized/Monitor.scala
```diff
@@ -0,0 +1,88 @@
...
+      timers.cancelAll()
+
+    case Monitor.TriggerHeartbeat =>
+      log.info("triggering a heartbeat")
+      timers.cancel(singleHbKey)
+      heartbeatGenRef ! Monitor.GenerateAndSendHeartbeat(connectionMediatorRef)
+
+    case Right(jsonRpcMessage: JsonRpcRequest) =>
+      jsonRpcMessage.getParams match {
+        case _: ParamsWithMap => /* Actively ignoring this specific message */
```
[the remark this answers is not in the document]
K1li4nL: Again, I am worried about cluttering the logs, the server will potentially receive a lot of these so their logs might grow very fast

be2-scala/src/test/scala/ch/epfl/pop/decentralized/MonitorSuite.scala
K1li4nL: The test goes as follows: 

`monitor should send heartbeats only when servers are connected`

1. Initiate monitor by telling it about connected servers
2. expect a heartbeat
3. disconnect the servers
4. expect no heartbeat
5. connect servers again 
6. expect heartbeats

Excluding these last steps 5 and 6 make the test look incomplete to me (what if monitor has a bug that prevent it to start again after some servers came back). I want the test to guarantee that monitor does its job _whenever and only when servers are connected_.  And it looks redundant to have a second test just doing that last bit.

Do you think the name : 
`monitor should send heartbeats if and only if servers are connected` 

Would make it clearer ?

Ajkunas: Okay I see your point and I agree that the second test would be redundant!
The new name could be a bit clearer in my opinion. :)

Ajkunas (approved):

PR #1501, opened 2023-04-12
Add server and client specific path 

K1li4nL:
This pr introduces change to the path clients use to connect. 
With this pr, the following paths are available: 

`/server`: For servers to connect to
`/client`: For clients, this path replaces `/` we had until now. 

For now, client and servers are treated the same way no matter the path used to connect, this will change in the future.

### DIFF
--- a/be2-scala/README.md
+++ b/be2-scala/README.md
@@ -43,11 +43,12 @@ where the configuration lives. This can be updated if needed.
 # Snapshot of application.config
 
 ch_epfl_pop_Server {
-	http {
-		interface = "127.0.0.1"
-		port = "8000"
-		path = "client"
-	}
+  http {
+      interface = "127.0.0.1"
+      port = "8000"
+      client-path = "client"
+      server-path = "server"
+    }
 }
  ```
 
--- a/be2-scala/README.md
+++ b/be2-scala/README.md
@@ -46,7 +46,7 @@ ch_epfl_pop_Server {
 	http {
 		interface = "127.0.0.1"
 		port = "8000"
-		path = ""
+		path = "client"
 	}
 }
  ```
@@ -93,7 +93,7 @@ environment:
 
 ```sh
 # install sdkman
-curl -s "https://get.sdkman.io" | bash 
+curl -s "https://get.sdkman.io" | bash
 # install java zulu 11
 sdk install java 11.0.18-zulu
 # install sbt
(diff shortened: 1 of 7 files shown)


be2-scala/README.md
```diff
@@ -46,7 +46,7 @@ ch_epfl_pop_Server {
 	http {
 		interface = "127.0.0.1"
 		port = "8000"
-		path = ""
+		path = "client"
```
Jharaxus, line 49: The final update is not completely reflected: it misses `client-path = "client"` and `server-path = "server"`

be2-scala/src/main/scala/ch/epfl/pop/Server.scala
```diff
@@ -46,16 +46,24 @@ object Server {
...
 
-      def publishSubscribeRoute: RequestContext => Future[RouteResult] = path(config.path) {
-        handleWebSocketMessages(PublishSubscribe.buildGraph(pubSubMediatorRef, dbActorRef, messageRegistry)(system))
+      // Setup routes
+      def publishSubscribeRoute: RequestContext => Future[RouteResult] = {
+        path(config.clientPath) {
+          handleWebSocketMessages(PublishSubscribe.buildGraph(pubSubMediatorRef, dbActorRef, messageRegistry)(system))
+        } ~ path(config.serverPath) {
+          handleWebSocketMessages(PublishSubscribe.buildGraph(pubSubMediatorRef, dbActorRef, messageRegistry)(system))
+        }
```
Jharaxus, line 55: As both routes are the same they can be concatenated using `(path(config.clientPath) | path(config.serverPath)) {handleWebSocketMessages(...)}`

Jharaxus (changes requested):
Two small nitpicks, otherwise good job!

be2-scala/src/main/scala/ch/epfl/pop/Server.scala
K1li4nL: Thats only temporary, the next steps are to change `buildGraph` so it behaves slightly differently based on the path used. More precisely: When `/server` is used, the clientActorRef will be sent to the `ConnectionMediator` (pr #1503 )
And the clientActor will also tell `ConnectionMediator` when it disconnects. It took me some time to figure out the `~`, I figured this way already have our template ready.

be2-scala/README.md
K1li4nL: Good catch, done !

be2-scala/src/main/scala/ch/epfl/pop/Server.scala
Jharaxus: Ok that makes sense !

Jharaxus (approved):

Ajkunas (approved):
LGTM !

Issue #1478, opened 2023-03-28
[BUG] Rescanning old participants upon reopening creates duplicates


matteosz:
### Description (Actual behavior)
After the reopening of a roll call, if the camera points and scans an already scanned attendee, he will be displayed twice. Additionally, trying to close the roll call after this operation leads to an error: 

![Screenshot (31)](https://user-images.githubusercontent.com/83547952/228370842-25680575-b651-42a7-81d4-1f87a0fbe0ac.png)

### Expected behavior
The list of attendees shouldn't have duplicates and roll call has to be closed properly.

### How to reproduce
1. Create a roll call
2. Scan a token
3. Close the roll call
4. Re-open the roll call
5. Re-scan the token

### Version & Environment
This bug was reproduced on:
- Commit: 84f24960ac1162dbc682ca0f06ebce02b54a3fde

##### Front-ends:
- [X] Fe1-Web (Chrome)

PR #1502, opened 2023-04-12
Fix bug duplicates list of attendees when re-opening roll call

dayan9265:
This PR fixes the issue #1478, where there could be duplicate attendees shown on screen. The bug simply comes from the fact that there was no check if there was duplicates when creating the new list of `attendeePopTokens`. There is also two added tests that fails with the previous implementation. (The code to ensure that there is no duplicates is not really "clean", so any idea to make it cleaner would be welcomed :) )

### DIFF
--- a/fe1-web/src/features/rollCall/components/AttendeeList.tsx
+++ b/fe1-web/src/features/rollCall/components/AttendeeList.tsx
@@ -6,6 +6,7 @@ import { View } from 'react-native';
 import { PoPIcon } from 'core/components';
 import { PublicKey } from 'core/objects';
 import { Color, Icon, List, Typography } from 'core/styles';
+import STRINGS from 'resources/strings';
 
 const AttendeeList = ({ popTokens }: IPropTypes) => {
   return (
@@ -15,7 +16,7 @@ const AttendeeList = ({ popTokens }: IPropTypes) => {
         content={
           <ListItem.Content>
             <ListItem.Title style={[Typography.base, Typography.important]}>
-              Attendees
+              {STRINGS.roll_call_attendees}
             </ListItem.Title>
           </ListItem.Content>
         }
--- a/fe1-web/src/features/rollCall/screens/__tests__/ViewSingleRollCall.test.tsx
+++ b/fe1-web/src/features/rollCall/screens/__tests__/ViewSingleRollCall.test.tsx
@@ -87,25 +87,32 @@ beforeEach(() => {
 
 describe('EventRollCall', () => {
   describe('render correctly', () => {
-    const testRender = (rollCall: RollCall, isOrganizer: boolean) => () => {
-      mockStore.dispatch(updateRollCall(rollCall.toState()));
+    const testRender =
+      (
+        rollCall: RollCall,
+        isOrganizer: boolean,
+        attendeePopTokens: String[] | undefined = undefined,
+      ) =>
+      () => {
+        mockStore.dispatch(updateRollCall(rollCall.toState()));
 
-      const obj = render(
-        <Provider store={mockStore}>
-          <FeatureContext.Provider value={contextValue}>
-            <MockNavigator
-              component={ViewSingleRollCall}
-              params={{
-                eventId: rollCall.id.valueOf(),
-                isOrganizer,
-              }}
-            />
-          </FeatureContext.Provider>
-        </Provider>,
-      );
+        const obj = render(
+          <Provider store={mockStore}>
+            <FeatureContext.Provider value={contextValue}>
+              <MockNavigator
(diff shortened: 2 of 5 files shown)


MeKHell (changes requested):
The code needs to be simplified at one point. Otherwise it seems to be good.

fe1-web/src/features/rollCall/components/AttendeeList.tsx
```diff
@@ -15,7 +16,7 @@ const AttendeeList = ({ popTokens }: IPropTypes) => {
         content={
           <ListItem.Content>
             <ListItem.Title style={[Typography.base, Typography.important]}>
-              Attendees
+              {STRINGS.roll_call_attendees}
```
Tyratox, line 19: Nice catch!

Tyratox (commented):
Hmm honestly while this probably solves the problem I feel like it is not the best solution. In theory https://github.com/dedis/popstellar/blob/master/fe1-web/src/features/rollCall/screens/RollCallScanner.tsx#L114 should already check for duplicates and remove them.. Did you find you why this does not happen? Is some navigation parameter not passed correctly?

It might be because the navigation parameter is still set while `rollCall.attendees` is now filled too resulting in doubling the list?

https://github.com/dedis/popstellar/blob/master/fe1-web/src/features/rollCall/screens/ViewSingleRollCall.tsx#L46-L47

dayan9265:
> It might be because the navigation parameter is still set while `rollCall.attendees` is now filled too resulting in doubling the list?

Yes that was the issue, both `attendeePopTokens` and `rollCall?.attendees` contained the same pop tokens

>But I would have assumed that
https://github.com/dedis/popstellar/blob/master/fe1-web/src/features/rollCall/components/RollCallOpen.tsx#L93
already clears it 🤔 Or did you find something else?

I realize that my title is wrong ^^', the duplication happens when we click on `Scan PoPTokens` multiple times (and not when closing the roll call, so this code is not called). 

Maybe the problems come from here?
https://github.com/dedis/popstellar/blob/master/fe1-web/src/features/rollCall/screens/RollCallScanner.tsx#L223
Since it gives all the scanned tokens including the previously scanned token, is it better to modify this part to make it so it only give as a parameter the newly scanned tokens?

Tyratox:
Oh I see so what happens is
1) you scan some tokens `x` and `y`
2) go back to token list screen, this passes `x` and `y` as navigation parameters 
3) close roll call, stores `x` and `y` in `rollCall.attendees`
4) re-open roll call
5) scan new tokens. here it should not be possible to scan the duplicate tokens since `rollCall.attendees` is passed to the scanner and due to https://github.com/dedis/popstellar/blob/master/fe1-web/src/features/rollCall/screens/RollCallScanner.tsx#L114
6) when going back to the token list screen `rollCall.attendees` and `attendeePopTokensStrings` are the exact same set.

Now if I am not mistaken, we should be fine if we always used `attendeePopTokensStrings` if it is non-empty and `rollCall.attendees` otherwise? This works under the assumption that `attendeePopTokensStrings` are always more up-to-date if they exist and I think that should hold? The only time `attendeePopTokensStrings` does not contain what we want is when an existing roll call is looked at / re-opened. But in this case, it should be empty and if we scan more tokens it will be a superset?

i.e. something like

```typescript
const attendeePopTokens = useMemo(() =>
    (attendeePopTokensStrings ?
        attendeePopTokensStrings.map((k) => new PublicKey(k) : 
        rollCall?.attendees
    ) || []
 , [attendeePopTokensStrings, rollCall?.attendees]);
```

Or am I missing something?

MeKHell (dismissed):
LGTM !

fe1-web/src/features/rollCall/screens/__tests__/ViewSingleRollCall.test.tsx
```diff
@@ -120,5 +127,14 @@ describe('EventRollCall', () => {
...
     });
+
+    // test to verify that if same attendee present multiple times, only one is displayed (here: attendee1)
+    describe('no duplicate attendees', () => {
+      it('opened roll calls', testRender(mockRollCallOpened, false, ['attendee1', 'attendee2']));
+      it(
+        're-opened roll calls',
+        testRender(mockRollCallReopened, false, ['attendee1', 'attendee2']),
+      );
+    });
```
Tyratox, line 138: Not sure if these tests still accomplish what the comment describes?

Tyratox:
Otherwise LGTM :)

Tyratox (approved):

Tyratox:
Sorry I was dumb, it makes after the discussion in the meeting :)

Issue #1451, opened 2023-03-14
[BUG] Message from others LAO are processed


MeKHell:
### Description (Actual behavior)
When a LAO is created, the FE1 process some messages from other LAOs that should not be processed at this stage.
![image](https://user-images.githubusercontent.com/42808302/224757648-4a6c21b7-2913-40b8-9749-70e36df3a265.png)

### Expected behavior
When a catch-up call is sent, only the messages of the current LAO should be processed.

### How to reproduce
1. Open or create a LAO on a server (LAO1)
2. Disconnect from LAO1
3. Open or create a second LAO on the same server (LAO2)
4. Create some traffic on LAO2 by completing a Roll Call and an election.
5. Disconnect from LAO2
6. Connect in LAO1
7. You will have warning in your console as on the image above

### Version & Environment
This bug was reproduced on:
- Commit: bc23763f

##### Front-ends:
- [x] Fe1-Web (Firefox 110.0.1  & Opera 96.0.4693.31)
- [ ] Fe2-Android (please specify if phone or emulation, and Android version)
- [ ] Not applicable

##### Back-ends:
- [ ] Be1-Go
- [ ] Be2-Scala
- [x] Not applicable

PR #1504, opened 2023-04-13
No more Warnings from wrong processing of messages

MeKHell:
This PR mainly addresses #1451.
The problem was that the message were sent to be witnessed even if they were not from the LAO was connected in. The fix will now keep the message until the correct LAO is opened.

In addition, I fixed the scanning problem (at least what I could test on my own) : It was very hard (sometimes impossible) to scan some pop tokens during roll calls.
This is fixed by reducing the space used by the circle in the middle. The computation provided in the comment should have been sufficient, but maybe did not take into account the mouse or the square that were partially hidden...


Tyratox (approved):
LGTM! Great work adding the regression test!

PR #1505, opened 2023-04-16
Added description and location to the RollCall UI

matteosz:
This PR includes in the `RollCallFragment` the description and the location of the rollcall itself, in order to be consistent with the FE1.

Changes:
- Inserted the description and location in the roll call UI, using an expandible text
- Changed the description to be serializable (removed the `transient` modifier)
- Fixed a subtle bug in the `RollCallBuilder` that was mixing location and description

New UI behaviour:
- Such metadata is hidden when the user is an attendee and the roll call is open to leave space for the QR
- The description is displayed only if not-empty

Screenshots:


fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/event/rollcall/RollCallFragment.java
```diff
@@ -357,6 +380,24 @@ private void retrieveAndDisplayPublicKey() {
     binding.rollCallPkQrCode.setImageBitmap(myBitmap);
   }
 
+  /** Callback function for the card listener to expand and shrink a text box */
+  private void handleExpandArrow(android.widget.ImageView arrow, android.widget.TextView text) {
+    float newRotation;
+    int visibility;
+    // If the arrow is pointing up, then rotate down and make visible the text
+    if (arrow.getRotation() == 0f) {
+      newRotation = 180f;
```
simone-kalbermatter, line 389: It would be better to define the 100f and 0f as constants like `UP` and `DOWN`

simone-kalbermatter (changes requested):
Very nice work, well documented and tested. Just a small comment. In general, is there a reason for using the expandable text? For example, location should normally not a be a long text.

matteosz:
> Very nice work, well documented and tested. Just a small comment. In general, is there a reason for using the expandable text? For example, location should normally not a be a long text.

Yeah, I know, but there's no actual limit on the amount of characters, so in general it can be arbitrarly long. But mainly I like this kind of approach where the UI is more customizable and the user can decide what to show 😄

simone-kalbermatter (approved):
LGTM!

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/event/rollcall/RollCallFragment.java
```diff
@@ -357,6 +379,24 @@ private void retrieveAndDisplayPublicKey() {
     binding.rollCallPkQrCode.setImageBitmap(myBitmap);
   }
 
+  /** Callback function for the card listener to expand and shrink a text box */
+  private void handleExpandArrow(android.widget.ImageView arrow, android.widget.TextView text) {
```
Warhedgehog, line 383: Why is the complete type needed ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/Constants.java
```diff
@@ -56,4 +56,7 @@ public class Constants {
 
   /** Number of milliseconds in a day */
   public static final long MS_IN_A_DAY = 1000 * 60 * 60 * 24L;
+
+  /** Constants to rotate up or down along the x axis */
+  public static final float ORIENTATION_UP = 0f, ORIENTATION_DOWN = 180f;
```
Warhedgehog, line 61: Bit of nitpicking but I am not a fan of multiple variables declaration in one line. It's not great for readability IMO and is not something you'll see a lot in this codebase.

fe2-android/app/src/main/res/layout/roll_call_fragment.xml
```diff
@@ -133,25 +133,131 @@
...
+        <androidx.constraintlayout.widget.ConstraintLayout
+          android:layout_width="match_parent"
+          android:layout_height="wrap_content"
+          android:paddingVertical="@dimen/padding_standard">
+
+          <ImageView
+            android:id="@+id/roll_call_location_arrow"
+            android:layout_width="@dimen/roll_call_arrow_side"
+            android:layout_height="@dimen/roll_call_arrow_side"
+            android:layout_marginEnd="@dimen/roll_call_card_lateral_margin"
```
Warhedgehog, line 161: Would it make sense to reuse some margin values of other displayed expand arrows ? I'm thinking about the ones either in the `EventListFragment` or the ones in the `DigitalCashHistoryFragment` (Though those are defined in `history_transaction_layout.xml`)
The idea is to keep as coherent as a possible a UI throughout the app (though you've probably noticed that it is very much a work in progress on this app 😄)

Warhedgehog (changes requested):
Great work, well documented and tested. Since the quality is already high, I've allowed myself to be picky with my comments :)

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/Constants.java
matteosz: Sure I'll split them!

fe2-android/app/src/main/res/layout/roll_call_fragment.xml
matteosz: Yeah it would be probably much better. I'll unify those values 👍🏼

Warhedgehog (approved):
LGTM

PR #1506, opened 2023-04-17
Handle incoming message reprocessing for unrecoverable failures

### LINKED ISSUE #1376: Unhandled messages are reprocessed infinitely
Currently, to avoid any issue created by message reordering, any unhandled message is placed in a queue to be reprocessed 5 seconds later.

This allows the following scenario to function correctly :
The `create lao` and `create roll call` messages were reordered, we receive `create roll call` first.

1. The processing of  `create roll call` fails, as the related LAO does not exist yet. It is placed in the queue to be reprocessed later.
2. The LAO is correctly created
3. The roll call is created properly now that the LAO exists.

The issue is : currently, there is no distinction between a recoverable failure and an unrecoverable one. So any failure will be reprocessed.

Also, an attacker could easily craft messages that will always fail and fill the user's resources by sending them on the network.

matteosz:
This PR resolves #1376 .

It uses a counter of reprocessing for each message failed to handle, such that messages with higher attempts than a threshold are discarded.


simone-kalbermatter (approved):
LGTM!

Warhedgehog (approved):
Great Work ! A suggestion. Define the reprocessing delay as a variable (the same as  `MAX_REPROCESSING`) and adapt `LAONetworkManagerTest` accordingly

Warhedgehog (approved):
LGTM

PR #1507, opened 2023-04-19
Fix starting server documentation

MariemBaccari:
This PR is a small fix to the starting server documentation by simply replacing `./pop organizer` (which is not used anymore) by `./pop server`.


Anghul (approved):
LGTM

PR #1510, opened 2023-04-20
Karate tests for decentralized communication

simone-kalbermatter:
This PR adds karate tests for the following messages:
- updateLao: there were no tests for this message so far and I added some to get used to how the karate tests work
- heartbeat: test that heartbeat messages are received, but so far no checks on the content
- getMessagesById: test that getMessagesById is sent in response to a hearbeat message with missing message ids, but so far no check on the content

These tests are still very basic and just a first step to check that the basic functionality and the testing system works. For now I am using the mock frontend to send and receive heartbeats as this seems enough for the current tests, but it might be changed in the future.


tests/karate/src/test/java/karate-config.js
```diff
@@ -17,14 +17,14 @@ function fn() {
   if (env === 'go') {
     // customize
     config.host = '127.0.0.1';
-    config.port = 9000;
-    config.path = 'organizer/client';
+    config.port = 9001;
```
matteosz, line 20: Is it the Go server port 9001? Because I usually connect on 9000

matteosz (approved):
LGTM! Just a question

tests/karate/src/test/java/karate-config.js
simone-kalbermatter: I changed this because Mariem said to receive the hearbeat messages you need to be connected to the server port. For the Scala be however, Kilian said it's enough to change /client to /server for this. We can test if this work and maybe get it more coherent between be1 and be2.

matteosz: Clear, thank you!

tests/karate/src/test/java/be/LAO/update.feature
```diff
@@ -0,0 +1,66 @@
+@env=go,scala
+Feature: Update a LAO
+
+  Background:
+    # This is feature will be called to test Update LAO messages
```
GabrielFleischer, line 5: Typo ? 😄

GabrielFleischer, line 5: Typo ?

tests/karate/src/test/java/be/heartbeat/heartbeat.feature
```diff
@@ -0,0 +1,39 @@
...
+    # This also sets up a valid lao by sending a create lao message, subscribing to the lao and sending a catchup message
+    * call read('classpath:be/utils/server.feature')
+    * call read('classpath:be/mockFrontEnd.feature')
+    * call read('classpath:be/constants.feature')
+    * call read('classpath:be/utils/simpleScenarios.feature@name=valid_lao')
+    * string channel = "/root/p_EYbHyMv6sopI5QhEXBf40MO_eNoq7V_LygBd4c9RA="
+
+    # After lao creation, wait (10 seconds for now) and check that a heartbeat message was received
+  Scenario: Server should send heartbeat messages
+    Given wait(10)
```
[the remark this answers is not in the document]
simone-kalbermatter: I reduced it a bit to 6 seconds, however right now the hearbeat messages are sent every 4 seconds, so I can't really go lower than this. The intervall might be still reduced in the future if we find that we need more heartbeats. I see the issue, but I don't think there will be so many tests that it causes problems.

tests/karate/src/test/java/common/net/MultiMsgWebSocketClient.java
```diff
@@ -115,6 +116,25 @@ public String getBackendResponseWithElectionResults(String data){
     throw new IllegalArgumentException("No election results where received");
   }
 
+  /**
+   * Returns all messages received that are of a given type of method
+   * @param method the type of method
+   * @return list of all received messages of with that method
```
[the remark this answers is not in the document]
simone-kalbermatter: Changed the comment to make it more concise.

simone-kalbermatter:
@Warhedgehog about renaming the frontend, I saw that it was a pretty big change, so I left it like this for the moment. If things work well this way, I will rename it later.

Warhedgehog (approved):
LGTM

PR #1516, opened 2023-04-23
Auth server Websocket communication

Nazianzenov:
- websocket server on endpoint /response
- websocket client on webpage, written in JS
- added test to check errors, with a fake client
- changed html directory "qrcode" to "resources"


be1-go/popcha/resources/popcha.html
```diff
@@ -0,0 +1,88 @@
...
+        socket.onopen = (_) => {
+            console.log("Starting webSocket connection with address: "+ "{{.WebSocketAddr}}");
+            console.log("redirect URI: " + "{{.RedirectHost}}");
+        }
+        // when receiving the parameters, construct the redirect URI
+        socket.onmessage = (event) => {
+            let redir = "{{.RedirectHost}}";
+            if (redir.startsWith("localhost")) {
+                redir = "http://" + redir
+            }
```
MeKHell, line 17: It might  be good to remove this later (since https is part of our threat model's assumptions), but it is ok to have it for now.

fe1-web/jest/setup.js
```diff
@@ -13,7 +13,7 @@ jest.mock('@rneui/base/dist/Icon');
 jest.mock('react-blockies', () => 'blockies');
 
 // qr codes are svgs and thus generate huge snapshot files
-jest.mock('core/components/QRCode.tsx', () => 'qrcode');
+jest.mock('core/components/QRCode.tsx', () => 'resources');
```
MeKHell, line 16: I do not know why this is changed in this PR, but it should stay the same

```suggestion
jest.mock('core/components/QRCode.tsx', () => 'qrcode');
```

MeKHell (changes requested):
Only reviewed the HTML part (and at the same time the fe1 part since I am working on fe1).

fe1-web/jest/setup.js
Nazianzenov: i changed it as i wanted to refactor the directory name from qrcode to resources. I think the name is more descriptive that way. It might have affected other locations in the fe1 code though

MeKHell: The change has not to do with the QR Codes location at all, it only changes how the QR codes are rendered during jest tests.
Instead of computing and storing the whole QR code in the snapshots, this small text is used as a placeholder to say "Hey there is some QR Code here". This change just changes the placeholder, thus leading to a fail of all screens that contains QR Codes.

It is indeed more descriptive, but particularly useless since it is a part only used by testing units. If this change should still be committed, then the snapshots should be updated too.

be1-go/popcha/resources/popcha.html
MeKHell: Or mkcert might be a solution: [](https://github.com/FiloSottile/mkcert/releases)

Nazianzenov: I will look into it, thanks

fe1-web/jest/setup.js
Nazianzenov: Ohh, no problem then i think my IDE did a wrong refactoring. I will make sure to revert that.

be1-go/popcha/server.go
```diff
@@ -328,6 +344,56 @@ func (as *AuthorizationServer) generateQRCode(w http.ResponseWriter, req *http.R
 	return nil
 }
 
+// helper method used for handling redirect uri by the Webpage websocket. Requests on the /response endpoint are
+// solely. handled through websocket. Because of the protocol, the first websocket to send a request will be the JS
```
MariemBaccari, line 348: Small typo here "solely. handled"

be1-go/popcha/server.go
```diff
@@ -328,6 +344,56 @@ func (as *AuthorizationServer) generateQRCode(w http.ResponseWriter, req *http.R
...
+// one.
+func (as *AuthorizationServer) responseEndpoint(w http.ResponseWriter, r *http.Request) {
```
MariemBaccari, line 350: For someone who's not familiar with the protocol, it's a bit hard to understand what the method does exactly as it has many nested conditions etc ... Writing one or two helper methods for it (for example one for the first if/else and another one for the rest) with descriptive names and documentations can make it more readable.

MariemBaccari (changes requested):
The go part looks mostly clean to me except one readability comment that should be adressed before I approve.

MeKHell (approved):
Since fe1 is no more modified and the JavaScript part looks good, I give my approval for the parts that I am concerned with. This is not an approval for the Go part.

MariemBaccari (dismissed):
Looks good to me now !

be1-go/popcha/server_test.go
```diff
@@ -144,13 +163,43 @@ func TestAuthRequestFails(t *testing.T) {
...
 
+	lastEntry = logTester.LastEntry()
+	lastEntry.ExpLevel(zerolog.ErrorLevel)
+	lastEntry.ExpMsg("Error while verifying the parameters of the request")
 	bodyBytes = helperValidateGetAndParseBody(t, res)
-
 	// we require that there are 5 missing arguments
 	helperMissingArgs(t, bodyBytes, 5)
 
+	time.Sleep(200 * time.Millisecond)
```
jbsv, line 173: isn't there another way to wait ? sleep produces long tests. Plus all the non deterministic problems.

be1-go/popcha/server_test.go
```diff
@@ -168,14 +219,180 @@ func helperMissingArgs(t *testing.T, body []byte, n int) {
...
+	resType := ResTypeMulti
+
+	// create the URL of the PopCHA webpage
+	u := createAuthRequestURL(nonce, clientID, scope, laoID, redirectURI, resType, state)
+
+	_, err = http.Get(u)
+	require.NoError(t, err)
+	l.Info().Msg(u)
+
+	time.Sleep(2 * time.Second)
```
jbsv, line 253: avoid or keep it as small as possible

jbsv (commented):
I know it is sometimes hard to avoid without a huge rework, but try to not use sleep.

be1-go/popcha/server_test.go
Nazianzenov: I can definitely remove most of those.

Anghul (approved):
LGTM

PR #1519, opened 2023-04-25
Add server peers list

K1li4nL:
This pr adds a list of peers in the scala-backend config file in order for tell the server where to connect for decentralized communications. 

To safely test config files, a new startup flag was added: -Dtest
This argument allows RuntimeEnvironment to work with `server-peers-list-TEST.conf` instead of the `server-peers-list.conf`, this will prevent the tests from completely breaking the original file while being able to test the api as-is.

### DIFF
--- a/be2-scala/src/main/scala/ch/epfl/pop/config/RuntimeEnvironment.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/config/RuntimeEnvironment.scala
@@ -41,15 +41,7 @@ object RuntimeEnvironment {
   }
 
   private def testMode: Boolean = {
-    val testParam = sp("test")
-    if (testParam != null && testParam.trim.nonEmpty) {
-      testParam.toLowerCase match {
-        case "true" => true
-        case _      => false
-      }
-    } else {
-      false
-    }
+    sp("test") != null
   }
 
   private lazy val confDir: String = getConfDir
--- a/be2-scala/src/main/scala/ch/epfl/pop/config/RuntimeEnvironment.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/config/RuntimeEnvironment.scala
@@ -33,7 +33,7 @@ object RuntimeEnvironment {
 
     val virtualMachineParam = "scala.config"
     val pathConfig = sp(virtualMachineParam)
-    if (pathConfig != null && !pathConfig.trim.isEmpty) {
+    if (pathConfig != null && pathConfig.trim.nonEmpty) {
       pathConfig.trim
     } else {
       throw new RuntimeException(s"-D$virtualMachineParam was not provided.")
--- a/be2-scala/src/main/scala/ch/epfl/pop/config/RuntimeEnvironment.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/config/RuntimeEnvironment.scala
@@ -48,7 +48,7 @@ object RuntimeEnvironment {
   private lazy val appConfFile = confDir + File.separator + "application.conf"
 
   lazy val isTestMode: Boolean = testMode
-  lazy val serverPeersList: String =
+  lazy val serverPeersListPath: String =
     if (isTestMode) {
       confDir + File.separator + "server-peers-list-TEST.conf"
     } else {
@@ -63,7 +63,7 @@ object RuntimeEnvironment {
   def readServerPeers(): List[String] = {
     val source =
       try {
-        fromFile(serverPeersList)
+        fromFile(serverPeersListPath)
       } catch {
         case ex: Throwable =>
           println("RuntimeEnvironment: " + ex.toString)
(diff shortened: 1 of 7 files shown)


be2-scala/src/test/scala/ch/epfl/pop/config/RuntimeEnvironmentSuite.scala
```diff
@@ -0,0 +1,27 @@
...
+      "ws://127.0.0.1:9000/client"
+    )
+
+    testWriteToServerPeersConfig(addressList)
+
+    // Verify it works as expected
+    readServerPeers() should equal(addressList.tail)
+
+    // Delete the test file
+    deleteTestConfig()
```
Jharaxus, line 25: When left after the test assertion, the config file will not be deleted if the test fails

be2-scala/src/test/scala/ch/epfl/pop/config/RuntimeEnvironmentSuite.scala
```diff
@@ -0,0 +1,27 @@
...
+
+class RuntimeEnvironmentSuite extends FunSuite {
+  test("readServerPeers() should only return the addresses that match the regex from the schema") {
+
+    // Values to add in the file
+    val addressList = List(
+      "http://127.0.0.1/",
+      "ws://127.0.0.1:7000/client",
+      "ws://127.0.0.1:8000/client",
+      "ws://127.0.0.1:9000/client"
```
Jharaxus, line 16: nitpick: you could also add a `wss`-like address in the check

be2-scala/src/main/scala/ch/epfl/pop/json/MessageDataProtocol.scala
```diff
@@ -153,20 +153,22 @@ object MessageDataProtocol extends DefaultJsonProtocol {
             lao.convertTo[Hash],
             frontend.convertTo[PublicKey],
             address,
-            peers.map(_.convertTo[String]).toList
+            peers.map(jsValue => jsValue.asJsObject.getFields(PARAM_ADDRESS).map(_.convertTo[String])).toList.flatten
```
Jharaxus, line 156: You could have a slightly more meaningful error if you explicitly checked that each jsValue is a string (JsString) as expected :
```scala
peers.map(jsValue => jsValue.asJsObject.getFields(PARAM_ADDRESS) match {
  case JsString(address) => address
  case _ => throw new IllegalArgumentException(s"Can't parse json value $jsValue to get an address")
}).toList       
```

be2-scala/src/test/scala/ch/epfl/pop/config/RuntimeEnvironmentTestingHelper.scala
```diff
@@ -0,0 +1,29 @@
+package ch.epfl.pop.config
+
+import java.io.{BufferedWriter, File, FileWriter}
+
+object RuntimeEnvironmentTestingHelper {
+
+  private def ensureTestingMode(): Unit = {
+    if (!RuntimeEnvironment.isTestMode) {
+      throw new ExceptionInInitializerError("Argument \"-Dtest\" missing")
```
Jharaxus, line 9: nitpick: you could add in the error message that it is mandatory during testing

Jharaxus (changes requested):
Good job it looks very neat, especially the `serverPeersListPath` usage to separate tests from live usages !

be2-scala/src/main/scala/ch/epfl/pop/config/RuntimeEnvironment.scala
```diff
@@ -33,15 +33,47 @@ object RuntimeEnvironment {
...
+    if (isTestMode) {
+      confDir + File.separator + "server-peers-list-TEST.conf"
+    } else {
+      confDir + File.separator + "server-peers-list.conf"
+    }
 
   lazy val appConf: Config = ConfigFactory.parseFile(new File(appConfFile))
 
+  // Regex from dataGreetLao.json
+  private val addressPattern = "^(ws|wss)://.*(:d{0,5})?/.*$"
```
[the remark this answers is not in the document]
K1li4nL: The greetLao regex was changed in #1554, maybe we should put it in public in `RuntimEnvironment` so that it is  consistent across the project and easy to update ?

be2-scala/src/main/scala/ch/epfl/pop/json/MessageDataProtocol.scala
K1li4nL: Good point ! Fixed.

be2-scala/src/main/scala/ch/epfl/pop/config/RuntimeEnvironment.scala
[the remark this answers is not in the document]
K1li4nL: Update: I say for now we leave this regex alone, I think a pr need to be made to fix greet_lao and greet_server regex as currently every address that do not contain `/` after the optional port are rejected....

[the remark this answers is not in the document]
Jharaxus: Ok let's fix this at once with the final regex version!

Jharaxus (approved):

Ajkunas (approved):
Nicely done :)

PR #1521, opened 2023-04-25
Second part of the PoPCHA example Client

MeKHell:
This PR will add the second part of the example client of the PoPCHA.
It adds the authentication and the proper app part.

The app currently does not verify the signature of the provided tokens.

All the computation is done on the server side, to simplify the app as much as possible.

The `run.sh` script sets up the server's virtual environment and launches it.

`run.sh` optionally takes two arguments (the public server address and the public server port). If **both** arguments are passed, then the script overwrites the config with the provided arguments (this also overwrites the client id).

Here is a screenshot of the app:
image


Nazianzenov (approved):
LGTM

demo/popcha-client/web/sonar-project.properties
```diff
@@ -0,0 +1,19 @@
...
+sonar.projectKey=dedis_popstellar_popchawebclient
+
+sonar.python.version=3
+sonar.dynamicAnalysis=reuseReports
+sonar.core.codeCoveragePlugin=cobertura
+sonar.python.coverage.reportPaths=*coverage*.xml
+
+# Path of the test and coverage reports
+sonar.go.tests.reportPaths=./report.json
+sonar.go.coverage.reportPaths=./coverage.xml
```
pierluca, line 11: you probably don't want to refer to `Go` tests and coverage.
Check the docs for the correct params
https://docs.sonarqube.org/9.9/analyzing-source-code/test-coverage/python-test-coverage/

demo/popcha-client/web/sonar-project.properties
```diff
@@ -0,0 +1,19 @@
+sonar.organization=dedis
+sonar.projectKey=dedis_popstellar_popchawebclient
+
+sonar.python.version=3
+sonar.dynamicAnalysis=reuseReports
+sonar.core.codeCoveragePlugin=cobertura
+sonar.python.coverage.reportPaths=*coverage*.xml
```
pierluca, line 7: this looks like it should probably be below :)

demo/popcha-client/web/sonar-project.properties
```diff
@@ -0,0 +1,19 @@
+sonar.organization=dedis
+sonar.projectKey=dedis_popstellar_popchawebclient
+
+sonar.python.version=3
```
pierluca, line 4: You might want to put this along with other python-specific params, to have a clean config file :)

demo/popcha-client/web/src/main.py
```diff
@@ -0,0 +1,222 @@
...
+            config_file.seek(0)
+            config_file.write(json.dumps(config))
+
+
+app = Flask("Example_authentication_server",
+            template_folder = path.join(path.dirname(__file__), "templates")
+            )
+csrf = CSRFProtect()
+csrf.init_app(app)
+on_startup()
```
pierluca, line 117: why not move this to the `run()` function ?

pierluca (commented):
Just a few minor comments :-)

.github/workflows/deploy.yaml
```diff
@@ -231,3 +231,26 @@ jobs:
...
+    env:
+      base_folder: demo/popcha-client/web    
+
+    steps:
+      - uses: actions/checkout@v3
+
+      # Use rsync to deploy the new website (no build is necessary)
+      - name: Sync
+        env:
+            dest: "popdemo@${{secrets.POPDEMO_DEPLOY_ADDR}}:demo/popcha-client/web/"
```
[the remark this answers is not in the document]
MeKHell: Ok, I will batch it in the next commit

.github/workflows/ci.yaml
```diff
@@ -283,6 +283,49 @@ jobs:
...
+      - name: Setup python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.9'
+
+      - name: Setup pip dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install ruff pytest pytest-cov freezegun
+          pip install -r requirements.txt
```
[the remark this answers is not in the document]
MeKHell: I tried to use it first but ran into some trouble, I will give it a look again.

demo/popcha-client/web/run.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+echo "Creating Virtual environment"
+python -m venv venv;
+echo "Entering Virtual environment"
+source venv/bin/activate;
+echo "Installing dependencies"
+pip install -r requirements.txt;
+if [ $# -eq 2 ] ; then
+    echo "Creating config file (This will reset your client id)"
+    printf "{\"client_id\":\"\", \"host_url\": \"$1\", \"host_port\": $2}" > data/config.json
```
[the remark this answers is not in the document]
MeKHell: If it is empty, it will be generated on launch by the app. Since the configuration would not change, the first time `run.sh` should be called with two arguments  (or the `config.json` should be edited).

.github/workflows/ci.yaml
```diff
@@ -283,6 +283,49 @@ jobs:
...
+
+      - name: Lint with ruff
```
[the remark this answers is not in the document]
Tyratox: Me neither 😁

demo/popcha-client/web/src/main.py
```diff
@@ -0,0 +1,222 @@
...
+    authenticationProvider = Authentication()
+    providers = load_providers()
+    authenticationProvider.providers = filter_providers(providers)
+    config_path = path.normpath(
+        path.join(
+            path.dirname(__file__),
+            "../data/config.json"
+            )
+        )
+    with open(config_path, "r+") as config_file:
```
Tyratox, line 110: What if the file doesn't exit?

demo/popcha-client/web/src/main.py
```diff
@@ -0,0 +1,222 @@
...
+# Define the global variables
+config: dict = {}
+core_app: CounterApp
+authenticationProvider: Authentication
+app = Flask("Example_authentication_server",
+            template_folder = path.join(path.dirname(__file__), "templates")
+            )
+
+
+def check_config(config_file: IO) -> bool:
```
Tyratox, line 26: I would try to limit the number of places where you access global variables. It makes functions easier to understand. Here for instance you could accept a dictionary object, modify it and return the modified dictionary object. The caller then updated the global variable / write the data back to disk.

Also I think it would not hurt logging some message saying "Hey I updated your config file"

demo/popcha-client/web/src/main.py
```diff
@@ -0,0 +1,222 @@
...
+    no connectivity tests.
+    :param provider: The provider to check
+    :return: True if the provider is valid
+    """
+    valid_lao_id = (("lao_id" in provider) and (len(provider["lao_id"]) == 44)
+                    and provider["lao_id"].endswith("="))
+    valid_domain = (("domain" in provider) and ('/' not in provider["domain"])
+                    and not provider["domain"].startswith("http"))
+    valid_public_key = len(provider["public_key"]) > 0
+    return valid_lao_id and valid_domain and valid_public_key
```
Tyratox, line 63: The formatting here is not very readable 😅 Also for some "magic" numbers (44) it would help adding a short comment explaining what the check does.

demo/popcha-client/web/src/main.py
```diff
@@ -0,0 +1,222 @@
...
+
+
+# Step1: Returns the homepage
+@app.get("/")
+def root() -> str:
+    """
+    Get the homepage.
+    :return: The homepage HTML
+    """
+    providers_string = [f'{provider.get("lao_id")}@{provider.get("domain")}'
```
Tyratox, line 127:
```suggestion
    provider_strings = [f'{provider.get("lao_id")}@{provider.get("domain")}'
```

Might look a detail but has a different meaning. Is it one string or multiple :)

demo/popcha-client/web/src/main.py
```diff
@@ -0,0 +1,222 @@
...
+
+
+# Step2: Process user connection data prepare the OIDC request
+@app.route("/authenticate")
+def authentication() -> Response:
+    """
+    Redirect the user to the PoPCHA based authentication server
+    :return: A response which includes a redirect to the original website
+    """
+    provider_id: int = request.args.get("serverAndLaoId", -1, type = int)
```
Tyratox, line 146: What is the -1 doing here? A default value? Could you also use a named parameter?

demo/popcha-client/web/src/main.py
```diff
@@ -0,0 +1,222 @@
...
+def authentication_callback() -> Response:
+    """
+    Redirects the user after the authentication callback has been verified
+    :return: A response which redirects the user to the homePage if the login
+    is not valid or to a new "app" page if the login answer is valid
+    """
+    user_id: str = authenticationProvider.validate_args(
+            request.args, config[
+                "client_id"], )
+    if user_id is not None:
```
Tyratox, line 169: Looks as if it should be a try / except rather than a "is None" check?

demo/popcha-client/web/src/main.py
```diff
@@ -0,0 +1,222 @@
...
+    else:
+        return html
+
+
+# WARNING: This code is used since it is an example app, but it is UNSAFE
+@app.route("/add_provider")
+def add_provider():
+    """
+    !!! DANGER ZONE !!!
+    WARNING: This call is unsafe and allows to easily add new providers.html. It
```
Tyratox, line 195:
```suggestion
    WARNING: This call is unsafe and allows to easily add new providers. It
```

demo/popcha-client/web/src/main.py
```diff
@@ -0,0 +1,222 @@
...
+        )
+    with open(provider_path) as provider_file:
+        return json.loads(provider_file.read())
+
+
+def on_startup() -> None:
+    """
+    Prepare the data needed by the client authentication server (This
+    server). This includes the list of authorized
+    Open ID providers.html, as well as basic config information such as the
```
Tyratox, line 94:
```suggestion
    Open ID providers, as well as basic config information such as the
```

demo/popcha-client/web/src/main.py
```diff
@@ -0,0 +1,222 @@
...
+# WARNING: This code is used since it is an example app, but it is UNSAFE
+@app.route("/add_provider")
+def add_provider():
+    """
+    !!! DANGER ZONE !!!
+    WARNING: This call is unsafe and allows to easily add new providers.html. It
+    is intended for example / showcase servers that do not provide security.
+    """
+    args = request.args
+    valid_request = ("domain" in args and "lao_id" in args
```
Tyratox, line 199:
```suggestion
    is_valid_request = ("domain" in args and "lao_id" in args
```

Tyratox (commented):
I added quite a few comments, many of them are regarding the code style. I tried to write a lot of comments because there is no python code in the repo yet and we should decide how we want the code to look like :)

I did not yet get to review the tests though!

demo/popcha-client/web/src/main.py
MeKHell: I will, and yes it is a default value.

demo/popcha-client/web/data/providers.json
```diff
@@ -1,14 +1,17 @@
...
     "domain": "valid2.server.example:8000",
-    "lao_id": "SecondLaoID"
+    "lao_id": "5ie-bfxMwsqCtZRY1WuQzhturCoQ3MIvCyX5yAhsyvQ=",
+    "public_key": "ddd"
   },
   {
     "domain": "http://invalid.server.example:8000",
-    "lao_id": "ThirdLaoID"
+    "lao_id": "ThirdLaoID",
+    "public_key": "ddd"
```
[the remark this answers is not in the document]
MeKHell: Yes, this is done on purpose. At the time I did it I did not know the format of the key, I will add correct dummies now.

demo/popcha-client/web/src/authentication.py
```diff
@@ -0,0 +1,99 @@
...
+from werkzeug.datastructures import MultiDict
+
+
+class Authentication:
+    """
+    Class that allows users to authenticate
+    """
+
+    def __init__(self):
+        self.providers: list = []
```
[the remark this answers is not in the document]
MeKHell: They are not defined by a proper type but only a basic dict. But I will still add it.

demo/popcha-client/web/src/authentication.py
```diff
@@ -0,0 +1,99 @@
...
+            "response_mode": "query",
+            "response_type": "id_token",
+            "client_id": client_id,
+            "redirect_uri": f"https://{host_server}:{str(host_port)}/cb",
+            "scope": "openid profile",
+            "login_hint": lao_id,
+            "nonce": nonce,
+            "state": state
+            }
+        return f"https://{auth_server}/authorize?{parse.urlencode(parameters)}"
```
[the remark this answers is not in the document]
MeKHell: Done

demo/popcha-client/web/src/authentication.py
```diff
@@ -0,0 +1,99 @@
...
+            return
+        nonce_data = self.login_states.pop(args.get("state"))
+        server_pub_key = self.public_key_from_iss(nonce_data[1])
+        try:
+            token: dict = jwt.decode(
+                jwt = args.get("id_token", type = str),
+                key = server_pub_key,
+                algorithms = ['RS256'],
+                audience = client_id
+                )
```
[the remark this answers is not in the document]
MeKHell: Just checked and both options are valid in PEP8. I checked in this file, and it seems that I always aligned the ending bracket with the argument indentation (see lines 52 and 103)

demo/popcha-client/web/src/authentication.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def public_key_from_iss(self, iss: str) -> str:
+        """
+        Returns the public key of a registered server.
+        :param iss: The registered server
+        :return: The public key
+        """
+        return ([p for p in self.providers if p.get("domain") == iss][0]).get(
+                "public_key"
+                )
```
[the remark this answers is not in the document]
MeKHell: Way better. I did not know `next`

demo/popcha-client/web/src/counterApp.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def __init__(self):
+        self.app_nonces: dict = {}  # Maps nonce to user
+        self.app_data: dict = {}  # Maps users to counter values
+
+    def get_new_login_params(self, user_id: str) -> dict[str, str]:
+        """
+        Returns a dict specifying the parameter to use for the app.
+        :param user_id: The id of the newly logged-in user
+        :returns: The url where the user should be redirected
```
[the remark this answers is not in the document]
MeKHell: Update in the function role.

demo/popcha-client/web/run.sh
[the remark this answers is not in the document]
MeKHell: After some discussions, since the server does not implement persistency and it is not scheduled to implement it (It is a test server after all), the current behavior should not be a problem. Since the client won't be able to reuse its client id.

demo/popcha-client/web/src/main.py
MeKHell: The run function has been removed. Legacy dev code.

demo/popcha-client/web/src/authentication.py
```diff
@@ -0,0 +1,99 @@
...
+        """
+
+        missing_arg: bool = ("id_token" not in args or "token_type" not in
+                             args or "state" not in args)
+        if missing_arg:
+            return None
+
+        if args.get("state") not in self.login_states.keys():
+            return
+        nonce_data = self.login_states.pop(args.get("state"))
```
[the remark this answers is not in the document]
MeKHell: Yes, all of your concerns should be fixed at a later stage. Since it is only an example server and the token creation time is already checked while the token is decoded, I do not think this is that important. In addition, an attacker could easily add new providers to this website through the `/add_provider` get request.

MeKHell:
I will revert the fe1 file to its original state tomorrow morning. If you have any comment you can already review the code.

demo/popcha-client/web/src/main.py
MeKHell: Then the process will imedidatly crash with an `FileNotFoundError: [Errno 2] No such file or directory:` error. According to this : [https://stackoverflow.com/questions/15562446/how-to-stop-flask-application-without-using-ctrl-c](https://stackoverflow.com/questions/15562446/how-to-stop-flask-application-without-using-ctrl-c), there is no clean easy way to stop the server on failure, so I think that the current behavior is enough.

demo/popcha-client/web/src/authentication.py
```diff
@@ -0,0 +1,118 @@
...
+    def validate_args(self, args: MultiDict[str, str], client_id: str) \
+            -> str | None:
+        """
+        Validate the args as if they were coming from OpenID Authentication
+        request
+        :param args: A MultiDict containing the arguments. Similar to a dict
+        but allows multiple values for a key. Implemented as dict[str, list[str]].
+        See https://werkzeug.palletsprojects.com/en/2.3.x/datastructures/#werkzeug.datastructures.MultiDict
+        :param client_id: The ID of this client
+        :return: The unique user identifier if logged in, None otherwise
```
Nazianzenov, line 73: i don't know if it is possible to have a specific error returned if there is a problem (instead of a general "None")

Nazianzenov (approved):
LGTM, clean code, a small comment on a method signature

demo/popcha-client/web/src/authentication.py
MeKHell: The idea of this function is to return the ID of the user. I think it is better to return `None` since it is the current user's ID based on the provided arguments.

demo/popcha-client/web/src/authentication.py
```diff
@@ -0,0 +1,118 @@
...
+        """
+
+        missing_arg: bool = ("id_token" not in args or "token_type" not in
+                             args or "state" not in args)
+        if missing_arg:
+            return None
+
+        if args.get("state") not in self.login_states:
+            return
```
Xelowak, line 82: Is there a reason why you omitted the `None` here?

demo/popcha-client/web/src/counterApp.py
```diff
@@ -0,0 +1,68 @@
+"""
+This class contains the function related to the application data persitance
```
Xelowak, line 2:
```suggestion
This class contains the function related to the application data persitence
```

demo/popcha-client/web/src/authentication.py
```diff
@@ -0,0 +1,118 @@
...
+    """
+    Class that allows users to authenticate
+    """
+
+    def __init__(self, providers: list[dict[str, str]] = None):
+        if providers is None:
+            providers = []
+
+        self.providers: list[dict[str, str]] = providers
+        self.login_states: dict[str, (str, str, float)] = {}
```
Xelowak, line 29: I think it might be nice to add a tiny bit of documentation to know why the format of these parameters is done this way.

demo/popcha-client/web/src/counterApp.py
```diff
@@ -0,0 +1,68 @@
...
+from werkzeug.datastructures import MultiDict
+
+
+class CounterApp:
+    """
+    An app that allows a user to count
+    """
+
+    def __init__(self):
+        self.app_nonces: dict[str, (str, float)] = {}  # Maps nonce to user
```
Xelowak, line 18: Then why is there a float? Is a user defined by the pair `(str, float)`? It is quite unclear to me 😅 .

Xelowak (approved):
Well done!
I do not have all the details of your implementation in mind given that I don't know much about PoPCha, but I tried to make useful comments for people like me that don't know how it works in details.

As we discussed today, I'm approving so that you can merge it for tomorrow's release. However. do remember to address those comments and any other comment that may come up  after my review.

demo/popcha-client/web/src/main.py
Tyratox: I thought it might be nice to show an error but I guess `FileNotFoundError` is explicit enough :)

demo/popcha-client/web/src/main.py
```diff
@@ -0,0 +1,260 @@
...
+@app.route("/cb")
+def authentication_callback() -> Response:
+    """
+    Redirects the user after the authentication callback has been verified
+    :return: A response which redirects the user to the homePage if the login
+    is not valid or to a new "app" page if the login answer is valid
+    """
+    user_id: str = authenticationProvider.validate_args(
+            request.args, config[
+                "client_id"], )
```
Tyratox, line 212: autopep is killing me 😂

demo/popcha-client/web/src/main.py
```diff
@@ -0,0 +1,260 @@
...
+    provider = {
+        "domain": args["domain"],
+        "lao_id": args["lao_id"],
+        "public_key": args["public_key"]
+        }
+    authenticationProvider.providers.append(provider)
+    return redirect("/")
+
+
+on_startup()
```
Tyratox, line 260:
```suggestion
if __name__ == "__main__":
    on_startup()
```
? :)

Tyratox (commented):
LGTM, just a few minor comments :) Great work

PR #1522, opened 2023-04-25
Message validation for elections

simone-kalbermatter:
This PR makes some improvements to the MessageValidator class to make it more flexible and adds some more checks to it:
- Checking lists for duplicates (based on hashCode)
- Check that timestamps are not too old

Usage of MessageValidator is also extended to the following messages:
- ElectionSetup
- ElectionKey
- ElectionOpen
- CastVote
- ElectionEnd
- ElectionResult


fe2-android/app/src/test/unit/java/com/github/dedis/popstellar/model/network/method/message/data/election/ElectionQuestionTest.java
```diff
@@ -61,6 +65,12 @@ public void electionQuestionGetterReturnsCorrectBallotOptions() {
     assertThat(ELECTION_QUESTION.getBallotOptions(), is(BALLOT_OPTIONS));
   }
 
+  @Test
+  public void testEquals() {
+    assertThat(QUESTION1.equals(QUESTION2), is(false));
+    assertThat(QUESTION2.equals(QUESTION3), is(true));
```
matteosz, line 71: Could it be more clean to just use `assertNotEquals`/`assertEquals`?

fe2-android/app/src/test/unit/java/com/github/dedis/popstellar/model/network/method/message/data/election/ElectionOpenTest.java
```diff
@@ -0,0 +1,102 @@
...
+  public void equalsTest() {
+    ElectionOpen electionOpen2 = new ElectionOpen(LAO_ID, ELECTION_ID, OPENED_AT);
+    assertEquals(ELECTION_OPEN, electionOpen2);
+    assertEquals(ELECTION_OPEN.hashCode(), electionOpen2.hashCode());
+
+    String random = Lao.generateLaoId(ORGANIZER, CREATION + 1, NAME);
+    assertNotEquals(ELECTION_OPEN, new ElectionOpen(random, ELECTION_ID, OPENED_AT));
+    assertNotEquals(ELECTION_OPEN, new ElectionOpen(LAO_ID, random, OPENED_AT));
+    assertNotEquals(ELECTION_OPEN, new ElectionOpen(LAO_ID, ELECTION_ID, CREATION - 1));
+    assertEquals(false, ELECTION_OPEN.equals(null));
```
matteosz, line 72: Here also it seems a bit overcomplicated, simply `assertNotEquals(ELECTION_OPEN, null)` could be more intuitive perhaps

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/data/ElectionHandler.java
```diff
@@ -178,13 +185,25 @@ public void handleElectionEnd(HandlerContext context, ElectionEnd electionEnd)
...
+    String laoId = castVote.getLaoId();
+    if (!laoRepo.containsLao(laoId)) {
+      throw new UnknownLaoException(laoId);
+    }
+
+    // Election id validity is checked with this
     Election election = electionRepository.getElectionByChannel(channel);
+
+    if (election.getCreation() > castVote.getCreation()) {
+      throw new DataHandlingException(castVote, "vote cannot be older than election creation");
```
matteosz, line 204: Here vote and election creation are switched

matteosz (changes requested):
Impressive work 😄 ! Just few comments on small details

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/data/ElectionHandler.java
simone-kalbermatter: Are you sure? I want to check that the election creation comes before the vote creation.

matteosz: Yes the `if` condition is right, I was referring to the error string

simone-kalbermatter: I think both are correct, the error is thrown if the vote is older than the election creation.

matteosz (approved):
LGTM!

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/MessageValidator.java
```diff
@@ -26,6 +26,12 @@ public static MessageValidatorBuilder verify() {
 
   public static class MessageValidatorBuilder {
 
+    // Defines how old messages can be to be considered valid, keeping it non-restrictive here for
+    // now
+    public static final long VALID_DELAY = 100000000;
+    private final long CURRENT_TIME = Instant.now().getEpochSecond();
```
Warhedgehog, line 32: While this is not problematic right now this could lead to unintuitive behaviours in the future if people use this class without realising it was designed so. To be more defensive, it would be bette to define `validPastTime` as a method so that `currentTime` is always correct

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/MessageValidator.java
```diff
@@ -44,36 +50,36 @@ public MessageValidatorBuilder checkValidLaoId(
...
-     * to be in Unix epoch time (UTC). Checks that time is valid and that the first time comes
-     * before the second time.
+     * Helper method to check that times in a Data are ordered. The time values provided are assumed
+     * to be in Unix epoch time (UTC).
      *
-     * @param beforeTime before time
-     * @param afterTime after time
-     * @throws IllegalArgumentException if a time is invalid or if the first time comes after the
-     *     second
+     * @param times time values to be checked
```
Warhedgehog, line 75: You should specify in the method documentation that the times should be ordered and in what order

Warhedgehog (changes requested):
Great work, well documented and tested. A couple things

Warhedgehog (approved):
LGTM

PR #1524, opened 2023-04-26
Solve QRCode not readable on roll call

MeKHell:
This PR should fix the problem where some QR Codes were not readable.

The computation leading to this change is provided as a commentary in the QRCode.tsx file.

This PR will now warn users that go over the 25% covered limit (even if the QR Code might still work).

Here is an image that shows the new size of the circle. 
image

### DIFF
--- a/fe1-web/src/core/components/QRCode.tsx
+++ b/fe1-web/src/core/components/QRCode.tsx
@@ -15,20 +15,23 @@ const styles = StyleSheet.create({
   },
   overlay: {
     /*
-      we are only allowed to cover < 30% (lets say 25% to be safe) for
+      We are only allowed to cover < 30% of the data (lets say 25% to be safe) for
       error correction level H (https://www.qrcode.com/en/about/error_correction.html)
-      assuming the qr code is a square, this means we can cover a width of 50%
-      and a height of 50%. centering this gives us 25% margin on every side
+      Following https://www.maketecheasier.com/assets/uploads/2019/07/qr-code-anatomy-overview.png.webp
+      and https://www.qrcode.com/en/about/version.html for version 7 only 1583 squares of 2025 are used to store data.
+      This leads to 25 * 1583 = 2025 * x where x is the max covered size in terms of QrCode surface percentage.
+      x = 19,543 % = 0.19543. Since we need the result for a line given the squared proportion we can cover
+      sqrt(0.19543) = 0.442 => (100-44.2) / 2 = 27.9
 
-      The above computation might not take into account the area covered by the mouse
-      and some squares that are partially hidden by the overlay (MeKHell, 13.04.2023)
+      The above computation takes into account the area covered by
+      position, alignment and timing patterns; and version and format information; which are not data
     */
     position: 'absolute',
     padding: '4',
-    top: '26%',
-    left: '26%',
-    right: '26%',
-    bottom: '26%',
+    top: '27.9%',
+    left: '27.9%',
+    right: '27.9%',
+    bottom: '27.9%',
     display: 'flex',
     justifyContent: 'center',
     alignItems: 'center',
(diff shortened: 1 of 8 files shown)


fe1-web/src/core/components/__tests__/QrCode.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+      ),
+    );
+
+    global.console.warn = tmpConsoleWarn;
+  });
+
+  it('looks good with short text and an overlay', () => {
+    const { toJSON } = render(<QRCode value={shortText} overlayText="OverlayText" />);
+    expect(toJSON()).toMatchSnapshot();
+  });
```
dayan9265, line 37: In my opinion, it seems a bit weird to test the visual of a case we would like to avoid (short text with overlay text)

dayan9265 (approved):
Looks good to me. Always a bit hard to estimate with these QR codes if now there will be no more problem when scanning, but let's hope so ^^. I added a small comment for one of the test, maybe the test does make sense in your opinion and then no need to change it ;)

fe1-web/src/core/components/QRCode.tsx
```diff
@@ -52,6 +55,13 @@ const QRCode = ({ value, visible, overlayText }: IPropTypes) => {
   // highest error correction if there is an overlay
   const errorCorrectionLevel = overlayText ? 'H' : 'L';
 
+  // Warns that the overlay is too big if the text is too small
+  if (overlayText && value && value.length < 59) {
+    console.warn(
+      'An overlay text has been added on a QRCode whose represents a too short text (length < 50)',
+    );
+  }
```
Tyratox, line 63: What about forcing overlayText to false in this case and logging the warning / showing some text below the qr code that indicates the issue?

fe1-web/src/core/components/__tests__/QrCode.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+  (message?: any, ...optionalParams: any[]): void;
+  (...data: any[]): void;
+  (message?: any, ...optionalParams: any[]): void;
+  (message?: any, ...optionalParams: any[]): void;
+};
+describe('QrCode ', () => {
+  // It is useless to test the contrary since we would wait and validate instantly even it would be wrong
+  it('throws a warning, short text with overlay', () => {
+    tmpConsoleWarn = global.console.warn;
+    global.console.warn = jest.fn();
```
Tyratox, line 22: You might want to check out `spyOn` for this :) The you also don't need the `tmpConsoleWarn`.

https://jestjs.io/docs/jest-object#jestspyonobject-methodname
https://stackoverflow.com/a/56677834/2897827

fe1-web/src/core/components/__tests__/QrCode.test.tsx
Tyratox: Haha I agree but with my comment above this test would make sense with a slightly different description. Maybe @MeKHell already had something like this in mind :)

Tyratox (approved):
LGTM! 😄 Maybe have a quick look at my comments and decide whether they make sense :)

fe1-web/src/core/components/__tests__/QrCode.test.tsx
MeKHell: I already tried it but maybe forgot to use global. (It did not work)

fe1-web/src/core/components/__tests__/QrCode.test.tsx
MeKHell: The idea of this test is :

- Make sure that even if we do not guarantee the QR Code is readable, it still looks good in terms of UI. (For the QR Code readability, we test it through warning test)
- Have a test that checks the blue overlay circle

fe1-web/src/core/components/__tests__/QrCode.test.tsx
MeKHell: Okay, the problem that I had is I did not wait for the warning.

fe1-web/src/core/components/QRCode.tsx
MeKHell: The QR might still work. So I am not sure it is a good idea.

Tyratox: It might but there is no guarantee. And I would certainly prioritize functionality over having a nice looking circle but this is up to you 😅

Tyratox (approved):

dayan9265 (approved):
LGTM👍

PR #1525, opened 2023-04-27
Add basic http-get and authenticate request handlers

Jharaxus:
- Adds architecture for handling http get requests by the server.
- Adds basic handling of authenticate requests from a client


Jharaxus:
Reopened after adding latest specification changes.

be2-scala/src/main/scala/ch/epfl/pop/authentication/Authenticate.scala
```diff
@@ -0,0 +1,99 @@
...
+      _ <- verifyResponseType(params.response_type)
+      _ <- verifyScope(params.scope)
+      _ <- verifyRedirectUri(params.redirect_uri)
+      result <- verifyResponseMode(params.response_mode)
+    } yield result
+  }
+
+  private def verifyResponseType(response_type: String): VerificationState = {
+    val expectedResponseType = "id_token token"
+    if (response_type == expectedResponseType) Right(())
```
K1li4nL, line 58: nitpick alert: I think a line return would make it more readable. 
```suggestion
    if (response_type == expectedResponseType) 
       Right(())
```

be2-scala/src/main/scala/ch/epfl/pop/authentication/Authenticate.scala
```diff
@@ -0,0 +1,99 @@
...
+    else
+      Left("unsupported_response_type" -> s"expected \"$expectedResponseType\" but received \"$response_type\"")
+  }
+
+  private def verifyScope(scope: String): VerificationState = {
+    val expectedScope = "openid"
+    val scopes = scope.split(" ")
+    if (scopes.contains(expectedScope)) Right(())
```
K1li4nL, line 66: nitpick alert: I think a line return would make it more readable. 
```suggestion
    if (scopes.contains(expectedScope))
       Right(())
```

be2-scala/src/main/scala/ch/epfl/pop/authentication/Authenticate.scala
```diff
@@ -0,0 +1,99 @@
...
+    else
+      Left("invalid_scope" -> s"expected scope to contain \"$expectedScope\" but received \"$scope\"")
+  }
+
+  private def verifyRedirectUri(uri: String): VerificationState = {
+    // Checks that the uri is a http or https format uri
+    // see https://stackoverflow.com/questions/3809401/what-is-a-good-regular-expression-to-match-a-url for regex source
+    val httpRegex =
+      "https?:\\/\\/(www\\.)?[-a-zA-Z0-9@:%._\\+~#=]{1,256}\\.[a-zA-Z0-9()]{1,6}\\b([-a-zA-Z0-9()@:%_\\+.~#?&//=]*)"
+    if (uri.matches(httpRegex)) Right(())
```
K1li4nL, line 76: nitpick alert: I think a line return would make it more readable. 
```suggestion
    if (uri.matches(httpRegex))
       Right(())
```

be2-scala/src/main/scala/ch/epfl/pop/authentication/Authenticate.scala
```diff
@@ -0,0 +1,99 @@
...
+    else
+      Left("invalid_request" -> s"expected http or https url format for redirect uri")
+  }
+
+  private def verifyResponseMode(mode: Option[String]): VerificationState = {
+    val supportedModes = List("query", "fragment")
+    if (mode.isEmpty || supportedModes.contains(mode.get)) Right(())
```
K1li4nL, line 83: nitpick alert: I think a line return would make it more readable. 
```suggestion
    if (mode.isEmpty || supportedModes.contains(mode.get)) 
      Right(())
```

be2-scala/src/main/scala/ch/epfl/pop/authentication/Authenticate.scala
```diff
@@ -0,0 +1,99 @@
...
+        "response_type",
+        "client_id",
+        "redirect_uri",
+        "scope",
+        "state".optional,
+        "response_mode".optional,
+        "login_hint",
+        "nonce"
+      ) {
+        (response_type, client_id, redirect_uri, scope, state, response_mode, login_hint, nonce) =>
```
K1li4nL, line 28: I think `client_id`,  `state`, `login_hint` and the `nonce` are left unchecked, wouldn't be a good idea to do so (if possible) or is that something bound to be done later maybe ?

be2-scala/src/main/scala/ch/epfl/pop/authentication/Authenticate.scala
K1li4nL, line 1: I  think this file deserve some more documentation / comments in general

be2-scala/src/test/scala/ch/epfl/pop/authentication/AuthenticateSuite.scala
```diff
@@ -0,0 +1,222 @@
...
+import org.scalatest.funsuite.AnyFunSuite
+import org.scalatest.matchers.should.Matchers
+
+class AuthenticateSuite extends AnyFunSuite with Matchers with ScalatestRouteTest {
+
+  private val routeName = "authenticate/"
+
+  private val responseType = "id_token token"
+  private val clientID = "abc"
+  private val redirectUri = "https://wikipedia.org"
```
K1li4nL, line 16: nitpick: I think the domain `example.com` exists for these use case, though I don't think Wikipedia would come back at us...

be2-scala/src/test/scala/ch/epfl/pop/authentication/AuthenticateSuite.scala
```diff
@@ -0,0 +1,222 @@
...
+      val subParams = mandatoryParams.filter(_._1 != name)
+      val request = buildRequest(subParams)
+
+      request ~> route ~> check {
+        status shouldBe Found
+        getAttributeValue(response, "error") shouldBe Some("invalid_request")
+        getAttributeValue(response, "error_description") shouldBe Some(s"Missing parameters: [$name]")
+      }
+    }
+  }
```
K1li4nL, line 140: A lot seems to be going there, I am not sure to follow, some comments would help ;)

K1li4nL (changes requested):
Overall the code is concise and good looking, well done ! 
Beside my nitpicks, I think some documentation would be nice ;)

be2-scala/src/main/scala/ch/epfl/pop/authentication/Authenticate.scala
Jharaxus: It is actually intended as they are left for the client (github for example using popcha for authentication) to use: client_id, state (optional) and nonce are entirely client-defined --> there are no ways to check them, neither any properties that we would like to enforce (nonce has a client-side property only). State actually should be a valid lao_id, but this doesn't need to be checked here as the user's pop app will do it to provide better flexibility, and we wouldn't actually have any means to check that now without heavily breaking the current encapsulation anyways (but again it is intended not to make that check here).

be2-scala/src/main/scala/ch/epfl/pop/authentication/Authenticate.scala
Jharaxus: Done! I commented all the public scope methods/classes and refactored a bit some parts to make them clear enough without comments.

be2-scala/src/test/scala/ch/epfl/pop/authentication/AuthenticateSuite.scala
Jharaxus: Wow I didn't know that domain actually existed for real 😳

be2-scala/src/test/scala/ch/epfl/pop/authentication/AuthenticateSuite.scala
Jharaxus: I simplified the code a bit and changed some variables names to make it clearer

K1li4nL (approved):
LGTM !

be2-scala/src/main/scala/ch/epfl/pop/authentication/Authenticate.scala
```diff
@@ -0,0 +1,112 @@
...
+      result <- verifyResponseMode(params.response_mode)
+    } yield result
+  }
+
+  private def verifyResponseType(response_type: String): VerificationState = {
+    val expectedResponseType = "id_token token"
+    if (response_type == expectedResponseType)
+      Right(())
+    else
+      Left("unsupported_response_type" -> s"expected \"$expectedResponseType\" but received \"$response_type\"")
```
Ajkunas, line 70: It is a general comment: I suggest that you declare the type of errors as constant values. 
You can then reuse them in your test suites.

Ajkunas (changes requested):
Very clean and concise code :) We easily understand what happens in the file. Still do not hesitate to add some comments here and there. 
Otherwise, nothing really important to add/change. I left a suggestion and I am open to discuss it :)

be2-scala/src/test/scala/ch/epfl/pop/authentication/AuthenticateSuite.scala
```diff
@@ -0,0 +1,207 @@
...
+      clientID,
+      redirectUri,
+      scope,
+      "",
+      "",
+      loginHint,
+      nonce
+    )
+
+    println(request)
```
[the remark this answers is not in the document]
Jharaxus: Indeed, I will remove it !

be2-scala/src/main/scala/ch/epfl/pop/authentication/Authenticate.scala
Jharaxus: That makes sense! I put them directly in the Authenticate object as they are not enough general-purpose to justify a separate file for them (and there are only 3).

Ajkunas (approved):
LGTM

PR #1529, opened 2023-04-29
Remove catchup and broadcasts between servers

MariemBaccari:
This PR removes catchups and broadcasts between servers on the go backend as they are not useful anymore since the merge of the heartbeat and getMessagesById features.
It adapts or deletes some previous tests that assume catchups and broadcasts between servers.
Finally, it removes the globalInbox and leaves only the hubInbox since we only need that one now that we don't have broadcasts between servers anymore.

On the heartbeat / getMessagesById feature side it:
- Adds sending a heartbeat message after receiving a publish message 
- Changes the delay of sending a heartbeat message to 30 seconds 
- Removes the updateRecord method as it appears to be useless

### DIFF
--- a/be1-go/channel/mod.go
+++ b/be1-go/channel/mod.go
@@ -44,13 +44,13 @@ func NewSockets() Sockets {
 	}
 }
 
-// Sockets provides thread-funcionalities around a socket store.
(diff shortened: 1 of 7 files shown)


Anghul (approved):
Nice work

PR #1530, opened 2023-04-30
Feature: connect components together for decentralized communications

K1li4nL:
This pr is connecting all the individual components we have been working on so far to allow for servers to servers communication. At this stage only rollcall are working properly. 

The commit history should reflect well the intent behind the code, hence I believe the review  should be easier to do by going commits by commits.

### DIFF
--- a/be2-scala/src/main/scala/ch/epfl/pop/pubsub/ClientActor.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/pubsub/ClientActor.scala
@@ -3,6 +3,7 @@ package ch.epfl.pop.pubsub
 import akka.actor.{Actor, ActorLogging, ActorRef, Props}
 import akka.event.LoggingReceive
 import akka.pattern.AskableActorRef
+import ch.epfl.pop.decentralized.ConnectionMediator
 import ch.epfl.pop.model.objects.Channel
 import ch.epfl.pop.pubsub.ClientActor._
 import ch.epfl.pop.pubsub.PubSubMediator._
@@ -20,6 +21,10 @@ final case class ClientActor(mediator: ActorRef, connectionMediatorRef: ActorRef
 
   private val mediatorAskable: AskableActorRef = mediator
 
+  // Tell connectionMediator we are online
+  if (isServer)
+    connectionMediatorRef ! ConnectionMediator.NewServerConnected(self)
+
   private def messageWsHandle(event: ClientActorMessage): Unit = event match {
     case ClientAnswer(graphMessage) => wsHandle.fold(())(_ ! graphMessage)
   }
@@ -30,7 +35,10 @@ final case class ClientActor(mediator: ActorRef, connectionMediatorRef: ActorRef
           log.info(s"Connecting wsHandle $wsClient to actor ${this.self}")
           wsHandle = Some(wsClient)
 
-        case DisconnectWsHandle => subscribedChannels.foreach(channel => mediator ! PubSubMediator.UnsubscribeFrom(channel, this.self))
+        case DisconnectWsHandle =>
+          if (isServer)
+            connectionMediatorRef ! ConnectionMediator.ServerLeft(self)
+          subscribedChannels.foreach(channel => mediator ! PubSubMediator.UnsubscribeFrom(channel, this.self))
 
         case ClientActor.SubscribeTo(channel) =>
           val ask: Future[PubSubMediatorMessage] = (mediatorAskable ? PubSubMediator.SubscribeTo(channel, this.self)).map {
--- a/be2-scala/src/main/scala/ch/epfl/pop/pubsub/ClientActor.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/pubsub/ClientActor.scala
@@ -22,8 +22,9 @@ final case class ClientActor(mediator: ActorRef, connectionMediatorRef: ActorRef
   private val mediatorAskable: AskableActorRef = mediator
 
   // Tell connectionMediator we are online
-  if (isServer)
+  if (isServer) {
     connectionMediatorRef ! ConnectionMediator.NewServerConnected(self)
+  }
 
   private def messageWsHandle(event: ClientActorMessage): Unit = event match {
     case ClientAnswer(graphMessage) => wsHandle.fold(())(_ ! graphMessage)
@@ -36,9 +37,10 @@ final case class ClientActor(mediator: ActorRef, connectionMediatorRef: ActorRef
           wsHandle = Some(wsClient)
 
         case DisconnectWsHandle =>
-          if (isServer)
-            connectionMediatorRef ! ConnectionMediator.ServerLeft(self)
           subscribedChannels.foreach(channel => mediator ! PubSubMediator.UnsubscribeFrom(channel, this.self))
+          if (isServer) {
+            connectionMediatorRef ! ConnectionMediator.ServerLeft(self)
+          }
 
         case ClientActor.SubscribeTo(channel) =>
           val ask: Future[PubSubMediatorMessage] = (mediatorAskable ? PubSubMediator.SubscribeTo(channel, this.self)).map {
--- a/be2-scala/src/main/scala/ch/epfl/pop/pubsub/ClientActor.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/pubsub/ClientActor.scala
@@ -13,7 +13,7 @@ import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.{Await, Future}
 import scala.util.Failure
 
-final case class ClientActor(mediator: ActorRef) extends Actor with ActorLogging with AskPatternConstants {
+final case class ClientActor(mediator: ActorRef, connectionMediatorRef: ActorRef, isServer: Boolean) extends Actor with ActorLogging with AskPatternConstants {
 
   private var wsHandle: Option[ActorRef] = None
   private val subscribedChannels: mutable.Set[Channel] = mutable.Set.empty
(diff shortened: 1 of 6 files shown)


be2-scala/src/main/scala/ch/epfl/pop/pubsub/PublishSubscribe.scala
```diff
@@ -5,30 +5,43 @@ import akka.actor.{ActorRef, ActorSystem}
...
     implicit builder: GraphDSL.Builder[NotUsed] =>
       {
         import GraphDSL.Implicits._
 
-        val clientActorRef: ActorRef = system.actorOf(ClientActor.props(mediatorActorRef))
+        val clientActorRef: ActorRef = system.actorOf(ClientActor.props(mediatorActorRef, connectionMediatorRef, isServer))
         dbActorRef = dbActorRefT
 
+        if (isServer)
+          connectionMediatorRef ! ConnectionMediator.NewServerConnected(clientActorRef)
```
Jharaxus, line 36: I don't think that a message indicating that a new server connected should be emitted in a function called `buildGraph`.
It looks to me that either there should be a more meaningful place to call that portion of code, or build graph now starts to be doing a bit to many things at the same time (and would at least need a proper documentation indicating what it does and in which more global context it is supposed to be called).
Actually it is called during server's initialization and in `ConnectionMediator` when receiving a `ConnectTo` message, so that makes sense that it matches every time a new server actor is created, but we are getting quite far from the initial function intent of "building a DSL graph" so I would still root for a call somewhere else.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/PublishSubscribe.scala
```diff
@@ -65,8 +84,12 @@ object PublishSubscribe {
         methodPartitioner.out(portPipelineError) ~> merger
         methodPartitioner.out(portParamsWithMessage) ~> hasMessagePartition ~> merger
         methodPartitioner.out(portParamsWithChannel) ~> hasChannelPartition ~> merger
+        methodPartitioner.out(portParamsWithMap) ~> hasMapPartition ~> merger
+        methodPartitioner.out(portResponseHandler) ~> responsePartition ~> merger
 
-        merger ~> jsonRpcAnswerGenerator ~> jsonRpcAnswerer ~> output
+        merger ~> broadcast
+        broadcast.out(0) ~> jsonRpcAnswerGenerator ~> jsonRpcAnswerer ~> output
+        broadcast.out(1) ~> monitorSink
```
Jharaxus, line 92: You could add variables such as `portBroadcastSuccess`, `portBroadcastFailure` and `totalBroadcastPorts` to remove magic numbers and make their intent clearer.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/Validator.scala
```diff
@@ -28,29 +28,29 @@ object Validator {
...
+        case _: Unsubscribe     => validateUnsubscribe(jsonRpcRequest)
+        case _: Heartbeat       => graphMessage // No check necessary
+        case _: GetMessagesById => graphMessage // No check necessary
+        case _                  => Left(validationError(jsonRpcRequest.id))
       }
-    case Right(jsonRpcResponse: JsonRpcResponse) => Left(PipelineError(
-        ErrorCodes.SERVER_ERROR.id,
-        "Unsupported action: MethodValidator was given a response message",
-        jsonRpcResponse.id
-      ))
```
Jharaxus, line 42: Was it necessary to remove that?

Jharaxus (changes requested):
Great job going through all the project's architecture to put all the pieces you added so far together!
Just 2/3 little comments and it should be ready to go :)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/PublishSubscribe.scala
K1li4nL: You are right, this piece of code has followed me since week 4 or something I haven't thought about its sense since. Definitely not a great place. How about putting it in the ClientActor directly ? The clientActor is already the one telling ConnectionMediator when it leaves so it might as well tell it it arrived.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/PublishSubscribe.scala
K1li4nL: These number have no meaning really, both pipe output the same data. I think Akka even lets me remove them, I will check that.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/Validator.scala
K1li4nL: get_messages_by_id responses wouldn't get through otherwise. The json answers our backend is able to parse is fairly limited, basically: `result: 0` + catchup (list of messages) + now the maps channels -> list(messages)
So if the answer is outside of these 3, it can't parse it and an error is thrown. Finally if it is one of these 3 then `GetMessagesByIdResponseHandler` will deal with it, either silently dropping them or treating the maps.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/PublishSubscribe.scala
K1li4nL: taken care of in commit: 0ad00eea2447aa4c53d75c9556aee01db6306881

be2-scala/src/main/scala/ch/epfl/pop/pubsub/PublishSubscribe.scala
K1li4nL: taken care of in commit: a8f08f609d32e42aabbc2159fbdc8efab80f9602

be2-scala/src/main/scala/ch/epfl/pop/pubsub/ClientActor.scala
```diff
@@ -25,6 +25,10 @@ final case class ClientActor(mediator: ActorRef, connectionMediatorRef: ActorRef
     case ClientAnswer(graphMessage) => wsHandle.fold(())(_ ! graphMessage)
   }
 
+  // Tell connectionMediator we are online
+  if (isServer)
+    connectionMediatorRef ! ConnectionMediator.NewServerConnected(self)
```
Jharaxus, line 30: Nitpick but this could be put above `private def messageWsHandle(event: ClientActorMessage): Unit = ...` to avoid having a "constructor instruction" appear out of nowhere between two method definitions.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/Validator.scala
Jharaxus: Ok thanks for the explanation, it indeed makes sense not to reject them now that we can receive responses from other servers!
However, I believe it would be clearer to keep the explicit case handling and always accept it (as you did for `Heartbeat` and `GetMessagesById`), both for `validateMethodContent` and `validateMessageContent`. In fact, I don't see the point of validating messages that are not explicitly handled, I would find it safer to reject messages if they don't match the expected patterns (which would then force us to explicitly handle jsonRpcResponse messages, even if trivially).

Jharaxus: I couldn't add the comments under the relevant lines as they were not affected by these commits, so here are the comments I wanted to add to the previous idea:
- line 41 of Validator.scala: `case _ => graphMessage`
- line 55 of Validator.scala: `case graphMessage @ _ => graphMessage`
As explained above, for safety reasons shouldn't they be rejected (`Left(...)`) if they are not handled? (It's not what your commit exactly is about, but this could be a good moment to change that if needed)

Jharaxus (commented):
A little nitpick but the rest is fine :)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/Validator.scala
K1li4nL: These are all good ideas, accept by default are not great, far from it. But it starts to introduce a lot in the same pr, I feel like this should be in its own refactoring work in order to avoid tangling together too many changes. What do you think ?

Jharaxus: You're right, I'll do an issue about it and a simple pr should solve it !

be2-scala/src/main/scala/ch/epfl/pop/pubsub/ClientActor.scala
K1li4nL: Fixed !

be2-scala/src/main/scala/ch/epfl/pop/pubsub/PublishSubscribe.scala
K1li4nL: Fixed !

Jharaxus (approved):

be2-scala/src/main/scala/ch/epfl/pop/pubsub/PublishSubscribe.scala
```diff
@@ -52,7 +60,9 @@ object PublishSubscribe {
         val hasChannelPartition = builder.add(ParamsWithChannelHandler.graph(clientActorRef))
 
         val merger = builder.add(Merge[GraphMessage](totalPorts))
+        val broadcast = builder.add(Broadcast[GraphMessage](2))
```
Ajkunas, line 69: For consistency, I think the "2" should be a variable with its corresponding name.

Ajkunas (changes requested):
Good job ! Thank you for your clean commits :) 
Honestly, only few nitpicks, otherwise it is ready to be approved.

Ajkunas (approved):
All good :) I will let you ping @pierluca to force the merging of the PR.

PR #1532, opened 2023-05-01
Refactor Karate Part 1

simone-kalbermatter:
This PR makes the Karate tests more readable by removing all `JSON.stringify` calls.


matteosz (approved):
LGTM! Nice refactoring, the method `JSON.stringify` was just impacting on the tests readability

Warhedgehog (commented):
So the `JSON.stringify` was completely useless ?

simone-kalbermatter:
@Warhedgehog Yes, the `JSON.stringify` was useless, it was probably just used because it was easier to work with strings directly instead of converting json maps in `MultiMsgWebSocketClient`. For example, because it extends `WebSocketClient`, it inherits a `send(String msg)` function.

Warhedgehog (approved):
LGTM

Issue #1508, opened 2023-04-20
[BUG] Very high cpu usage 


MeKHell:
### Description (Actual behavior)
High CPU usage when we have others LAO unprocessed messages.

### How to reproduce
1. Have multiple open LAOS and open one.

### Version & Environment
This bug was reproduced on:
- Commit: 1.9 release
##### Front-ends:
- [x] Fe1-Web (all versions)
- [ ] Fe2-Android (please specify if phone or emulation, and Android version)
- [ ] Not applicable

##### Back-ends:
- [ ] Be1-Go
- [ ] Be2-Scala
- [ ] Not applicable

### Impact
Indicate here, based on your understanding, what the possible consequences of this bug are.
[e.g. The app becomes unusable, as data gets permanently corrupted and it crashes on every subsequent start]

### Possible root cause
The Processing of the witness message after the Issue removing warning messages.

pierluca:
@MeKHell has this issue been wholly fixed in #1535 ? 

FYI, if you indicate in a PR something like `fixes #1508` GitHub will automatically close the issue for you.
Here's the list of keywords: https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests

MeKHell:
I used the only keyword that I thought was in but did not. :sweat_smile: It is indeed solved.

PR #1535, opened 2023-05-02
Fix performance issues

MeKHell:
This PR solves #1508. 
The idea is to keep in memory the message and use the mandatory first pass to map each message ID to the correct LAO ID. This will now only create the extended message from the ID a maximum of 2 times per message. 

Before this change, the extended message creation was every time a new message arrived, for each message that belongs to other LAOs.


dayan9265 (approved):
Thanks for fixing the buggy test from my PR ^^, otherwise the code looks good :)

fe1-web/src/features/witness/network/WitnessStoreWatcher.ts
```diff
@@ -152,21 +157,31 @@ export const makeWitnessStoreWatcher = (
       return;
     }
 
-    const messagesToWitness = currentAllIds.filter(
+    let messagesToWitness = currentAllIds.filter(
       (msgId) =>
         !currentUnprocessedIds.includes(msgId) &&
         (!previousAllIds.includes(msgId) || previousUnprocessedIds.includes(msgId)),
     );
```
Tyratox, line 164: This is also a possible source of performance issues as it is $O(n^2)$, maybe we can improve this at a later point :)

Tyratox (approved):
LGTM!

PR #1536, opened 2023-05-02
json schema for authentication messages in PoPCHA

Nazianzenov:
json schema describing an authentication message transmitted from the PoP App to the PoP Backend

### DIFF
--- a/protocol/query/method/message/data/dataAuthenticateUser.json
+++ b/protocol/query/method/message/data/dataAuthenticateUser.json
@@ -1,0 +1,46 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://raw.githubusercontent.com/dedis/popstellar/master/protocol/query/method/message/data/dataSendAuthMessage.json",
+  "description": "Authenticate a user in PoPCHA protocol",
+  "type": "object",
+  "properties": {
+    "object": {
+      "const": "popcha"
+    },
+    "action": {
+      "const": "authenticate"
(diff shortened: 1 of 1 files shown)


protocol/query/method/message/data/dataAuthenticateUser.json
```diff
@@ -0,0 +1,46 @@
...
+        "description": "typeof http response",
+        "type": "string"
+    },
+    "popcha_address": {
+        "description": "host of the popcha authorization server",
+        "type": "string"
+    }
+  },
+  "additionalProperties": false,
+  "required": ["object", "action","client_id","nonce","identifier","identifier_proof","state","response_mode","popcha_address"]
```
dayan9265, line 45: We still need to see if we want to make the `response_mode` and `state` required

dayan9265 (changes requested):
The JSON looks nice 👍 , you can try to merge main with this branch to have the tests working

protocol/query/method/message/data/dataAuthenticateUser.json
MeKHell: I think they won't be...

Nazianzenov: indeed.

MeKHell (commented):
LGTM. But I cannot accept it since I am not working with it. I let @dayan9265 accept it for me, if it is also good for him.

Anghul (approved):

PR #1540, opened 2023-05-04
Write CreateLao messages on /root and SetupElection on /root/lao_id

K1li4nL:
CreateLao message are expected to be received on `/root`, until now the scala backend wrote them under `/root/lao_id`. This  behavior causes problems with `heartbeat` and `get_messages_by_id` since they reflect the internal organisations of messages. When other server receive  our `CreateLao` under the `/root/lao_id` channel they  reject it. 

The slight difficulty of moving `CreateLao` messages in `/root` is that they still need to be included in catchups, clients won't connect without them, hence the need of the specialized functions added in this pr. 

SetupElection message is expected to be under `/root/lao_id`, some functions were added to make that possible. Without this change heartbeat will send this message under `/root/lao_id/election_id` which won't be accepted by other server as it goes against the protocol.

### DIFF
--- a/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandler.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandler.scala
@@ -3,7 +3,6 @@ package ch.epfl.pop.pubsub.graph.handlers
 import akka.pattern.AskableActorRef
 import ch.epfl.pop.json.MessageDataProtocol.{KeyElectionFormat, resultElectionFormat}
 import ch.epfl.pop.model.network.JsonRpcRequest
-import ch.epfl.pop.model.network.method.message.data.ObjectType
 import ch.epfl.pop.model.network.method.message.data.election.VersionType._
 import ch.epfl.pop.model.network.method.message.data.election._
 import ch.epfl.pop.model.objects.ElectionChannel._
@@ -51,7 +50,6 @@ class ElectionHandler(dbRef: => AskableActorRef) extends MessageHandler {
       electionId: Hash = data.id
       electionChannel: Channel = Channel(s"${rpcMessage.getParamsChannel.channel}${Channel.CHANNEL_SEPARATOR}$electionId")
       keyPair = KeyPair()
-      _ <- dbActor ? DbActor.CreateChannel(electionChannel, ObjectType.ELECTION)
       _ <- dbActor ? DbActor.WriteSetupElectionMessage(electionChannel, message)
       _ <- dbActor ? DbActor.CreateElectionData(rpcMessage.extractLaoId, electionId, keyPair)
     } yield (data, electionId, keyPair, electionChannel)
--- a/be2-scala/src/test/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandlerTest.scala
+++ b/be2-scala/src/test/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandlerTest.scala
@@ -49,7 +49,10 @@ class ElectionHandlerTest extends TestKit(ActorSystem("Election-DB-System")) wit
     val dbActorMock = Props(new Actor() {
       override def receive: Receive = {
         // You can modify the following match case to include more args, names...
-        case DbActor.WriteAndPropagate(_, _) | DbActor.ChannelExists(_) | DbActor.CreateChannel(_, _) =>
+        case DbActor.WriteAndPropagate(_, _)
+            | DbActor.ChannelExists(_)
+            | DbActor.CreateChannel(_, _)
+            | DbActor.WriteSetupElectionMessage(_, _) =>
           system.log.info(f"Received a message")
           system.log.info("Responding with a Nack")
           sender() ! Status.Failure(DbActorNAckException(1, "error"))


be2-scala/src/test/scala/ch/epfl/pop/storage/DbActorSuite.scala
```diff
@@ -493,6 +493,58 @@ class DbActorSuite extends TestKit(ActorSystem("DbActorSuiteActorSystem")) with
...
 
+  test("CreateLao messages can be written and read back in catchup on lao channel") {
+    // arrange
+    val initialStorage: InMemoryStorage = InMemoryStorage()
+    val dbActor: AskableActorRef = system.actorOf(Props(DbActor(mediatorRef, MessageRegistry(), initialStorage)))
+    val channel = Channel(CHANNEL_NAME)
+
+    // Write some message using WriteCreateLaoMessage
+    val writeAsk = dbActor ? DbActor.WriteCreateLaoMessage(channel, MESSAGE)
+    val writeAnswer = Await.result(writeAsk, duration)
```
BelghmiAmine, line 504: Since sending a WriteCreateLaoMessage can throw a DbActorNackException, maybe you could replace calls to Await.result by Await.ready in order to query the dbActor in a cleaner way. See : https://stackoverflow.com/questions/41170280/difference-await-ready-and-await-result

BelghmiAmine (changes requested):

be2-scala/src/test/scala/ch/epfl/pop/storage/DbActorSuite.scala
K1li4nL: I am not convinced, this is a unit test meant to pass, hence no errors are expected. The point is that the test failed and something need fixing. I don't see the point of handling errors since it will end the same way: Test failed

BelghmiAmine (dismissed):

be2-scala/src/main/scala/ch/epfl/pop/storage/DbActor.scala
```diff
@@ -45,6 +45,21 @@ final case class DbActor(
...
+  @throws[DbActorNAckException]
+  private def writeCreateLao(channel: Channel, message: Message): Unit = {
+    createChannel(channel, ObjectType.LAO)
+    storage.write((storage.DATA_KEY + storage.CREATE_LAO_KEY + channel.toString, message.message_id.toString()))
+    write(Channel.ROOT_CHANNEL, message)
+  }
+
+  @throws[DbActorNAckException]
+  private def readCreateLao(channel: Channel): Option[Message] = {
+    storage.read(storage.DATA_KEY + storage.CREATE_LAO_KEY + channel.toString) match {
```
Ajkunas, line 57: Now that I think about it, the `read` function defined inside `DbActor.scala` should be commented a bit: we should precise that we use this function to read messages except the `CreateLao` ones (smth like this).

Ajkunas (changes requested):
Only some nitpicks, othw all good :)

be2-scala/src/main/scala/ch/epfl/pop/storage/DbActor.scala
```diff
@@ -45,6 +45,48 @@ final case class DbActor(
...
 
+  @throws[DbActorNAckException]
+  private def writeCreateLao(channel: Channel, message: Message): Unit = {
+    createChannel(channel, ObjectType.LAO)
+    storage.write((storage.DATA_KEY + storage.CREATE_LAO_KEY + channel.toString, message.message_id.toString()))
+    write(Channel.ROOT_CHANNEL, message)
+  }
+
+  @throws[DbActorNAckException]
+  private def readCreateLao(channel: Channel): Option[Message] = {
```
Ajkunas, line 56: Now that I think about it, maybe the `read` function should be a bit commented (used for all messages except the CreateLao ones). It seems obvious since we add a new function `readCreateLao`, but some new students could still miss it. :)

be2-scala/src/main/scala/ch/epfl/pop/storage/DbActor.scala
```diff
@@ -278,6 +335,27 @@ final case class DbActor(
...
+
+    case WriteSetupElectionMessage(channel, message) =>
+      log.info(s"Actor $self (db) received a WriteSetupElectionMessage request")
+      Try(writeSetupElectionMessage(channel, message)) match {
+        case Success(_) => sender() ! DbActorAck()
+        case failure    => sender() ! failure.recover(Status.Failure(_))
+      }
+
+    case ReadSetupElectionMessage(channel) =>
+      log.info(s"Actor $self (db) received a WriteSetupElectionMessage request")
```
Ajkunas, line 353:
```suggestion
      log.info(s"Actor $self (db) received a ReadSetupElectionMessage request")
```

be2-scala/src/test/scala/ch/epfl/pop/storage/DbActorSuite.scala
```diff
@@ -493,6 +493,109 @@ class DbActorSuite extends TestKit(ActorSystem("DbActorSuiteActorSystem")) with
...
+    failingMessage should equal(None)
+  }
+
+  test("SetupElection messages should only be written on main lao channel") {
+    // arrange
+    val initialStorage: InMemoryStorage = InMemoryStorage()
+    val dbActor: AskableActorRef = system.actorOf(Props(DbActor(mediatorRef, MessageRegistry(), initialStorage)))
+    val channel = Channel("/root/lao_id/election_id")
+
+    // Write some message using WriteCreateLaoMessage
```
Ajkunas, line 554:
```suggestion
    // Write some message using WriteSetupElectionMessage
```

be2-scala/src/test/scala/ch/epfl/pop/storage/DbActorSuite.scala
```diff
@@ -493,6 +493,109 @@ class DbActorSuite extends TestKit(ActorSystem("DbActorSuiteActorSystem")) with
...
+    val writeAsk = dbActor ? DbActor.WriteSetupElectionMessage(channel, MESSAGE)
+    val writeAnswer = Await.result(writeAsk, duration)
+
+    // assert
+    writeAnswer shouldBe a[DbActor.DbActorAck]
+
+    // Message written should appear only in the root channel
```
Ajkunas, line 561: In the lao channel, no ?

be2-scala/src/test/scala/ch/epfl/pop/storage/DbActorSuite.scala
```diff
@@ -493,6 +493,109 @@ class DbActorSuite extends TestKit(ActorSystem("DbActorSuiteActorSystem")) with
...
+    failingMessage should equal(None)
+  }
+
+  test("ReadSetupElectionMessage() reads correctly from /root/lao_id/election_id channels") {
+    // arrange
+    val initialStorage: InMemoryStorage = InMemoryStorage()
+    val dbActor: AskableActorRef = system.actorOf(Props(DbActor(mediatorRef, MessageRegistry(), initialStorage)))
+    val channel = Channel("/root/lao_id/election_id")
+
+    // Write some message using WriteCreateLaoMessage
```
Ajkunas, line 582:
```suggestion
    // Write some message using WriteSetupElectionMessage
```

be2-scala/src/test/scala/ch/epfl/pop/storage/DbActorSuite.scala
```diff
@@ -493,6 +493,109 @@ class DbActorSuite extends TestKit(ActorSystem("DbActorSuiteActorSystem")) with
...
+    val writeAsk = dbActor ? DbActor.WriteSetupElectionMessage(channel, MESSAGE)
+    val writeAnswer = Await.result(writeAsk, duration)
+
+    // assert
+    writeAnswer shouldBe a[DbActor.DbActorAck]
+
+    // Message written should appear only in the root channel
```
Ajkunas, line 589: The comment does not match the following code.

Ajkunas (changes requested):
Some comments to change, othw all good

be2-scala/src/test/scala/ch/epfl/pop/storage/DbActorSuite.scala
K1li4nL: yes...

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandler.scala
```diff
@@ -51,9 +51,8 @@ class ElectionHandler(dbRef: => AskableActorRef) extends MessageHandler {
       electionId: Hash = data.id
       electionChannel: Channel = Channel(s"${rpcMessage.getParamsChannel.channel}${Channel.CHANNEL_SEPARATOR}$electionId")
       keyPair = KeyPair()
-      _ <- dbActor ? DbActor.WriteAndPropagate(rpcMessage.getParamsChannel, message)
       _ <- dbActor ? DbActor.CreateChannel(electionChannel, ObjectType.ELECTION)
```
Ajkunas, line 54: Is this `CreateChannel` necessary ? Since, it seems that you use `createChannel` function inside the `WriteSetupElectionMessage`

Ajkunas (changes requested):
I have a small comment. Otherwise, it is good to go :)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandler.scala
K1li4nL: Good point ! Fixed.

Ajkunas (approved):

PR #1541, opened 2023-05-04
[In progress] Popcha scanner screen

dayan9265:


fe1-web/src/features/popcha/screens/__tests__/PopchaScanner.test.tsx
```diff
@@ -0,0 +1,214 @@
...
+        </FeatureContext.Provider>,
+      );
+      const scannerButton = getByTestId('popcha_scanner_button');
+      fireEvent.press(scannerButton);
+      fireScan(url.toString());
+      await waitFor(() => {
+        expect(mockToastShow).toHaveBeenCalledTimes(0);
+        expect(toJSON()).toMatchSnapshot();
+      });
+    });
```
dayan9265, line 195: This test currently does not work as planned since in the snapshot it seems the scanner is still open, I am not sure why though

fe1-web/src/features/popcha/network/messages/PopchaAuthMsg.ts
```diff
@@ -0,0 +1,67 @@
...
+    this.nonce = msg.nonce;
+
+    if (!msg.client_id) {
+      throw new ProtocolError("Undefined 'client_id' parameter encountered during 'PopchaAuthMsg'");
+    }
+    this.client_id = msg.client_id;
+
+    if (!msg.identifier) {
+      throw new ProtocolError(
+        "Undefined 'identifier_' parameter encountered during 'PopchaAuthMsg'",
```
MeKHell, line 39: "identifier" instead of "identifier_" ?

fe1-web/src/features/popcha/network/messages/__tests__/PopchaAuthMsg.test.ts
```diff
@@ -0,0 +1,111 @@
+import { Hash, ProtocolError } from '../../../../../core/objects';
```
MeKHell, line 1: Fix import path

fe1-web/src/features/popcha/screens/PopchaScanner.tsx
```diff
@@ -0,0 +1,199 @@
...
+    // Check if all required arguments are present
+    for (const arg of requiredArguments) {
+      if (!urlArg.has(arg)) {
+        showErrorMessage(`Missing argument ${arg}`);
+        return false;
+      }
+    }
+
+    // Check if the response respects openid standard
+    if (urlArg.get('response_type') !== 'id_token token') {
```
MeKHell, line 88: We switched to id_token only.

fe1-web/src/features/popcha/screens/PopchaScanner.tsx
```diff
@@ -0,0 +1,199 @@
...
+            if (success) {
+              setTextScanned(data);
+              setShowScanner(false);
+            }
+          })
+        }>
+        <View style={styles.container}>
+          <View>
+            <Text style={Typography.paragraph}>Hello, here is your laoID: {laoId}</Text>
+            <text>{textScanned}</text>
```
MeKHell, line 168: Still needed ? What does it provide ? Or at least having an informal text might be better.

MeKHell (changes requested):
Few comments.

fe1-web/src/features/popcha/screens/PopchaScanner.tsx
```diff
@@ -0,0 +1,199 @@
...
+
+    if (!(urlArg.get('scope')!.includes('openid') && urlArg.get('scope')!.includes('profile'))) {
+      showErrorMessage('Invalid scope');
+      return false;
+    }
+
+    if (urlArg.has('response_mode')) {
+      if (
+        !(
+          urlArg.get('response_mode')!.includes('query') ||
```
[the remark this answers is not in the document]
dayan9265: It's because the `get('response')` could return null, so it is to assert it is non null

dayan9265:
I realized that there were some issues with the communication implementation, I will fix this and reopen the pr once it is done

fe1-web/src/features/popcha/network/PopchaMessageApi.ts
```diff
@@ -0,0 +1,46 @@
...
+ */
+export const sendPopchaAuthRequest = (
+  client_id: Hash,
+  nonce: string,
+  popcha_address: string,
+  state: string | null,
+  response_mode: string | null,
+  laoId: Hash,
+  // TODO: hook currently throwing error
+  generateToken: (laoId: Hash, clientId: Hash | undefined) => Promise<PopToken>,
```
dayan9265, line 29: There is currently a problem with the hook generateToken: I can't use in this function, due the error [Hooks can only be called inside the body of a function component](https://react.dev/warnings/invalid-hook-call-warning)
I am not sure of a nice way to fix this

dayan9265:
There is also an error with the tests since the json file is not yet merged with master, I import it even if it is currently not existing (the json should soon be merged)

fe1-web/src/features/popcha/screens/PopchaScanner.tsx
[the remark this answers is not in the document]
MeKHell: Ok thanks

fe1-web/src/features/popcha/screens/PopchaScanner.tsx
MeKHell: Yes `<Text>` is indeed better.

MeKHell (approved):
One minor comment abov, otherwise LGTM.

fe1-web/src/features/popcha/network/PopchaMessageApi.ts
MeKHell: The best way is to compute the promise in the component and to pass it to the sendPopchaAuthRequest as an arg. But I am not sure if it will work, since you do not have the client_id before scanning the QR Code

fe1-web/src/features/popcha/screens/PopchaScanner.tsx
```diff
@@ -0,0 +1,199 @@
...
+      </QrCodeScanner>
+    </>
+  );
+};
+
+export default PopchaScanner;
+
+export const popchaScannerScreen: PopchaFeature.LaoScreen = {
+  id: STRINGS.navigation_lao_popcha,
+  Icon: makeIcon('scan'),
```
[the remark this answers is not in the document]
dayan9265: True I changed the icon to this:
image

fe1-web/src/features/popcha/network/PopchaMessageApi.ts
dayan9265: Yes that is the problem I have, even if I generate before calling the `sendPopchaAuthRequest`, I still get the same issue. I have some ideas to fix this but I am not too sure it works / it is good practice, I will ask for this on Wednesday

fe1-web/src/features/popcha/network/index.ts
```diff
@@ -0,0 +1,17 @@
...
+import { PopchaAuthMsg } from './messages/PopchaAuthMsg';
+
+export const configureNetwork = (configuration: PopchaConfiguration) => {
+  configuration.messageRegistry.add(
+    ObjectType.POPCHA,
+    ActionType.AUTH,
+    // We cannot subscribe on this channel, so we should not receive any message
+    (_: ProcessableMessage) => {
+      console.log(`Received unprocessed popcha auth message${JSON.stringify(_)}`);
+      return true;
```
Xelowak, line 13: Then would you consider it as an error? If it is never supposed to happen from a security perspective, it may be better to throw an error.

In either way, you can just completely drop the "ProcessableMessage" type since you're not using it:
```suggestion
    // We cannot subscribe on this channel, so we should not receive any message
    () => {
      console.log(`Received unprocessed popcha auth message ${JSON.stringify(_)}`);
      return true;
```

fe1-web/src/features/popcha/network/messages/PopchaAuthMsg.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  public readonly identifier: Hash;
+
+  public readonly identifier_proof: Hash;
+
+  public readonly state: string | undefined;
+
+  public readonly response_mode: string | undefined;
+
+  public readonly popcha_address: string;
```
Xelowak, line 25: It could be nice to add comments to understand what is really contained in the message. It may be obvious for you, but may not be for the people that do not know about the Popcha's design. `state` , `response_mode` or `identifier_proof` are not really self-explained in my opinion.

fe1-web/src/features/popcha/screens/PopchaScanner.tsx
```diff
@@ -0,0 +1,198 @@
...
+          </View>
+        </View>
+      </QrCodeScanner>
+    </>
+  );
+};
+
+export default PopchaScanner;
+
+export const popchaScannerScreen: PopchaFeature.LaoScreen = {
```
Xelowak, line 193:
```suggestion
export const PopchaScannerScreen: PopchaFeature.LaoScreen = {
```
To stay consistent.

fe1-web/src/features/popcha/screens/__tests__/PopchaScanner.test.tsx
Xelowak: Are you sure to close the scanner correctly?
Maybe you could try to mock the QR code closing function and then check that it has been called.

It's quite hard for me to understand what is not working from just reading the code.

Xelowak (commented):
Very nice work! I appreciate the fact that you put effort into testing everything you did.
I made a couple of comments but overall it's lookking good.

As a general testing remark, it could be nice to ensure that the `it` makes sense. It is there to help you make an easy sentence! For example, prefer `it('renders correclty when scanner is closed'), () => { ... })` than  `it('closed scanner renders correctly', () => { ... })`. It also makes more sense when the test fails and this string is displayed.

Your `PopchaScanner` component will be refined if I understood correctly, so don't forget to add all displayed text in `strings.ts` and use the string constants.

fe1-web/src/features/popcha/network/PopchaMessageApi.ts
dayan9265: oh I see that I haven't removed the comment, but now it works fine. It was an issue on how I defined the hook in the popcha hooks, but now that it is fixed I will remove the comment

fe1-web/src/features/popcha/screens/PopchaScanner.tsx
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    if (urlArg.get('login_hint') !== laoId.toString()) {
+      showErrorMessage('Invalid lao id');
+      console.log(`Scanned lao id: ${urlArg.get('login_hint')}, current lao id: ${laoId}`);
+      return false;
+    }
+
+    return true;
+  };
```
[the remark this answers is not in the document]
dayan9265: It is true that some tests are redundant, but I feel having having the variable `requiredArguments` shows clearly what are the expected arguments, and then there's check for some of these variables (but not all). I will try to think of a way to make it more nice, but currently I don't have a nice solution in mind

fe1-web/src/features/popcha/hooks/__tests__/PopchaHooks.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+      const { result } = renderHook(() => PopchaHooks.useGenerateToken(), {
+        wrapper,
+      });
+      await result
+        // call with mock params
+        .current({} as Hash, undefined)
+        .then((token) => expect(token).toEqual(mockPopToken))
+        .catch(assert.fail);
+    });
+  });
```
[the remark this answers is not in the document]
dayan9265: Yes indeed the tests could be more clear / clean, I modified the file to have test similar to the ones in `RollCallHooks.test.tsx`

fe1-web/src/features/popcha/screens/__tests__/PopchaScanner.test.tsx
dayan9265: The problem was that in the snapshot file, I could see that the scanner didn't close, but it seems that the problem was I was not waiting long enough for. So I modified in the following way:

```Typescript
      await waitFor(() => {
        // wait for scanner to close
        expect(getByText(STRINGS.popcha_open_scanner)).toBeTruthy();
        expect(mockToastShow).toHaveBeenCalledTimes(0);
        expect(toJSON()).toMatchSnapshot();
      });
```

fe1-web/src/features/popcha/screens/PopchaScanner.tsx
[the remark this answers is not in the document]
Xelowak: Anyway, I won't block this PR for this. If you have an idea that comes up, don't hesitate to refactor!

Xelowak (approved):
I a tiny comment, but feel free to merge if you want it ready.

fe1-web/src/features/popcha/screens/__tests__/PopchaScanner.test.tsx
```diff
@@ -0,0 +1,221 @@
...
+        <FeatureContext.Provider value={contextValue}>
+          <PopchaScanner />
+        </FeatureContext.Provider>,
+      );
+      const scannerButton = getByTestId('popcha_scanner_button');
+      fireEvent.press(scannerButton);
+      fireScan(url.toString());
+      await waitFor(() => {
+        // wait for scanner to close
+        expect(getByText(STRINGS.popcha_open_scanner)).toBeTruthy();
```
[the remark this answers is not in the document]
dayan9265: I used `getByText` because I want to wait for the "open scanner" to appear, I think a `getByTestId` wouldn't work since the button exists when the scanner is open (that was my initial issue with this test, I was only getting a snapshot test and it would take it with the scanner still open -> wouldn't wait for it to close)

PR #1549, opened 2023-05-05
Add connection retries and config file for starting the server

### LINKED ISSUE #1539: [BUG] Multiserver connection fails when the first server hasn't started yet
### Description 
Considering two servers:
- Server 1 
- Server 2 which has the address of server 1 (and possibly other servers) as a parameter
Server 2 will try to connect to server 1 once and if that fails, it will also fail (on go).  

### Expected behavior
It has yet to be discussed and defined by the protocol but we can look into implementing retries.

### How to reproduce
1. Start a server with parameters for other servers' addresses without starting them 
2. See error `failed to connect to server: failed to dial to ws://address/server`

##### Front-ends:
- [X] Not applicable

##### Back-ends:
- [X] Be1-Go
- [ ] Be2-Scala

### Workaround
You can make sure that server 1 has already started before starting server 2. 

### Impact
For now, there is a delay in the deployed versions that makes sure that server 1 is started before server 2 but from a usability perspective it is not ideal to have a server crash because of this reason.  

### Possible root cause
On the go backend, in `cli/mod.go`, the server simply has a loop that iterates over the other servers' addressed and tries to connect to them then fails directly if one of them doesn't succeed.

MariemBaccari:
This PR: 
- Adds the possibility to use a json config file to start the server using the flag `-cf` which overrides all other flag if it is present. If it is not present, the server will start using the usual flags from the command line.
- Allows adding other servers, if we are using a config file, after the server is started since we listen to changes using `fsnotify`. The server will try to connect to them as soon as they are added.
- Implements a exponential backoff strategy for connecting to the servers to fix the bootstrapping issue ( Closes #1539 ).
- Adds this new starting option to the documentation 

I added two config files at the root of the project that have the same settings as the default flag values that can be locally modified. I also added them to the gitignore for that reason. 

The PR doesn't take into account servers being removed from the config file during execution or servers shutting down (and possibly reconnecting) during execution. That will be handled in another PR.

(PS: the small modification in the `socket.go` file isn't related to what I am implementing. It just fixes an error that staticcheck raises for deprecated methods.)


Nazianzenov (dismissed):
Looks good to me. I didn't check very deeply the json config files. Go code is clean.

be1-go/cli/mod.go
```diff
@@ -110,6 +161,59 @@ func Serve(cliCtx *cli.Context) error {
 	return nil
 }
 
+// serverConnectionLoop tries to connect to the remote servers following an exponential backoff strategy
+// it also listens for updates in the other-servers field and tries to connect to the new servers
+func serverConnectionLoop(h hub.Hub, wg *sync.WaitGroup, done chan struct{}, otherServers []string, updatedServersChan chan []string, connectedServers *map[string]bool) {
```
Anghul, line 166: This function needs a way to stop it when the server is closed, for example via a close channel.

be1-go/cli/mod.go
```diff
@@ -20,34 +23,61 @@ import (
...
-	if clientPort == serverPort {
-		return xerrors.Errorf("client and server ports must be different")
-	}
-	otherServers := cliCtx.StringSlice("other-servers")
+	startedWithConfigFile := false
 
-	pk := cliCtx.String("public-key")
+	// by default, start using a config file
+	serverConfig, err = startWithConfigFile(configFilePath)
+	if err != nil {
```
Anghul, line 65: I personnaly would just fail the start of the server if there are any error while reading the config file, and use flags only if there is no config file set.
But if you think your way is best, you don't need to change it.

Anghul (changes requested):
Two small comments, but good job.

be1-go/cli/mod.go
MariemBaccari: I think your idea is better indeed. It should fail fast in this case. I pushed a commit 30 minutes ago to address your comments. It shows on the branch but not the PR ... (probably because of github issues). I'll re-request a review when everything is back to normal.

Anghul (approved):
LGTM, good job

PR #1551, opened 2023-05-05
Implement witnessing pt.1

matteosz:
This PR:
- introduces in the application the possibility for the clients to show the public key in form of QR code in order to be scanned as witnesses by the organizer
- adds the possibility for the organizer to scan witnesses before creating the LAO, such that witnesses are statically defined at the start and consensus is not required

Some UI screenshots:


simone-kalbermatter (changes requested):
The code looks good to me, just a comment about the UI. On the my Lao screen, the witness button should probably not be more highlighted than the join button. Now it is the only blue button and also has the QR code icon. It would probably be better if both buttons are blue. Maybe the QR code icon could be left away completely or be given to both buttons, but I don't really see a case for why one button should have it and the other not. If there is not enough space on the screen horizontally, the buttons could also be aligned vertically.

matteosz:
> The code looks good to me, just a comment about the UI. On the my Lao screen, the witness button should probably not be more highlighted than the join button. Now it is the only blue button and also has the QR code icon. It would probably be better if both buttons are blue. Maybe the QR code icon could be left away completely or be given to both buttons, but I don't really see a case for why one button should have it and the other not. If there is not enough space on the screen horizontally, the buttons could also be aligned vertically.

Just changed as you suggested, you can find an updated screenshot in the description

simone-kalbermatter (dismissed):
LGTM!

Warhedgehog (commented):
Code wise it looks great to me. For the UI, I'm gonna be against Simone's opinion. For buttons, a principle is to have only one high-emphasis button, unless there's a really good reason. Here, joining an LAO will be much more common than creating one. So I think the `Create` button should not be blue.

jbsv:
Please don't forget to reduce the code smells. Most of them, if not all, are easy to fix ;-)

matteosz:
> Code wise it looks great to me. For the UI, I'm gonna be against Simone's opinion. For buttons, a principle is to have only one high-emphasis button, unless there's a really good reason. Here, joining an LAO will be much more common than creating one. So I think the `Create` button should not be blue.

👍🏼 I changed the UI and updated the screenshot in the description

Warhedgehog (approved):
LGTM 👍

Issue #1134, opened 2022-06-21
Enable lao#greet check


Tyratox:
At the moment the authenticity check for `lao#greet` messages are disabled which means any server can claim to be the back-end of any front-end. As long as there is only one single organizer server this is fine but as soon as there are multiple back-ends in a single LAO, this check must be enabled. Otherwise, malicious organizers or witnesses can claim to be the back-end of a the organizer's front-end enabling them to send forged `election#result` or `election#key` messages.

For the check to work, both front-ends has to (automatically) sign `lao#greet` messages sent by **their** back-end that list them as their front-end. This requires witness signatures to work and must allow attendees other than witnesses to submit them for `lao#greet` messages. After that, the checks have to be implemented in the front-ends.

For fe-1 the check is already implemented [here](https://github.com/dedis/popstellar/blob/master/fe1-web/src/features/lao/network/LaoHandler.ts#L93-L106), it just has to be enabled.

PR #1554, opened 2023-05-07
Add server greet schema

MariemBaccari:
This PR aims to add a new json schema that allows servers to share their endpoints and public key to each other.
It's a new query `greet_server` that contains the server's public key, the address at which clients can connect and the one at which servers can connect. 
It should be first sent by the server that's making the connection. The server that receives it should send a `greet_server` back

This will allow us to make the greet lao work since we're missing the peers' addresses.


docs/protocol.md
```diff
@@ -204,6 +209,32 @@ Here are the different methods that can be called:
...
 * GetMessagesById
 
+### Greeting a server
+🧭 **RPC Message** > **Query** > **Greet Server**
+
+By executing a greet server action, a server can send its public key as well as the 
+addresses of its client and server endpoints to another server.
+
+This action should be executed by the server that opens a connection with another server. The 
+latter should then reply with a `GreetServer` answer containing its own information.
```
K1li4nL, line 219: Now that I think about it, shouldn't the answer rather be in a `result`, otherwise they might start sending each other `GreetServer`'s until the end of time ?

protocol/examples/query/greet_server/wrong_greet_server_invalid_client_endpoint_address.json
```diff
@@ -0,0 +1,10 @@
+{
+  "jsonrpc": "2.0",
+  "id": 7,
+  "method": "greet_server",
+  "params": {
+    "public_key": "J9fBzJV70Jk5c-i3277Uq4CmeL4t53WDfUghaK0HpeM=",
+    "client_address": "wss://popdemo.dedis.ch/client",
```
K1li4nL, line 7: I don't think this endpoint is invalid in the sense that the url has the correct shape to me. Urls on the default ports should be accepted I think, I see that they made the same choice on the `greet_lao`, is there a reason for it ? 

Though, by accepting both default and specified port it might make the regex a bit uglier

protocol/examples/query/greet_server/wrong_greet_server_invalid_server_endpoint_address.json
```diff
@@ -0,0 +1,10 @@
+{
+  "jsonrpc": "2.0",
+  "id": 7,
+  "method": "greet_server",
+  "params": {
+    "public_key": "J9fBzJV70Jk5c-i3277Uq4CmeL4t53WDfUghaK0HpeM=",
+    "client_address": "wss://popdemo.dedis.ch:9000/client",
+    "server_address": "wss://popdemo.dedis.ch/server"
```
K1li4nL, line 8: same as above

K1li4nL (changes requested):
Beside these few points it looks good to me, good job !

docs/protocol.md
MariemBaccari: I might be wrong about the scala backend (so correct me if I am) but on the go side the implementation would be much easier if the answer was a query because adding a new type of `result` would be a mess since it expects an array of messages or a map between a channel and an array of messages for now. We already keep track of the queries sent so when we receive one we just check if we sent a query with this `id `and either send a `greet_server `or we don't depending on that condition.

protocol/examples/query/greet_server/wrong_greet_server_invalid_client_endpoint_address.json
MariemBaccari: I also wasn't (and still am not) sure about why it was the case for the greet_lao. @pierluca could you perhaps clarify that point ?

docs/protocol.md
K1li4nL: I don't think we keep track of any queries on scala, I guess we will start now ;)

pierluca: I am a bit confused as to your reasoning here. Are you sure what you want is not a JSON-RPC Notification ?

It seems your message is not really an RPC, you're not calling a function and expecting a return value.

You do expect "an answer" (from a logic perspective), but the "connected-to" server would already know what to send in his own `GreetServer` - and could, in principle, send it of his own accord without breaking the protocol.

Basically, it's a bit like the heartbeat and get_messages_by_id you're familiar with. If you know the application protocol, you know that a heartbeat may trigger a get_messages_by_id, but one is not the answer to the other (at the JSON-RPC level), 

Also, you don't have a choice from a JSON-RPC perspective, if you do a query, you expect an answer object which contains a `result` field. Otherwise, you're breaking the JSON-RPC protocol. The good news (for both of you) is that it seems you don't really need a query at the JSON-RPC layer 😅

docs/protocol.md
```diff
@@ -204,6 +209,32 @@ Here are the different methods that can be called:
...
+    "id": 7,
+    "method": "greet_server",
+    "params": {
+        "public_key": "J9fBzJV70Jk5c-i3277Uq4CmeL4t53WDfUghaK0HpeM=",
+        "client_address": "wss://popdemo.dedis.ch:9000/client",
+        "server_address": "wss://popdemo.dedis.ch:9001/server"
+    }
+}
+```
+
```
pierluca, line 237: don't forget to include the 
```
<details>
<summary>
💡 See the full specification
</summary>
...
```

bit :-)

protocol/examples/query/greet_server/wrong_greet_server_invalid_client_endpoint_address.json
pierluca: I agree, default ports should be acceptable and I suppose in the greet_lao it was a question of convenience at the time, or the default port scenario might not have been spotted.

Thankfully marking something as optional in a regex only makes it marginally uglier 😅

protocol/query/method/greet_server.json
```diff
@@ -0,0 +1,46 @@
...
+        },
+        "server_address": {
+          "description": "[String] canonical address of the server with a protocol prefix and the port number for connecting as a server",
+          "type": "string",
+          "pattern": "^.*:\\/\\/.*:\\d{0,5}\\/.*$"
+        },
+        "client_address": {
+          "description": "[String] canonical address of the server with a protocol prefix and the port number for connecting as a client",
+          "type": "string",
+          "pattern": "^.*:\\/\\/.*:\\d{0,5}\\/.*$"
```
pierluca, line 30: are you sure you don't want to enforce `ws` and `wss` URI schemes only ?

docs/protocol.md
MariemBaccari: I think you're right, I wasn't aware that by doing a query we were systematically expecting an `answer` object. 
If I understood properly, a JSON-RPC Notification would simply be this same schema but without an ID field correct ? In that case , I can remove it and specify in the documentation that a `GreetServer` will trigger another `GreetServer` (if we didn't already send one to that server).

protocol/query/method/greet_server.json
MariemBaccari: Yes, thank you I added that.

protocol/examples/query/greet_server/wrong_greet_server_invalid_client_endpoint_address.json
MariemBaccari: I added them as optional now but we're still working on the regex to enforce the number of digits in the port number.

docs/messageData.md
```diff
@@ -466,9 +466,9 @@ Last but not least, the greeting message contains a list of peers that tells cli
             "$comment": "Note: the string is encoded in Base64"
         },
         "address": {
-            "description": "Canonical address of the server with a protocol prefix and the port number",
+            "description": "Canonical address of the server with a protocol prefix and (optionally) the port number",
             "type": "string",
-            "pattern": "^.*:\\/\\/.*:\\d{0,5}\\/.*$"
+            "pattern": "^(ws|wss):(\\/\\/)[a-zA-Z0-9|.|-|_)+\\.[a-zA-Z]+((:[0-9]{1,5}(\\/.*)?)|(\\/).*|)$"
```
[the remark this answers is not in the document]
K1li4nL: I don't know if you copy pasted the regex, but if you did and it doesn't work, it is expect. In a json string we have to double escape so that `\/`  is treated as whole in the string instead of being interpreted as an escaped `/` , or something along these lines. Hence the double escape have to be converted to single escape on website and such. 
The CI currently fails for another reason. The regex I copied from the website I was testing it on came with some funny invisible characters....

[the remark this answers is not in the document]
matteosz: Yeah I didn't consider the double escape problem of the json string, so it looks good then 👍

K1li4nL (approved):
LGTM !

Tyratox (commented):
Happy to see the that the lao greet message is getting closer to becoming fully operational. Might be worth to keep https://github.com/dedis/popstellar/issues/1134 in mind as well. It is not important for this PR but in general if you are now working on the server to server communication.

I agree with @pierluca 's comments regarding the JSON-RPC query / notification :) 
Moreover I added two small comments. Otherwise LGTM 😄

docs/protocol.md
```diff
@@ -204,6 +209,81 @@ Here are the different methods that can be called:
...
+RPC
+
+```json5
+// ../protocol/examples/query/greet_server/greet_server.json
+
+{
+    "jsonrpc": "2.0",
+    "method": "greet_server",
+    "params": {
+        "public_key": "J9fBzJV70Jk5c-i3277Uq4CmeL4t53WDfUghaK0HpeM=",
```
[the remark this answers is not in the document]
MariemBaccari: I see that the sender field is in messages that are related to the LAO. However, there might be a few cases where we would need the public key of the other servers before a LAO is created and after the moment a connection is established. One example from the discussions during the decentralized communications meeting was (in the case where two servers open a connection which each other instead of only one) deciding which connection to drop could depend on the public key. What do you think ?

docs/messageData.md
```diff
@@ -466,9 +466,9 @@ Last but not least, the greeting message contains a list of peers that tells cli
             "$comment": "Note: the string is encoded in Base64"
         },
         "address": {
-            "description": "Canonical address of the server with a protocol prefix and the port number",
+            "description": "Canonical address of the server with a protocol prefix and (optionally) the port number",
```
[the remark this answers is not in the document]
MariemBaccari: Thank you, I just pushed that. Let me know if it's okay now :)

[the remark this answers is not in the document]
Tyratox: I think it would have also been fine just in the `messageData.md` as text but this is even more precise now 👍 :)

docs/protocol.md
```diff
@@ -11,6 +11,7 @@
   - [Concatenation for hashing](#concatenation-for-hashing)
 - [JSON RPC (low-level)](#json-rpc-low-level)
   - [Query](#query)
+    - [Greeting a server](#greeting-a-server)
```
Tyratox, line 14: Given the discussion with @pierluca this should no longer be in the query section I assume?

docs/protocol.md
```diff
@@ -204,6 +209,81 @@ Here are the different methods that can be called:
...
 * GetMessagesById
 
+### Greeting a server
+🧭 **RPC Message** > **Query** > **Greet Server**
+
+By sending a greet server message, a server can transmit its public key as well as the 
+addresses of its client and server endpoints to another server.
+
+This message should be sent by the server that opens a connection with another server. It should trigger
+a `GreetServer` message from the server that receives the connection containing its own information.
```
Tyratox, line 219: Where is this message sent (i.e. on what channel, the default one?)

docs/protocol.md
```diff
@@ -204,6 +209,81 @@ Here are the different methods that can be called:
...
+### Greeting a server
+🧭 **RPC Message** > **Query** > **Greet Server**
+
+By sending a greet server message, a server can transmit its public key as well as the 
+addresses of its client and server endpoints to another server.
+
+This message should be sent by the server that opens a connection with another server. It should trigger
+a `GreetServer` message from the server that receives the connection containing its own information. 
+
+RPC
```
Tyratox, line 221: Typo?:)

docs/protocol.md
MariemBaccari: It's not sent to any channel similarly to the `heartbeat `and `getMessagesById` messages.

docs/protocol.md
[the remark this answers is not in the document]
MariemBaccari: They're not sent on any channel so I think in that sense they're not using the pub/sub model and, if I understood the documentation properly, they shouldn't contain that field. We have other messages that operate in the same way (heartbeats and getMessagesById which I mentionned in another comment).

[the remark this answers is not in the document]
Tyratox: I see, that explains my confusion!
Yes I guess then my comments don‘t make sense / you must include the public key.

But I did not realize these messages are completely different from all the others. I would reccomend putting them in a completely separate section in the documentation to make this clear. (including the heartbeat and getMessagesById)

docs/protocol.md
Tyratox: Yes you‘re right, the structure is not ideal here .. 🤔 As I have written above I would put heartbeat somehwere else because it and the other server-server messages are wildely different. And yes you are right, broadcast should probably also be in a separate notification section..? (and similarly in the protocol there should be a notification.json) Or do you think it is fine as it is?

I think I need to quickly chat with pierluca / the other TAs :)

MariemBaccari: In terms of correctness, you're right I think the heartbeats, broadcasts and greet server messages should be in separate notification.json. I think that changing this now would require the subsystems (at least the go backend from my pov) to change some things in the way they implement these. 

What I suggest is keeping it like this for now so that we can start with fixing the lao properly and perhaps opening an issue (if with pierluca and the other TAs you decided that it should indeed be in a notification.json) then do the refactoring in another PR because it will likely break multiple CI jobs. 

If you agree with this, I can still put them apart in the documentation in the meanwhile.

docs/protocol.md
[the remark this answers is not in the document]
Tyratox: Sorry I was confused, I think I now unterstand!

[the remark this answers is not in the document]
Tyratox: Since there is now payload, there won‘t be any mid-level protocol of course. Sorry, not sure what my mind was doing.

docs/protocol.md
Tyratox: Yes, that was me being dumb sorry!

docs/protocol.md
Tyratox: Oh and I got your reply a bit delayed.

Yes that sounds reasonable! 😄

MariemBaccari: Putting them in a Notification section would still imply creating a `../protocol/notification/notification.json` file tho right ? Or are you just talking about the documentation in `protocol.md` ? I'm sorry, it's a bit confusing ...

MariemBaccari: Sorry I also got your reply (about getting my reply delayed) ... delayed. Is the PR okay now ?

Tyratox: I liked your suggestion of at least separating them in the documentation for now but yes, the PR is fine for me 😄

Tyratox (approved):
Formal approval as well 😄
Sorry that I took so long to understand 😂

PR #1555, opened 2023-05-08
Implement social media reactions

matteosz:
This PR introduces the reactions for social media also in FE2. 
The three possible reactions as in FE1 are: 👍🏼 👎🏼 ❤️ 
As in FE1: 
- 👍🏼 and 👎🏼 are mutually exclusive
- ❤️ reaction is indepedent from 👍🏼 👎🏼 (so it's not exclusive)

The reactions are sent to the channel social/reactions, which the client subscribes to after the close roll call.
The client has the visibility on the reactions and can interact with them only if it has a valid pop token.

Recording of the UI behaviour:

[recording](https://user-images.githubusercontent.com/83547952/236790052-fc6174cf-436e-483c-9804-65ee157aa08f.webm)


simone-kalbermatter (changes requested):
Looks very good to me, code is clean, well documented and tested! Just a detail for the UI: could you change the colors of the icons to match the blue app theme (or maybe black) please?

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/data/ReactionHandler.java
```diff
@@ -0,0 +1,70 @@
...
+  private final LAORepository laoRepo;
+  private final SocialMediaRepository socialMediaRepo;
+
+  @Inject
+  public ReactionHandler(LAORepository laoRepo, SocialMediaRepository socialMediaRepo) {
+    this.laoRepo = laoRepo;
+    this.socialMediaRepo = socialMediaRepo;
+  }
+
+  public void handleAddReaction(HandlerContext context, AddReaction addReaction)
```
simone-kalbermatter, line 33: I just noticed that documenation is missing here, also for `handleDeleteReaction`

matteosz:
> Looks very good to me, code is clean, well documented and tested! Just a detail for the UI: could you change the colors of the icons to match the blue app theme (or maybe black) please?

Sure, I updated the UI using the app theme blue color. 
Here there are some screenshots:

Warhedgehog (approved):
Great work! I've seen no point needing improvement 👍

simone-kalbermatter (approved):
LGTM!

PR #1558, opened 2023-05-09
Add qrcode generation for popcha challenge in be2-scala

Jharaxus:
The result obtained can be observed by adding
```scala
println(f"Example of authentication request: \n" +
  s"http://${config.interface}:${config.port}/${config.authenticationPath}?" +
  "response_type=id_token%20token&client_id=10&redirect_uri=https%3A%2F%2Fclient.example.org%2Fcb" +
  "&scope=openid%20profile&state=abcd&nonce=42&login_hint=42")
```
in `Server.scala` line 99 and going on the link printed in the terminal.


K1li4nL (changes requested):
Beside the small library issue and the question it looks good to me, well done ! 
I thought for a bit about testing qrcode generation but I haven't got much ideas so... I suppose manual testing is enough

be2-scala/src/main/scala/ch/epfl/pop/authentication/QRCodeChallengeGenerator.scala
```diff
@@ -0,0 +1,76 @@
...
+
+    HttpEntity(ContentTypes.`text/html(UTF-8)`, challengePage)
+  }
+
+  private def fromMatrixToHTML(matrix: ByteMatrix): String = {
+    if (matrix.getWidth != matrix.getHeight)
+      return "invalid qrcode matrix"
+
+    val svgHtmlOpenTag =
+      s"<svg width=\"$qrcodeTotalSize\" height=\"$qrcodeTotalSize\"\n" +
```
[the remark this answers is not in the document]
Jharaxus: Yes they are hard-coded and do not scale with the screen size. I'll see with the web front-end team if they have any idea regarding how to improve that, but this is not the final version of the page template anyway so it should be fixed later on.

be2-scala/src/main/scala/ch/epfl/pop/authentication/QRCodeChallengeGenerator.scala
```diff
@@ -0,0 +1,76 @@
...
+    */
+  def generateChallengeContent(content: String): ResponseEntity = {
+    val encodedContent = Encoder.encode(content, ErrorCorrectionLevel.H)
+    val htmlQRCode = fromMatrixToHTML(encodedContent.getMatrix)
+
+    val templateFile = Source.fromFile(templateFileName)
+    val lines = for {
+      line <- templateFile.getLines()
+      substitutedLine = line.replace(webTemplateQRCodeTemplate, htmlQRCode)
+    } yield substitutedLine
```
K1li4nL, line 34: This should be my last comment before approval, I believe you forgot to close the stream

K1li4nL (changes requested):

be2-scala/src/main/scala/ch/epfl/pop/authentication/QRCodeChallengeGenerator.scala
Jharaxus: fixed !

K1li4nL (approved):
LGTM !

Ajkunas (approved):
Clean code !

PR #1559, opened 2023-05-10
Add greet server and peers to greet lao

MariemBaccari:
This PR implements the greet server on the go backend and adds the peers addresses' to the greet lao.

When a server opens a connection with another server, it sends a greet server message to it containing its info (public key, client and server full addresses). When a server receives a greet server message, it either greets back or doesn't do anything if it already greeted said server. 

It also adds the possibility of specifying the client and server addresses instead of computing them directly so that the correct ones (which include ws:// or wss://) are displayed in the greet lao and greet server messages. It can also be convenient during deployment. If they are not provided, "ws://" is assumed which is convenient when using the servers locally.

There's a high number of files that were modified but it's mainly because I needed to add an argument and a function to the Hub which allows passing the client and server addresses directly to the hub but required modifying all the `fakeHub` instances in the test classes.


be1-go/hub/standard_hub/message_handling.go
```diff
@@ -494,6 +494,32 @@ func (h *Hub) handleGetMessagesById(socket socket.Socket,
...
+	var greetServer method.GreetServer
+
+	err := json.Unmarshal(byteMessage, &greetServer)
+	if err != nil {
+		return xerrors.Errorf("failed to unmarshal greetServer message: %v", err)
+	}
+
+	h.Lock()
+	// store information about the server
+	h.peersInfo[socket.ID()] = greetServer.Params
```
Nazianzenov, line 507: Could be good to add methods for hub, like "AddPeersInfo", to avoid locking and unlocking every time.

be1-go/hub/standard_hub/mod.go
```diff
@@ -568,6 +617,18 @@ func (h *Hub) NotifyNewChannel(channelID string, channel channel.Channel, sock s
 	h.Unlock()
 }
 
+func (h *Hub) GetPeersInfo() []method.ServerInfo {
```
Nazianzenov, line 620: Good method!

Nazianzenov (approved):
Looks good to me. Given  my comments, i would suggest to be careful with locks, and try to have helper methods to use them in, in order to avoid having locks everywhere and potentially forgetting one. But otherwise LGTM, clean code!

Anghul (approved):
Nothing to add to the comments already done, good jub

PR #1560, opened 2023-05-10
Refactoring of the auth server

Nazianzenov:
* changing the signature of the constructor, adding the path of the html template as parameter, removing the endpoint parameter.
* refactoring of the constant, making them private
* adding comments for clarity
* adding a new parameter "response mode", optional
* testing the validity of this parameter


Anghul (approved):
LGTM

PR #1568, opened 2023-05-16
Fix witness signature checking and add pending signatures 

MariemBaccari:
This PR aims to fix the witness signature checking by adding the decoding operation before the check. 
It also adds the possibility of accepting signatures for messages we haven't yet received and adding them when they arrive. For that purpose, it modifies the hub interface for the channels so that we can notify the hub when we receive a new signature. That way, it can also be added to the hub inbox which is used for sending the heartbeats.


Nazianzenov (dismissed):
Looks good to me. Besides the refactoring or the NotifyWitnessMessage, the test and pending signature code looks clean.

Anghul (dismissed):
LGTM

Anghul (approved):

PR #1569, opened 2023-05-16
Remove the cash, and notifications from the fe1 home screen

MeKHell:
This very tiny PR will remove the digital cash and notification features from the home screen. 
The notification is currently only used by witnessing, therefor it is not useful and part of release.
image

### DIFF
--- a/fe1-web/src/features/index.ts
+++ b/fe1-web/src/features/index.ts
@@ -22,7 +22,6 @@ export function configureFeatures() {
   // configure features
   const eventConfiguration = events.configure();
   const walletConfiguration = wallet.configure();
-  const digitalCashConfiguration = digitalCash.configure();
 
   const notificationConfiguration = notification.configure();
   const laoConfiguration = lao.configure({ registry: messageRegistry });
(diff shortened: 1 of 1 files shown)


fe1-web/src/features/index.ts
```diff
@@ -22,7 +22,6 @@ export function configureFeatures() {
   // configure features
   const eventConfiguration = events.configure();
   const walletConfiguration = wallet.configure();
-  const digitalCashConfiguration = digitalCash.configure();
 
   const notificationConfiguration = notification.configure();
```
dayan9265, line 26: Is there a reason to have kept the configuration of the notification?

MeKHell: I removed only the necessary code and the one that was not used anymore. digitalCashConfiguration is still used even id it is not necessary

Tyratox: Ah I see, `digitalCashConfiguration` was not used anywhere else 👍. I might consider leaving `digitalCash.configure();` without assigning it to a variable so that the front-end doesn't break if unknown messages are received but this is up to you.

Tyratox (approved):
LGTM:)

PR #1570, opened 2023-05-16
Add config file watch service

K1li4nL:
This pr adds the ability for the system to react on the fly to changes in the config files added in #1519. 

Limitations: Without greet_server implementation we are still susceptible to having double connections (A to B and B to A, over two websockets).

### DIFF
--- a/be2-scala/src/main/scala/ch/epfl/pop/decentralized/Monitor.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/decentralized/Monitor.scala
@@ -115,8 +115,8 @@ private class FileMonitor(mediatorRef: ActorRef) extends Runnable {
           if (serverPeersListPath.endsWith(event.context().toString)) {
             mediatorRef ! ConnectionMediator.ConnectTo(readServerPeers())
           }
-          watchKey.reset()
         }
+        watchKey.reset()
       }
     } catch {
       case _: InterruptedException =>
--- a/be2-scala/src/main/scala/ch/epfl/pop/decentralized/Monitor.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/decentralized/Monitor.scala
@@ -4,12 +4,16 @@ import akka.NotUsed
 import akka.actor.{Actor, ActorLogging, ActorRef, Props, Timers}
 import akka.event.LoggingReceive
 import akka.stream.scaladsl.Sink
+import ch.epfl.pop.config.RuntimeEnvironment.{readServerPeers, serverPeersListPath}
 import ch.epfl.pop.decentralized.Monitor.TriggerHeartbeat
 import ch.epfl.pop.model.network.JsonRpcRequest
 import ch.epfl.pop.model.network.method.ParamsWithMap
 import ch.epfl.pop.pubsub.graph.GraphMessage
 
+import java.nio.file.{Path, WatchService}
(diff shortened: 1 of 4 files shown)


be2-scala/src/main/scala/ch/epfl/pop/decentralized/Monitor.scala
```diff
@@ -86,3 +94,35 @@ object Monitor {
...
+      while (!Thread.currentThread().isInterrupted) {
+        // Blocks until an event happen
+        val watchKey = watchService.take()
+
+        // For any event, read the file and send it
+        for (event <- watchKey.pollEvents().asScala.toList) {
+          if (serverPeersListPath.endsWith(event.context().toString)) {
+            mediatorRef ! ConnectionMediator.ConnectTo(readServerPeers())
+          }
+          watchKey.reset()
```
Jharaxus, line 118: Do you need to do that? The events are already fetched and the watchKey will be deleted by the end of this while loop's scope anyway...

be2-scala/src/test/scala/ch/epfl/pop/decentralized/MonitorSuite.scala
```diff
@@ -99,4 +104,40 @@ class MonitorSuite extends TestKit(ActorSystem("MonitorSuiteActorSystem")) with
...
+    mockConnectionMediator.send(monitorRef, ConnectionMediator.Ping())
+
+    // Expect no message as long as the server peers list is untouched
+    mockConnectionMediator.expectNoMessage(timeout)
+
+    val newContent = List("some", "strings")
+    testWriteToServerPeersConfig(newContent)
+
+    mockConnectionMediator.expectMsgType[ConnectionMediator.ConnectTo](timeout)
+    deleteTestConfig()
```
Jharaxus, line 122: I don't think the test config file will be deleted if the test failed in the previous assertion

Jharaxus (changes requested):
Good job, the tests of `FileMonitor` are very nice !

be2-scala/src/main/scala/ch/epfl/pop/decentralized/Monitor.scala
K1li4nL: As I understood it, this line is mandatory to put the key back from `signalled` into the `ready` state. Even if the variable holding the ref goes out of scope, the key it points to seems to be living as long as the watch service is not cancelled. I am very not comfortable with this... Here is the doc:

be2-scala/src/test/scala/ch/epfl/pop/decentralized/MonitorSuite.scala
K1li4nL: Fixed !

be2-scala/src/main/scala/ch/epfl/pop/decentralized/Monitor.scala
Jharaxus: Ok that makes more sense now!
However I think this should be moved out of the `for` loop then

K1li4nL: I hadn't seen that I think... yes indeed

Jharaxus (approved):

Ajkunas (approved):
LGTM

Issue #1374, opened 2023-02-13
LAO Object handles features that it should not


GabrielFleischer:
The LAO object has too many responsibilities, and they should be decoupled from it.

The remaining features left in the Lao class are :
 - Witnesses
 - Consensus

matteosz:
The witnessing part has been removed, the only feature left now is consensus

PR #1574, opened 2023-05-17
Handle witness signatures

simone-kalbermatter:
This PR adds a handler for witness signature messages. Witness messages are now ordered by newest on top to oldest on the bottom. After signing a witness message, the sign button is disabled and shows the text _signed_. The witness signatures are now visible in the signatures dropdown menu.

I also changed the description of the election setup witness message to display all election questions.


fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/data/WitnessingHandler.java
```diff
@@ -0,0 +1,62 @@
...
+    PublicKey witnessPk = context.getSenderPk();
+
+    // Check that the sender of the message is a witness
+    if (!laoView.getWitnesses().contains(witnessPk)) {
+      throw new InvalidWitnessException(witnessMessageSignature, witnessPk);
+    }
+
+    // Check that a witness message with the given message id exists in the lao
+    MessageID messageID = witnessMessageSignature.getMessageId();
+    WitnessMessage witnessMessage = laoView.getWitnessMessages().get(messageID);
```
matteosz, line 45: As specified in the issue #1374 the LAO should not handle the witness messages, so a custom repository would be far better, but it's your choice whether to postpone it for a future PR or do it now

matteosz (changes requested):
Very good work! Just few suggestions

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/witness/WitnessMessageListViewAdapter.java
```diff
@@ -111,8 +112,20 @@ public View getView(int position, View convertView, ViewGroup parent) {
...
+      Set<PublicKey> witnessSignatures = witnessMessage.getWitnesses();
+
+      if (witnessSignatures.contains(laoViewModel.getPublicKey())) {
+        // The user already signed the message
+        binding.signMessageButton.setEnabled(false);
+        binding.signMessageButton.setText("Signed");
+      } else {
+        binding.signMessageButton.setEnabled(true);
+        binding.signMessageButton.setText("Sign");
+        View.OnClickListener listener =
```
matteosz, line 124: Last thing that I forgot, could you use string resources here? As a general principle each displayed string should be an Android resource

matteosz (changes requested):

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/data/WitnessingHandler.java
simone-kalbermatter: Yes I saw that, but decided to keep the larger refactoring of adding the witness repo for another PR.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/Lao.java
```diff
@@ -77,7 +77,7 @@ public Lao(
     this.organizer = organizer;
     this.modificationId = modificationId;
     this.witnesses = new HashSet<>(witnesses);
-    this.witnessMessages = new HashMap<>(witnessMessages);
+    this.witnessMessages = new LinkedHashMap<>(witnessMessages);
```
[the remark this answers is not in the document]
simone-kalbermatter: The `LinkedHashMap` works for ordering the messages, the problem was caused by dropping signed messages to the bottom of the list. This caused inconsistencies depending on if the witness messages were signed one after another or in a random order. I changed the order now to be fixed, regardless wether the message was signed or not. The oldest messages are now on the bottom, and newest on top.

matteosz (approved):
LGTM!

Warhedgehog (approved):
Great Work, well documented and tested !

PR #1571, opened 2023-05-16
Add persistence layers to the application #2

matteosz:
This PR follows #1528 to address the issue #1383. 

In details, here I implemented a persistence layer for:
- ElectionRepository
- RollcallRepository

Also, there's an improvement fix for some tests and the disposables of the asynchronous calls.

PR #1575, opened 2023-05-18
Add persistence layers to the application #3 

### LINKED ISSUE #1383: Add a persistence layer in the repositories
The current persistence system is straightforward.

It stores the wallet seed, current server address and subscribed channels.
When the app loads data from storage, it will retrieve and process every message.

This is very heavy, and storing the actual state of the app rather than how to get to it is necessary.

The easiest way to do that is to add a persistence layer in the repositories that stores/loads data.

(A caching layer might be a great addition to avoid unnecessary io accesses)

matteosz:
This PR follows #1571 and finally resolves #1383. 

In details, here I implemented a persistence layer for:
- MeetingRepository
- SocialMediaRepository
- DigitalCashRepository

The only missing things to be persisted remain the WitnessMessages and the ConsensusMessages, for which still there's no a repository , so it will be implemented in the future.


simone-kalbermatter (dismissed):
LGTM!

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/serializer/database/JsonChirpSerializer.java
```diff
@@ -0,0 +1,40 @@
+package com.github.dedis.popstellar.model.network.serializer.database;
+
+import com.github.dedis.popstellar.model.objects.Chirp;
+import com.github.dedis.popstellar.model.objects.security.MessageID;
+import com.github.dedis.popstellar.model.objects.security.PublicKey;
+import com.google.gson.*;
+
+import java.lang.reflect.Type;
+
+public class JsonChirpSerializer implements JsonSerializer<Chirp>, JsonDeserializer<Chirp> {
```
GabrielFleischer, line 10: I might be missing something, but couldn't you just use the feature of Gson allowing automated parsing ?

https://sites.google.com/site/gson/gson-user-guide#TOC-Object-Examples

Basically, Gson will extract the json elements directly from the fields of a class.
This is bi-directonal and can be fine-tuned with annotations
https://sites.google.com/site/gson/gson-user-guide#TOC-JSON-Field-Naming-Support

You can also exclude a field by marking it transient.

This comment also applies to the other serializers. Probably not all, but simple objects don't need a dedicated serializer.

GabrielFleischer (changes requested):
Very good job !

I had never seen this library in action and it looks very cool !

I left a small comment : 
I think you can make a better use of Gson's magic and simplify a lot of the parsing code.

This is a bit picky, but I couldn't find anything else 😄

PR #1577, opened 2023-05-18
Fix long chirps being cut off scren

dayan9265:
This PR fixes the problem with long chirp with no spaces and no special characters (such as / f.ex) to not be completely shown onscreen. It also fixes the problem with the trashcan not being gray when offline. Here is how it looks like now:

image


Tyratox:
As mentioned yesterday I think it would be good to move the polyfill to `core/platform`, rename it to something more general & import it in App.tsx. Otherwise LGTM 😄

MeKHell (approved):
LGTM

fe1-web/src/features/social/components/ChirpCard.tsx
```diff
@@ -264,6 +265,7 @@ const ChirpCard = ({ chirp, isFirstItem, isLastItem }: IPropTypes) => {
                     name="delete"
                     testID="delete_chirp"
                     onPress={() => setShowDeleteConfirmation(true)}
+                    disabled={reactionsDisabled.delete}
                     size="small"
```
Tyratox, line 269: Good catch! 😁

Tyratox (approved):
LGTM 😁

PR #1580, opened 2023-05-20
Refactoring Akka graph construction

K1li4nL:
The point of this pr is to prepare the arrival of the new `greet_server` json. The old partitioning did not allow for easy addition of new `JsonRpcRequest`. 
This refactoring helps in two ways: 
- By partitioning  using `MethodType` we can get rid of some boilerplate code in the other handlers
- By renaming `ParamsWithChannelHandler` to `ParamsHandler`, more general handlers can be added  without breaking consistency too much (`greet_server` should fit well in there, especially since it will also take a `clientActorRef` as parameter)

### DIFF
--- a/be2-scala/src/main/scala/ch/epfl/pop/pubsub/PublishSubscribe.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/pubsub/PublishSubscribe.scala
@@ -10,7 +10,7 @@ import ch.epfl.pop.decentralized.Monitor
 import ch.epfl.pop.model.network.MethodType._
 import ch.epfl.pop.model.network.{JsonRpcRequest, JsonRpcResponse}
 import ch.epfl.pop.pubsub.graph._
-import ch.epfl.pop.pubsub.graph.handlers.{GetMessagesByIdResponseHandler, ParamsWithChannelHandler, ParamsWithMapHandler, ParamsWithMessageHandler}
+import ch.epfl.pop.pubsub.graph.handlers.{GetMessagesByIdResponseHandler, ParamsHandler, ParamsWithMapHandler, ParamsWithMessageHandler}
 
 object PublishSubscribe {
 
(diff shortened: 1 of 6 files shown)


be2-scala/src/main/scala/ch/epfl/pop/pubsub/PublishSubscribe.scala
```diff
@@ -95,9 +96,12 @@ object PublishSubscribe {
           /* partitioner port numbers */
           val portPipelineError = 0
           val portParamsWithMessage = 1
```
Jharaxus, line 98: nitpick: Maybe you could add separate ports for the remaining Method types to have more consistency in the ports, and then simply do `val portPublish = portParamsWithMessage ` and `val portBroadcast = portParamsWithMessage` to use them later in the partition in place of `portParamsWithMessage`

Jharaxus (changes requested):
Great `flatten` of the existing graph 🔥

be2-scala/src/main/scala/ch/epfl/pop/pubsub/PublishSubscribe.scala
K1li4nL: I tried doing that but it seems to add inconsistency later on when gluing the components together since we have to use `portParamsWithMessage` anyway (line 141). It would mean that `portBroadcast` and `portPublish` would look unused, I think it makes the code less clear.

Jharaxus: Indeed that may look more confusing instead of helping

Jharaxus (approved):

be2-scala/src/main/scala/ch/epfl/pop/pubsub/PublishSubscribe.scala
```diff
@@ -95,9 +96,12 @@ object PublishSubscribe {
...
-          val portParamsWithChannel = 2
-          val portParamsWithMap = 3
-          val totalPorts = 4
+          val portSubscribe = 2
```
Ajkunas, line 99: Suggestion: maybe you can use define a Map , e.g. `portValues`. It still remains clear when you will use it in the code.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/PublishSubscribe.scala
```diff
@@ -106,16 +110,27 @@ object PublishSubscribe {
...
+                }
+
+              case _ => portPipelineError // Pipeline error goes directly in merger
             }
           ))
 
           val hasMessagePartition = builder.add(ParamsWithMessageHandler.graph(messageRegistry))
-          val hasChannelPartition = builder.add(ParamsWithChannelHandler.graph(clientActorRef))
-          val hasMapPartition = builder.add(ParamsWithMapHandler.graph(dbActorRef))
+          val subscribePartition = builder.add(ParamsHandler.subscribeHandler(clientActorRef))
```
Ajkunas, line 129: Same suggestion here. 
**Disclaimer**: both are good, and I am not even sure if it will make the code clearer. it is just a good way to avoid having too much variables.

Ajkunas (commented):
Good job :))

PR #1582, opened 2023-05-21
Implement a witnessing repository and a witnessing policy

matteosz:


simone-kalbermatter (dismissed):
LGTM! The popup is a good solution to not always have to navigate to the witnessing tab. And the mnemonic words are a nice idea to make the messages ids human readable 🙂

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/data/LaoHandler.java
```diff
@@ -52,10 +53,11 @@ public LaoHandler(
    * @param context the HandlerContext of the message
    * @param createLao the message that was received
    */
-  @SuppressLint("CheckResult") // for now concerns Consensus which is not a priority this semester
+  // for now concerns Consensus which is not a priority this semester
```
Warhedgehog, line 56: This was about the above annotation, so the comment is not needed anymore

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/data/ElectionHandler.java
```diff
@@ -247,6 +226,30 @@ public void handleCastVote(HandlerContext context, CastVote castVote)
     }
   }
 
+  /**
+   * Simple way to handle a election key, add the given key to the given election
```
Warhedgehog, line 230: **an** election key :)

Warhedgehog (changes requested):
Impressive work. Just the two comments :)

Warhedgehog (approved):
LGTM

PR #1583, opened 2023-05-21
add greetServer high level support

BelghmiAmine:
This pull request adds high-level support for parsing greetServer jsons.


be2-scala/src/main/scala/ch/epfl/pop/model/network/MethodType.scala
```diff
@@ -15,6 +15,7 @@ object MethodType extends Enumeration {
   val HEARTBEAT: Value = MatchingValue("heartbeat")
   val GET_MESSAGES_BY_ID: Value = MatchingValue("get_messages_by_id")
 
+  val GREET_SERVER: Value = MatchingValue("greet_server")
```
K1li4nL, line 18: Absolute nitpick: i think we shoud regroup the MethodType together, the line return don't serve much

K1li4nL (commented):
just a minor refactoring nitpick, otherwise LGTM !

Ajkunas (approved):

Issue #1027, opened 2022-05-17
Election results not sent in catchup 


Tesa27:
- idea is to store election result on election channels when broadcast by go server
    - this is problematic since it is defined that the server is not supposed to create 
       messages, so election results should only be broadcast but not sorted by the backend. This causes 
       problems if someone joins an election too late, the catchup will not include the registered votes.
       An idea would be to recomputed the election results on every catchup but this can cause an overhead in 
       terms of computation and even network traffic.
    - Temporary solution is to store results on lection channel even if backend is not supposed to create messages and 
       is supposed to be more passive and creating messages is active.

Robingoumaz:
Also with the implementation of the new features in e-voting lao#greet and election#key messages, the backend sends messages.

Issue #866, opened 2022-03-07
Send election#result messages after a catchup


Tyratox:
This is mostly a copy of PR #848 as it seems as a more complex solution will be implemented than the suggested fix.

---

When working on https://github.com/dedis/popstellar/issues/703 I noticed that election result messages are not sent to a new client connecting after sending a catchup message. The reason is that the election#result messages are not stored when broadcasting them.

Compare
https://github.com/dedis/popstellar/blob/ab162f1cb2122f10dafbac003979abbfee4ac862/be1-go/channel/election/mod.go#L343-L344

and

https://github.com/dedis/popstellar/blob/ab162f1cb2122f10dafbac003979abbfee4ac862/be1-go/channel/election/mod.go#L384-L385

to

https://github.com/dedis/popstellar/blob/ab162f1cb2122f10dafbac003979abbfee4ac862/be1-go/channel/election/mod.go#L430-L435.

---

> @Warhedgehog: Per the slack message, without further clarifications, I do not think it is a desirable change since the documentation says that the inbox storage is for incoming message

> Noémien on Slack: IIRC we decided to store everything in the inbox, and filter out unnecessary messages when sending a catchup. In this case implementing a persistence layer would only require to persist the inbox, and replay the inbox content to recover the state of an organiser.

> @Tyratox It seems then that this comment should be updated as well? Should I include that in this PR or will you do this on a different branch & PR?

> @nkcr: Some questions that arose in the Go-meeting:
> - is it normal that the election result is only broadcast to the clients ?
> - if not, we should also broadcast the message to the servers, and this is where the inbox will be populated (we assume that if we broadcast to the servers, we will receive our own message)
> - in this case there is no need to store the message in the inbox there

> @Tyratox: Good points! I do not have a full overview over the project but it seems like a good solution as other servers, for example the witnesses, probably should get the results as well. I will ask in the evoting channel for confirmation and update this reply as soon as I get an answer.
> In any case it should just be made sure that this also happens if there is no other server.

> [Louis-Henri](https://app.slack.com/team/UMRHRP1K5) [vor 18 Stunden](https://dedis.slack.com/archives/C03442XUBCK/p1646239534435329?thread_ts=1646234776.541329&cid=C03442XUBCK)
> Election Result I believe should be signed by the witnesses to say that we witness and agree to the Election result
> Before it is sent to the clients. Otherwise clients are unsure whether to accept the result or not.
> This would require witness signing between the organizers and witnesses which relies on server-2-server communication.

PR #1594, opened 2023-05-24
Fix the catchup on election results issue 

MariemBaccari:
After reading through the discussions in #1027 and #866, I think the solution of storing the election results in the election channel inbox is relevant now with respect to the current state of the system since the backends now create and send messages. 
The scala backend writes the results on its database (we don't have one on the go backend but the equivalent would be storing it in the inboxes).  
Also, the channel inboxes are only used for the catchup so their purpose should be to provide all the necessary information about the channel to the new subscribers. 
In this PR, I simply add storing the election results message in the inbox after broadcasting it to the clients and remove the "incoming messages" specification from the inbox documentation.

### DIFF
--- a/be1-go/channel/election/mod.go
+++ b/be1-go/channel/election/mod.go
@@ -642,8 +642,8 @@ func (c *Channel) broadcastElectionResult() error {
 		WitnessSignatures: []message.WitnessSignature{},
 	}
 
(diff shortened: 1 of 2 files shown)


Nazianzenov (approved):
Not a lot of modifs, LGTM!

Anghul (approved):
LGTM, good call for the comment change

PR #1595, opened 2023-05-24
Fix offline wrongly redirect

dayan9265:


Tyratox (approved):
🥳 Very cool to see this fixed! Sorry for introducing this bug 🫣

Issue #1597, opened 2023-05-24
[BUG] BE1 - some tests randomly failing


dayan9265:
### Description (Actual behavior)
When doing a pull request with a job not related the the be1, the CI will sometimes fail on some tests of the BE1

### Expected behavior
CI passing for BE1 tests

### How to reproduce
Create a pull request

### Version & Environment
This bug was reproduced on:
- Commit: 06860647510fbcd5370bd611a33878b62f857cd6

##### Front-ends:
- [ ] Fe1-Web (please include browser's names & version)
- [ ] Fe2-Android (please specify if phone or emulation, and Android version)
- [x] Not applicable

##### Back-ends:
- [x] Be1-Go
- [ ] Be2-Scala
- [ ] Not applicable

### Workaround
Relaunch the test

### Impact
Makes the process to create a successful PR longer

### Possible root cause
here are the error messages I get:
Ubuntu:
![image](https://github.com/dedis/popstellar/assets/82887324/8b30b8f8-b227-4e2d-8fa2-2b6aed33890f)
Windows:
![image](https://github.com/dedis/popstellar/assets/82887324/52822d81-8e9c-4722-9494-34da3a7f13af)

PR #1599, opened 2023-05-24
Fixing issue #1597

Nazianzenov:
Reference : (https://github.com/dedis/popstellar/issues/1597)

Cause: tests of the popcha server were using the same port, causing unstable tests due to the servers not closing properly in test n before restarting in test n+1.

Solution: each test has a different port, allowing to be completely independent from other tests.


Anghul (approved):
LGTM

PR #1600, opened 2023-05-25
Setup environment for server-server tests

simone-kalbermatter:
Set up the environment so that heartbeat tests are only run when connected as a server and all other tests when connected as a frontend.


matteosz (dismissed):
LGTM!

GabrielFleischer (dismissed):
LGTM !

pierluca (approved):

PR #1604, opened 2023-05-25
Fine-tune Decentralized communications

K1li4nL:
Two things in this pr: 

`GetMessagesByIdResponseHanlder`:
The validation has changed so that instead of retrying a certain amount of time per channel, each message has its own counter. 

`ConnectionMediator`: 
Prevent it to connect to the same urls upon change in the config. The behavior is not perfect as for now there is no link between clientActorRef and their urls. `GreetServer` will be helpful in that regard.

### DIFF
--- a/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/GetMessagesByIdResponseHandler.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/GetMessagesByIdResponseHandler.scala
@@ -22,7 +22,15 @@ object GetMessagesByIdResponseHandler extends AskPatternConstants {
   private val MAX_RETRY_PER_MESSAGE = 5
   private val SUCCESS = 0
 
-  // This function packs each message into a publish before pushing them into the pipeline
+  /** Validate and replay on the system each message received in the get_messages_by_id result
+    *
+    * @param messageRegistry
+    *   The system message registry to use in the validation pipeline
+    * @param system
+    *   Implicit actor system to use the given validator
+    * @return
+    *   Left if some messages couldn't be validated after MAX_RETRY_PER_MESSAGE times, Right for success
+    */
   def responseHandler(messageRegistry: MessageRegistry)(implicit system: ActorSystem): Flow[GraphMessage, GraphMessage, NotUsed] = Flow[GraphMessage].map {
     case Right(JsonRpcResponse(_, Some(resultObject), None, _)) =>
       resultObject.resultMap match {
@@ -43,8 +51,17 @@ object GetMessagesByIdResponseHandler extends AskPatternConstants {
     case value @ _ => value
   }
 
-  // This function will try to digest the set of messages for an entire channel
-  // It will go channel by channel until it finished processing messages or MAX_RETRY_PER_MESSAGE * total_number_of_message is reached
+  /** Will try to digest each GraphMessage until their retry-counter reaches 0 or they all get validated
+    *
+    * @param receivedResponse
+    *   The wrapped messages with their counter
+    * @param validatorFlow
+    *   The validator to push the messages through
+    * @param system
+    *   Implicit actor system to use the given validator
+    * @return
+    *   true if all messages could be validated, otherwise false
+    */
   @tailrec
   private def passThroughPipeline(
       receivedResponse: List[(GraphMessage, Int)],
@@ -73,7 +90,17 @@ object GetMessagesByIdResponseHandler extends AskPatternConstants {
     passThroughPipeline(failedMessages.reverse, validatorFlow)
   }
 
-  // Push a message corresponding through the pipeline
+  /** Push a message through the given validator
+    *
+    * @param validator
+    *   The pipeline verifying the message
+    * @param message
+    *   The message to verify
+    * @param system
+    *   Implicit actor system to use the given validator
+    * @return
+    *   The result of the validator on the given message, Left for failure, Right for success
+    */
   private def messagesThroughPipeline(
       validator: Flow[GraphMessage, GraphMessage, NotUsed],
       message: GraphMessage
@@ -85,7 +112,13 @@ object GetMessagesByIdResponseHandler extends AskPatternConstants {
     }
   }
 
-  // Must return a list sorted by channel length
+  /** Wrap messages in Publish RPCs along with a counter
+    *
+    * @param map
+    *   A map of channels to their set of messages
+    * @return
+    *   A list of GraphMessage - Int pairs where the GraphMessage are sorted in increasing order of their channel length and the Int is the number of retry left for each message
+    */
   private def wrapMsgInPublish(map: Map[Channel, Set[Message]]): List[(GraphMessage, Int)] = {
     var publishedList = map.foldLeft(List.empty[JsonRpcRequest])((acc, elem) =>
       acc ++ elem._2.map(msg =>
--- a/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/GetMessagesByIdResponseHandler.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/GetMessagesByIdResponseHandler.scala
@@ -12,24 +12,24 @@ import ch.epfl.pop.pubsub.graph.{ErrorCodes, GraphMessage, PipelineError, pretty
 import ch.epfl.pop.pubsub.{AskPatternConstants, MessageRegistry, PublishSubscribe}
 
 import scala.annotation.tailrec
+import scala.util.Success
 import scala.concurrent.Await
 
 //This object's job is to handle responses it receives from other servers after sending a heartbeat.
 // When receiving the missing messages, the server's job is to write them on the database.
 object GetMessagesByIdResponseHandler extends AskPatternConstants {
 
-  private final val MAX_RETRY_PER_MESSAGE = 5
-  private final val SUCCESS = 0
+  private val MAX_RETRY_PER_MESSAGE = 5
+  private val SUCCESS = 0
 
   // This function packs each message into a publish before pushing them into the pipeline
   def responseHandler(messageRegistry: MessageRegistry)(implicit system: ActorSystem): Flow[GraphMessage, GraphMessage, NotUsed] = Flow[GraphMessage].map {
     case Right(JsonRpcResponse(_, Some(resultObject), None, _)) =>
       resultObject.resultMap match {
         case Some(resultMap) =>
-          val receivedResponse: Map[Channel, Set[GraphMessage]] = wrapMsgInPublish(resultMap)
+          val receivedResponse = wrapMsgInPublish(resultMap)
           val validator = PublishSubscribe.validateRequests(ActorRef.noSender, messageRegistry)
-          val maxTrials = MAX_RETRY_PER_MESSAGE * receivedResponse.foldRight(0)((elem, acc) => acc + elem._2.size)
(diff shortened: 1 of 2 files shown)


be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/GetMessagesByIdResponseHandler.scala
```diff
@@ -43,65 +44,60 @@ object GetMessagesByIdResponseHandler extends AskPatternConstants {
...
     }
-
-    failedGraphMessages
   }
 
-  private def wrapMsgInPublish(map: Map[Channel, Set[Message]]): Map[Channel, Set[GraphMessage]] = {
-    map.map { case (channel, set) =>
-      channel -> set.map(msg =>
-        Right(JsonRpcRequest(
+  // Must return a list sorted by channel length
```
Ajkunas, line 88: Try to be more precise in your comments: a list of what?

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/GetMessagesByIdResponseHandler.scala
```diff
@@ -12,23 +12,24 @@ import ch.epfl.pop.pubsub.graph.{ErrorCodes, GraphMessage, PipelineError, pretty
...
+import scala.util.Success
 import scala.concurrent.Await
 
 //This object's job is to handle responses it receives from other servers after sending a heartbeat.
 // When receiving the missing messages, the server's job is to write them on the database.
 object GetMessagesByIdResponseHandler extends AskPatternConstants {
 
-  private final val MAX_RETRY = 5
-  private final val SUCCESS = 0
+  private val MAX_RETRY_PER_MESSAGE = 5
```
Ajkunas, line 22: Why did you remove the `final`? (in line 22 too)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/GetMessagesByIdResponseHandler.scala
```diff
@@ -43,65 +44,60 @@ object GetMessagesByIdResponseHandler extends AskPatternConstants {
...
+              if (retry == 1) {
+                println("Errors: " + err.toString + "\n On:\n" + prettyPrinter(graphMessage))
+              }
+            case _ => /* DO NOTHING */
           }
         }
     }
 
-    passThroughPipeline(failedMessages, validatorFlow, remainingAttempts - 1)
+    passThroughPipeline(failedMessages.reverse, validatorFlow)
```
Ajkunas, line 73: Why do you need to reverse your list of messages ?

Ajkunas (changes requested):
Most of the comments are nitpicks or questions. Othw all good :)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/GetMessagesByIdResponseHandler.scala
K1li4nL: I figured the `final` didn't add anything given that `val` is used and these are private variables.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/GetMessagesByIdResponseHandler.scala
K1li4nL: I unfortunately don't have any data to back this up but:
Since we are ordering the channels based on the length, I  wanted to keep their order. (since the we add the `graphMessage` on the top of the failed list, the order is reversed) 

I don't know performance wise if one way is better than the other though...

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/GetMessagesByIdResponseHandler.scala
K1li4nL: I added proper documentation

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/GetMessagesByIdResponseHandler.scala
```diff
@@ -42,66 +51,86 @@ object GetMessagesByIdResponseHandler extends AskPatternConstants {
...
+  /** Wrap messages in Publish RPCs along with a counter
+    *
+    * @param map
+    *   A map of channels to their set of messages
+    * @return
+    *   A list of GraphMessage - Int pairs where the GraphMessage are sorted in increasing order of their channel length and the Int is the number of retry left for each message
+    */
+  private def wrapMsgInPublish(map: Map[Channel, Set[Message]]): List[(GraphMessage, Int)] = {
+    var publishedList = map.foldLeft(List.empty[JsonRpcRequest])((acc, elem) =>
+      acc ++ elem._2.map(msg =>
```
Jharaxus, line 124: I think you could use a `flatMap` instead of `foldLeft` + an accumulator

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/GetMessagesByIdResponseHandler.scala
```diff
@@ -42,66 +51,86 @@ object GetMessagesByIdResponseHandler extends AskPatternConstants {
...
           MethodType.PUBLISH,
-          new Publish(channel, msg),
+          new Publish(elem._1, msg),
           Some(0)
-        ))
+        )
       )
-    }
+    )
+    publishedList = publishedList.sortWith((e1, e2) => e1.getParamsChannel.channel.length <= e2.getParamsChannel.channel.length)
```
Jharaxus, line 133: Is this ordering part of the decentralized specification? I find an ordering by the channel name's length quite strange... Is there some specific intent behind that?

Jharaxus: On a scala side, you could avoir using `var` for `publishedList` by simply returning
```scala
publishedList.sortWith(...)
.map(...)
```

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/GetMessagesByIdResponseHandler.scala
Jharaxus: Is there really a need to keep that ordering all the way through the simulation pipeline?

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/GetMessagesByIdResponseHandler.scala
```diff
@@ -42,66 +51,86 @@ object GetMessagesByIdResponseHandler extends AskPatternConstants {
...
-          // only log a during last attempt
-          if (remainingAttempts == 1) {
-            println("Errors: " + failedMsgErrorMap.map {
-              case (msg, error) => error.toString + "\n On:\n" + prettyPrinter(msg)
-            })
+      case (graphMessage, retry) =>
+        if (retry > 0) {
+          messagesThroughPipeline(validatorFlow, graphMessage) match {
+            case Left(err) =>
+              failedMessages = (graphMessage, retry - 1) :: failedMessages
```
Jharaxus, line 80: Nitpicks: You could probably use the operator `::=` that you made me discover a couple of lines above :)
On a side note, maybe using a mutable collection from `scala.collection.mutable` would be better than using a `var` for performance purposes

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/GetMessagesByIdResponseHandler.scala
```diff
@@ -42,66 +51,86 @@ object GetMessagesByIdResponseHandler extends AskPatternConstants {
...
   @tailrec
   private def passThroughPipeline(
-      receivedResponse: Map[Channel, Set[GraphMessage]],
-      validatorFlow: Flow[GraphMessage, GraphMessage, NotUsed],
-      remainingAttempts: Int
+      receivedResponse: List[(GraphMessage, Int)],
+      validatorFlow: Flow[GraphMessage, GraphMessage, NotUsed]
   )(implicit system: ActorSystem): Boolean = {
-    if (receivedResponse.isEmpty || remainingAttempts <= 0) {
+    if (receivedResponse.isEmpty || receivedResponse.forall(elem => elem._2 <= 0)) {
```
Jharaxus, line 70: The second check `receivedResponse.forall(elem => elem._2 <= 0)` is actually not really necesssary as messages with `retry == 0` are ignored in the following `foreach(...)`, so if all the responses received have 0 retry left, the next `passThroughPipeline` will have an empty list of responses and will therefor still return (so it would avoid iterating over the entire list of responses 2 times per call instead of only 1)

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/GetMessagesByIdResponseHandler.scala
```diff
@@ -12,23 +12,32 @@ import ch.epfl.pop.pubsub.graph.{ErrorCodes, GraphMessage, PipelineError, pretty
...
+    *   Implicit actor system to use the given validator
+    * @return
+    *   Left if some messages couldn't be validated after MAX_RETRY_PER_MESSAGE times, Right for success
+    */
   def responseHandler(messageRegistry: MessageRegistry)(implicit system: ActorSystem): Flow[GraphMessage, GraphMessage, NotUsed] = Flow[GraphMessage].map {
     case Right(JsonRpcResponse(_, Some(resultObject), None, _)) =>
       resultObject.resultMap match {
         case Some(resultMap) =>
-          val receivedResponse: Map[Channel, Set[GraphMessage]] = wrapMsgInPublish(resultMap)
+          val receivedResponse = wrapMsgInPublish(resultMap)
```
Jharaxus, line 38: nitpick: maybe you could pass the number of retries as a parameter to `wrapMsgInPublish` as you used to do with `passThroughPipeline`: otherwise, I feel like it kinds of come from nowhere as `wrapMsgInPublish` as a name doesn't really carry the intent of allocating a number of retry to each message...

Jharaxus (dismissed):
Nice code, it indeed makes more sense to have a retry per message !

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/GetMessagesByIdResponseHandler.scala
K1li4nL: Absoluetely not in the protocol. The intent is as follows: 
In general shorter channel name are less specific and so should be treated first. 

`/root`: should be treated first as it contains the lao creation messages
`/root/lao_id`: should be treated right after since it contains the rollcalls and all.

It might not be very elegant and there are exceptions but the retries seem to be enough to deal with them

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/GetMessagesByIdResponseHandler.scala
K1li4nL: I removed since its utility beside being confusing is questionable.

Ajkunas (approved):
LGTM

PR #1610, opened 2023-05-26
bug fix karate testing

BelghmiAmine:
This pull request fixes a couple of bugs detected by the karate tests : 

1. Opening the election with open time before election setup results in an error.
2. Casting a valid vote before election opening time should fail.
3. Sending an election end with creation time before election setup should return an error.

### DIFF
--- a/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/ElectionValidator.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/ElectionValidator.scala
@@ -150,8 +150,8 @@ sealed class ElectionValidator(dbActorRef: => AskableActorRef) extends MessageDa
 
         val setupElectionTimeStamp = Await.ready(channel.getSetupMessage(dbActorRef), duration).value.get match {
           case Success(setupElection: SetupElection) => setupElection.created_at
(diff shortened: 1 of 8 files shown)


Ajkunas (changes requested):
Reviewed a bit the PR :) I just have some comments concerning some implementations to make it more concise, othw all good

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/ElectionValidator.scala
```diff
@@ -148,12 +154,23 @@ sealed class ElectionValidator(dbActorRef: => AskableActorRef) extends MessageDa
 
         val electionId = channel.extractChildChannel
 
+        val setupElectionTimeStamp = extractSetupElectionTimestamp(channel, rpcMessage)
+        if (setupElectionTimeStamp.isEmpty) {
```
Ajkunas, line 158: Same comment as below. 
I would suggest you to use a match case instead.

Ajkunas (changes requested):
Last nitpicks and it's ready to be approved !

pierluca (approved):
Approving blindly :)

Ajkunas (approved):

PR #1611, opened 2023-05-27
Fix CI tests and general concurrency issues + avoid spam of lao connecting

### LINKED ISSUE #1608: [BUG] CI tests of FE2 sometimes get stuck
When executing the tests on the CI it happens sometimes that the fe2 tests get stuck.

### Expected behavior
It shouldn't happen.

### Version & Environment
I've gone through the github action history to see when this behaviour first appeared, and it was during the PR #1575, so the problem could be in those modifications .

##### Front-ends:
- [ ] Fe1-Web
- [X] Fe2-Android
- [ ] Not applicable

##### Back-ends:
- [ ] Be1-Go
- [ ] Be2-Scala
- [X] Not applicable

### Workaround
Usually canceling the workflow and restarting the test fixes this.

### Impact
In terms of deployment this could slow a bit the process.

### Possible root cause
Looking at the logs, the lock happens generally when testing the repositories, so my best guess is that there is a deadlock, since the tests are executed in parallel. Also in the repos we used synchronized methods, so this can be totally plausible

### LINKED ISSUE #1609: [BUG] Joining a LAO spams the ConnectingActivity
The ui keeps refreshing a new connecting activity when joining a LAO.

### Expected behavior
The connecting activity should be launched only once.

### How to reproduce
1. Click on Join
2. Point the camera to a valid LAO QR
3. See connecting UI keeps refreshing

### Version & Environment
This bug was reproduced on:
- Commit: 74ac62845c7e7d6f69e1a6bb9d568aa8b10506d3

##### Front-ends:
- [ ] Fe1-Web
- [X] Fe2-Android
- [ ] Not applicable

##### Back-ends:
- [ ] Be1-Go
- [ ] Be2-Scala
- [X] Not applicable

### Impact
Just visually seems buggy, but there's no actual impact on functionality.

### Possible root cause
In my opinion this is due to the scanner that actually scans a QR multiple times even in few seconds, thus this creates multiple connecting activity. The solution could be to enforce the scanner to quit after it has captured a valid qr

matteosz:
This PR is a bug-fixing PR.

- It resolves #1608 by replacing synchronization of methods with a more elegant and efficient usage of concurrent thread-safe data structures.

>**Note**: with these changes now tests are executing in less time (around 2min 30sec against the previous 6min), so the application seems way faster and responsive.

- It improves the documentation of the repositories.

- It resolves #1609 by preventing the connecting activity to be spammed several times upon connection of a lao.

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/RollCallRepository.java
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/RollCallRepository.java
@@ -16,6 +16,7 @@ import com.github.dedis.popstellar.utility.error.UnknownRollCallException;
 import com.github.dedis.popstellar.utility.error.keys.NoRollCallException;
 
 import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
@@ -58,7 +59,7 @@ public class RollCallRepository {
   }
 
   /**
-   * This either updates the roll call in the repository or adds it if absent
+   * This either updates the roll call in the repository or adds it if absent.
    *
    * @param rollCall the roll call to update/add
    */
@@ -71,11 +72,10 @@ public class RollCallRepository {
     }
     Timber.tag(TAG).d("Updating roll call on lao %s : %s", laoId, rollCall);
 
-    RollCallEntity rollCallEntity = new RollCallEntity(laoId, rollCall);
     // Persist the rollcall
     disposables.add(
         rollCallDao
-            .insert(rollCallEntity)
+            .insert(new RollCallEntity(laoId, rollCall))
             .subscribeOn(Schedulers.io())
             .observeOn(AndroidSchedulers.mainThread())
             .subscribe(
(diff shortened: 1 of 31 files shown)


simone-kalbermatter (dismissed):
LGTM, well documented!

fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/RollCallRepository.java
```diff
@@ -103,17 +103,35 @@ public Observable<RollCall> getRollCallObservable(String laoId, String persisten
...
+   * @return the roll call matching the identifier in the lao
+   * @throws UnknownRollCallException if no roll call with the provided id could be found
+   */
   public RollCall getRollCallWithPersistentId(String laoId, String persistentId)
       throws UnknownRollCallException {
     return getLaoRollCalls(laoId).getRollCallWithPersistentId(persistentId);
   }
 
+  /**
+   * This function retrieves a roll call from the repository given its ephemeral identifier.
```
Warhedgehog, line 120: I'd prefer state-dependant identifier

Warhedgehog (changes requested):
Just the small detail in the comment. Great work !

Warhedgehog (approved):
LGTM

PR #1612, opened 2023-05-28
Hide meeting

dayan9265:


Tyratox:
Depends on the goals you have with the release branch but what is the reason for not disabling the meeting feature but rather hide the option to create meetings from the UI? (Meetings created by other organizers would still show up)

But I like that you added the comment `release:` so that you can later easily search for these changes :)

Tyratox (approved):

PR #1613, opened 2023-05-28
Extended multi connection to multiple greet laos

matteosz:
This PR extends the current multiconnection to also connect to servers which join later.

As the backends have implemented a dynamic server configuration, they are going to send a new GreetLao when a new peer joins. So far, the current multiconnection was only connecting to the peers in the very first greet lao, maintaining static the servers set. By this change now we process each greetlao and we connect to all the peers we have not been connected yet, allowing to connect to new servers joining the network.


simone-kalbermatter (approved):
LGTM, well documented and tested!

Warhedgehog (approved):
LGTM

PR #1614, opened 2023-05-29
Refactor RuntimeEnvironment

K1li4nL:
Given the growing usage of `RuntimeEnvironment` in the system this refactoring should tidy things a bit

### DIFF
--- a/be2-scala/src/main/scala/ch/epfl/pop/config/RuntimeEnvironment.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/config/RuntimeEnvironment.scala
@@ -16,7 +16,42 @@ import scala.sys.SystemProperties
   */
 object RuntimeEnvironment {
 
+  // Regex from dataGreetLao.json
+  private val addressPattern = "^(ws|wss)://.*(:d{0,5})?/.*$"
+
   private lazy val sp = new SystemProperties()
+  private lazy val confDir: String = getConfDir
+  private lazy val appConfFile = confDir + File.separator + "application.conf"
+
+  lazy val appConf: Config = ConfigFactory.parseFile(new File(appConfFile))
+  lazy val serverConf: ServerConf = ServerConf(appConf)
+  lazy val ownClientAddress = f"ws://${serverConf.interface}:${serverConf.port}/${serverConf.clientPath}"
+  lazy val ownServerAddress = f"ws://${serverConf.interface}:${serverConf.port}/${serverConf.serverPath}"
+  lazy val ownAuthAddress = f"http://${serverConf.interface}:${serverConf.port}/${serverConf.authenticationPath}"
+  lazy val isTestMode: Boolean = testMode
+
+  lazy val serverPeersListPath: String =
+    if (isTestMode) {
+      confDir + File.separator + "server-peers-list-mock.conf"
+    } else {
+      confDir + File.separator + "server-peers-list.conf"
+    }
+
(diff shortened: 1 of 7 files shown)


Ajkunas (dismissed):
LGTM

be2-scala/src/main/scala/ch/epfl/pop/config/RuntimeEnvironment.scala
```diff
@@ -76,4 +54,39 @@ object RuntimeEnvironment {
...
+      println("Starting the server without any previous persistent state")
+
+      // removing database folder
+      val directory = new Directory(new File(dbPath))
+      if (directory.deleteRecursively()) {
+        println("Removed old database folder")
+      }
+    }
+
+    val virtualMachineParam = "scala.config"
```
Jharaxus, line 70: nitpick: as this is an important value (used as a parameter to launch the program), it could be turned into an attribute instead of being inlined in the function for better visibility

Jharaxus: The same goes for `dbPathParam` and "/database" below, and `sp("clean")` above

be2-scala/src/main/scala/ch/epfl/pop/config/application.conf
```diff
@@ -17,12 +17,14 @@ akka {
 }
 
 #POP Server config default
+#Example of external address: wss://pop-demo.online, will default to ws://interface:port if empty
```
Jharaxus, line 20: Is that a safe behavior to have? The pop specification expect communication over TLS for security, so I don't believe this is a safe default value...

Jharaxus (commented):
Very nice refactoring and modularization !

be2-scala/src/main/scala/ch/epfl/pop/config/application.conf
K1li4nL: This address is only relevant in case of reverse proxy usage. As far as I can tell the system doesn't support TLS natively, we can only rely on external solution. This default behavior when the field is empty is what we actually do today. Unless we add native support of TLS, I don't think we can have safe defaults, you are correct.

Ajkunas (approved):

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/LaoHandler.scala
```diff
@@ -27,8 +26,7 @@ case object LaoHandler extends MessageHandler {
         socialChannel: Channel = Channel(s"$laoChannel${Channel.SOCIAL_MEDIA_CHIRPS_PREFIX}")
         reactionChannel: Channel = Channel(s"$laoChannel${Channel.REACTIONS_CHANNEL_PREFIX}")
         // we get access to the canonical address of the server
-        config = ServerConf(appConf)
-        address: Option[String] = Some(f"ws://${config.interface}:${config.port}/${config.clientPath}")
+        address: Option[String] = Some(ownClientAddress)
```
pierluca, line 29: since this is used in GreetLao, shouldn't it use the external address ?

K1li4nL: Absoluetely ! Fixed.

pierluca (approved):

PR #1615, opened 2023-05-29
Add greet server logic

BelghmiAmine:
This pull request adds the greetServer logic.


be2-scala/src/main/scala/ch/epfl/pop/decentralized/ConnectionMediator.scala
```diff
@@ -37,28 +38,22 @@ final case class ConnectionMediator(
...
       )
 
-    case ConnectionMediator.NewServerConnected(serverRef) =>
-      log.info("Received new server")
-      if (serverSet.isEmpty)
-        monitorRef ! Monitor.AtLeastOneServerConnected
-
-      serverSet += serverRef
-
     case ConnectionMediator.ServerLeft(serverRef) =>
```
Ajkunas, line 47: I see that in line 67, you removed the `ConnectionMediator.`, because I guess it is not necessary to put it. For consistency, you should also remove this one, or put it in both cases.

be2-scala/src/main/scala/ch/epfl/pop/pubsub/ClientActor.scala
```diff
@@ -86,11 +97,24 @@ final case class ClientActor(mediator: ActorRef, connectionMediatorRef: ActorRef
         case _ => log.error("UNKNOWN MESSAGE TO CLIENT ACTOR: " + m)
       }
   }
+
+  private def triggerGreetServer(): Unit = {
```
Ajkunas, line 101: Commenting a bit this function could be useful for future students :)

Ajkunas (changes requested):
I suggest you comment a bit the line of codes to make it clearer for the next students :)

pierluca (dismissed):
Approving blindly (no time to review)

pierluca (approved):
Re-approving blindly

PR #1617, opened 2023-05-29
Add circular handling of GetMessagesById answer

MariemBaccari:
We noticed that the go servers fail to catch up properly through a GetMessagesById if they connect some time after the creation of the LAO. Since the message processing is linear and only goes through the list once, it fails if, for example, the message for an election creation is processed before the lao creation.
To fix that, I implemented a simple circular way of handling the answer. The server now goes through the list enough times until it succeeds or reaches maxium number of retries (=10). 

When testing with 3 servers, we noticed that there are a lot of errors logged. That is because a server can send a GetMessagesById with the same message ids to multiple servers. When it receives the result, it processes the messages correctly the first time. It then receives the result a second time from another server and logs errors because the message was already received. I think that in the context of decentralized communication, it makes more sense to log that as an info rather than an error since it is virtually unavoidable and doesn't cause any harm as far as I know. 

The PR also removes sending empty results to messages that aren't json rpc requests (messages that don't have an id field).

### DIFF
--- a/be1-go/hub/standard_hub/message_handling.go
+++ b/be1-go/hub/standard_hub/message_handling.go
@@ -3,7 +3,6 @@ package standard_hub
 import (
 	"encoding/base64"
 	"encoding/json"
-	"github.com/rs/zerolog/log"
 	"popstellar/crypto"
 	jsonrpc "popstellar/message"
 	"popstellar/message/answer"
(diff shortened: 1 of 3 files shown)


Nazianzenov (approved):
LGTM!

Anghul (approved):
LGTM

Issue #1448, opened 2023-03-13
[BUG] Private keys leak on catchup on be2-scala


K1li4nL:
### Description (Actual behavior)
The private election channel can be subsrcibed to and sending a catchup on it will leak the private keys of the backend. 

![image](https://user-images.githubusercontent.com/79536516/224720012-a5544938-f1d1-4dba-b880-034564400cd0.png)

### Expected behavior
The backend should reject such rpc and not leak keys. 

### How to reproduce
1. Run the backend and create elections (then close them)
2. Go to https://hoppscotch.io/realtime/websocket 
3. connect to the backend via websocket 
4. Send a catchup (see protocol for the json format, or pic above) to channel `/root/private/electionid`
5. Notice the leak in the server response

### Version & Environment
This bug was reproduced on: 
- Master, Commit: 8e382cea06475a84c83c693489898f1f3f3f9fc3

##### Front-ends:
- [ ] Fe1-Web (please include browser's names & version)
- [ ] Fe2-Android (please specify if phone or emulation, and Android version)
- [x] Not applicable

##### Back-ends:
- [ ] Be1-Go
- [x] Be2-Scala
- [ ] Not applicable

##### Environment (as applicable):
Not applicable

### Impact
The election secret ballot is not secret anymore, anyone can pass themselves as the server by signing messages...etc

Ajkunas:
As I see it : 
- We should focus on making the function `Catchup` of the dbActor more secure: should we simply check if a channel is `private` before catching up messages ? should we ensure that only the organizer can catchup those messages ? 
- Ofc we should change the content generated by  the `AnswerGenerator` so that we do not leak private informations

PR #1620, opened 2023-05-31
Add server Public / Private key pair

K1li4nL:


Jharaxus (commented):
Nice work to finally bring a public/private key pair to scala !
I still have an open question on this matter, is it safe enough to store the private key in the database in the same way that we do for all other messages that go through the pop server?
I actually couldn't think of a better way, and browsing the internet didn't give me anything satisfactory, but I still find quite weak storing such important information as plaintext in our database...
On the other hand, your solution is elegant, works very well on restarts and scales across servers automatically as it is not hardcoded, so it already has several nice properties which don't necessarily make it an urgent matter to find a more secure solution

K1li4nL:
> Nice work to finally bring a public/private key pair to scala ! I still have an open question on this matter, is it safe enough to store the private key in the database in the same way that we do for all other messages that go through the pop server? I actually couldn't think of a better way, and browsing the internet didn't give me anything satisfactory, but I still find quite weak storing such important information as plaintext in our database... On the other hand, your solution is elegant, works very well on restarts and scales across servers automatically as it is not hardcoded, so it already has several nice properties which don't necessarily make it an urgent matter to find a more secure solution

You are absoluetely right ! This is not ideal, we will have to trust the server safety regarding access to the disk, unless we start playing with HSMs... 
A question that we should ask though: Is there a way to lead the server to leak the keys by crafting some message ? (see issue #1448 ). I have not tested if the same can be done with the solution here, since it is not on a channel I don't think so but we never know...

One thing that bothers me here is that with this implementation, it looks very easy to shoot ourself in the foot. Ideally we would have a tight API that prevents the private key to be revealed instead of just handing the key and hoping it doesn't end in some error log over the network... 
One idea was to have some singleton dealing with all the key signing operation in order to have all the sensitive operations done in a particular place in the code. But having this singleton made it harder to have the persistency working.

Ajkunas (approved):

pierluca:
I share those concerns, but there aren't that many options:
we either integrate it into the DB (which conveniently is self-contained) or we need to put the keys in a separate file. 
the latter is probably safer (for all the reasons you mentioned) but it's not as convenient. 

To future-proof things a bit, may I suggest using some key after the prefix? e.g.`ServerPublicKey:Default`, to ensure that if we need other keys, we can easily reuse that prefix :)

Random thought: do you know if any of the Unicode characters allows to do the equivalent of a "backspace" ? That could be a way to "attack" the database, if it's possible to "backspace" through the prefix and substitute it. In my understanding DEL shouldn't and that would be the most likely one, but ... I'm not familiar enough with the Unicode control characters to state this confidently 🤔

K1li4nL:
> I share those concerns, but there aren't that many options: we either integrate it into the DB (which conveniently is self-contained) or we need to put the keys in a separate file. the latter is probably safer (for all the reasons you mentioned) but it's not as convenient.
> 
It could be helpful for the organizer though, they would have a way of knowing their server's public key at least. It would also make it easier to build a stricter API  that way. Instead of just giving the keys away, we could have some object that is self-contained, it would read the key file by itself and have the functions to sign data without having to reveal the keys. 

> To future-proof things a bit, may I suggest using some key after the prefix? e.g.`ServerPublicKey:Default`, to ensure that if we need other keys, we can easily reuse that prefix :)
> 
Done !
> Random thought: do you know if any of the Unicode characters allows to do the equivalent of a "backspace" ? That could be a way to "attack" the database, if it's possible to "backspace" through the prefix and substitute it. In my understanding DEL shouldn't and that would be the most likely one, but ... I'm not familiar enough with the Unicode control characters to state this confidently

I have taken a quick look at it, there seems to be unicode  `\b` for that which I wasn't able to use (the db says it doesn't have the channel `/root/\b\b....PREFIX:`. But that doesn't prove anything

pierluca:
> It could be helpful for the organizer though, they would have a way of knowing their server's public key at least. It would also make it easier to build a stricter API that way. Instead of just giving the keys away, we could have some object that is self-contained, it would read the key file by itself and have the functions to sign data without having to reveal the keys.

Good point. It was too early in the morning for me to spot that 😆 

Do you want to do that or not ? And if yes, in this or a subsequent PR ? 
Just to know when to "approve" this one :)

K1li4nL:
> Do you want to do that or not ? And if yes, in this or a subsequent PR ? Just to know when to "approve" this one :)

For all the point mentionned, I think it should be done that way. But I don't know when I will have the time to get it done, and we need this pr so that @BelghmiAmine can finish the GreetServer. I will open an issue so that it doesn't get forgotten and kindly accept your offer on the approval ;)

pierluca (approved):

PR #1631, opened 2023-06-02
Witnessing policy #2

matteosz:
This PR addresses the open issue left by the previous PR (#1582).

It simply provides a way of storing and making persistent pending objects (i.e. objects which have been received as message but not yet reached the threshold of witness signatures). So, now the user can exit the application and the pending objects will be restored.


simone-kalbermatter (approved):
LGTM!

fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/WitnessingRepository.java
Warhedgehog, line 1: Could you add some documentation to the class please ? Mainly the general witnessing policy and how persistence impacts it. I think this would be rather complex to understand for a newcomer

Warhedgehog (changes requested):
Great work ! I've not reviewed the test yet but for the rest I just have the sole request :)

Warhedgehog:
Also, I think that it is now quite necessary to update the android documentation, in a separate PR, for the witnessing and the persistence parts

matteosz:
> Also, I think that it is now quite necessary to update the android documentation, in a separate PR, for the witnessing and the persistence parts

Yeah, I was planning to update it now that the work is essentially finished

Warhedgehog (approved):
LGTM !

PR #1633, opened 2023-06-03
Partially fix mobile scanner

MeKHell:
This PR aims to fix the QR Code scanner on mobile devices.
According to my tests, it seems that the `requestCameraPermissionAsync` blocks the rear camera while `getCameraPermissionAsync` blocks the front camera.

Therefore, this PR disables the ability to switch camera, and does not use `requestCameraPermissionAsync`.

Finally, this fix only works on Chromium based mobile browser. This fix does not work on Firefox mobile.


MeKHell:
We also need to edit the web polyfill to allow the camera to get the stream (and not just the request). Thus, we need to close the stream provided by the `getMediaDevices`.

fe1-web/src/core/components/QrCodeScanner.tsx
```diff
@@ -57,62 +57,47 @@ const styles = StyleSheet.create({
  * (2022-12-05, Tyratox) Issue to track this: https://github.com/dedis/popstellar/issues/1306
  */
 const QrCodeScanner = ({ showCamera, children, handleScan }: IPropTypes) => {
-  const [permission, requestPermission] = Camera.useCameraPermissions();
+  // On future releases this might be changed to allow switching cameras. But for now this is not working. (2023-06-03)
```
Tyratox, line 60: I would add your username in the comment too, this was the format we usually used in the past. Is a bit quicker to see who put the comment than using git blame :)

```suggestion
  // On future releases this might be changed to allow switching cameras. But for now this is not working. (2023-06-03, MeKHell)
```

Tyratox (approved):
I find it odd that `requestPermission` is not required but I just verified that the Chrome and Firefox Desktop Browsers still work sooooo... ✅😁

Tyratox (approved):

MeKHell:
> I find it odd that `requestPermission` is not required.

The way `getPermission` is implemented in JS is almost the same as the `requestPermission`. I think this fix might break when we will try to export to Android and IOS.

PR #1634, opened 2023-06-03
Refactor Karate Part 3

simone-kalbermatter:
This PR refactors the following features: 

- electionSetup.feature
- electionOpen.feature
- castVote.feature
- electionEnd.feature
- update.feature
- transaction.feature
- hearbeat.feature 

The hearbeat feature was split into 2 features, heartbeat and getMessagesById and for both some more tests were added.

The `Transaction` class currently only supports one transaction and does not keep track of previous transactions. This causes one failing test in `transaction.feature`:  Transfer valid amount should work. I created an issue for this here:  [#1645](https://github.com/dedis/popstellar/issues/1645)


matteosz (approved):
Very nice and clean work, also very well documentated! 😄

tests/karate/src/test/java/be/model/Transaction.java
```diff
@@ -0,0 +1,211 @@
...
+   * @return the transaction id computed based on the previous inputs and outputs.
+   */
+  public String computeId() {
+    // Make a list all the string in the transaction
+    List<String> collectTransaction = new ArrayList<>();
+    // Add them in lexicographic order
+
+    // Inputs
+    for (Input currentTxin : inputs) {
+      System.out.println("computing input: " + currentTxin);
```
GabrielFleischer, line 103: Is this some debug print or logs ?

It this is debug, I suggest you remove it.

If this is a log, you should probably be more verbose.

GabrielFleischer (changes requested):
Very great work !

I found some small things here and there, nothing too big.

tests/karate/src/test/java/be/model/Transaction.java
simone-kalbermatter: It was a debug print, removed it :)

tests/karate/src/test/java/be/decentralizedCom/getMessagesById.feature
```diff
@@ -0,0 +1,54 @@
...
+
+   # Check that after sending a heartbeat message with invalid message ids, the server does not request the messages
+  Scenario: Server should not request messages for invalid lao ids
+    Given def invalidMessageIds = []
+    And eval invalidMessageIds.push('invalid message id')
+    And eval heartbeat.params[lao.channel] = invalidMessageIds
+    When mockServer.send(heartbeat)
+    And def getMessagesByIdMessages = mockServer.getMessagesByMethod('get_messages_by_id')
+    Then assert getMessagesByIdMessages.length == 0
+
```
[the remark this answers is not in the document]
simone-kalbermatter: I think this would require a modification in the code to return the actual message id used to send a valid message. It's computed in `JsonConverter` and I don't see an easy way to access it here. So I won't do it in this PR, but later if I find some time : )

GabrielFleischer (approved):
LGTM

PR #1636, opened 2023-06-04
Fix high CPU usage + time not working

dayan9265:
This PR fixes issue with a simple fix: I think the issue was that when the token was generated the state would change ->  page would re-render -> token would be generated again -> re-render -> infinite loop

I don't know why the quality is so bad, but we can see the time updating:
Untitled video - Made with Clipchamp (1)


Tyratox (approved):
Oh wow, great work getting to the bottom of this issue!! 🥳🥳🥳

PR #1639, opened 2023-06-05
Fix overflow special character

dayan9265:
This small PR fixes the issue with characters such as Zalgo characters overflowing, now the screen looks like this:

image

Now these Zalgo characters are cropped, I tried using "big" characters that are used to check they are still correctly displayed and they are not cropped, so normally this shouldn't affect any "normal" text.


Tyratox (approved):
Very nice fix, I agree 😁

PR #1640, opened 2023-06-05
Highlight personal token rollcall + sort rollcall

dayan9265:
This PR adds the changes of @Tyratox to make sure that the list is sorted when sending it to the back-end, currently if it is not the case when receiving a non sorted list, it just creates a warning. Later when back-end will make sure it is sorted, we could make it throw if it is not sorted.
This PR also highlights personal token, to make it easier for the user to know his token and also if he is in the current roll call, currently it looks like this (in light blue being my token):
image


fe1-web/src/features/rollCall/objects/RollCall.ts
```diff
@@ -84,56 +84,79 @@ export class RollCall {
...
+        console.warn(
+          'Attendee list is not sorted alphabetically, be careful there is a risk of de-anonymization',
+        );
+      }
+    }
     this.attendees = obj.attendees;
 
-    this.openedAt = obj.openedAt;
-    this.closedAt = obj.closedAt;
+    // these fields are optional and do not have to satisfy any checks
```
MeKHell, line 158:
```suggestion
    // This field is optional and do not have to satisfy any checks
```

MeKHell (changes requested):
In general, I do not see any visual feedback on the UI when the list is not listed. Is it not one of the goal of this PR too ?
I think that the warning is not sufficient, since almost nobody logs its terminal.

dayan9265:
> In general, I do not see any visual feedback on the UI when the list is not listed. Is it not one of the goal of this PR too ? I think that the warning is not sufficient, since almost nobody logs its terminal.

Xelowak (approved):
LGTM

MeKHell (approved):
LGTM

PR #1641, opened 2023-06-05
Fix empty lines in ballot disappearing

dayan9265:
This simple PR fixes the fact that the ballot option would disappear if the line is empty, which could surprise the user, now it can have the empty line and there would be a message to say the user can't submit (message was already there)

image


Tyratox:
Is this fix not problematic as the last entry will always be empty (since it automatically creates a new empty one)?

Xelowak (approved):
Nice change in terms of UX!

PR #1643, opened 2023-06-05
Leaving roll call screen forgets attendees while scanning

### LINKED ISSUE #1629: [BUG] Leaving roll call screen forgets all previously scanned attendees
After scanning some attendees, leaving the screen and coming back to the process of scanning attendees, all previously scanned attendees will be forgotten

### Expected behavior
Previously scanned attendees should be remembered

### How to reproduce
1. Create roll call
2. Scan some attendees
3. Leave the screen (go the the social screen f.ex)
4. Go back to the roll call
5. Close the roll call
6. No attendees are inside the roll call

### Version & Environment
This bug was reproduced on:
- Commit: 71152f31d54630416f52d55b68ea3aea5575ebe5 (release branch)

##### Front-ends:
- [x] Fe1-Web (Firefox 113.0.2)
- [ ] Fe2-Android (haven't tested the behavior)
- [ ] Not applicable

##### Back-ends:
- [ ] Be1-Go
- [ ] Be2-Scala
- [x] Not applicable

 - OS: [Windows 11]
 - Browser: [Firefox 113.0.2]

### Workaround
Scan again the attendees

### Impact
Can create confusion and organizer needs to re-scan some attendees

### Possible root cause

MeKHell:
Fix #1629.

The goal of this PR is to keep in memory the scanned Pop tokens, even when the user leaves the opened roll call. See the referenced issue for the current behavior.

This PR: 
 - does not send any new message to the server
 - saves the scanned tokens into the Redux store as soon as they are passed to the single roll call screen

### DIFF
--- a/fe1-web/src/features/rollCall/components/RollCallOpen.tsx
+++ b/fe1-web/src/features/rollCall/components/RollCallOpen.tsx
@@ -55,7 +55,7 @@ const RollCallOpen = ({
   const [popToken, setPopToken] = useState('');
   const [hasWalletBeenInitialized, setHasWalletBeenInitialized] = useState(hasSeed());
 
-  const updateAttendeeList = () => {
+  const updateAttendeeList = useCallback(() => {
     // If it comes from the scanner then we take this list, otherwise we try if it has already been set
     const attendeesList = scannedPopTokens || rollCall.attendees || [];
     setPopTokens(attendeesList);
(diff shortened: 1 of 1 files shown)


dayan9265 (approved):
It looks good to me 👍

Tyratox:
I know this might be a nice fix but can cause all sorts of problems. At the moment all of fe-1 is designed in a way that the redux store is only updated with respect to network messages, not any local (non-persistent) changes.
Given this principle is very fundamental I would encourage you to look for another solution that does not require updating the redux store. This can be as simple as showing a warning when leaving the view that the changes will be lost.

@Xelowak Do you have a better idea or do you see this differently?

pierluca:
I fully agree with @Tyratox. This breaks the whole abstraction created by the architecture/design.

If redux is the way you want/need to go, you'd at least need to separate the temporary data in its own space, separate from the actual data we got from the network. This however introduces its own complexity.

MeKHell:
This bug will be fixed by @dayan9265 by adding a pop-up button with a message.

Issue #1576, opened 2023-05-18
[BUG] Roll Call with only organizer doesn't behave consistently 


MariemBaccari:
### Description (Actual behavior)
As an organizer, when you open a roll call and close it, your token will not be taken into account. However, if you open a roll call, click on the scan attendees button then close the roll call, your token will be taken into account which leads to inconsistencies in the behavior.

### Expected behavior
This has to be defined but it should be the same behavior. If a one-person LAO should be authorized, the expected behavior would be adding the organizer's token in the "scanned attendees" independently of whether the button is clicked. 

### How to reproduce
1. Create a LAO and a roll call 
2. Open the roll call 
3. Close it directly 
4. Notice that no scanned attendees tokens are shown

6. Create a new roll call (in the same or a different LAO)
7. Open the roll call
8. Click on the scan attendees button  
9. Close the roll call 
10. Notice that one scanned attendee token (the organizer's) is shown 

### Version & Environment
This bug was reproduced on: ALL-V1.12

##### Front-ends:
- [x] Fe1-Web (Chrome)
- [ ] Fe2-Android (please specify if phone or emulation, and Android version)

##### Environment (as applicable):
 - Device: [Laptop]
 - OS: [Windows 10]
 - Browser: [Chrome 113.0.5672.94]

### Workaround
We can click on the scan attendees button before closing the roll call if we want to have a one person LAO. 

### Impact
Minor as we mostly have pop parties with many attendees.

dayan9265:
I am not sure of a nice way to fix this, because:
- if we make that as soon as a roll call is created by the organizer, his token is added in the roll call, then at the start there would be only one token displayed (the one of the organizer), and so his token id would not be anonymous 
- if we refuse roll call with only one person, it would make the whole process of testing (when debugging or simply for new user who wants to explore the app alone) harder since it would require a second device to join the roll call

I think the current state is good, since if the organizer wants to be alone in his roll call he can do that, but he needs to do it voluntarily (by opening and closing the roll call)

MeKHell:
I think that a solution would be to add the organizer as soon as we open the roll call (not creating it) in the ViewSingleRollCall screen and not in the scanner one, it might have the advantages of both world. 

 - It is consistent, and the organizer does not need to do a useless action to add its own token.
 - If the organizer does not want to be added to the roll call, then he can not be the organizer even with the current behavior.
[...]

PR #1647, opened 2023-06-06
Fix inconsistency alone rollcall 

dayan9265:
This small PR fixes the issue with single roll call organizer, now it behaves as described by @MeKHell in the related issue (issue #1576)

### DIFF
--- a/fe1-web/src/features/rollCall/components/RollCallOpen.tsx
+++ b/fe1-web/src/features/rollCall/components/RollCallOpen.tsx
@@ -55,7 +55,7 @@ const RollCallOpen = ({
   const [popToken, setPopToken] = useState('');
   const [hasWalletBeenInitialized, setHasWalletBeenInitialized] = useState(hasSeed());
 
-  const updateAttendeeList = () => {
+  const updateAttendeeList = useCallback(() => {
     // If it comes from the scanner then we take this list, otherwise we try if it has already been set
     const attendeesList = scannedPopTokens || rollCall.attendees || [];
     setPopTokens(attendeesList);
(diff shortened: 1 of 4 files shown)


MeKHell:
I think that it might be better to remove the same behavior on the scanner since we will already add it on the roll call screen.

dayan9265:
> I think that it might be a good idea to remove the same behavior on the scanner since we will already add it on the roll call screen.

Good point, I removed this in the scanner

MeKHell (dismissed):
Fe comments if you have time.

Tyratox (approved):
Good catch!! 😄

Tyratox:
Oh wait, I approved the wrong PR, let me do an actual review sorry!

fe1-web/src/features/rollCall/components/RollCallOpen.tsx
```diff
@@ -48,10 +50,35 @@ const RollCallOpen = ({
...
+
+    setPopTokens(attendeesList);
+
+    // Stores the already scanned tokens in case we go to a new screen (different from the scanner)
+    const updatedRollCall = {
+      ...rollCall.toState(),
+      attendees: attendeesList.map((key) => key.toState()),
+    };
+
+    dispatch(updateRollCall(updatedRollCall));
```
Tyratox, line 74: This is problematic for the same reason described here: https://github.com/dedis/popstellar/pull/1643#issuecomment-1579422050

In fe-1, the redux store should only ever be updated because of network messages..

Tyratox (changes requested):

fe1-web/src/features/rollCall/screens/__tests__/__snapshots__/RollCallScanner.test.tsx.snap
```diff
@@ -576,7 +576,7 @@ exports[`RollCallOpened renders correctly when no scan 1`] = `
                                   >
                                     Scanned attendees
                                     : 
-                                    -1
+                                    0
```
Xelowak, line 579: This indeed makes more sense 😄.

dayan9265:
Close since it's using parts of code that we won't keep

PR #1649, opened 2023-06-06
Make popcha screen cleaner

dayan9265:
I made the popcha screen a bit cleaner (still pretty empty ^^) and more useful, now:
- Instead of just displaying the url on successful scan, it displays a success toast msg
- Possibility to add manually the url 
- Welcome text changed a bit (and now it disappears when opening the scanner)

To have an idea of the visual:

Opening screen:
image

Scanner screen:
image

Confirmation modal screen:
image


fe1-web/src/features/popcha/screens/PopchaScanner.tsx
```diff
@@ -33,7 +37,7 @@ const PopchaScanner = () => {
   const generateToken = PopchaHooks.useGenerateToken();
 
   const [showScanner, setShowScanner] = useState(false);
-  const [textScanned, setTextScanned] = useState('');
+  const [showInputModal, setInputModalIsVisible] = useState(false);
```
MeKHell, line 40:
```suggestion
  const [isInputModalVisible, setIsInputModalVisible] = useState(false);
```
I think it is better if the names are the same with a set in front of the setter.

fe1-web/src/features/popcha/screens/PopchaScanner.tsx
MeKHell, line 1: There is a lot of logic (in the UI part) based on the value of `showScanner`. Maybe a refactoring splitting both more clearly might be interesting.

MeKHell (commented):
Few comments, but the screens themselves looks very good.

fe1-web/src/features/popcha/screens/PopchaScanner.tsx
dayan9265: Indeed it could make it more clear, now it is completely separated depending on the value of `showScanner`

fe1-web/src/features/popcha/screens/PopchaScanner.tsx
```diff
@@ -148,42 +164,88 @@ const PopchaScanner = () => {
...
               <PoPTouchableOpacity
                 testID="popcha_scanner_button"
                 onPress={() => setShowScanner(!showScanner)}>
                 <Text style={[Typography.base, Typography.accent]}>
-                  {showScanner ? STRINGS.popcha_close_scanner : STRINGS.popcha_open_scanner}
+                  {STRINGS.popcha_open_scanner}
                 </Text>
               </PoPTouchableOpacity>
             </View>
+            <View />
```
Xelowak, line 248: ?

dayan9265:
> Also, you might have to reduce the complexity of `PoPChaScanner`

I agree that is shown as too complex, but I am not too sure how I should do it, should I put the data verification in an another file (if yes where should I put it)?

fe1-web/src/features/popcha/screens/PopchaScanner.tsx
```diff
@@ -148,42 +164,88 @@ const PopchaScanner = () => {
...
           sendAuthRequest(data).then((success) => {
             if (success) {
-              setTextScanned(data);
               setShowScanner(false);
+              showSuccessMessage(STRINGS.popcha_success_authentication);
             }
           })
         }>
+        <View style={styles.container}>
+          <View />
```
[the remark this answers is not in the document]
MeKHell: Can't you use flex `space-around` to achieve what you want ?

MeKHell (approved):
Added a small comment about your view.

fe1-web/src/features/popcha/screens/PopchaScanner.tsx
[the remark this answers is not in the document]
dayan9265: not sure how it would change to use `space-around` in this case?

pierluca:
@dayan9265 it looks like the logic (`sendAuthRequest` , `verifyScannedInfo`) is pretty much UI-independent. 
You could rewrite those to either succeed or throw on failure, which would make your life easier within the component.
Instead of having an inlined `.then` that needs to handle the `success` bool, you could just have a tiny function that wraps everything around a try/catch (or `then/catch`, depending on whether you use the async or promise syntax).

dayan9265:
> @dayan9265 it looks like the logic (sendAuthRequest , verifyScannedInfo) is pretty much UI-independent and could be moved outside the component. [...]

Indeed it would make the code inside the scanner easier to read, I have moved the logic inside an another file (`functions/network.ts`) and used throws instead of boolean

fe1-web/src/features/popcha/functions/network.ts
Tyratox, line 1: Maybe you could put these functions in `PopchaMessageApi` as well?

fe1-web/src/features/popcha/screens/__tests__/PopchaScanner.test.tsx
Tyratox, line 1: Really nice tests! 🤩

Tyratox (approved):
Great work! 😄 I would address the code smells aka the non-null assertions and the try/catch, then I am certainly happy :) From the rest of the comments see what you think makes sense. In case you need it merged quickly, I already approved it. Let me know if you need another review round / an approval :)

fe1-web/src/features/popcha/screens/PopchaScanner.tsx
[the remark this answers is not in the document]
MeKHell: If the top third needs to be empty, then it won't work. But using some % margins with flex-grow might work, I guess. But it might be even less readable than the current state.

fe1-web/src/features/popcha/functions/network.ts
dayan9265: I feel like maybe it would put too much code in the network file, I like the separation of:
- verification of valid URL in the /function folder
- creation of the arguments and sending the message in the /network folder
I don't know if it makes sense to you?

fe1-web/src/features/popcha/components/PopchaScannerClosed.tsx
```diff
@@ -0,0 +1,56 @@
...
+      </QrCodeScanner>
+    </>
+  );
+};
+
+const propTypes = {
+  // eslint-disable-next-line react/forbid-prop-types
+  containerStyle: PropTypes.any,
+  // eslint-disable-next-line react/forbid-prop-types
+  topMarginStyle: PropTypes.any,
```
dayan9265, line 44: For the style, I couldn't find a way to avoid the "eslint-disable", so I did something similar to the `PoPTouchableOpacity`

pierluca (approved):
Approving blindly to allow a merge anytime

PR #1651, opened 2023-06-07
Fix build info over left menu

### LINKED ISSUE #1592: [BUG] Release version number overlaps side menu
This happens on mobile and on desktop if the window is too small.

MeKHell:
Fix #1592

This PR puts the Build infos in the header background, so that they do not overlap anymore with the left menu.
Without menu:
image

With menu: 
image


Xelowak (approved):
LGTM

PR #1653, opened 2023-06-07
Update go documentation 

MariemBaccari:
This PR updates the go documentation by adding the new messages to the description and a new flowchart. Some subtle details aren't show on the flowchart to avoid making it unreadable but it should be enough to get a good idea of the big picture.


Nazianzenov (approved):
LGTM

Anghul (approved):
LGTM, small error in the flowchart where you wrote "an query" in one step, in case you want to fix it

PR #1655, opened 2023-06-07
Fix pb with already deleted chirp

dayan9265:
This small PR fixes the small problem of error msg shown in terminal ("reaction is undefined"), this was simply due that after deleting the reaction if the user receives again the delete info message, it would look for this reaction but since it was deleted, it no longer exists. To fix I simply check that the reaction is not undefined, before continuing


MeKHell (approved):
LGTM! So this simple change completely solve the problem ?

Tyratox (approved):
Nice, LGTM 😄

PR #1656, opened 2023-06-07
Fix roll call issues

MeKHell:
This issue addresses the following bugs: 
 - Unable to close rollCall after reopening. This was due to the fact we were using only the scannedTokens array to display.
 - Add the organizer token as soon as the roll call is open.


fe1-web/src/features/rollCall/screens/__tests__/ViewSingleRollCall.test.tsx
```diff
@@ -135,6 +143,30 @@ describe('EventRollCall', () => {
...
+              />
+            </FeatureContext.Provider>
+          </Provider>,
+        );
+        fireEvent.press(getByTestId('roll_call_close_button'));
+        expect(requestCloseRollCall).toHaveBeenCalledTimes(1);
+        expect(requestCloseRollCall).toHaveBeenCalledWith(
+          expect.anything(),
+          expect.anything(),
+          ATTENDEES.map((attendee) => PublicKey.fromState(attendee)),
```
pierluca, line 167: this doesn't really test the absence of duplicates, does it ?

fe1-web/src/features/rollCall/components/RollCallOpen.tsx
```diff
@@ -71,20 +80,13 @@ const RollCallOpen = ({
...
-  }, [toast, navigation, rollCall, scannedPopTokens]);
+  }, [toast, navigation, rollCall, allAttendees]);
 
   const onCloseRollCall = useCallback(async () => {
-    // get the public key as strings from the existing rollcall
-    const previousAttendees = rollCall.attendees || [];
-    // add the create a set of all attendees (takes care of deduplication)
-    const allAttendees = new Set([...previousAttendees, ...(scannedPopTokens || [])]);
-    // convert it back to a list
-    const attendeesList = [...allAttendees];
```
pierluca, line 86: how do you ensure deduplication without this logic? 🤔

fe1-web/src/features/rollCall/screens/__tests__/ViewSingleRollCall.test.tsx
```diff
@@ -135,6 +143,30 @@ describe('EventRollCall', () => {
         're-opened roll calls',
         testRender(mockRollCallReopened, false, ['attendee1', 'attendee2']),
       );
+      it('while closing roll call', () => {
```
dayan9265, line 146: name not really explicit of what it is testing

dayan9265 (approved):
Except small detail in test name, it looks good to me

fe1-web/src/features/rollCall/components/RollCallOpen.tsx
```diff
@@ -52,6 +52,15 @@ const RollCallOpen = ({
...
   const [hasWalletBeenInitialized, setHasWalletBeenInitialized] = useState(hasSeed());
+  const allAttendees = useMemo(() => {
+    if (isOrganizer && popToken !== '') {
+      if (scannedPopTokens && scannedPopTokens.length > 0) {
+        return scannedPopTokens;
+      }
+      return [PublicKey.fromState(popToken)];
+    }
+    return scannedPopTokens || [];
+  }, [isOrganizer, popToken, scannedPopTokens]);
```
Tyratox, line 63: I would at least add a comment that you want to add the organizer's key if it is not already present. Otherwise this looks a bit confusing as you implicitly assume that the organizer key is always present if `scannedPopTokens.length > 0` (which is okay but might no be clear without any previous knowledge)

fe1-web/src/features/rollCall/components/RollCallOpen.tsx
MeKHell: Because we control the scanner. `ScannedPopTokens ` is non-empty, only if we come back from the scanner, which already has the scanned attendees. Then the scanner takes to only add a new token if it is not already in the list.

Tyratox: I think this is now handled by always passing around the full list of tokens and the scanner preventing to scan any duplicates, right @MeKHell? 😅

fe1-web/src/features/rollCall/screens/__tests__/ViewSingleRollCall.test.tsx
MeKHell: It does since we do not use `expect.arrayContains` but give the exact array.

fe1-web/src/features/rollCall/screens/__tests__/ViewSingleRollCall.test.tsx
MeKHell: The name is concatenated with the above `describe`. Wich leads to `no duplicate attendees while closing roll call`.

Tyratox (commented):
One small question and I agree with @dayan9265 @pierluca comments that it is unclear what is tested in `'while closing roll call'`.
Otherwise LGTM:)

fe1-web/src/features/rollCall/screens/__tests__/ViewSingleRollCall.test.tsx
pierluca: I must be missing something, but I'm going to trust you (and talk later 😅 )

pierluca (dismissed):
Approving blindly, but merge conflicts still need to be resolved.

fe1-web/src/features/rollCall/components/RollCallOpen.tsx
MeKHell: Yes

PR #1659, opened 2023-06-08
Fix sorting problem

dayan9265:
This PR addresses the issue of sorting not working as expected with uppercase and lowercase, it comes from the fact that the `localCompare()` doesn't work the same as `<` with strings with lowerCase and upperCase letters. (I initially used `localCompare`, because SonarCloud didn't like just writing .sort()). So now I always compare using `localCompare` and I added a test that failed with the previous code, to hopefully catch this mistake if it happens again ^^


MeKHell (approved):
LGTM!

Tyratox:
You should be able to use `.sort()` but I think what SonarCloud doesn't like is if you do `return array.sort()` because that also sorts the original array. In theory you should be able to do `return [...array].sort()`, or, if  you are okay with sorting the original array:

```typescript
array = ...

array.sort()
return array
```

The one thing I am worried about when using `localCompare` is that it might return different results depending on the system's locale in some cases.

pierluca:
Sonar complains about `.sort()` for valid reasons that are not applicable here (because we're dealing with base64 data, thus ASCII strings)
See https://sonarcloud.io/organizations/dedis/rules?open=typescript%3AS2871&rule_key=typescript%3AS2871

I think it'd be fine marking this as a false positive in Sonar. What do you think ?

dayan9265:
> In any case, it should be clarified in the specification if you're sorting alphabetically (in a case-sensitive way) or sorting by underlying binary data (i.e. capital letters come first). I'm sure @K1li4nL would love to know this, because it'll radically change his implementation 😁

I added the sorting order in the json schema of dataCloseRollCall

Xelowak (approved):
LGTM!
I also marked the bugs as "Info" in SonarCloud.

PR #1660, opened 2023-06-08
Block invalid poptoken

dayan9265:
This PR blocks invalid pop token when scanning (this avoid bugs with wrong channel created, and then when trying to connect to these channel it fails, which could lead to the app thinking it is offline)

PS: the updated snap test is due to the current state of the release not being correct

### DIFF
--- a/fe1-web/src/features/digital-cash/screens/SendReceive/SendReceive.tsx
+++ b/fe1-web/src/features/digital-cash/screens/SendReceive/SendReceive.tsx
@@ -349,7 +349,7 @@ export const SendReceiveHeaderRight = () => {
 
           <View>
             <QRCode
-              value={serializedPopToken || ''}
+              value={serializedPopToken}
               overlayText={STRINGS.digital_cash_wallet_qrcode_text}
             />
           </View>
--- a/fe1-web/src/features/digital-cash/screens/SendReceive/SendReceive.tsx
+++ b/fe1-web/src/features/digital-cash/screens/SendReceive/SendReceive.tsx
@@ -8,6 +8,7 @@ import {
   TouchableOpacity,
   TouchableWithoutFeedback,
 } from 'react-native-gesture-handler';
+import { useToast } from 'react-native-toast-notifications';
 import { useSelector } from 'react-redux';
 
 import {
@@ -26,6 +27,7 @@ import { DigitalCashParamList } from 'core/navigation/typing/DigitalCashParamLis
 import { Hash, PublicKey } from 'core/objects';
 import { ScannablePopToken } from 'core/objects/ScannablePopToken';
 import { Color, Icon, ModalStyles, Spacing, Typography } from 'core/styles';
+import { FOUR_SECONDS } from 'resources/const';
 import STRINGS from 'resources/strings';
 
 import { DigitalCashHooks } from '../../hooks';
@@ -290,6 +292,8 @@ export const SendReceiveHeaderRight = () => {
   const route = useRoute<NavigationProps['route']>();
   const laoId = DigitalCashHooks.useCurrentLaoId();
 
+  const toast = useToast();
+
   const { rollCallId: serializedRollCallId, isCoinbase } = route.params;
   const rollCallId = useMemo(
     () => (serializedRollCallId ? new Hash(serializedRollCallId) : undefined),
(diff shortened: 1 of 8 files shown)


fe1-web/src/features/rollCall/screens/RollCallScanner.tsx
Tyratox, line 1: At some time in the past we had something like that already .. 

https://github.com/dedis/popstellar/blob/6a6367459ae4e6dfe3eb1509f6eb6f4a5d00b00c/fe1-web/src/features/rollCall/screens/RollCallOpened.tsx#L34

It might be nice to put this validation into the constructor of ScannablePopToken?

Tyratox (approved):
LGTM! 😄We already had something like this in the past, I marked it above. Feel free to check it out or just merge this one. I think the other regex was even a bit more strict because the pop tokens are all of the same size :)

fe1-web/src/features/rollCall/screens/RollCallScanner.tsx
```diff
@@ -111,9 +111,13 @@ const RollCallOpened = () => {
 
   const addAttendeePopToken = useCallback(
     (popToken: string) => {
+      // if popToken is not in base64 format, do not add it
+      if (!popToken.match(/^[a-zA-Z0-9_-]+={0,3}$/) || popToken.length % 4 !== 0) {
+        throw new Error('Invalid popToken format');
+      }
+
```
MeKHell, line 118: I think it is better if this is moved to at least another function or the `ScannablePopToken.fromJson` which is used line 144

dayan9265:
It was indeed a better idea to put the verification inside ScannablePopToken, so I moved it. (I let the tests inside the scanner, since we want to make sure that wrong tokens are refused)

dayan9265:
Okay it seems that there's now a risk when calling the ScannablePopToken generator with empty strings (such as in the RollCallOpen), I saw there is an another place where it could be the case:

https://github.com/dedis/popstellar/blob/release-2023-spring/fe1-web/src/features/digital-cash/screens/SendReceive/SendReceive.tsx#LL332C1-L334C15

@Tyratox since this is code that you had written, do you know if it can cause an issue?

fe1-web/src/features/rollCall/components/RollCallOpen.tsx
```diff
@@ -182,7 +182,11 @@ const RollCallOpen = ({
           <Text style={Typography.paragraph}>{STRINGS.roll_call_open_attendee}</Text>
           <View>
             <QRCode
-              value={ScannablePopToken.encodePopToken({ pop_token: popToken })}
+              value={
+                popToken !== ''
+                  ? ScannablePopToken.encodePopToken({ pop_token: popToken })
+                  : undefined
```
pierluca, line 188: although I don't see how, this technically could throw.
Would it make sense to do this operation beforehand and wrap it like you did in digital cash ?

dayan9265: this could throw since the popToken is initially empty, and it is changed through a promise (`generateToken`), so for a brief moment the popToken will be an empty string
I will wrap it earlier 👍

fe1-web/src/features/digital-cash/screens/SendReceive/SendReceive.tsx
```diff
@@ -300,6 +304,22 @@ export const SendReceiveHeaderRight = () => {
 
   const popToken = useMemo(() => rollCallToken?.token.publicKey.valueOf() || '', [rollCallToken]);
 
+  const serializedPopToken = useMemo(() => {
+    try {
+      return ScannablePopToken.encodePopToken({ pop_token: popToken });
+    } catch {
+      return null;
+    }
```
Xelowak, line 312: Why don't you return an empty string there directly then (in catch)?
It seems that you return null to convert it in an empty string after if it's null.

Xelowak (dismissed):
LGTM! 
I made a small comment but I won't be blocking you right now so feel free to merge.

fe1-web/src/features/digital-cash/screens/SendReceive/SendReceive.tsx
dayan9265: The idea was that if the popToken is null, it displays nothing (the return null after the toast show). In reality the `|| ''` condition will never be used (I have removed it now)

pierluca (approved):
Approving blindly. Can't review right now.

Xelowak (approved):
It's ok for me.

PR #1661, opened 2023-06-08
Fix new chirp modal

MeKHell:
This PR fixes the issues with the modal not closing after chirp submission. 

It also adds a new error message, when the user inputs an invalid string. (This is better than a modal window, since we need some live feedback here).
EDIT: We do not see it on the picture, but the input box contains whitestring text.
image

Finally, it adds few tests.

### DIFF
--- a/fe1-web/src/features/social/components/TextInputChirp.tsx
+++ b/fe1-web/src/features/social/components/TextInputChirp.tsx
@@ -38,11 +38,14 @@ const styles = StyleSheet.create({
     borderRadius: Border.radius,
     borderWidth: Border.width,
   } as TextStyle,
-  buttonView: {
+  bottomRow: {
     flexDirection: 'row',
-    alignItems: 'center',
-    alignSelf: 'flex-end',
     marginTop: Spacing.x1,
+  } as ViewStyle,
+  buttonView: {
+    marginLeft: 'auto',
+    flexDirection: 'row',
+    alignContent: 'flex-end',
   } as ViewStyle,
   charsLeft: {
     marginRight: Spacing.x1,
@@ -58,6 +61,7 @@ const TextInputChirp = (props: IPropTypes) => {
     onChangeText,
     disabled: alwaysDisabled,
     currentUserPublicKey,
+    errorMessage,
     testID,
   } = props;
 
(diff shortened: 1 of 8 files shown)


dayan9265 (dismissed):
It looks good to me, detail: I think it would have been nicer to include the text inside the whitebox of the chirp (either next to the number or completely on the left), but that's a detail, it is nice to display a message to inform the user 👍

fe1-web/src/features/social/components/__tests__/NewChirp.test.tsx
```diff
@@ -81,9 +82,51 @@ describe('NewChirp', () => {
...
+    fireEvent.changeText(getByTestId('new_chirp_input'), mockText);
+    fireEvent.press(getByTestId('new_chirp_publish'));
+
+    expect(getByTestId('confirm-modal-confirm')).toBeDefined();
+    expect(toJSON()).toMatchSnapshot();
+
+    // Accept the modal message
+    fireEvent.press(getByTestId('confirm-modal-confirm'));
+
+    expect(await findByText('300')).not.toBeNull();
```
Tyratox, line 111: The `300` is not ideal but I don't have a better idea right this second sooo 👍

fe1-web/src/features/social/components/__tests__/NewChirp.test.tsx
```diff
@@ -81,9 +82,51 @@ describe('NewChirp', () => {
...
+          </SocialMediaContext.Provider>
+        </FeatureContext.Provider>
+      </Provider>,
+    );
+
+    const mockText = '   \t\n \r\n  ';
+
+    fireEvent.changeText(getByTestId('new_chirp_input'), mockText);
+
+    expect(queryByText(STRINGS.social_media_empty_chirp)).not.toBeNull();
```
Tyratox, line 129: This is actually an elegant solution with the string constant 😄

Tyratox (approved):
I agree with @dayan9265 but otherwise LGTM :)

pierluca:
Related but separate concern: "trim" to remove whitespace is a technical term and it's unlikely a user would understand what the modal is saying. I'd suggest making the modal more explicit in that regard, something along the lines of "It seems you've used more spacing characters than needed. To improve its readability, the following message will be sent: " 
I'm not entirely convinced by my phrasing, it's just a rough draft, but you get the idea...

fe1-web/src/features/social/components/__tests__/NewChirp.test.tsx
MeKHell: Tried a new one, but not necessary. better

dayan9265 (approved):
Nice changes ! LGTM 👍

fe1-web/src/features/social/components/TextInputChirp.tsx
```diff
@@ -94,19 +98,22 @@ const TextInputChirp = (props: IPropTypes) => {
           onFocus={() => setFocused(true)}
           onBlur={() => setFocused(false)}
         />
-        <View style={styles.buttonView}>
-          <View style={styles.charsLeft}>
-            <Text style={textIsRed ? [Typography.base, Typography.error] : [Typography.base]}>
-              {charsLeft.toString()}
-            </Text>
+        <View style={styles.bottomRow}>
+          {errorMessage && <Text style={[Typography.base, Typography.error]}>{errorMessage}</Text>}
```
Xelowak, line 102: Is this && really needed? I mean, if the next is undefined, it will not show anything anyway, right?

Xelowak (approved):
Looks good to me, I just did a minor comment.
Since you may need it for tomorrow, I prefer to not block you and approve.

So now there's this red text appearing + a modal when you try to send the chirp?

fe1-web/src/features/social/components/TextInputChirp.tsx
MeKHell: Won't the Text component still be built, resulting in empty divs ?

MeKHell:
No, there is a red error message if the input would be trimmed to an empty string and a modal with the trimmed message otherwise.

PR #1662, opened 2023-06-08
Fix meeting past date

dayan9265:
This PR fixes bug where meeting that had not ended would appear as past event, this happened because there was no verification on the end time (other then it was defined)

This works for all types of event expect Meetings (since for the other events, it is set when the organizer ends the event), but meetings are not ended by the organizer

This fix is really not clean, and in the future a better way to handle this should be found, one possibility would be that meetings like other events ends when organizer decides to, or we would need an additional field to know if an event has ended.

(I also had to use hardcoded 'MEETING' since I can not import in this module Meeting.id)

### DIFF
--- a/fe1-web/src/features/events/functions/events.ts
+++ b/fe1-web/src/features/events/functions/events.ts
@@ -29,9 +29,7 @@ export const categorizeEventsByTime = (time: Timestamp, events: EventState[]) =>
     // TODO: should find a way to do this without relying on the event type
     // for example, by making that events ends when organizers mark them as ended (same as the others)
     if (
-      (e.eventType === 'MEETING' &&
-        e.end &&
-        e.end <= t) ||
+      (e.eventType === 'MEETING' && e.end && e.end <= t) ||
       (e.eventType !== 'MEETING' && e.end)
     ) {
       pastEvents.push(e);
--- a/fe1-web/src/features/events/functions/events.ts
+++ b/fe1-web/src/features/events/functions/events.ts
@@ -24,7 +24,7 @@ export const categorizeEventsByTime = (time: Timestamp, events: EventState[]) =>
 
   events.forEach((e: EventState) => {
     // ended events are either:
-    // - meetings that have ended more than {CURRENT_EVENTS_THRESHOLD_HOURS} hours ago
+    // - meetings that have ended
     // - other events that have been marked as ended (e.end is set)
     // TODO: should find a way to do this without relying on the event type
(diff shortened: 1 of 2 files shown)


fe1-web/src/features/events/functions/events.ts
```diff
@@ -23,8 +23,17 @@ export const categorizeEventsByTime = (time: Timestamp, events: EventState[]) =>
...
-    if (e.end) {
+    // ended events are either:
+    // - meetings that have ended more than {CURRENT_EVENTS_THRESHOLD_HOURS} hours ago
+    // - other events that have been marked as ended (e.end is set)
+    // TODO: should find a way to do this without relying on the event type
+    // for example, by making that events ends when organizers mark them as ended (same as the others)
+    if (
+      (e.eventType === 'MEETING' &&
+        e.end &&
+        e.end <= t - CURRENT_EVENTS_THRESHOLD_HOURS * 60 * 60) ||
```
Tyratox, line 34: Probably just
```suggestion
        e.end <= t) ||
```
?

Current events is for events where the start time is in [now, now + CURRENT_EVENTS_THRESHOLD_HOURS], i.e. events that are about to start.

fe1-web/src/features/events/functions/events.ts
```diff
@@ -23,8 +23,15 @@ export const categorizeEventsByTime = (time: Timestamp, events: EventState[]) =>
...
-    // if end time is set, the event has ended
-    if (e.end) {
+    // ended events are either:
+    // - meetings that have ended
+    // - other events that have been marked as ended (e.end is set)
+    // TODO: should find a way to do this without relying on the event type
+    // for example, by making that events ends when organizers mark them as ended (same as the others)
+    if (
+      (e.eventType === 'MEETING' && e.end && e.end <= t) ||
+      (e.eventType !== 'MEETING' && e.end)
```
pierluca, line 33: why not just keep the `e.end && e.end <= t` logic ? 
In a way isn't it weird when a roll call or an election immediately jumps into "past events" the moment it's closed (i.e. the results are available) ?

dayan9265: I guess it would make sense, I initially didn't want to do this, because it would change the general behavior we're used to regarding roll call or election.

MeKHell: I agree with @pierluca. It is better if we have the same behaviour for all event types.

MeKHell (approved):
LGTM!

Xelowak (approved):
LGTM!

PR #1665, opened 2023-06-09
Welcome message on event screen

dayan9265:
This PR simply adds a welcome message if no events were created (messages are the same as the one displayed by front-end 2)


Xelowak (approved):
NIce addition!

PR #1667, opened 2023-06-10
Handle social channel creation error for invalid pop_token

K1li4nL:
When the rollcall closes and social channels are created for attendees, no guarantees are given regarding their pop token beside them being valid base 64 strings. The empty string is also considered valid, but when creating the channel: `/root/lao_id/social/[empty string]` we get an unhandled exception because this is not a valid channel. This pr deal with that by simply ignoring such invalid channels.

### DIFF
--- a/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
@@ -125,7 +125,7 @@ class RollCallHandler(dbRef: => AskableActorRef) extends MessageHandler {
           Some((generateSocialChannel(rpcRequest.getParamsChannel, attendee), ObjectType.CHIRP))
         } catch {
           case _: Throwable =>
-            println(s"Failed to create attendee channel for pop token: ${attendee.base64Data.data}")
+            println(s"Failed to generate attendee social channel for pop token: \"${attendee.base64Data.data}\"")
             None
         }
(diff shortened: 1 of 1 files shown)


be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/RollCallHandler.scala
```diff
@@ -119,8 +119,15 @@ class RollCallHandler(dbRef: => AskableActorRef) extends MessageHandler {
...
     val data: CloseRollCall = message.decodedData.get.asInstanceOf[CloseRollCall]
-    val listAttendeeChannels: List[(Channel, ObjectType.ObjectType)] = data.attendees.map {
-      attendee => (generateSocialChannel(rpcRequest.getParamsChannel, attendee), ObjectType.CHIRP)
+    val listAttendeeChannels: List[(Channel, ObjectType.ObjectType)] = data.attendees.flatMap {
+      attendee =>
+        try {
+          Some((generateSocialChannel(rpcRequest.getParamsChannel, attendee), ObjectType.CHIRP))
+        } catch {
+          case _: Throwable =>
+            println(s"Failed to create attendee channel for pop token: ${attendee.base64Data.data}")
```
pierluca, line 128: I would add quotes around the data, it might be more explicit if you're trying to display an empty string :-)

pierluca (approved):

pierluca (approved):
Approving blindly. Can't review right now.

PR #1670, opened 2023-06-13
Fixed wrong LAO ID for authentication response.

Nazianzenov:
also fixed wrong encoding for nonce in the authentication message

### DIFF
--- a/be1-go/channel/authentication/mod.go
+++ b/be1-go/channel/authentication/mod.go
@@ -132,7 +132,7 @@ func (c *Channel) Publish(publish method.Publish, socket socket.Socket) error {
 }
 
 // Catchup is used to handle a catchup message.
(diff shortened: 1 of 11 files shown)


MariemBaccari (approved):
LGTM

Tyratox (approved):
fe1 part looks good to me although I do not know why you switch from simply a string to to base64 encoding said string?

Nazianzenov:
> fe1 part looks good to me although I do not know why you switch from simply a string to to base64 encoding said string?

During test, the signature of the nonce string was never valid. Encoding it into base64 seems to have fixed it. It probably comes from the fact that in Go, to verify the signature, we have to cast the string into a byte array. This cast is not necessary with the base64 decoding.

PR #1672, opened 2023-06-14
Implement blacklist

MariemBaccari:
This PR implements a simple blacklist that's used when handling the answer to getMessagesById to avoid blocking the hub. 
If it fails to handle a message after 10 tries it will put its id in the blacklist and not ask for it / avoid processing it again.

### DIFF
--- a/be1-go/hub/standard_hub/mod.go
+++ b/be1-go/hub/standard_hub/mod.go
@@ -95,6 +95,12 @@ type Hub struct {
 
 	// peersGreeted stores the peers that were greeted by the socket ID
 	peersGreeted []string
+
+	// blacklist stores the IDs of the messages that failed to be processed by the hub
+	// the server will not ask for them again in the heartbeat
+	// and will not process them if they are received again
+	// @TODO remove the messages from the blacklist after a certain amount of time by trying to process them again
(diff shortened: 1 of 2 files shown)


be1-go/hub/standard_hub/mod.go
```diff
@@ -95,6 +95,12 @@ type Hub struct {
 
 	// peersGreeted stores the peers that were greeted by the socket ID
 	peersGreeted []string
+
+	// blacklist stores the IDs of the messages that failed to be processed by the hub
+	// the server will not ask for them again in the heartbeat
+	// and will not process them if they are received again
+	// @TODO remove the messages from the blacklist after a certain amount of time by trying to process them again
```
Nazianzenov, line 102: there is a TODO here, i suppose it is left for future work?

Nazianzenov (changes requested):
LGTM!

Nazianzenov (approved):

be1-go/hub/standard_hub/mod.go
MariemBaccari: Yes !

Anghul (approved):
LGTM

PR #1674, opened 2023-06-20
fixed response mode handling

Nazianzenov:
* response mode can be selected between fragment and query
* the server was not handling correctly the fragment response mode, and was always using the query one.


MariemBaccari (approved):
LGTM

jbsv (approved):
Looks Good To Me

PR #1679, opened 2023-06-20
Properly handle authentication message to complete popcha authentication

Jharaxus:
- Adds implementation for `Authentication` message handling
- Adds public/private key rsa pair and rsa pair generation script in `src/security` folder
- Adds dummy public/private key rsa pairs in `src/security` and `src/security/test` folders until they are automatically generated during the deployment
- Adds `-Dscala.security` mandatory parameter to indicate the server's keys folder location
- Adds `SecurityModuleActor` to handle logic related to server's private keys in a (to a certain extent) secure way (the private keys never leave the security module)


be2-scala/src/security/test/private_key.pem
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
pierluca, line 1: do you really want to commit these keys ?

be2-scala/src/security/generateKeys.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
```
pierluca, line 1: You might want to add some documentation about this (and the rest) in the docs.

pierluca (commented):
This is not an in-depth review, I just had a very quick look at a couple of files.

This probably breaks Karate tests. You might want to have a look at those as well.

be2-scala/src/test/scala/ch/epfl/pop/authentication/WebSocketResponseHandlerSuite.scala
```diff
@@ -0,0 +1,40 @@
+package ch.epfl.pop.authentication
+
+import akka.http.scaladsl.testkit.{ScalatestRouteTest, WSProbe}
+import ch.epfl.pop.config.RuntimeEnvironment
+import org.scalatest.funsuite.AnyFunSuite
+import org.scalatest.matchers.should.Matchers
+
+class WebSocketResponseHandlerSuite extends AnyFunSuite with Matchers with ScalatestRouteTest {
```
Ajkunas, line 8: You should include the Testkit with the name of the class of test for more clarity (as it is done in the other files) :)

be2-scala/src/test/scala/ch/epfl/pop/storage/FakeSecurityModuleActor.scala
```diff
@@ -0,0 +1,36 @@
...
+import akka.actor.Actor
+import akka.event.LoggingReceive
+import ch.epfl.pop.storage.FakeSecurityModuleActor.{rsaPrivateKey, rsaPublicKey, rsaPublicKeyPem}
+import ch.epfl.pop.storage.SecurityModuleActor.{ReadRsaPublicKey, ReadRsaPublicKeyAck, ReadRsaPublicKeyPem, ReadRsaPublicKeyPemAck, SignJwt, SignJwtAck}
+import com.auth0.jwt.algorithms.Algorithm
+
+import java.security.interfaces.{RSAPrivateKey, RSAPublicKey}
+import java.security.{KeyPair, KeyPairGenerator}
+
+final case class FakeSecurityModuleActor() extends Actor {
```
Ajkunas, line 12: I am not sure of the relevance of this entire fake actor. You repeat some of the functionalities. Is there not a simpler way to do the similar things?

Ajkunas (dismissed):
There are still some points to clarify, and some things that you maybe could do in a simpler way, but otherwise it is well done. See the corresponding comments of @pierluca and myself. 
ps: I briefly went through the tests, thus I could have missed something.

be2-scala/src/main/scala/ch/epfl/pop/storage/SecurityModuleActor.scala
```diff
@@ -0,0 +1,110 @@
...
+import java.security.interfaces.{RSAPrivateKey, RSAPublicKey}
+import java.security.spec.{PKCS8EncodedKeySpec, X509EncodedKeySpec}
+import scala.io.Source
+import scala.util.{Success, Try}
+
+/** Actor to use for logic involving the server's private keys. Only the public keys can exit the module, on the other hand private keys must stay here and can only be used through specific endpoint
+  * @param keysFolderPath
+  *   folder where the keys are stored at
+  */
+final case class SecurityModuleActor(keysFolderPath: String) extends Actor with ActorLogging {
```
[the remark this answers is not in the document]
Jharaxus: For now there is only one file that has a "security" behavior so I believed it was easier and less heavy to put it in the already existing storage module (it is in a sense a "secured storage"), but indeed that could be a good idea once there are more "security" files !

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/validators/PopchaValidator.scala
```diff
@@ -41,20 +38,13 @@ sealed class PopchaValidator(dbActorRef: => AskableActorRef) extends MessageData
...
           _ <- checkAttendee(rpcMessage, sender, channel, dbActorRef, validationError(s"User doesn't belong to the requested lao $laoId"))
-          _ <- checkIdentifierProof(
+          result <- checkIdentifierProof(
             rpcMessage,
             authenticate.identifier,
             authenticate.identifierProof,
             authenticate.nonce,
             validationError("Failed to verify the identifier proof with the given identifier/nonce pair")
           )
-          result <- checkNoDuplicateIdentifiers(
```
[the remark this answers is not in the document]
Jharaxus: Not really, it was removed here because the check was moved to `PopchaHandler` (in `registerAuthentication` to be precise) to avoid redundant checks and calls to the DB that would otherwise need to be duplicated in both the handler and the verifier.

be2-scala/src/test/scala/ch/epfl/pop/storage/SecurityModuleActorSuite.scala
```diff
@@ -0,0 +1,41 @@
...
+import com.auth0.jwt.algorithms.Algorithm
+import org.scalatest.BeforeAndAfterAll
+import org.scalatest.funsuite.AnyFunSuiteLike
+import org.scalatest.matchers.should.Matchers
+
+import scala.concurrent.Await
+
+class SecurityModuleActorSuite extends TestKit(ActorSystem("SecurityModuleActorSystem")) with AnyFunSuiteLike with Matchers with BeforeAndAfterAll with AskPatternConstants {
+
+//  private val testSecurityDirectory = "/Users/hugo/Documents/EPFL/BA6/PoP/popstellar/be2-scala/src/security/test/"
```
K1li4nL, line 18: We should probably delete that line

be2-scala/src/test/scala/ch/epfl/pop/storage/FakeSecurityModuleActor.scala
Jharaxus: Indeed I removed it from the project as it would also be hard to maintain because it could not inherit `SecurityModuleActor` which is sealed for security.

be2-scala/src/security/generateKeys.sh
Jharaxus: It is now done !

be2-scala/src/security/test/private_key.pem
Jharaxus: They are now removed by default and need to be generated before running the server (as documented in the `readme` now).

be2-scala/src/test/scala/ch/epfl/pop/storage/SecurityModuleActorSuite.scala
Jharaxus: Indeed 😅

be2-scala/src/main/scala/ch/epfl/pop/authentication/PopchaWebSocketResponseHandler.scala
```diff
@@ -0,0 +1,82 @@
...
+import akka.{Done, NotUsed}
+import ch.epfl.pop.config.ServerConf
+
+import scala.concurrent.{Future, Promise}
+
+/** Websocket handler that, given two websocket connections on a common endpoint, forwards a single message from the second websocket connected to the first:
+  *   - Connection requests are expected to lend on endpoints with format "/laoId/authentication/clientId/nonce". This uniquely determines a connection by its socketId (laoId, clientId, nonce).
+  *   - The first websocket connected on a given endpoint is assumed to be waiting for a message. It will be provided by the second websocket to connect on that same endpoint (at any point later in time).
+  *   - Once a message is received from the second websocket connected on a given endpoint, it is forwarded to the first websocket connected and both connections are then closed.
+  */
```
K1li4nL, line 18: I think the code speaks for itself and the doc should just give a higher level explanation, what do you think ?

```suggestion
/** This websocket handler receives the JWT from the PopchaHandler via websocket and forwards it to the client webpage connected via websocket as well.
  */
```

K1li4nL (changes requested):
Good job, this pr looks ready to me ! I just have a tiny suggestion on the doc. Once the CI gets fixed I think we should be good to merge.

be2-scala/src/main/scala/ch/epfl/pop/authentication/PopchaWebSocketResponseHandler.scala
Jharaxus: Yes that makes sense !

K1li4nL (approved):
With the karate config updated, this pr is ready to be merged, LGTM !

PR #1681, opened 2023-06-21
replaced all relative paths for files by absolute, raw github URLS.

Nazianzenov:
Relative paths for retrieving resource such as the QR Code HTML page, or the key-pair int he /crypto package, might not work when the back-end is deployed, as the location of the binary can break those paths.Introducing URLs as absolute paths might give a constant, reliable way of retrieving the resources.


pierluca (commented):
This seems very weird. Are you sure this is what you want to do? 🤔

Nazianzenov:
> This seems very weird. Are you sure this is what you want to do? thinking

Using relative path doesnt seem to work on the deployed server. I couldn't find any other solution to have a constant way to retrieve the resources...

PR #1690, opened 2023-09-23
Fe2 CI got stuck

### LINKED ISSUE #1608: [BUG] CI tests of FE2 sometimes get stuck

matteosz:
This PR resolves #1608

### DIFF
--- a/fe2-android/app/src/test/unit/java/com/github/dedis/popstellar/repository/WitnessingRepositoryTest.java
+++ b/fe2-android/app/src/test/unit/java/com/github/dedis/popstellar/repository/WitnessingRepositoryTest.java
@@ -5,30 +5,50 @@ import android.app.Application;
 import androidx.test.core.app.ApplicationProvider;
 import androidx.test.ext.junit.runners.AndroidJUnit4;
 
-import com.github.dedis.popstellar.di.AppDatabaseModuleHelper;
 import com.github.dedis.popstellar.model.network.method.message.data.election.ElectionVersion;
 import com.github.dedis.popstellar.model.objects.*;
 import com.github.dedis.popstellar.model.objects.event.EventState;
 import com.github.dedis.popstellar.model.objects.security.MessageID;
 import com.github.dedis.popstellar.model.objects.security.PublicKey;
 import com.github.dedis.popstellar.repository.database.AppDatabase;
-import com.github.dedis.popstellar.repository.database.witnessing.PendingEntity;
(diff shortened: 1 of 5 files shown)


pierluca (dismissed):
This looks deceptively simple. I'm happy to approve, but I don't know if I'm missing some subtlety here 😅

pierluca:
The CI (still) seems broken :-/

K1li4nL (approved):
For what its worth: LGTM, I never developped on android but what I am reading make sense, at least mocking these elements should hopefully make it easier to debug if the problem  comes up again

fe2-android/app/src/test/unit/java/com/github/dedis/popstellar/repository/WitnessingRepositoryTest.java
```diff
@@ -67,11 +87,20 @@ public class WitnessingRepositoryTest {
...
           new ArrayList<>());
 
+  @Rule(order = 0)
+  public final MockitoRule mockitoRule = MockitoJUnit.rule();
+
   @Before
   public void setUp() {
     Application application = ApplicationProvider.getApplicationContext();
-    appDatabase =
-        AppDatabaseModuleHelper.getAppDatabase(ApplicationProvider.getApplicationContext());
```
simone-kalbermatter, line 74: I don't really understand this line, was the problem that it was not using a mock?

matteosz: The Room database instance created was shared among several thread instances running the tests in parallel, thus creating concurrency issues

PR #1691, opened 2023-09-24
Patch for using SDK 17 on the CI

matteosz:
SDK11 will stop working soon on the CI


K1li4nL (dismissed):
LGTM, will be doing the same for scala soon

pierluca:
@K1li4nL 's comment makes sense, I have to say 😅

K1li4nL (approved):
I quickly checked the versions of the package + the CI built and ran everything correctly, I think we are good to go

PR #1692, opened 2023-09-25
CI java update

K1li4nL:
Java 11 will soon go unsupported on github CI. Java 17 should be good for at least 3 more years. The java requirement have been updated in the doc so that the CI env is closer to the local one.

### DIFF
--- a/be2-scala/README.md
+++ b/be2-scala/README.md
@@ -95,8 +95,8 @@ environment:
 ```sh
 # install sdkman
 curl -s "https://get.sdkman.io" | bash
-# install java zulu 17
-sdk install java 17.0.8-zulu
+# install java temurin 17
+sdk install java 17.0.8-tem
 # install sbt
 sdk install sbt
 # check that everything went well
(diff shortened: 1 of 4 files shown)


be2-scala/README.md
```diff
@@ -95,13 +95,13 @@ environment:
 ```sh
 # install sdkman
 curl -s "https://get.sdkman.io" | bash
-# install java zulu 11
-sdk install java 11.0.18-zulu
+# install java zulu 17
+sdk install java 17.0.8-zulu
```
pierluca, line 99: general question:
do we want to stick with the zulu distribution ?
https://sdkman.io/jdks

K1li4nL: There is this website: https://whichjdk.com

Given their opinion we should go preferably with Temurin but Zulu seems fine. As the website explains multiple companies are behind the Eclipse Foundation where Azul are the only one behind Zulu. Not depending on a single company is better so we can try switching to it, what do you think ?

pierluca: Agreed !

matteosz (approved):

PR #1694, opened 2023-10-03
Improve hub locking

MariemBaccari:
This PR aims to:
- Upgrade the go backend to go 1.21 
- Fix a small bug that caused double connections between the client and the server (because of the servers sharing twice the same peers in the greet lao if they have a double connection between each other)
- Provide new data structures that allow finer-grain locking of the hub. The hub is currently entirely locked very often for processing messages which can be avoided by replacing the maps by thread-safe structs.
- Creates new package "hub_state" to store the structures that hold the hub's state to improve the main file's readability


be1-go/hub/standard_hub/mod.go
```diff
@@ -295,7 +266,7 @@ func (h *Hub) SendGreetServer(socket socket.Socket) error {
 
 	socket.Send(buf)
 
-	h.peersGreeted = append(h.peersGreeted, socket.ID())
+	h.peers.AddPeerGreeted(socket.ID())
```
jbsv, line 269: can the Send() fail ? if yes, is it still ok to AddPeerGreated() ?

be1-go/hub/standard_hub/mod.go
```diff
@@ -480,7 +448,7 @@ func (h *Hub) handleMessageFromServer(incomingMessage *socket.IncomingMessage) e
 func (h *Hub) handleIncomingMessage(incomingMessage *socket.IncomingMessage) error {
 	defer h.workers.Release(1)
 
-	h.log.Info().Str("msg", string(incomingMessage.Message)).Msg("handle incoming message")
+	h.log.Info().Str("msg", string(incomingMessage.Message)).Msgf("handle incoming message")
```
jbsv, line 451: Msgf might not be necessary without variables

jbsv (commented):
Great job. A few questions though.

be1-go/hub/standard_hub/mod.go
MariemBaccari: You're right, I added that for debugging and forgot to remove it before pushing, thanks for pointing it out. I fixed it in my commit.

be1-go/hub/standard_hub/mod.go
MariemBaccari: The Socket interface doesn't define any return value for it so the function Send() can't really fail by itself. 
It sends the message to the WritePump() goroutine which logs errors for calls to NextWriter() or Close() but, from what I see, it doesn't handle the errors following the call to Write(). I can open an issue for this so that it's dealt with in the future. There are many parts in the code where we just assume that Send() doesn't fail.

be1-go/hub/standard_hub/hub_state/ids_by_channel.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+}
+
+// Add adds a message id to the table
+func (i *IdsByChannel) Add(channel string, id string) {
+	i.Lock()
+	defer i.Unlock()
+	messageIds, channelStored := i.table[channel]
+	if !channelStored {
+		i.table[channel] = append(i.table[channel], id)
```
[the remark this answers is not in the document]
MariemBaccari: It's covered in the Test_Handle_Heartbeat test where we add a messageId for a channel that's not in the table. It seems that when we call append on a non existing channel, a new entry is created.

be1-go/hub/standard_hub/mod.go
jbsv: yes please open an issue referencing this

be1-go/hub/standard_hub/hub_state/channels_by_id.go
```diff
@@ -0,0 +1,44 @@
...
+	"maps"
+	"popstellar/channel"
+	"sync"
+)
+
+// ChannelsById provides a thread-safe structure that stores channel ids with their corresponding channels
+type ChannelsById struct {
+	sync.RWMutex
+	table map[string]channel.Channel
+}
```
pierluca, line 13: This struct looks very much like a thread-safe map.
Have you considered creating a generic instead ?

```suggestion
type ThreadSafeMap[K comparable, V any] struct {
	sync.RWMutex
	m map[K]V
}

type ChannelsById ThreadSafeMap[string, channel.Channel]
```

be1-go/hub/standard_hub/hub_state/channels_by_id.go
```diff
@@ -0,0 +1,44 @@
...
+
+// GetAll returns a copy of the table
+func (c *ChannelsById) GetAll() map[string]channel.Channel {
+	c.RLock()
+	defer c.RUnlock()
+	channelsCopy := make(map[string]channel.Channel)
+	maps.Copy(channelsCopy, c.table)
+	return channelsCopy
+}
```
pierluca, line 44: All of these functions could be replaced by a `ThreadSafeMap` API, i.e. `Add`, `Get`, `Copy`

It would still be readable, I believe, and you could still add aliases where it's really necessary (e.g. implementing `GetChannel`, if you really want it, which in turn would call `Get`)

be1-go/hub/standard_hub/hub_state/ids_by_channel.go
```diff
@@ -0,0 +1,59 @@
...
+	"golang.org/x/exp/slices"
+	"maps"
+	"sync"
+)
+
+// IdsByChannel provides a thread-safe structure that stores a channel id with its corresponding message ids
+type IdsByChannel struct {
+	sync.RWMutex
+	table map[string][]string
+}
```
pierluca, line 13: this could then become 
```suggestion
type IdsByChannel ThreadSafeMap[string,[]string]
```

be1-go/hub/standard_hub/hub_state/ids_by_channel.go
```diff
@@ -0,0 +1,59 @@
...
+
+// NewIdsByChannel creates a new IdsByChannel structure
+func NewIdsByChannel() IdsByChannel {
+	return IdsByChannel{
+		table: make(map[string][]string),
+	}
+}
+
+// Add adds a message id to the table
+func (i *IdsByChannel) Add(channel string, id string) {
```
pierluca, line 23: Conveniently, using type embedding would allow you to override the `Add` of the underlying ThreadSafeMap to put this extra logic.

be1-go/hub/standard_hub/hub_state/channels_by_id.go
```diff
@@ -0,0 +1,44 @@
...
+	channel, ok := c.table[channelId]
+	return channel, ok
+}
+
+// GetAll returns a copy of the table
+func (c *ChannelsById) GetAll() map[string]channel.Channel {
+	c.RLock()
+	defer c.RUnlock()
+	channelsCopy := make(map[string]channel.Channel)
+	maps.Copy(channelsCopy, c.table)
```
pierluca, line 42: does this do a deep copy ? Or is the new map containing pointers to the original channel ? It would be good to document this behaviour

pierluca (commented):
I've reviewed about half of it, I still have a few files to go through, but here's the first batch of comments :-)

be1-go/hub/standard_hub/message_handling.go
```diff
@@ -590,10 +574,8 @@ func (h *Hub) handleReceivedMessage(socket socket.Socket, messageData message.Me
 		return xerrors.Errorf(publishError, err)
 	}
 
-	h.Lock()
 	h.hubInbox.StoreMessage(publish.Params.Message)
-	h.addMessageId(publish.Params.Channel, publish.Params.Message.MessageID)
-	h.Unlock()
+	h.messageIdsByChannel.Add(publish.Params.Channel, publish.Params.Message.MessageID)
```
pierluca, line 578: just wondering (it wasn't better before) - should these two things be updated atomically or is the software going to be okay if there's, say, 5 seconds between one and the other happening ?

be1-go/hub/standard_hub/mod.go
```diff
@@ -198,12 +174,10 @@ func (h *Hub) Start() {
 					}
 				}()
 			case id := <-h.closedSockets:
-				h.RLock()
-				for _, channel := range h.channelByID {
+				for _, channel := range h.channelByID.GetAll() {
```
pierluca, line 177: This means that technically if a channel is added while this is running, we won't unsubscribe.
Is there a mechanism to protect against this ? (just asking :-)) 

Otherwise you could also let `channelByID` manage the mutex and instead implement a ForEach function, such that you can call this like: 

```
h.channelByID.ForEach(func(c Channel){
  c.Unsubscribe(id, method.Unsubscribe{})
})
```

and ForEach could be holding the lock :)

be1-go/hub/standard_hub/mod.go
```diff
@@ -495,12 +463,8 @@ func (h *Hub) handleIncomingMessage(incomingMessage *socket.IncomingMessage) err
...
 // sendGetMessagesByIdToServer sends a getMessagesById message to a server
 func (h *Hub) sendGetMessagesByIdToServer(socket socket.Socket, missingIds map[string][]string) error {
-	h.Lock()
-	defer h.Unlock()
-
-	queryId := h.queries.nextID
-	baseValue := false
-	h.queries.state[queryId] = &baseValue
+	queryId := h.queries.GetNextID()
+	h.queries.SetQueryState(queryId, false)
```
pierluca, line 467: shouldn't this function be executed _after_ the `json.Marshal` succeeds?  (and right before the `AddQuery` ?
Also, should the query state be set automatically if/when we do an `AddQuery` ?

be1-go/hub/standard_hub/mod_test.go
```diff
@@ -890,21 +891,21 @@ func Test_Handle_Answer(t *testing.T) {
 	})
 	require.Error(t, sock.err, "rpc message sent by a client should be a query")
 	sock.err = nil
-	require.False(t, queryState)
+	require.False(t, *hub.queries.GetQueryState(1))
```
pierluca, line 894: these calls really make me think it shouldn't be a pointer, we could just return bool + err :-)

pierluca (changes requested):

be1-go/hub/standard_hub/message_handling.go
```diff
@@ -189,21 +189,16 @@ func (h *Hub) handleAnswer(senderSocket socket.Socket, byteMessage []byte) error
...
 	}
 
 	if *val {
-		h.queries.Unlock()
 		return xerrors.Errorf("query %v already got an answer", answerMsg.ID)
 	}
 
-	*h.queries.state[*answerMsg.ID] = true
-	h.queries.Unlock()
+	h.queries.SetQueryState(*answerMsg.ID, true)
```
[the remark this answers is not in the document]
MariemBaccari: I agree it looks terrible. It might've made sense somehow when the code was first introduced (a few semesters ago) but I can't find the reason. Especially since I added Set and Get functions now it's definitely better if I also move the logic inside.

be1-go/hub/standard_hub/mod.go
MariemBaccari: Yes, good catch. I'll change that and also make the state set automatic and leave the function for changing the state only upon reception.

be1-go/hub/standard_hub/hub_state/queries.go
```diff
@@ -0,0 +1,60 @@
+package hub_state
+
+import (
+	"popstellar/message/query/method"
+	"sync"
```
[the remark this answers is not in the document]
MariemBaccari: Interesting ... 👀  I'll add it so that it executes during the lint :)

be1-go/hub/standard_hub/hub_state/channels_by_id.go
MariemBaccari: I believe the new map contains pointers to the original channels. I will add that in the documentation.

be1-go/hub/standard_hub/message_handling.go
MariemBaccari: They should be updated atomically indeed. As a matter of fact, when I was rereading the Inbox code and usage, I realized that we had two different Inbox instances in the hub that essentially serve the same purpose. I think I can do some refactoring in that regards (by creating a new HubInbox type that embeds Inbox but has more features (Channels and MessagesById that only the Hub needs) in another PR.

be1-go/hub/standard_hub/hub_state/ThreadSafeMap.go
```diff
@@ -0,0 +1,44 @@
...
+func (i *ThreadSafeMap[K, V]) Set(key K, val V) {
+	i.Lock()
+	defer i.Unlock()
+	i.table[key] = val
+}
+
+// GetTable returns a shallow copy of the map
+func (i *ThreadSafeMap[K, V]) GetTable() map[K]V {
+	i.Lock()
+	defer i.Unlock()
```
pierluca, line 32: Shouldn't this be `RLock` ?

be1-go/hub/standard_hub/hub_state/ThreadSafeMap.go
```diff
@@ -0,0 +1,44 @@
...
+	tableCopy := make(map[K]V, len(i.table))
+	for key, val := range i.table {
+		tableCopy[key] = val
+	}
+	return tableCopy
+}
+
+func (i *ThreadSafeMap[K, V]) IsEmpty() bool {
+	i.Lock()
+	defer i.Unlock()
```
pierluca, line 42: `RLock` too ?

be1-go/hub/standard_hub/mod.go
```diff
@@ -117,23 +117,27 @@ func NewHub(pubKeyOwner kyber.Point, clientServerAddress string, serverServerAdd
...
+		schemaValidator: schemaValidator,
+		stop:            make(chan struct{}),
+		workers:         semaphore.NewWeighted(numWorkers),
+		log:             log,
+		laoFac:          laoFac,
+		serverSockets:   channel.NewSockets(),
+		hubInbox:        *inbox.NewInbox(rootChannel),
+		rootInbox:       *inbox.NewInbox(rootChannel),
+		queries:         state.NewQueries(),
+		messageIdsByChannel: state.MessageIds{
```
pierluca, line 136: you could also create a utility function `state.NewMessageIdsMap()` to encapsulate this.

be1-go/hub/standard_hub/mod.go
```diff
@@ -117,23 +117,27 @@ func NewHub(pubKeyOwner kyber.Point, clientServerAddress string, serverServerAdd
...
-		serverSockets:       channel.NewSockets(),
-		hubInbox:            *inbox.NewInbox(rootChannel),
-		rootInbox:           *inbox.NewInbox(rootChannel),
-		queries:             state.NewQueries(),
-		messageIdsByChannel: state.NewIdsByChannel(),
-		peers:               state.NewPeers(),
-		blacklist:           make([]string, 0),
+		channelByID: state.Channels{
+			ThreadSafeMap: state.NewThreadSafeMap[string, channel.Channel](),
+		},
```
pierluca, line 122: you could also create a utility function `state.NewChannelMap()` to encapsulate this.

be1-go/hub/standard_hub/hub_state/queries.go
```diff
@@ -43,12 +49,23 @@ func (q *Queries) GetNextID() int {
...
 
-	q.state[id] = &state
+	currentState, ok := q.state[id]
+
+	if !ok {
+		return xerrors.Errorf("query with id %d not found", id)
+	}
+
+	if currentState {
+		return xerrors.Errorf("query with id %d already answered", id)
```
pierluca, line 64: Just realized something, completely outside the scope of this PR.
This error propagates all the way to the socket.
So if another peer (for whatever reason) sends the "answer" message twice, they will get an error in return saying "we've already gotten your answer".
Is this something that would need to be shared with the other peer, or just logged ? 
What would the other peer be expected to do - retry ? That would likely make things worse (infinite loop of errors / retries)
It seems manageable (in this case) to implement some degree of idempotency.

be1-go/hub/standard_hub/hub_state/queries.go
```diff
@@ -20,17 +22,21 @@ type Queries struct {
 // NewQueries creates a new queries struct
 func NewQueries() Queries {
 	return Queries{
-		state:                  make(map[int]*bool),
+		state:                  make(map[int]bool),
```
pierluca, line 25: I just realized something, also completely outside the scope of this PR.
Does this map grow forever and ever as the application keeps running ? 
When do we clear out old IDs to free up some resources ? 😅 
We might need to add some mechanism for that.

be1-go/hub/standard_hub/hub_state/peers.go
```diff
@@ -34,17 +36,17 @@ func (p *Peers) AddPeerInfo(socketId string, info method.ServerInfo) {
 func (p *Peers) AddPeerGreeted(socketId string) {
 	p.Lock()
 	defer p.Unlock()
-	if slices.Contains(p.peersGreeted, socketId) {
+	if slices.Contains(maps.Keys(p.peersGreeted), socketId) {
 		return
 	}
```
pierluca, line 41:
```suggestion
	if _, ok := p.peersGreeted[socketId]; ok {
		return
	}
```

This could be implemented in a much faster way like this.
But do we even need it ?
The operation `p.peersGreeted[socketId] = struct{}{}` is idempotent, we might as well just go through it every time.

```suggestion
```

be1-go/channel/lao/mod.go
```diff
@@ -703,16 +703,12 @@ func (c *Channel) createAndSendLAOGreet() error {
 		return xerrors.Errorf("failed to marshal the organizer key: %v", err)
 	}
 
-	peers := []messagedata.Peer{}
+	peersInfo := c.hub.GetPeersInfo()
```
pierluca, line 706: Just taking a step back, why wouldn't `GetPeersInfo()` already return a unique list of peers ? 
Do we ever need it to return duplicates ? I might very well be missing something here and there could be a perfectly good reason, so think of this as a genuine question and not an invitation to change "because @pierluca said so" 😅

be1-go/hub/standard_hub/hub_state/MessageIds.go
```diff
@@ -0,0 +1,32 @@
...
+// Add adds a message id to the slice of message ids of the channel
+func (i *MessageIds) Add(channel string, id string) {
+	i.Lock()
+	defer i.Unlock()
+	messageIds, channelStored := i.table[channel]
+	if !channelStored {
+		i.table[channel] = append(i.table[channel], id)
+		return
+	}
+	alreadyStoredId := slices.Contains(messageIds, id)
```
pierluca, line 21: Here we have two conflicting needs, we want messageIds stored sequentially (i.e. order matters) but this access is O(n) and it won't scale well as the number of messages grows.
We could trade memory for speed, and have a secondary map that contains maps as values, and allows us to check in O(1) if an ID is included or not.
Now that you have a `MessageIds` struct, nothing prevents you from optimizing the heck out of it 😁 

I assume this is not the only place where O(n) operations happen, so maybe this is also best addressed in a separate PR, probably when we look at persisting this to disk as well :)

pierluca (commented):
Much better, and it's allowed me to spot a few other issues for future PRs :) 
We're on the right path !

be1-go/hub/standard_hub/hub_state/ThreadSafeMap.go
MariemBaccari: Yes ... good catch

be1-go/hub/standard_hub/hub_state/queries.go
MariemBaccari: I agree that we should just log it because we wouldn't expect the peer to do anything. I think it could be a nice first issue so I'll add it to the repo. Regarding idempotency, would that also apply to when we try to create a channel that was already created ? Currently we send a "duplicate resource" error but how do we differentiate between cases that warrant idempotency and those that don't ?

Unrelated to that but I also noticed that we don't always respect the [protocol](https://github.com/dedis/popstellar/blob/master/docs/protocol.md#rpc-answer-error) when it comes to sending errors back. Since we have many "levels" of returning in go, the nature of the error gets lost by the time it arrives to the method that's supposed to create the error to send back so we don't have the right code. Thankfully, we already have methods designed to create the right error with the right code but those aren't always used for some reason. I'll investigate the individual cases and open PRs for them. They're easy fixes but also a good way to get to know the codebase so they'll be good first issues as well.

be1-go/channel/lao/mod.go
MariemBaccari: You're right, I moved the checks to the `GetAllPeersInfo()` in `Peers` to make it return a unique list of server info.

be1-go/hub/standard_hub/hub_state/queries.go
MariemBaccari: The map does indeed grow forever ... I just created #1707 to open the discussion about this so that it's better documented. Let me know what you think !

be1-go/hub/standard_hub/hub_state/MessageIds.go
MariemBaccari: I agree that there's a lot of room for optimization here. I opened #1708 to keep track of the issue and discuss it when we start working on persistence.

pierluca (approved):
As discussed, there are still some performance improvements possible, but this is already so much better than in the past.
Great job !

PR #1700, opened 2023-10-14
Popcha on FE2

matteosz:
Keep up with FE1 and implement PoPCHA also on FE2


fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/security/Base64URLData.java
```diff
@@ -37,6 +35,10 @@ private static byte[] decode(String data) {
     return Base64.getUrlDecoder().decode(data);
   }
 
+  public static String encode(String data) {
+    return encode(data.getBytes());
```
pierluca, line 39: does `getBytes()` assume a particular encoding, e.g. UTF-8 ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/qrcode/QrScannerFragment.java
```diff
@@ -66,7 +61,8 @@ public View onCreateView(
     scanningViewModel =
         scanningAction.obtainScannerViewModel(requireActivity(), popViewModel.getLaoId());
 
-    if (scanningAction != ScanningAction.ADD_LAO_PARTICIPANT) {
+    if (scanningAction != ScanningAction.ADD_LAO_PARTICIPANT
+        && scanningAction != ScanningAction.ADD_POPCHA) {
```
pierluca, line 65: would it make sense to have `displayCounter` in the fragment bundle ? 
Basically, should this corner case be handled by a long series of `if not`, or can we just let the application calling the fragment decide if they want a counter or not ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/security/KeyManager.java
```diff
@@ -88,6 +84,19 @@ public PoPToken getPoPToken(LaoView laoView, RollCall rollCall) throws KeyExcept
...
+   *
+   * @param laoId to generate the PoP Token from
+   * @param clientId to generate the PoP Token from
+   * @return the generated PoP Token
+   * @throws KeyGenerationException if an error occurs during key generation
+   * @throws UninitializedWalletException if the wallet is not initialized with a seed
+   */
+  public PoPToken getLongTermPoPToken(String laoId, String clientId) throws KeyException {
+    return wallet.generatePoPToken(laoId, Hash.hash(clientId));
+  }
```
pierluca, line 98: I have a nomenclature / typing problem here.
Is it really a `long term PoP token` or an entirely different object, say a `long term pseudonymous auth identity` ? And if so, shouldn't this have its own type (i.e., not `PoPToken`) ?

fe2-android/app/src/test/unit/java/com/github/dedis/popstellar/utility/MessageValidatorTest.java
```diff
@@ -189,4 +190,43 @@ public void testValidEmoji() {
     assertThrows(
         IllegalArgumentException.class, () -> validator.isValidEmoji("random string", field));
   }
+
+  @Test
+  public void testValidPopCHAUrl() {
```
pierluca, line 195: Entirely outside the scope of this PR, but in general it's better to have separate tests for separate conditions being checked.

E.g., instead of both testing `valid` and `valid2` within this test, you'd have a `testValidPopCHAurlLocalhost` , `testValidPopCHAurlPublicDomain` , and then maybe `testInvalidPopCHAurlMissingClientId`, etc.

This has the effect that when the CI cries about a test having failed, you immediately know which condition failed, without having to track down the stack trace and failing condition.

fe2-android/app/src/test/unit/java/com/github/dedis/popstellar/model/qrcode/PoPCHAQRCodeTest.java
```diff
@@ -0,0 +1,71 @@
...
+  @Test
+  public void extractResponseMode() {
+    assertEquals(RESPONSE_MODE, POPCHA_QR_CODE.getResponseMode());
+  }
+
+  @Test
+  public void extractHost() {
+    assertEquals(ADDRESS, POPCHA_QR_CODE.getHost());
+  }
+}
```
pierluca, line 71: you don't have any non-happy path tests that check that these functions crash if the URL is invalid. Would you want to at least add one ? Just thinking of how we could limit the damage in case of invalid data.

fe2-android/app/src/test/unit/java/com/github/dedis/popstellar/model/network/method/message/data/popcha/PoPCHAAuthenticationTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+  @Test
+  public void testToString() {
+    String expected =
+        String.format(
+            "PoPCHAAuthentication{clientId='%s', nonce='%s', identifier='%s', identifierProof='%s', state='%s', responseMode='%s', popchaAddress='%s'}",
+            CLIENT_ID, NONCE, IDENTIFIER, IDENTIFIER_PROOF, STATE, RESPONSE_MODE, ADDRESS);
+    assertEquals(expected, POPCHA_AUTHENTICATION.toString());
+  }
+}
```
pierluca, line 98: and/or you could test this here as well.

pierluca (commented):
This is a partial review, only 15/23 files reviewed, but I won't have time to do more for now :)

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/objects/security/Base64URLData.java
matteosz: It uses the default charset of the platform, which is usually UTF-8, but to be sure I'll force the charset specifying it as param

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/qrcode/QrScannerFragment.java
matteosz: Yes, that is a nice choice design-wise, I'll add it

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/security/KeyManager.java
matteosz: Yes, that long term auth token object would also be a key pair essentially, but I agree it's cleaner having a proper type for it

fe2-android/app/src/test/unit/java/com/github/dedis/popstellar/model/network/method/message/data/popcha/PoPCHAAuthenticationTest.java
matteosz: I'll add the non-happy tests for the PoPCHAQRCode object rather this (PoPPCHAAuthentication), since the latter is a message that the client itself craft and send over the network, so the non-happy path if it's the case is already intercepted before

fe2-android/app/src/test/ui/robolectric/com/github/dedis/popstellar/ui/lao/popcha/PoPCHAHomeFragmentTest.java
```diff
@@ -0,0 +1,183 @@
...
+          ROLLCALL_ID,
+          ROLLCALL_ID,
+          ROLLCALL_NAME,
+          CREATION_TIME,
+          TIMESTAMP_1,
+          TIMESTAMP_2,
+          EventState.CLOSED,
+          attendees,
+          "bc",
+          "");
```
pierluca, line 81: considering almost all the constants here are not (really) used in the test, but only in the setup, and I imagine (maybe mistakenly) they might be copied from other tests where they're similarly not really used:

Could this logic be put in a test utility class to generate this stuff for all tests ? 
and/or
Could these constants be removed and expressed as succinctly as reasonably possible in the test setup itself ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/LaoActivity.java
```diff
@@ -68,6 +66,8 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
 
     witnessingViewModel = obtainWitnessingViewModel(this, laoId);
 
+    obtainPoPCHAViewModel(this, laoId).disableConnectingFlag();
```
pierluca, line 69: why `disableConnectingFlag` ? It's somewhat surprising to read this here, as someone who's somewhat unfamiliar with the codebase.

pierluca (commented):
A few questions, but mostly LGTM :)

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/LaoActivity.java
matteosz: Indeed, double checking I think it's unnecessary because when switching LAO the view model is recreated and the initial state of the flag is already false :)

fe2-android/app/src/test/ui/robolectric/com/github/dedis/popstellar/ui/lao/popcha/PoPCHAHomeFragmentTest.java
matteosz: Yes, that could be improved 😄. I need to validate where in all the tests this is reused to refactor it

matteosz: Okay, looking better at the code refactoring this goes out of the scope of this PR as it involves a lot of changes, so I'll leave that for future improvements. Instead I will contain as possible the constants used here and express them in the setup

jbsv (approved):
LGTM

pierluca (approved):

PR #1701, opened 2023-10-15
Constants for feature paths

simone-kalbermatter:
Paths to features and scenarios that are used across features are now defined as constants in `constants.feature`. 
This makes it easier to change paths in the future, as they will only have to be changed in one place. It also improves the readability of calls to `call read(...)` .

### DIFF
--- a/tests/karate/src/test/java/be/features/utils/constants.feature
+++ b/tests/karate/src/test/java/be/features/utils/constants.feature
@@ -12,3 +12,20 @@ Feature: Constants
 
     * def rootChannel = '/root'
     * def random = Java.type('be.utils.RandomUtils')
+
+    # Paths to util features
+    * def utilsPath = 'classpath:be/features/utils/'
+    * def serverFeature = utilsPath + 'server.feature'
+    * def mockClientFeature = utilsPath + 'mockClient.feature'
+
+    # Paths to scenarios defined in simpleScenarios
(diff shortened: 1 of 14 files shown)


tests/karate/src/test/java/be/features/utils/constants.feature
```diff
@@ -12,3 +12,20 @@ Feature: Constants
...
+
+    # Paths to util features
+    * def utilsPath = 'classpath:be/features/utils/'
+    * def serverFeature = utilsPath + 'server.feature'
+    * def mockClientFeature = utilsPath + 'mockClient.feature'
+
+    # Paths to scenarios defined in simpleScenarios
+    * def simpleScenario = 'classpath:be/features/utils/simpleScenarios.feature@name='
+    * def createLaoScenario = simpleScenario + 'valid_lao'
+    * def createRollCallScenario = simpleScenario + 'valid_roll_call'
```
jbsv, line 24: I find `create` and `valid` quite different and a bit confusing. Shouldn't the `valid_roll_call` be renamed `create_roll_call` to avoid this confusion ? Other proposition ?

jbsv (commented):
Maybe some naming improvement ?

tests/karate/src/test/java/be/features/utils/constants.feature
simone-kalbermatter: Good point, I forgot to change this.

matteosz (approved):
LGTM

PR #1702, opened 2023-10-15
Extend Karate tests for decentralized communication

simone-kalbermatter:
This PR:
- Refactors `JsonConverter` and `MultiMsgWebSocketClient` to make them easier to understand,
- Makes message ids of publish messages accessible in feature files. This improves checks on the content of messages returned by the tested components and can be used to craft better test messages.
- Adds a test case for the _getMessagesById_ message that checks that a server does not request messages in a hearbeat that it already received.

### DIFF
--- a/tests/karate/src/test/java/be/features/decentralizedCom/getMessagesById.feature
+++ b/tests/karate/src/test/java/be/features/decentralizedCom/getMessagesById.feature
@@ -10,7 +10,8 @@ Feature: Request messages by id from other servers
     * call read(serverFeature)
     * call read(mockClientFeature)
     * def mockServer = call createMockClient
-    * def lao = mockServer.createValidLao()
+    * def mockFrontend = call createMockClient
+    * def lao = mockFrontend.createValidLao()
 
     # Create the template for heartbeat message
     # This is used in combination with 'eval' to dynamically resolve the channel keys in the heartbeat JSON
@@ -24,7 +25,7 @@ Feature: Request messages by id from other servers
 
     # This call executes all the steps to create a valid lao on the server before every scenario
     # (lao creation, subscribe, catchup)
-    * call read(createLaoScenario) { organizer: '#(mockServer)', lao: '#(lao)' }
+    * call read(createLaoScenario) { organizer: '#(mockFrontend)', lao: '#(lao)' }
 
   # Check that after sending a heartbeat message with unknown message id, the server responds with a
   # getMessagesByID requesting this message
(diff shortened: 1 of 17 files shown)


tests/karate/src/test/java/common/net/MultiMsgWebSocketClient.java
```diff
@@ -41,113 +42,114 @@ public MultiMsgWebSocketClient(WebSocketOptions options, Logger logger, MessageQ
...
-  public MessageBuffer getBuffer() {
-    return queue;
+  /**
+   * JSON messages defined inside features are interpreted as maps from String to Object.
+   * This method is called directly inside features and is just a wrapper around the send method of WebsocketClient (that takes Strings).
+   * @param messageData the message to send as a JSON map
+   *                    (for example: subscribe and catchup in simpleScenarios.feature).
+   */
+  public void send(Map<String, Object> messageData){
+    this.send(mapToJsonString(messageData));
```
pierluca, line 52: I'd consider renaming `messageData` to something like `wsMessage` or `jsonRpcMsg` to indicate that we're talking about the lowest level.
`messageData` is already pretty (over)loaded as a concept in this project.

pierluca (commented):
LGTM.
Some elements could be improved here and there (e.g. more heartbeat tests / more complex scenarios), but as a foundation, this is already pretty nice 👍

simone-kalbermatter:
@pierluca Yes, this is only a foundation so far. I realized that splitting the server tests from the client tests was probably not the best idea for more complex test scenarios 😅 I might add them back together, but I want to do this in a seperate PR.

tests/karate/src/test/java/be/features/decentralizedCom/getMessagesById.feature
```diff
@@ -54,7 +55,40 @@ Feature: Request messages by id from other servers
...
+          "location": '#(validRollCall.location)',
+          "description": '#(validRollCall.description)',
+        }
+      """
+
+    And mockFrontend.publish(validCreateRollCall, lao.channel)
+    And json answer = mockFrontend.getBackendResponse(validCreateRollCall)
+    And def message_id = mockFrontend.getPublishMessageId(validCreateRollCall)
+    And def messageIds = []
+    And eval messageIds.push(message_id)
```
matteosz, line 85: Is it possible to initialize in-line?
`And def messageIds = [message_id]`

tests/karate/src/test/java/be/features/decentralizedCom/heartbeat.feature
```diff
@@ -10,22 +10,25 @@ Feature: Send heartbeats to other servers
...
+  # (The initial one would be a response to publishing lao creation)
   Scenario: Server should send heartbeat messages automatically after a time interval
-    Given wait(30)
+    Given wait(40)
 
-    When def heartbeatMessages = mockServer.getMessagesByMethod('heartbeat')
+    When def heartbeatMessages = mockServer.getHeartbeats()
 
-    Then assert heartbeatMessages.length > 0
+    Then assert heartbeatMessages.length > 1
```
matteosz, line 28: Would be better to check for the actual number? Or at least if we cannot determine it precisely we could add an upper threshold (just to be sure heartbeats are not spammed more than expected)

tests/karate/src/test/java/common/net/MultiMsgWebSocketClient.java
```diff
@@ -41,113 +42,114 @@ public MultiMsgWebSocketClient(WebSocketOptions options, Logger logger, MessageQ
...
+  /**
+   * JSON messages defined inside features are interpreted as maps from String to Object.
+   * This method is called directly inside features to send a complete publish message given the high-level message data to publish.
+   * @param highLevelMessageDataMap the high-level message to publish as a JSON map
+   *                                (for example: validCreateRollCall and badCreateRollCall in createRollCall.feature).
+   * @param channel the channel to publish on.
+   */
+  public void publish(Map<String, Object> highLevelMessageDataMap, String channel){
+    String highLevelMessageData = mapToJsonString(highLevelMessageDataMap);
+    int messageId = new Random().nextInt(Integer.MAX_VALUE);
```
matteosz, line 64: `Integer.MAX_VALUE` is already the default max value for nextInt so it's redundant

tests/karate/src/test/java/common/net/MultiMsgWebSocketClient.java
```diff
@@ -171,4 +173,66 @@ public void useWrongSignature() {
...
+    String msg = queue.take();
+
+    if (msg == null) logger.error("listen timed out");
+
+    return msg;
+  }
+
+  private PublishMessageIds getPublishMessageIds(Map<String, Object> highLevelMessageDataMap){
+    String highLevelMessageData = mapToJsonString(highLevelMessageDataMap);
+    PublishMessageIds message_ids = sentMessages.get(highLevelMessageData);
```
matteosz, line 195: nitpick: I'm not a great fan of snake case in Java 😬

matteosz (changes requested):
LGTM, just few comments

tests/karate/src/test/java/be/features/decentralizedCom/getMessagesById.feature
simone-kalbermatter: It seems like Karate has some problem parsing this as a JSON array, I get an error if I try to do it this way.

tests/karate/src/test/java/be/features/decentralizedCom/heartbeat.feature
simone-kalbermatter: I was thinking to keep it more open in case it changes in the future, but right now it can absolutely just check that the number is 2.

pierluca (approved):
LGTM.
Are you satisfied @matteosz  ?

matteosz (approved):
LGTM!

PR #1705, opened 2023-10-28
Disabled witnessing for fe2

matteosz:
Disable the witnessing functionality to be compatible with FE1, to be restored when also FE1 will implement it.


MariemBaccari (approved):
The changes look good to me. It could be useful to open an issue to document the fact that witnessing should be reactivated when fe1 gets there with some indication about how exactly to reactivate it and a reference to this PR in case the "knowledge" gets lost when other people take over the project.

PR #1710, opened 2023-11-11
Added automatic signature mechanism for organiser

matteosz:
This CR allows to re-enable the witnessing feature by implementing an automatic signature mechanism for the organiser. By having this mechanism in place, if no witness is added to the LAO (other than the organiser which is implicitly considered a witness) the behaviour between FE1 and FE2 is going to be coherent. Additionally, I've implemented a checkbox for fe2 when creating a lao which allows to intentionally disable the witnessing feature.

PR #1712, opened 2023-11-19
Added lint report for fe2

matteosz:
After a conversation about Timber-lint in the PR #1710 , here's an integration on the FE2 CI that generates a lint report and trigger a static code analysis based on it

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -227,26 +227,18 @@ jobs:
           # To run the CI with debug informations, add --info
           ./gradlew jacocoTestReport --parallel --build-cache
 
-      - name: Generate Lints
+      - name: Generate Sarif Lint Report
         shell: bash
         run: |
           # To run the CI with debug informations, add --info
           ./gradlew lintDebug --parallel --build-cache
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -227,6 +227,19 @@ jobs:
...
 
+      - name: Generate Sarif Lint Report
+        shell: bash
+        run: |
+          # To run the CI with debug informations, add --info
+          ./gradlew lintDebug --parallel --build-cache
+      
+      - name: Upload lint sarif
+        uses: github/codeql-action/upload-sarif@v2
+        if: success() || failure()
```
pierluca, line 238: why this `if` clause here but neither before nor after ?

matteosz: No particular reason as the order doesn't change the workflow

jbsv: instead of `success() || failure()` why not use `always()`?

jbsv (changes requested):
maybe a way to KIS

pierluca (approved):

PR #1726, opened 2023-12-28
Migration from Java to Kotlin

matteosz:
This PR updates the build.gradle to support Kotlin and migrates part of the codebase to Kotlin (utility and repository packages).


pierluca:
A couple of suggestions, maybe for subsequent PRs:
1. Setup `ktfmt` on the project
2. Setup `detekt` as the linter on the project
3. Consider setting up [Konsist](https://docs.konsist.lemonappdev.com/getting-started/readme) to enforce architectural constraints and at least emit warnings if they're violated. We've used dependency-cruiser on fe1 with some success to ensure a clean separation between the software core and the features, and between the features themselves.

fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/remote/GlobalNetworkManager.kt
```diff
@@ -0,0 +1,61 @@
...
+  val messageSender: MessageSender
+    get() {
+      checkNotNull(networkManager) { "The connection has not been established." }
+      return networkManager as MessageSender
+    }
+
+  override fun dispose() {
+    if (networkManager != null) {
+      networkManager!!.dispose()
+    }
```
pierluca, line 54: why not `networkManager?.dispose()` ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/remote/LAONetworkManager.kt
```diff
@@ -0,0 +1,304 @@
...
+        handleError(e)
+      } catch (e: UnknownRollCallException) {
+        handleError(e)
+      } catch (e: NoRollCallException) {
+        handleError(e)
+      } catch (e: UnknownElectionException) {
+        handleError(e)
+      } catch (e: UnknownWitnessMessageException) {
+        handleError(e)
+      }
```
pierluca, line 225: You don't need to be this verbose in Kotlin. You can just do the following

```
    } catch (e: Exception) {
        when (e) {
            is DataHandlingException,
            is UnknownLaoException,
            // ... and so on
            is UnknownWitnessMessageException-> handleError(e)
            else -> throw e
        }
    }
```

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/NetworkLogger.kt
```diff
@@ -0,0 +1,200 @@
...
+
+    // Always print the log to console
+    if (t != null) {
+      Log.println(
+          priority,
+          tag,
+          """
+     $message
+     $error
+     """
```
pierluca, line 38: the alignment here seems wrong

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/NetworkLogger.kt
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  private fun sendToServer(log: String) {
+    // Take the lock and send the log as UTF-8 encoded string
+    synchronized(lock) {
+      // If the websocket hasn't been already opened then open it
+      if (webSocket == null) {
+        connectWebSocket()
+        // Connect may fail
+        webSocket?.send(log)
```
pierluca, line 58: shouldn't this be outside of the `if` ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/error/DataHandlingException.kt
```diff
@@ -0,0 +1,30 @@
...
+    this.data = data
+  }
+
+  override val message: String?
+    get() =
+        """
+               Error while handling data : ${super.message}
+               data=$data
+               """
+            .trimIndent()
```
pierluca, line 23: alignment seems off again, you might want to check where `"""` shows up in the codebase.

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/PoPRXOperators.kt
```diff
@@ -0,0 +1,40 @@
...
+        override fun onSubscribe(d: Disposable) {
+          observer.onSubscribe(d)
+        }
+
+        override fun onNext(t: T & Any) {
+          observer.onNext(t)
+        }
+
+        override fun onError(e: Throwable) {
+          onError.accept(e)
```
pierluca, line 31: shouldn't this be logged in some way, though ? or is it logged somewhere else ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/PoPRXOperators.kt
```diff
@@ -0,0 +1,40 @@
+package com.github.dedis.popstellar.utility
+
+import io.reactivex.ObservableOperator
+import io.reactivex.Observer
+import io.reactivex.disposables.Disposable
+import java.util.function.Consumer
+
+object PoPRXOperators {
+  /**
+   * This function creates an observable operator that suppresses errors to let the underlying
```
pierluca, line 10:
```suggestion
   * This function creates an observable operator that suppresses errors to keep the underlying
```

fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/RollCallRepository.kt
```diff
@@ -0,0 +1,281 @@
...
+    disposables.add(
+        rollCallDao
+            .insert(RollCallEntity(laoId, rollCall))
+            .subscribeOn(Schedulers.io())
+            .observeOn(AndroidSchedulers.mainThread())
+            .subscribe({
+              Timber.tag(TAG).d("Successfully persisted rollcall %s", rollCall.persistentId)
+            }) { err: Throwable ->
+              Timber.tag(TAG).e(err, "Error in persisting rollcall %s", rollCall.persistentId)
+            })
```
pierluca, line 67: Are you sure this syntax is correct ? 
Isn't the second `{}` block supposed to be part of `subscribe` ?

fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/SocialMediaRepository.kt
```diff
@@ -0,0 +1,353 @@
...
+                                      // Do not retrieve deleted reactions
+                                      if (reaction.isDeleted) {
+                                        return@inner
+                                      }
+                                      // Load the reaction into the memory
+                                      addReaction(reaction)
+                                      Timber.tag(TAG)
+                                          .d("Retrieved from db reaction %s", reaction.id)
+                                    })
+                              }) { err: Throwable ->
```
pierluca, line 336: same question, are you sure this is correct and you don't mean to do the following?  🤔 
```suggestion
                              }, { err: Throwable ->
```

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/data/MeetingHandler.kt
```diff
@@ -0,0 +1,148 @@
...
+    private val witnessingRepo: WitnessingRepository
+) {
+  /**
+   * Process a CreateMeeting message.
+   *
+   * @param context the HandlerContext of the message
+   * @param createMeeting the message that was received
+   */
+  @Throws(UnknownLaoException::class)
+  fun handleCreateMeeting(context: HandlerContext, createMeeting: CreateMeeting) {
```
pierluca, line 33: a bit of vertical spacing would help readability in this function :)

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/data/MeetingHandler.kt
```diff
@@ -0,0 +1,148 @@
...
+     ${meeting.location}
+     
+     
+     """.trimIndent()
+      }Starts at :
+                   ${Date(meeting.startTimestampInMillis)}
+                   
+                   Finishes at :
+                   ${Date(meeting.endTimestampInMillis)}
+                   """
```
pierluca, line 117: here too the alignment is a bit of a mess

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/data/MeetingHandler.kt
```diff
@@ -0,0 +1,148 @@
...
+                   Mnemonic identifier :
+                   ${generateMnemonicWordFromBase64(meeting.id, 2)}
+                   
+                   ${
+        if (meeting.location.isEmpty()) "" else """
+     Location :
+     ${meeting.location}
+     
+     
+     """.trimIndent()
```
pierluca, line 111: put this in its own variable instead of embedding multi-line strings in multi-line strings, this is really not readable 😅

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/security/Hash.kt
```diff
@@ -0,0 +1,41 @@
...
+   * Hash some objects using SHA256. Concatenate the object's string representation following the
+   * protocol's directive. Then hash the obtained string
+   *
+   * @param strs : the strings to hash
+   * @return the hashed data or null if failed to hash
+   * @throws IllegalArgumentException if the data is null or empty
+   * @throws UnsupportedOperationException if SHA-256 MessageDigest is unavailable
+   */
+  @JvmStatic
+  fun hash(vararg strs: String?): String {
```
pierluca, line 23: here as well, a bit of vertical spacing wouldn't hurt.

newlines are really helpful in separating logical blocks in a function, improving readability and communicating intention. here there's clearly:
1. a verification of preconditions (the require of `strs.isNotEmpty()`) 
2. a try block, within which:
2.1. initialization of the digest
2.2. looping over string components
2.3. obtain the result and returning
3. a catch block

On a side note, while Kotlin allows to do `return try`, you're not really returning anything from the catch, so I would suggest only using the `return` keyword on line 35 when encoding the hash.

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/ActivityUtils.kt
```diff
@@ -0,0 +1,287 @@
...
+import java.security.MessageDigest
+import java.security.NoSuchAlgorithmException
+import java.util.Arrays
+import java.util.Collections
+import java.util.function.Consumer
+import java.util.function.Supplier
+import kotlin.math.abs
+import timber.log.Timber
+
+object ActivityUtils {
```
pierluca, line 38: Outside the scope of this PR:
this object seems to be getting quite bit and with a bunch of unrelated things, i.e. `buildBackButtonCallback` doesn't exactly seem to fit.

Could we break this into smaller, more narrowly-focused objects ?

fe2-android/app/schemas/com.github.dedis.popstellar.repository.database.AppDatabase/3.json
```diff
@@ -1,515 +0,0 @@
-{
```
pierluca, line 1: why do we not need this anymore ?

pierluca (changes requested):
Considering the size of this PR, we'll have to forego a full review as a quality gate and rely on testing instead, but I nonetheless randomly sampled a few files to review. 

There seem to be some common issues around the codebase that popped up in a few files. I suggest resolving those (not just where I marked them, but wherever those patterns appear), in particular to ensure readability and correctness.

You might also want to do a self-review (even just randomly sampling different files, e.g. 1-2 repositories, 1-2 utility files, 1-2 handlers, ...) and see if you spot other issues I may have missed.

fe2-android/app/schemas/com.github.dedis.popstellar.repository.database.AppDatabase/3.json
matteosz: This was for a cleanup, that schema was unused but I forgot to delete it, the up-to-date version is the 4th

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/NetworkLogger.kt
matteosz: Yeah it should, thanks!

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/error/DataHandlingException.kt
matteosz: Yeah, I just fixed every string that was replaced with """

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/PoPRXOperators.kt
matteosz: It's actually a class not used in the codebase, I don't know what was the purpose of its creation. But I still can insert a log there in case in the future we want to use it

fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/RollCallRepository.kt
matteosz: There's no difference, it's a trailing lambda. In short, if the last param of a function is a lambda, you can move it outside the parenthesis.
Here's some more context in a [StackOverflow post](https://stackoverflow.com/questions/60242136/rx-java-whats-the-difference-between-subscribe-vs-subscribe)

fe2-android/app/schemas/detekt/detekt-config.yml
pierluca, line 1: This file would benefit from a few comments, in particular when some stuff is disabled.
Is it temporarily disabled or forever ? If temporarily, until when ?
And perhaps more importantly, why ?

pierluca (changes requested):
It seems that your commit "merge consensus refactoring" actually reverted a lot of files that had been migrated to Kotlin into their Java version.
E.g. the models folder and the view models.

You might want to double check that. Other than that, it looks in much better shape now 👍

matteosz:
> It seems that your commit "merge consensus refactoring" actually reverted a lot of files that had been migrated to Kotlin into their Java version. E.g. the models folder and the view models.
> 
> You might want to double check that. Other than that, it looks in much better shape now 👍

Oh no, they weren't part of this refactoring, so they always were in their Java version. The refactored parts of the code are in the folders repository and utility only. Probably you saw there some references to the consensus repo that was indeed modified and converted to Kotlin.

pierluca (approved):
LGTM !

PR #1729, opened 2024-01-27
Kotlin migration - final part

matteosz:
This is the last part of Kotlin migration. It will migrate the remaining of the ui folder and the tests


pierluca:
Please make sure at the end of the migration that the move to Kotlin preserves the following:
1. The system actually works :-)
2. Sonar is (somewhat) happy about the codebase
3. The coverage is computed correctly

pierluca (approved):
As discussed, merging as-is and introducing bugfixes later if need.

PR #1730, opened 2024-01-30
Extend Karate tests for decentralized communication Part 2

simone-kalbermatter:
This PR:
- Simplifies the test environments back to only one Go and one Scala environment, while still allowing the creation of multiple mock frontends and backends. Mock frontends can now explicitly be created inside features with `createMockFrontend` and mock backends with `createMockBackend`.
- Added a new feature to test the greetServer message.
- Added tags to each scenario to make it possible to run individual scenarios on the command line (with `mvn test -Dkarate.options="--tags @scenarioTagName"`). The Karate plugin for Intellij put the functionality to do this in the IDE behind a paywall 😞


tests/karate/src/test/java/be/features/utils/server.feature
```diff
@@ -37,14 +37,27 @@ Feature: This feature starts a server and stops it after every scenario.
...
                     }
+
                     if(i >= MAX_CONNECTION_ATTEMPTS){
                       server.stop()
-                      karate.fail(`Failed waiting for ${wsURL}`)
+                      var error = ""
+                      if(!frontendPortReady)
+                        error += `Failed waiting for ${frontendWsURL}. `
+                    if(!backendPortReady)
+                        error += `Failed waiting for ${backendWsURL}.`
```
pierluca, line 59: indentation seems off

tests/karate/src/test/java/be/features/decentralizedCom/greetServer.feature
```diff
@@ -0,0 +1,85 @@
...
+  @greetServer3
+  Scenario: Server should not reply to a greetServer message with invalid client address (does not fit regex pattern)
+    Given def invalidGreetServer =
+        """
+          {
+            "jsonrpc": "2.0",
+            "method": "greet_server",
+            "params": {
+              "public_key": "#(mockBackend.publicKey)",
+              "client_address": "https://invalid:9000/client",
```
pierluca, line 58: why should `invalid` be invalid ? what's the regex pattern ? 
I'm asking because technically `invalid` could be a valid hostname in the local network.
A wrong scheme (other than `https`) or an invalid port (>65535) might make more sense.

pierluca (commented):
Minor remarks, but otherwise looks pretty good to me ! 
Also, so sad about the IntelliJ plugin :-(

tests/karate/src/test/java/be/features/decentralizedCom/greetServer.feature
simone-kalbermatter: The regex does not accept https (only ws or wss), so that is actually what's invalid here, not the host name. But this is not very clear, I'll change it to a clearly invalid scheme.

pierluca (dismissed):

pierluca (approved):

Issue #1722, opened 2023-12-14
[BUG] be2-Scala: Election broken


K1li4nL:
### Description (Actual behavior)
Election result do not appear after election end:
![image](https://github.com/dedis/popstellar/assets/79536516/efb9aa40-7e43-4302-9891-e423f892bfd0)

### Expected behavior
We expect election results to be visible when the election is closed. 

### How to reproduce
1. Run a rollcall and then run an election up to completion.

### Version & Environment
This bug was reproduced on: 
- Commit: f325a91a80115f3b9df16da1a405cb5f346cbee5

##### Front-ends:
Tested with Fe1 clients only
- [X] Fe1-Web (please include browser's names & version)
- [ ] Fe2-Android (please specify if phone or emulation, and Android version)
- [ ] Not applicable

##### Back-ends:
- [ ] Be1-Go
- [X] Be2-Scala
- [ ] Not applicable

### Workaround
None.

### Impact
Election pretty useless in that state.

### Possible root cause
This feature is most likely broken since pr #1675, in the rush of the semester end we did not properly test the impacts of such changes. 
The scala backend relied on storing the GreetLao in the db so that client catching up would receive the greet. The issue is that the message was broadcasted only when the server would receive a `create_lao`. But since the lao is freshly created, no one is subscribed to it yet, hence we 100% relied on the db to do the job.

K1li4nL:
### Potential solutions
A) Manually include the GreetLao in the catchup 
B) Change the protocol so that GreetLao are expected as responses to Subscribe request

### Drawbacks:
Solution A) is not very elegant but it would work and only require changes on the scala backend.
 
Solution B)  make logical sense as GreetLao are expected by the client when they subscribe to the channel. But implementation most likely require changes on all subsystems. This implementation can go in two ways:

1: GreetLao is sent as a separate message from the server to the client.
2: GreetLao is sent as a JsonRpcResponse from the server back to the client.

### Drawbacks again:
1: Might lead to bug if the client is catching up before it received the GreetLao as they might reject ElectionResults at this point.

2: Heavy changes to the protocol + frontends so that they can parse GreetLao inside JsonRpcResponses, but will avoid weird time-dependent behavior.

PR #1731, opened 2024-02-05
[BE2]: Fix elections

K1li4nL:
This pr is in reference to #1722, it fixes the elections by manually building and adding GreetLao to the catchup message list.


MariemBaccari (approved):
LGTM

PR #1732, opened 2024-02-05
Popcha flaky test fix 

MariemBaccari:
This PR aims to fix a flaky POPCHA test that sometimes fails when the CI runs by adding a `time.sleep` that allows the logs to update before the checks.

### DIFF
--- a/be1-go/go.mod
+++ b/be1-go/go.mod
@@ -26,6 +26,7 @@ require (
 )
 
 require (
+	github.com/BurntSushi/toml v1.2.1 // indirect
 	github.com/cpuguy83/go-md2man/v2 v2.0.0-20190314233015-f79a8a8ca69d // indirect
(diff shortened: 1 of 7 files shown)


jbsv (changes requested):
Be careful that the `time.Sleep()` might work on some machines but might still fail on others. To be sure that it works on most machines, you got to be generous with the delay. The problem with this approach is that you end up with unnecessary long tests, that would run way faster if the concurrency issue is fixed. On projects where you have a few hundreds of tests "fixed" with this bad habit (that's the experienced, lazy developer in me talking), tests might take a very long time to execute and even become a threat to a project.
The behavior of the logger is very annoying in this case, but may I request to either leave a comment on how to do it better (as in here: https://stackoverflow.com/questions/52935858/how-and-when-is-the-go-sdk-logger-flushed), or try to fix it with a flush ?

PR #1743, opened 2024-02-13
Fix PoPCHA flaky test

### LINKED ISSUE #1733: [BUG] TestAuthorizationServerWorkflow fails sometimes during CI runs
The test is flaky. It sometimes fails during CI runs.

### Expected behavior
The test should always pass 

### Version & Environment
This bug was reproduced in [this run](https://github.com/dedis/popstellar/actions/runs/7783203821/job/21221173895?pr=1731).

##### Front-ends:
- [x] Not applicable

##### Back-ends:
- [x] Be1-Go
- [ ] Be2-Scala
- [ ] Not applicable

### Possible root cause
The check that fails is an equality assertion in the logs. When the test fails, it prints the previous log which indicates that the expected log  wasn't printed yet. This could be fixed by adding a flush before the check to allow the synchronization of the logs.

MariemBaccari:
This PR aims to fix a flaky PoPCHA test that sometimes fails during the CI runs by adding flushes before checking the logs.
Closes #1733

### DIFF
--- a/be1-go/popcha/server_test.go
+++ b/be1-go/popcha/server_test.go
@@ -1,12 +1,14 @@
 package popcha
 
 import (
+	"bufio"
 	"bytes"
 	"fmt"
 	"io"
 	"math/rand"
 	"net/http"
 	"net/url"
+	"os"
 	"popstellar"
 	"popstellar/crypto"
 	"popstellar/hub"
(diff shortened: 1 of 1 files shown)


jbsv (approved):
LGTM, thanks

Issue #1748, opened 2024-02-20
AddPeerInfo can override existing server information 


MariemBaccari:
### Description (Actual behavior)
The `AddPeerInfo` function sets the information of the server without any validation which allows it to override the information. 

### Expected behavior
It should only raise an error if the call tries to override the information for a `socket.ID()`. 

##### Front-ends:
- [ ] Fe1-Web (please include browser's names & version)
- [ ] Fe2-Android (please specify if phone or emulation, and Android version)
- [x] Not applicable

##### Back-ends:
- [x] Be1-Go
- [ ] Be2-Scala
- [ ] Not applicable

### Possible root cause
We simply need to add a condition inside the function.

PR #1752, opened 2024-02-23
Prevent AddPeerInfo to override existing server information

arnauds5:
Addresses the issue #1748
The `AddPeerInfo` function is now checking if it already contains a value for the specified socketId before writing the server information.
If it is called with an existing socketId, it will throw an error.

### DIFF
--- a/be1-go/hub/standard_hub/hub_state/Peers.go
+++ b/be1-go/hub/standard_hub/hub_state/Peers.go
@@ -1,6 +1,7 @@
 package hub_state
 
 import (
+	"golang.org/x/xerrors"
 	"popstellar/message/query/method"
 	"sync"
 
@@ -26,10 +27,17 @@ func NewPeers() Peers {
 }
 
 // AddPeerInfo adds a peer's info to the table
-func (p *Peers) AddPeerInfo(socketId string, info method.ServerInfo) {
+func (p *Peers) AddPeerInfo(socketId string, info method.ServerInfo) error {
 	p.Lock()
 	defer p.Unlock()
+
+	_, ok := p.peersInfo[socketId]
+	if ok {
+		return xerrors.Errorf("peersInfo already contains [%s]", socketId)
+	}
+
 	p.peersInfo[socketId] = info
+	return nil
 }
 
 // AddPeerGreeted adds a peer's socket ID to the slice of peers greeted
--- a/be1-go/hub/standard_hub/message_handling.go
+++ b/be1-go/hub/standard_hub/message_handling.go
@@ -451,7 +451,10 @@ func (h *Hub) handleGreetServer(socket socket.Socket, byteMessage []byte) error 
 	}
 
 	// store information about the server
-	h.peers.AddPeerInfo(socket.ID(), greetServer.Params)
+	err = h.peers.AddPeerInfo(socket.ID(), greetServer.Params)
+	if err != nil {
+		return xerrors.Errorf("failed to add peer info: %v", err)
+	}
 
 	if h.peers.IsPeerGreeted(socket.ID()) {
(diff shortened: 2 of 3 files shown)


be1-go/hub/standard_hub/hub_state/Peers.go
```diff
@@ -26,10 +27,17 @@ func NewPeers() Peers {
...
 
 // AddPeerInfo adds a peer's info to the table
-func (p *Peers) AddPeerInfo(socketId string, info method.ServerInfo) {
+func (p *Peers) AddPeerInfo(socketId string, info method.ServerInfo) error {
 	p.Lock()
 	defer p.Unlock()
+
+	_, ok := p.peersInfo[socketId]
+	if ok {
+		return xerrors.Errorf("peersInfo already contains [%s]", socketId)
```
MariemBaccari, line 36: I think the logs would be a bit more readable if they included the peer's info than the socket id. Maybe something like "peersInfo already contains socketid *id* for server *peer's address* would be useful for debugging.

MariemBaccari (changes requested):
Good job, I suggest a minor change before approving.

be1-go/hub/standard_hub/hub_state/Peers.go
arnauds5: I changed it to also include the peer's info.

be1-go/hub/standard_hub/hub_state/Peers.go
```diff
@@ -26,10 +27,17 @@ func NewPeers() Peers {
...
 // AddPeerInfo adds a peer's info to the table
-func (p *Peers) AddPeerInfo(socketId string, info method.ServerInfo) {
+func (p *Peers) AddPeerInfo(socketId string, info method.ServerInfo) error {
 	p.Lock()
 	defer p.Unlock()
+
+	_, ok := p.peersInfo[socketId]
+	if ok {
+		return xerrors.Errorf("peersInfo already contains socketId %s for server %s", socketId, info)
+	}
```
sgueissa, line 37: You should change `info` inside the `xerrors.Errorf` by the value you get from `p.peersInfo[socketId]` because we want to say that the socket is already used by an other server.

sgueissa (changes requested):
Nice just one little thing to change

be1-go/hub/standard_hub/hub_state/Peers.go
sgueissa: Or something like 'peersInfo cannot add peer %s because socketId %s already exists'

arnauds5: Do you thing that including both values of info could be useful ?
for example "peersInfo cannot add peer *new_value* because  it already contains peer *current_value* for socketID *id*"

sgueissa: Yes but in this case you should only print the peerId without the servers informations for both

sgueissa: peersInfo cannot add new peer new_peerID because socket socketID is already used by peer current_peerID

arnauds5: peersInfo is a map of socketId to ServerInfo, maybe the following could be more readable ?
"cannot add *new_peer_info* because peersInfo[*id*] already contains *current_peer_info*"

sgueissa: I don't think that adding the peer information other than its public key is useful in this case.

sgueissa (approved):

MariemBaccari (approved):
LGTM

PR #1753, opened 2024-02-24
Work be1 stuart fix query already answered

sgueissa:
the `SetQueryReceived` function now doesn't return an error when receiving more than once an answer to one of its query and just logs it.

the `Test_Handle_Answer` test has also been fix to work with this bug fix.

### DIFF
--- a/be1-go/hub/standard_hub/mod_test.go
+++ b/be1-go/hub/standard_hub/mod_test.go
@@ -821,7 +821,7 @@ func Test_Handle_Answer(t *testing.T) {
 		c: &fakeChannel{},
 	}
 
-	hub, err := NewHub(keypair.public, "", "", nolog, fakeChannelFac.newChannel)
+	hub, err := NewHub(keypair.public, "", "", yeslog, fakeChannelFac.newChannel)
 	require.NoError(t, err)
 
 	result := struct {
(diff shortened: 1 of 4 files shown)


be1-go/hub/standard_hub/mod_test.go
```diff
@@ -821,7 +821,7 @@ func Test_Handle_Answer(t *testing.T) {
 		c: &fakeChannel{},
 	}
 
-	hub, err := NewHub(keypair.public, "", "", nolog, fakeChannelFac.newChannel)
+	hub, err := NewHub(keypair.public, "", "", yeslog, fakeChannelFac.newChannel)
```
emonnin-epfl, line 824: Since you don't test if the error has been logged or not, I'm not sure why you would enable logs during testing

emonnin-epfl (changes requested):
Everything seems good except one minor thing.

be1-go/hub/standard_hub/mod_test.go
sgueissa: I change that by writing my logs inside a buffer then checking it

emonnin-epfl (approved):

MariemBaccari (approved):
LGTM !

Issue #1747, opened 2024-02-20
Hub blacklist is not thread safe


MariemBaccari:
### Description (Actual behavior)
The hub's blacklist is a simple array of strings which is not thread safe since we implemented the fine grain locking of the hub as opposed to locking the entire hub at every step. 

### Expected behavior
The blacklist should be implemented as a thread safe structure.

##### Front-ends:
- [ ] Fe1-Web (please include browser's names & version)
- [ ] Fe2-Android (please specify if phone or emulation, and Android version)
- [x] Not applicable

##### Back-ends:
- [x] Be1-Go
- [ ] Be2-Scala
- [ ] Not applicable

### Impact
This can cause concurrency issues.

### Possible root cause
In the `Hub struct`, `blacklist` is implemented as `[]string`. We should use a thread safe structure instead and make the required changes in the parts of the code that use it.

PR #1754, opened 2024-02-24
Make []string slice 'blacklist' thread-safe

emonnin-epfl:
Addresses the issue #1747 

The []string slice 'blacklist' should now be thread-safe.

### DIFF
--- a/be1-go/hub/standard_hub/hub_state/ThreadSafeSlice.go
+++ b/be1-go/hub/standard_hub/hub_state/ThreadSafeSlice.go
@@ -1,0 +1,29 @@
+package hub_state
+
+import (
+	"golang.org/x/exp/slices"
+	"sync"
+)
+
+type ThreadSafeSlice[E comparable] struct {
+	sync.RWMutex
+	els []E
+}
+
+func NewThreadSafeSlice[E comparable]() ThreadSafeSlice[E] {
+	return ThreadSafeSlice[E]{
+		els: make([]E, 0),
+	}
+}
+
+func (i *ThreadSafeSlice[E]) Append(elems ...E) {
+	i.Lock()
+	defer i.Unlock()
+	i.els = append(i.els, elems...)
+}
+
+func (i *ThreadSafeSlice[E]) Contains(elem E) bool {
+	i.RLock()
+	defer i.RUnlock()
+	return slices.Contains(i.els, elem)
+}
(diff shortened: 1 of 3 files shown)


be1-go/hub/standard_hub/message_handling.go
```diff
@@ -468,11 +469,11 @@ func (h *Hub) handleGreetServer(socket socket.Socket, byteMessage []byte) error
 
 // getMissingIds compares two maps of channel Ids associated to slices of message Ids to
 // determine the missing Ids from the storedIds map with respect to the receivedIds map
-func getMissingIds(receivedIds map[string][]string, storedIds map[string][]string, blacklist []string) map[string][]string {
+func getMissingIds(receivedIds map[string][]string, storedIds map[string][]string, blacklist *hub_state.ThreadSafeSlice[string]) map[string][]string {
```
sgueissa, line 472: Do you really need to give a pointer for blacklist?

be1-go/hub/standard_hub/hub_state/ThreadSafeSlice.go
```diff
@@ -0,0 +1,29 @@
...
+	i.Lock()
+	defer i.Unlock()
+	i.els = append(i.els, elems...)
+}
+
+func (i *ThreadSafeSlice[E]) Contains(elem E) bool {
+	i.RLock()
+	defer i.RUnlock()
+	return slices.Contains(i.els, elem)
+}
```
sgueissa, line 29: Shouldn't you add a method to remove an element from the blacklist in case we need it later?

sgueissa (changes requested):
looks good just some minor things to change

be1-go/hub/standard_hub/hub_state/ThreadSafeSlice.go
emonnin-epfl: I think we should keep the code as simple as possible and add functions only when we need them

be1-go/hub/standard_hub/message_handling.go
emonnin-epfl: If I pass blacklist as a value, a copy of the Mutex will be made along with the struct, leading to unexpected behaviours

sgueissa: you are right mybad

sgueissa (approved):

MariemBaccari (approved):
LGTM, good job !

Issue #1751, opened 2024-02-22
Complete message validation


simone-kalbermatter:
To address issue #1375, the  `MessageValidator` class was implemented. 
The functions it contains are used to validate message data objects at their creation. 

Message data objects that are already validated:
- CreateLao
- GreetLao
- UpdateLao
- StateLao
- ElectionSetup
- ElectionKey
- ElectionOpen
- CastVote
- ElectionEnd
- ElectionResult

The goal is to eventually validate all messages. The  `MessageValidator` can be extended to add more validation functions.

PR #1758, opened 2024-02-26
Further work on message validation

Kaz-ookid:
Addressing #1751 

This PR will hold new message data object validation, to help the effort of complete message validation depicted in #1751.

Note to myself: remember to update the list of validated data in #1751, after this PR gets approved.


Kaz-ookid:
Hint for reviewers:

The construction of a Question is now validated. I also am checking the question handed in the ElectionQuestion for the exact same constraints. This means we are validating the question twice, thus reducing coverage (since we can't construct any invalid question to hand to the ElectionQuestion constructor). It may not be needed to validate the question in ElectionQuestion, but I stil did just in case. I am open to suggestion about it.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/election/ElectionQuestion.kt
```diff
@@ -2,22 +2,33 @@ package com.github.dedis.popstellar.model.network.method.message.data.election
...
-  @SerializedName(value = "ballot_options")
-  val ballotOptions: List<String> = Collections.unmodifiableList(question.ballotOptions)
+  @SerializedName(value = "ballot_options") val ballotOptions: List<String>
 
-  @SerializedName(value = "write_in") val writeIn: Boolean = question.writeIn
+  @SerializedName(value = "write_in") val writeIn: Boolean
+
+  init {
+    // questions are already validated when constructed, but never too much validation.
+    verify().isNotEmptyBase64(electionId, "election ID").validQuestions(listOf(question))
```
matteosz, line 24: You do have your point with 'never too much validation', but in this case we're executing exactly the same check twice, as the only injection point for questions performs the check already. In my opinion we can avoid doing that but feel free to leave it as is

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/election/ElectionQuestion.kt
```diff
@@ -67,7 +92,7 @@ class ElectionQuestion(electionId: String, question: Question) {
     }
 
     override fun hashCode(): Int {
-      return Objects.hash(title, votingMethod, ballotOptions, writeIn)
+      return Objects.hash(title, votingMethod, ballotOptions)
```
matteosz, line 95: Is there a reason why we're excluding the `writeIn` from the hash?

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/MessageValidator.kt
```diff
@@ -173,6 +175,19 @@ object MessageValidator {
...
+     * @param input the encrypted write-in to check
+     * @param field name of the field (to print in case of error)
+     * @throws IllegalArgumentException if the string is empty or not a URL-safe base64 encoding
+     */
+    fun isValidWriteIn(input: String?, field: String): MessageValidatorBuilder {
+      requireNotNull(input) { "Write-in is enabled but encrypted write-in is null" }
+      isNotEmptyBase64(input, field)
+      return this
+    }
+
```
matteosz, line 190: I think we should try to keep the MessageVerify as generic as possible, and I find this function a bit redundant and very specific to just a single use. I'm assuming that the purpose was to have a more custom message in case of null input, but if that's the case it could be done differently (like extend the isNotEmptyBase64 to support custom messages).

The `isNotEmptyBase64` function already performs this check for you (not null), so in the end this function is just a wrapper with no strong need for it.

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/MessageValidator.kt
```diff
@@ -219,6 +234,54 @@ object MessageValidator {
...
+     * @param ballotOptions the ballot options of the question
+     * @throws IllegalArgumentException if the question does not meet the criteria.
+     */
+    fun validQuestion(
+        title: String,
+        votingMethod: String,
+        ballotOptions: List<String>,
+    ): MessageValidatorBuilder {
+      stringNotEmpty(title, "question title")
+      val validVotingMethods = ElectionSetupFragment.VotingMethods.values().map { it.desc }
```
matteosz, line 266: This looks like a constant that can be extract as static field

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/MessageValidator.kt
```diff
@@ -219,6 +234,54 @@ object MessageValidator {
...
+    }
+
+    private fun validBallotOptions(ballotOptions: List<String>?): MessageValidatorBuilder {
+      requireNotNull(ballotOptions) { "Ballot options cannot be null" }
+      listNotEmpty(ballotOptions)
+      require(ballotOptions.size >= 2) { "There must be at least 2 ballot options" }
+      noListDuplicates(ballotOptions)
+      ballotOptions.map {
+        stringNotEmpty(it, "ballot option in place " + ballotOptions.indexOf(it))
+      }
```
matteosz, line 281: I don't actually get the usage of map here, as we don't actually want to map the result of the verification back in the list. I'd rather use a forEach() instead

fe2-android/app/src/test/unit/java/com/github/dedis/popstellar/model/network/method/message/data/election/CastVoteTest.kt
```diff
@@ -26,15 +27,21 @@ class CastVoteTest {
...
   private val writeInEnabled = false
   private val writeIn = "My write in ballot option"
+  private val vote1 = 1
+  private val vote2 = 2
+  private val vote1Base64 = Base64URLData(vote1.toString().toByteArray()).encoded
+  private val vote2Base64 = Base64URLData(vote2.toString().toByteArray()).encoded
 
   // Set up a open ballot election
   private val plainVote1 = PlainVote(questionId1, 1, writeInEnabled, writeIn, electionId)
   private val plainVote2 = PlainVote(questionId2, 2, writeInEnabled, writeIn, electionId)
```
matteosz, line 37: Now that you defined vote1 and vote2 as variable you could also modify plainVote1/2 to use such variables instead of the hardcoded 1 and 2

matteosz (changes requested):
Really good work! Just some picky comments

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/election/ElectionQuestion.kt
Kaz-ookid: I think it was when removing the `writeIn` in argument in `validQuestion` (figuring out that there's noting to check for a boolean here). I probably blindly removed it here thinking it was the `verify().validQuestion` line, since the order of arguments matched...

Nice catch thanks!

Kaz-ookid:
Thanks a lot @matteosz for your insightful review :+1:

matteosz:
Just a reminder also for future PRs (you can skip it for this one): try to keep commits number as low as possible. 
I just saw you did a commit `formatting` and then `runktmftm Format`, plus probably 16 commits for a 300 lines change is a bit too much (but it's just my subjective opinion obv).

Commits are a great mechanism for tracking history, but if multiple commits have the same purpose this simply increases the entropy of the codebase and "obfuscate" a good commit history. So try to squash together all the commits that serve to the same purpose.

Simply run `git rebase -i HEAD~<number_of_commits>` and then manually go through commits and decide whether to squash or keep them. Keep in mind that if you squash commits already pushed you'd need to force push, so be careful about it.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/election/ElectionQuestion.kt
```diff
@@ -2,22 +2,32 @@ package com.github.dedis.popstellar.model.network.method.message.data.election
...
 
-  @SerializedName(value = "ballot_options")
-  val ballotOptions: List<String> = Collections.unmodifiableList(question.ballotOptions)
+  @SerializedName(value = "ballot_options") val ballotOptions: List<String>
 
-  @SerializedName(value = "write_in") val writeIn: Boolean = question.writeIn
+  @SerializedName(value = "write_in") val writeIn: Boolean
+
+  init {
+    verify().isNotEmptyBase64(electionId, "election ID").validQuestions(listOf(question))
```
matteosz, line 23: Saw that you did a commit named `ElectionQuestion : removed overlapping Question validations` but it's actually an empty commit (just some formatting), so I guess you forgot to remove the validQuestions check here

Kaz-ookid: I actually removed the redundant checks performed in `validQuestions` (calling `validQuestion` for each Question in the list). validQuestions is still used, as it checks for duplicates and requires the list not to be empty. Maybe it makes it not relevant anymore to be a function in `MessageValidator` tho.

quadcopterman (approved):
Seems OK to me

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/election/ElectionQuestion.kt
matteosz: Yeah the point was: why are you still calling `.validQuestions(listOf(question))` in `ElectionQuestion`? The fact that the list is not empty and without duplicates is given by the construction of it (a list of 1 element)

Kaz-ookid: oh yes that makes a lot of sense, I overlooked

matteosz (dismissed):
LGTM!

matteosz (approved):

Issue #1680, opened 2023-06-21
Go backend should also send empty heartbeats


MariemBaccari:
### Description (Actual behavior)
For now, the Go backend only sends heartbeats if it has stored messages. 

### Expected behavior
To match the Scala behavior and to still show that the server is alive, the heartbeats should be sent even if they're empty. 

### Possible root cause
There's a condition that is checked before sending heartbeats.

PR #1760, opened 2024-02-27
remove non-empty condition on hubInbox for heartbeats

1florentin:
Just removes the check if the hubInbox is empty before sending heartbeats.
This is done in accordance to #1680 -- to match the behaviour of the scala backend.


arnauds5 (approved):
LGTM

Issue #1749, opened 2024-02-21
[BE2]: Kickstart peer-to-peer connection upon server start


K1li4nL:
For now the config file is `src/main/scala/ch/epfl/pop/config/server-peers-list.conf` is only watched for changes and read when such change is detected. Ideally the file should be read when the server starts.

The following class are of interest in solving this problem:

- `src/main/scala/ch/epfl/pop/decentralized/Monitor.scala`
- `src/main/scala/ch/epfl/pop/decentralized/ConnectionMediator.scala`

PR #1764, opened 2024-02-28
[BE2] #1749 Start connection upon start

DanielTavaresA:
- Added read instruction at the start of the FileMonitor which is created in response to ping at the start of connectionMediator


K1li4nL (approved):
LGTM !

Issue #1734, opened 2024-02-06
Failing test under specific conditions on FE2


matteosz:
### Description (Actual behavior)
The test `cannotChooseStartTimeTooFarInPast` ([ElectionSetupFragmentTest.kt](https://github.com/dedis/popstellar/blob/master/fe2-android/app/src/test/ui/robolectric/com/github/dedis/popstellar/ui/lao/event/election/ElectionSetupFragmentTest.kt)) fails consistently when executed between 00:00 and 00:09am.

### Expected behavior
The test shouldn't be failing anytime.

### How to reproduce
Execute this test in the specified time.

### Impact
Minimal to none impact (time window of the test failing is late in the night and narrow)

### Possible root cause
There are 2 possibilities to investigate:
1. The logic tested is not robust in the time going during 00:00 and 00:09. This means that the test discovered an implementation bug and it should be fixed
2. The test logic is flaky and doesn't reproduce an expected behaviour

PR #1766, opened 2024-02-29
Finally solved issue #1734

quadcopterman:


Kaz-ookid:
my bad for marking it as ready to review, forgot some tests are to come

quadcopterman:
This PR is another way to solve issue #1734 (much much simpler than my previous attempt) : just let the day comparison go to -1 when comparing the start date of an event (in case midnight is just between start date and now). If this event is too far in the past, it will be handeled by computeTimeInSeconds() as Instant also contains the date.
A test was added to make sure that a 1 day old date won't work and that this issue won't occur again

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/event/AbstractEventCreationFragment.kt
```diff
@@ -202,7 +202,9 @@ abstract class AbstractEventCreationFragment : Fragment() {
     startDateEditText?.setText("")
     startDate = null
 
-    if (compareWithNowByDay(newDate) < 0) {
+    // let the comparison go to -1 for cases where the time is just after midnight
+    // this is handled just fine by computeTimeInSeconds() as an Instant also contains the date
+    if (compareWithNowByDay(newDate) < -1) {
```
matteosz, line 207: Refactor opportunity: this check appears twice (here and line 224), so it'd be better to extract it into a variable

matteosz (commented):
LGTM! 

Just 1 small comment

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/event/AbstractEventCreationFragment.kt
quadcopterman: Yes, thanks for spotting that, I'll correct it

Kaz-ookid (approved):

matteosz (approved):
LGTM!

PR #1767, opened 2024-02-29
Refactor files name

sgueissa:
Renaming the mod.go files without touching the message package because it will need more work than just renaming the files.


emonnin-epfl (approved):
LGTM!

MariemBaccari (approved):
I can see some changes inside files but I assume it's because of some automatic IDE formatting issue. Assuming you just renamed the file, the PR looks good to me, good job!

PR #1770, opened 2024-03-03
PlainVote validation

Kaz-ookid:
addresses #1751

Adds validations et tests at construction of PlainVote.

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/election/PlainVote.kt
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/election/PlainVote.kt
@@ -32,11 +32,11 @@ class PlainVote : Vote {
       vote: Int?,
       writeInEnabled: Boolean,
       writeIn: String?,
-      electionId: String?,
+      electionId: String
   ) {
     verify()
         .isNotEmptyBase64(questionId, "question ID")
-        .isNotEmptyBase64(electionId ?: "", "election ID")
+        .isNotEmptyBase64(electionId, "election ID")
         .apply {
           if (writeInEnabled) {
             isNotEmptyBase64(writeIn, "write-in")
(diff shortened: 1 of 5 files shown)


Kaz-ookid:
ready for review :tada:

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/election/PlainVote.kt
```diff
@@ -31,14 +32,31 @@ class PlainVote : Vote {
       vote: Int?,
       writeInEnabled: Boolean,
       writeIn: String?,
-      electionId: String?
+      electionId: String?,
   ) {
+    verify()
+        .isNotEmptyBase64(questionId, "question ID")
+        .isNotEmptyBase64(electionId ?: "", "election ID")
```
matteosz, line 39: I don't see any reason why we should pass an empty string when it's null rather than just pass directly `electionId`.

If electionId is null in the `Election.generateElectionVoteId` function it will throw the illegal argument exception, so we can safely check if it's null right from the beginning in the MessageValidator

matteosz (commented):
LGTM, just a small comment

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/election/PlainVote.kt
Kaz-ookid: I remember I did this because IDE was screaming about it, but it's not anymore... Strange :man_shrugging:  
You are right tho, thanks :+1: 

Also is there a reason we had `String?` and not `String` for electionId? I did not want to touch it without knowing, but in my opinion we can just remove the `?` to ensure we are not given a null argument.

quadcopterman (approved):
Looks good to me

matteosz (approved):
Great work! 

Nit comment: try not to have commits like 'addressing comment from XX' but rather 'doing XX', as commits are often seen without their PR, so no one actually will know what the commit does without looking at the code change

Issue #847, opened 2022-02-25
Fe1-Web CI is slow / node_modules are not cached


pierluca:
### Description (Actual behavior)
The action `Setup npm dependencies` takes >90s on each run.

### Expected behavior
`Setup npm dependencies` should be extremely fast and only slow when dependencies change, i.e. when the `package-lock.json` file has changed and new dependencies need to be installed and the cache file re-created.

### Possible root cause
Currently, there's an action called `Setup node cache` in the [CI definition](https://github.com/dedis/popstellar/blob/master/.github/workflows/ci.yaml) for Fe1-Web, but it only caches the node environment and not the `node_modules` directory.
It might also be worthwhile to look at whether `Setup expo action` could be cached, as it takes ~30s.

### How to reproduce
Open any PR and look at the execution of the Fe1-Web actions.
e.g. https://github.com/dedis/popstellar/runs/5336261406?check_suite_focus=true#logs

### Version & Environment
This bug was reproduced on:
- Commit: https://github.com/dedis/popstellar/pull/844/commits/6fb168e044b1a8b9dec3b21a4388cb3f0ae818e7

##### Front-ends:
- [x] Not applicable

##### Back-ends:
- [x] Not applicable

##### Environment (as applicable):
 - GitHub Actions

### Workaround
Being patient :-)

### Impact
It takes a longer time to obtain feedback from GitHub actions (>5min instead of ~3min)

PR #1771, opened 2024-03-04
Fix CI node_modules caching issue #847

osm-alt:
Added the node_modules path to the caching action of the workflow for fe1-web.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -43,18 +43,28 @@ jobs:
       - name: Setup node cache
         uses: actions/cache@v2
         env:
-          cache-name: cache-node-modules
+          cache-name: cache-node-npm
         with:
           # npm cache files are stored in `~/.npm` on Linux/macOS
-          path: |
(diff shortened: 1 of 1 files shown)


osm-alt:
Removed caching of node_modules on MacOS due to my previous comment but left it for the other platforms

4xiom5 (dismissed):
Seems good

.github/workflows/ci.yaml
```diff
@@ -53,6 +53,16 @@ jobs:
             ${{ runner.os }}-build-
             ${{ runner.os }}-
 
+      - name: Setup node modules cache
+        if: ${{ runner.os != 'macOS'}}
```
Tyratox, line 57: Why exclude MacOs? Maybe add a comment?

Tyratox (changes requested):
Maybe add a small comment/explanation why macOS is excluded, otherwise LGTM!

Tyratox (approved):

PR #1786, opened 2024-03-20
Fixes #1784 Clear the previous PoP tokens before adding new ones

### LINKED ISSUE #1784: [BUG] PoP Tokens are always valid on be1-go
The Go backend does not verify that a PoP Token was from the latest rollcall but checked against all PoP tokens ever.
It is possible to vote using PoP tokens from an older rollcall on the Go backend.

### Expected behavior
Reject any vote if the PoP Token is not from the latest roll call (already done in Scala)

### How to reproduce
1. Participate in a rollcall_1 with token_1
2. Don't participate in the next rollcall_2
3. Vote on an election (with the token_1)
4. The vote will be counted

### Version & Environment
This bug was reproduced on:
- Commit: 8f30905436545d24b2ee85c8ce1d0d63e289cebf

##### Front-ends:
- [ ] Not applicable

##### Back-ends:
- [x] Be1-Go
- [ ] Be2-Scala
- [ ] Not applicable

### Workaround
The Scala backend already checks that only PoP tokens from the latest roll call are valid.

### Impact
It's possible to vote to an election without participating in the latest rollcall.
With a modified frontend, it can be possible to accumulate PoP tokens from multiple consecutive rollcalls and use them all on one election.

### Possible root cause
When an election channel is created, the attendees valid keys are copied from the current keys on the Lao channel. Theses keys are only updated by adding each attendee of a rollcall#close. But it is never cleared.
A solution to this problem could be to clear the attendees map of the Lao during a rollcall#close.

arnauds5:
fixes #1784 
When a valid lao#close is received, the attendees map in the lao is replace by an empty one, before the new keys are added.

### DIFF
--- a/be1-go/channel/lao/lao.go
+++ b/be1-go/channel/lao/lao.go
@@ -429,6 +429,7 @@ func (c *Channel) processRollCallClose(msg message.Message, msgData interface{},
 	c.rollCall.id = data.UpdateID
 	c.rollCall.state = Closed
 
+	c.attendees = make(map[string]struct{})
 	for _, attendee := range data.Attendees {
 		c.attendees[attendee] = struct{}{}
 


MariemBaccari (approved):
LGTM

Issue #1776, opened 2024-03-11
Highlight our token during Roll Calls


quadcopterman:
### Description (Actual behavior)
When doing a Roll Call the list of tokens is displayed (during the roll call for the organizer and after the roll call for every participant), but our token is not clearly distinguishable.

### Expected behavior
Being able to clearly see our token in the list (like it is done in fe1).

This was mentioned during a global meeting.

PR #1788, opened 2024-03-21
Highlighting Owned Token in Attendees List

quadcopterman:
I created an Adapter with a new getView() function. This function is then called when the tokens are displayed in a rollcall, changed how our token is displayed (to highlight only our token)
This solves issue #1776

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/event/rollcall/RollCallArrayAdapter.kt
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/event/rollcall/RollCallArrayAdapter.kt
@@ -1,12 +1,12 @@
 package com.github.dedis.popstellar.ui.lao.event.rollcall
 
 import android.content.Context
-import android.graphics.Color
-import android.graphics.Typeface
 import android.view.View
 import android.view.ViewGroup
 import android.widget.ArrayAdapter
 import android.widget.TextView
+import androidx.core.content.ContextCompat
+import com.github.dedis.popstellar.R
 import com.github.dedis.popstellar.model.objects.security.PoPToken
 
 class RollCallArrayAdapter(
@@ -22,8 +22,8 @@ class RollCallArrayAdapter(
         //highlights our token in the list
         val currentToken = getItem(position)
         if (myToken != null && currentToken == myToken.publicKey.encoded) {
-            (view as TextView).setTypeface(null, Typeface.BOLD)
-            view.setTextColor(Color.BLUE)
+            val colorAccent = ContextCompat.getColor(context, R.color.colorAccent)
+            (view as TextView).setTextColor(colorAccent)
         }
 
         return view
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/event/rollcall/RollCallArrayAdapter.kt
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/event/rollcall/RollCallArrayAdapter.kt
@@ -10,10 +10,10 @@ import com.github.dedis.popstellar.R
 import com.github.dedis.popstellar.model.objects.security.PoPToken
 
 class RollCallArrayAdapter(
-  private val context: Context,
-  private val layout: Int,
-  private val attendeesList: List<String>,
-  private val myToken: PoPToken?,
+    private val context: Context,
+    private val layout: Int,
+    private val attendeesList: List<String>,
+    private val myToken: PoPToken?,
 ) : ArrayAdapter<String>(context, layout, attendeesList) {
 
   override fun getView(position: Int, convertView: View?, parent: ViewGroup): View {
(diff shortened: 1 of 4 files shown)


Kaz-ookid (commented):
LGTM!

It does not seem like this change requires testing, but you should make sure of it and ask a TA first :)

I have been successfully testing it and saw no issue. Maybe we'll keep an eye on it during the next global meeting!

Also, a small note on the PR description - the title "Fix issue #1776" could be more descriptive. How about "Highlighting Owned Token in Attendees List"? It's always good to be clear about what the PR does. Plus, mentioning the issue number as #1776 in the description creates a clickable link to the issue, which is super handy for quick navigation.

Keep up the great work!

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/event/rollcall/RollCallArrayAdapter.kt
```diff
@@ -0,0 +1,31 @@
...
+) : ArrayAdapter<String>(context, layout, attendeesList) {
+
+    override fun getView(position: Int, convertView: View?, parent: ViewGroup): View {
+        val view = super.getView(position, convertView, parent)
+
+        //highlights our token in the list
+        val currentToken = getItem(position)
+        if (myToken != null && currentToken == myToken.publicKey.encoded) {
+            (view as TextView).setTypeface(null, Typeface.BOLD)
+            view.setTextColor(Color.BLUE)
```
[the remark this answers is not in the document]
matteosz: Yes absolutely agree with @Kaz-ookid, avoid using constants for this kind of things, as very often there are reusable components, and use resources instead

matteosz (changes requested):
LGTM, just 2 comments about using resources and testing

matteosz:
> LGTM!
> 
> It does not seem like this change requires testing, but you should make sure of it and ask a TA first :)
> 
> I have been successfully manually testing it and saw no issue. Maybe we'll keep an eye on it during the next global meeting!
> 
> Also, a small note on the PR description - the title "Fix issue #1776" could be more descriptive. How about "Highlighting Owned Token in Attendees List"? It's always good to be clear about what the PR does. Plus, mentioning the issue number as #1776 in the description creates a clickable link to the issue, which is super handy for quick navigation.
> 
> Also, make sure to mark this PR as "ready for review" when you think you are done ! 👍
> 
> Keep up the great work!

Actually, I'd say that all the changes, even small ones, need validation. Testing this seems also straightforward (create a roll call with 2 attendees and verify that own token is highlighted and the other not) so I'd suggest to go for it

quadcopterman:
Thanks for your comments, I'll change the mentioned points.
For manual testing, I can't do a roll call for the moment as my emulator doesn't have access to my camera for some reason... I will try to solve that today and then do the requested changes

matteosz:
> Thanks for your comments, I'll change the mentioned points. For manual testing, I can't do a roll call for the moment as my emulator doesn't have access to my camera for some reason... I will try to solve that today and then do the requested changes

 Actually I'd suggest unit tests rather than manual testing, also because the code coverage is not above threshold

fe2-android/app/src/test/ui/robolectric/com/github/dedis/popstellar/ui/lao/event/rollcall/RollCallArrayAdapterTest.kt
```diff
@@ -0,0 +1,67 @@
...
+
+    @Mock
+    private lateinit var mockView: View
+
+    private lateinit var adapter: RollCallArrayAdapter
+
+    val context = ApplicationProvider.getApplicationContext<Context>()
+
+    private val MY_PRIVATE_KEY =
+            Utils.hexToBytes("3b28b4ab2fe355a13d7b24f90816ff0676f7978bf462fc84f1d5d948b119ec66")
```
matteosz, line 31: It's fine for me having this hardcoded, as it doesn't impact the test coverage, but we have in the codebase a random generator for keys that we use in test, just pointing this for future work as it's gonna save you some time

matteosz (approved):
It's fine by me, but the test is just a unit test for the adapter, which is obviously okay. I'd have liked to have at least 1 integration test that mimics the users' behaviour in the UI though, so adding a test in the RollCallFragment test, where you create a rollcall there and verify everything works (correct token is highlighted), in that case you could have avoided having a specific unit test for the adapter. But it's just my personal preference, it kinda works the same either way :)

fe2-android/app/src/test/ui/robolectric/com/github/dedis/popstellar/ui/lao/event/rollcall/RollCallArrayAdapterTest.kt
quadcopterman: Oh ok, thanks for the information

Kaz-ookid:
LGTM!

Issue #1783, opened 2024-03-20
Fe-1 disconnects from all connections if only one connection dies


osm-alt:
### Description (Actual behavior)
Fe-1 disconnects from all connections if only one connection dies as shown in the following screenshot:
![image](https://github.com/dedis/popstellar/assets/66867492/ac719a63-7897-479d-ab3b-70fd3c59bae6)

### Expected behavior
It would be better to only terminate disconnect from the connection concerned and not all if there is more than one concurrent connection.

##### Front-ends:
- [x] Fe1-Web 
- [ ] Fe2-Android
- [ ] Not applicable

##### Back-ends:
- [ ] Be1-Go
- [ ] Be2-Scala
- [ ] Not applicable

### Impact
The user would disconnect if only one connection fails.

PR #1790, opened 2024-03-23
Fixes issue #1783

osm-alt:
In case a connection dies, it now only disconnects from that connection and not all connections (considering that there are multiple connections). When there are no remaining connections, it navigates the to the home screen.

### DIFF
--- a/fe1-web/src/core/network/NetworkManager.ts
+++ b/fe1-web/src/core/network/NetworkManager.ts
@@ -139,7 +139,10 @@ class NetworkManager {
 
     if (alwaysPersistConnection) {
       // always push the connection to the array so that we can re-connect later
-      this.connections.push(connection);
+      // if it doesn't already exist to avoid duplicates
+      if (!this.connections.includes(connection)) {
+        this.connections.push(connection);
+      }
     }
 
     // if the inital connection attempt failed, reject the promise
@@ -162,7 +165,7 @@ class NetworkManager {
     }
   }
 
-  public disconnectFrom(address: string, intentional = true): void {
+  public disconnectFrom(address: string, intentional = true): Number {
     if (!address) {
       throw new Error('No address provided in disconnectFrom');
     }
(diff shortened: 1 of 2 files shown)


ljankoschek:
Looks good to me!

ljankoschek (approved):
Looks good to me!

fe1-web/src/core/network/NetworkManager.ts
```diff
@@ -139,7 +139,10 @@ class NetworkManager {
 
     if (alwaysPersistConnection) {
       // always push the connection to the array so that we can re-connect later
-      this.connections.push(connection);
+      // if it doesn't already exist to avoid duplicates
```
Tyratox, line 142: Great improvement! Do you think it makes sense to use a set instead of an array?

fe1-web/src/core/network/NetworkManager.ts
```diff
@@ -162,14 +165,16 @@ class NetworkManager {
     }
   }
 
-  public disconnectFrom(address: string, intentional = true): void {
+  public disconnectFrom(address: string, intentional = true): Number {
```
Tyratox, line 168: Maybe add a tsdoc describing the return value? Also did you intentionally chose to return `Number` and not `number`? (https://stackoverflow.com/a/67155149)

Tyratox (changes requested):
Nice improvement! I think it would be nice to add a small tsdoc for the return value but also let me know what you think of my other remarks/questions :)

fe1-web/src/core/network/NetworkManager.ts
osm-alt: Arrays in Typescript have useful functions that Sets do not. 
I see that sets can be useful because of the uniqueness of their elements, but due to the above point, I'd say we keep it as an array. I don't think that checking if a connection already exists in the array is going to introduce much overhead considering that the array is probably not going to be that big.
What do you think?

fe1-web/src/core/network/NetworkManager.ts
osm-alt: The tsdoc can be useful to have.
No, it was not intentional and I just learned about the difference between them from the link you shared. Thank you. Will be switching to `number` instead.

Tyratox (approved):

Issue #1793, opened 2024-03-26
Enable copying of LAO's ID and Server in Invite section


Kaz-ookid:
When trying to join a LAO, we can either scan the QR or manually enter the LAO's ID and server.
On FE2, it is not possible to copy the ID/Server in order to share it.

We should be able to copy them in an easy way, maybe by adding a tiny "copy" button.

For reference, here is the Invite page from a LAO, where we should be able to copy the text from:
![image](https://github.com/dedis/popstellar/assets/100324323/0e1e9762-f56a-47e6-9c13-a7925ef02e78)

PR #1797, opened 2024-04-03
Enhance Invite Section with Copy Functionality for LAO Details

Kaz-ookid:
## Description
This pull request addresses issue #1793 by introducing a copy button next to the LAO ID and server address within the invite section. Additionally, the LAO ID and server address fields have been made selectable for manual copying.

## Changes
- Added copy buttons for the LAO ID and server address.
- Enabled text selection on the LAO ID and server address TextViews for manual copying.

## Impact
- Enhances user experience by simplifying the sharing process of LAO information.
- Facilitates quick sharing without the need to manually type the details.

image

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/InviteFragment.kt
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/InviteFragment.kt
@@ -1,15 +1,10 @@
 package com.github.dedis.popstellar.ui.lao
 
-import android.content.ClipData
-import android.content.ClipboardManager
-import android.content.Context
 import android.graphics.Color
 import android.os.Bundle
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
-import android.widget.TextView
-import android.widget.Toast
 import androidx.fragment.app.Fragment
 import com.github.dedis.popstellar.R
 import com.github.dedis.popstellar.databinding.InviteFragmentBinding
@@ -32,7 +27,7 @@ class InviteFragment : Fragment() {
 
   private lateinit var laoViewModel: LaoViewModel
   private lateinit var binding: InviteFragmentBinding
-  private lateinit var clipboardManager : GeneralUtils.ClipboardUtil
+  private lateinit var clipboardManager: GeneralUtils.ClipboardUtil
 
   override fun onCreateView(
       inflater: LayoutInflater,
@@ -68,8 +63,10 @@ class InviteFragment : Fragment() {
     }
 
     handleBackNav()
-    clipboardManager.setupCopyButton(binding.copyServerButton, binding.laoPropertiesServerText, "Server Address")
-    clipboardManager.setupCopyButton(binding.copyIdentifierButton, binding.laoPropertiesIdentifierText, "LAO ID")
+    clipboardManager.setupCopyButton(
+        binding.copyServerButton, binding.laoPropertiesServerText, "Server Address")
+    clipboardManager.setupCopyButton(
+        binding.copyIdentifierButton, binding.laoPropertiesIdentifierText, "LAO ID")
 
     return binding.root
   }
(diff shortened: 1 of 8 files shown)


fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/InviteFragment.kt
```diff
@@ -59,6 +63,8 @@ class InviteFragment : Fragment() {
     }
 
     handleBackNav()
+    setupCopyServerButton(binding)
+    setupCopyIdentifierButton(binding)
```
matteosz, line 67: I'd save the binding as a class attribute to avoid passing it around (it's just a preference but also how it's usually done in the codebase)

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/InviteFragment.kt
```diff
@@ -73,6 +79,35 @@ class InviteFragment : Fragment() {
...
+      val text = identifierTextView.text.toString()
+      copyTextToClipboard(text)
+      Toast.makeText(requireContext(), R.string.successful_copy, Toast.LENGTH_SHORT).show()
+    }
+  }
+
+  private fun copyTextToClipboard(token: String) {
+    val clipboard =
+        requireActivity().getSystemService(Context.CLIPBOARD_SERVICE) as ClipboardManager
+    val clip = ClipData.newPlainText(token, token)
```
matteosz, line 107: Here reading the doc the first param should be the text label that you're copying. So instead of reusing the token as its own label, I'd put an actual label that explains what has been copied (not its content)

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/InviteFragment.kt
```diff
@@ -73,6 +79,35 @@ class InviteFragment : Fragment() {
...
+  private fun setupCopyIdentifierButton(binding: InviteFragmentBinding) {
+    val identifierTextView = binding.laoPropertiesIdentifierText
+    val copyButton = binding.copyIdentifierButton
+
+    copyButton.setOnClickListener {
+      val text = identifierTextView.text.toString()
+      copyTextToClipboard(text)
+      Toast.makeText(requireContext(), R.string.successful_copy, Toast.LENGTH_SHORT).show()
+    }
+  }
```
matteosz, line 102: Refactoring opportunity: these 2 methods are almost the same, so you can extract a single method, which is more generic, and parametrise its usage to reduce code duplication

matteosz (changes requested):
Good job! Just few comments.

General suggestion: whenever you're modifying the UI (either layout or button behaviour) it would be nice to add to the PR description screenshots or screen recordings. That would help who's reviewing the PR as they wouldn't need to pull the code and run it to see the changes

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/InviteFragment.kt
Kaz-ookid: yes that's what I did first, but then I saw line 100 in [TokenFragment.kt](https://github.com/dedis/popstellar/blob/master/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/token/TokenFragment.kt) and went to do the same. Maybe I just should have trusted my critical sense :woman_shrugging: 

code was : 
```
  private fun copyTextToClipboard(token: String) {
    val clipboard =
        requireActivity().getSystemService(Context.CLIPBOARD_SERVICE) as ClipboardManager
    val clip = ClipData.newPlainText(token, token)
    clipboard.setPrimaryClip(clip)
  }
```

matteosz: Yes in an utils file would be perfect

matteosz (approved):
LGTM! Really good job

PR #1789, opened 2024-03-21
new message for gossip protocol

DanielTavaresA:
We want to add a new "rumor" message in order to have a message that represents an information that we want to spread across the network. A "rumor" message will represent this. 

It should contain :
- A unique rumor_ID so that a receiver can quickly check if it already received this rumor
- At least one key-value of : channel -> array of messages

This would allow to have a type of message to push new messages to other servers

PR #1799, opened 2024-04-06
jsonRPC handle of rumor msg in scala backend

DanielTavaresA:
Based on PR #1789 , handles and tests the new rumor jsonRPC message.

Can read an incoming rumor jsonRPC, convert it to a Rumor object containing its respective objects : - serverId : PublicKey, rumorId : Int, messages : Map[Channel, List[Messages]]. Can also do the reverse operation and write to a jsonRPC Rumor


be2-scala/src/main/scala/ch/epfl/pop/json/HighLevelProtocol.scala
```diff
@@ -198,6 +199,36 @@ object HighLevelProtocol extends DefaultJsonProtocol {
...
+  implicit object RumorFormat extends RootJsonFormat[Rumor] {
+    final private val PARAM_SENDER_PK: String = "sender_id"
+    final private val PARAM_RUMOR_ID: String = "rumor_id"
+    final private val PARAM_MESSAGES: String = "messages"
+
+    override def read(json: JsValue): Rumor = {
+      val jsonObject: JsObject = json.asJsObject
+      jsonObject.getFields(PARAM_SENDER_PK, PARAM_RUMOR_ID, PARAM_MESSAGES) match {
+        case Seq(senderPk @ JsString(_), rumorId @ JsNumber(_), rumors @ JsObject(_)) =>
+          val map = mutable.HashMap[Channel, Array[Message]]()
```
t1b00, line 211: I think you can reduce the complexity of this, especially the use of a mutable HashMap
Try:
val map: Map[Channel, Array[Message]] = rumors.asJsObject.fields.map {
  case (k: String, JsArray(v)) => Channel(k) -> v.map(_.convertTo[Message]).toArray
  case _                       => throw new IllegalArgumentException(s"Unrecognizable rumor in $json")
}

Tested it and it passes the tests

t1b00 (changes requested):
Just one detail on the use of a mutable HashMap which could be simplified (see comment), otherwise LGTM

be2-scala/src/main/scala/ch/epfl/pop/json/HighLevelProtocol.scala
```diff
@@ -198,6 +199,36 @@ object HighLevelProtocol extends DefaultJsonProtocol {
...
   }
 
+  implicit object RumorFormat extends RootJsonFormat[Rumor] {
+    final private val PARAM_SENDER_PK: String = "sender_id"
+    final private val PARAM_RUMOR_ID: String = "rumor_id"
+    final private val PARAM_MESSAGES: String = "messages"
+
+    override def read(json: JsValue): Rumor = {
+      val jsonObject: JsObject = json.asJsObject
+      jsonObject.getFields(PARAM_SENDER_PK, PARAM_RUMOR_ID, PARAM_MESSAGES) match {
```
onsriahi14, line 209: maybe here you can do json.asJsObject.getFields directly without the need to create the variable jsonObject

be2-scala/src/main/scala/ch/epfl/pop/model/network/method/Rumor.scala
```diff
@@ -0,0 +1,25 @@
+package ch.epfl.pop.model.network.method
+
+import ch.epfl.pop.json.HighLevelProtocol.RumorFormat
+import ch.epfl.pop.model.network.Parsable
+import ch.epfl.pop.model.network.method.message.Message
+import ch.epfl.pop.model.objects.{Channel, PublicKey}
+import spray.json._
+
+final case class Rumor(senderPk: PublicKey, rumorId: Int, messages: Map[Channel, Array[Message]]) extends Params {
```
K1li4nL, line 9: Is there a good reason to use a mutable array instead of an immutable list here ?  I believe we should stick to Scala preferred type to be able to better exploit `yield`, `foreach` or `match` in the rest of the code

be2-scala/src/test/scala/util/examples/json/rumor/exampleRumor.scala
K1li4nL, line 1: There is a `readJsonFromFile` function in `IOHelper.scala` that could prevent this copy of json ;)

K1li4nL: It would then allow you to test easily the wrong examples as well

K1li4nL (changes requested):
Coverage and a few comments / question before we can move on ;)

K1li4nL (approved):
Good job, looks good to me !

t1b00 (approved):

K1li4nL (approved):

Issue #1800, opened 2024-04-10
Federation Project: Authentication Protocol


1florentin:
## Description
As part of the federation sub-project we want to implement an authentication protocol to "link" two LAOs together. For this purpose we designed the following schematics and messages.

This issue is meant to serve as a central place to reference in subsequent pull requests (implementing the protocol) and for discussion.

### Schematic Illustration
<img align="center" width="800" src="https://github.com/dedis/popstellar/assets/20774003/906531fa-8f40-4b2e-891c-c0dcd4745083" alt="schematic">

### Sequential Illustration
<img align="left" src="https://github.com/dedis/popstellar/assets/20774003/28b3eb37-4d6c-43d7-9d8b-d9c5a6afd8ad"  width="500" alt="schematic">

The two organizers meet in person to exchange the *conn_details* message.

They also decide on a time slot to perform their next roll-calls---these have to happen simultaneously.
<br clear="left"/>
## TODOs (for this issue)
* Add message explainations
* Add explanation regarding trust boundaries between the two connected LAOs.

arnauds5:
With this protocol design, the exchange of conn_details happens only with QR-code, without any new message, is it correct ?
Maybe changing the name from conn_init/expect/etc to something like federation_init, could avoid any confusion with the basic server to server connection ?

The conn_init/conn_expect messages schemas would be something like the following ?

conn_init
```json
{
    "$schema": "http://json-schema.org/draft-07/schema#",
    "$id": "https://raw.githubusercontent.com/dedis/popstellar/master/protocol/query/method/message/data/connInit.json",
    "description": "Sent by an organizer client to its server, initiates a connection to a remote LAO",
    "type": "object",
    "properties": {
        "object": {
            "const": "conn"
        },
        "action": {
            "const": "init"
        },
        "lao_id": {
            "type": "string",
            "contentEncoding": "base64",
            "$comment": "ID of the remote LAO"
        },
        "server_address": {
            "type": "string",
            "$comment": "public address of the remote organizer server"
        },
        "public_key": {
            "type": "string",
            "contentEncoding": "base64",
            "$comment": "public key of the remote organizer"
        },
        "challenge": {
            "type": "object",
            "properties": {
                "value": {
                    "type": "integer",
                    "minimum": 0
                },
                "valid_until": {
                    "type": "integer",
                    "description": "[Timestamp] of the expiration time",
                    "minimum": 0
                }
            },
            "additionalProperties": false,
[...]

PR #1801, opened 2024-04-10
draft authentication protocol federation

arnauds5:
more information here https://github.com/dedis/popstellar/issues/1800

Add the following messages data:
- federation_challenge_request 
- federation_challenge: (value is a 32 bytes array encoded in hexadecimal, valid_until is the unix timestamp of expiration)
- federation_expect
- federation_init (may need to modify the challenge field to a signed message containing a challenge)

documentation and new schemas need to be added


onsriahi14:
Since we added the necessary messages for the authentication protocol, we can already merge them into work-federation-protocol.

onsriahi14 (approved):

PR #1804, opened 2024-04-12
Add challenge_request and challenge message JSON schemas

1florentin:
I added the *challenge_request* message, I talked about here https://github.com/dedis/popstellar/issues/1800#issuecomment-2051471188, also I added the *challenge* object as a standalone message---this makes it easier to transfer it between the organizers.

Documentation is still missing for the new one however (will be added shortly).


arnauds5 (approved):
looks good to me
Since we are not yet merging into master directly, the documentation could be added in a future PR (with the others messages and new schemas)

PR #1805, opened 2024-04-12
Match directory name with package name

onsriahi14:
Minor change in the directory name to avoid compilation errors in certain environments.


K1li4nL (approved):
LGTM

PR #1807, opened 2024-04-13
authentication protocol enhancement

onsriahi14:


onsriahi14:
Documentation or challenge and challenge_request added, support for missing messages added but messages still not defined.

arnauds5:
>and the federation_init currently only contains the challenge, not a signed challenge (it's the federation_init itself who is signed with the current schema)
One possibility for the challenge from S1 to S2, we could just forward the federation_init itself, not just the challenge, so that S2 can check that it is signed by Alice
Another possibility is to change the challenge field of federation_init to a method/message containing a challenge

PR #1811, opened 2024-04-16
Work protocol fix rumor message type

DanielTavaresA:
Small improvement to the rumor jsonRPC Schema. Before we used a Array[JsObject] to emulate a Map. But this behavior can be directly done having only a JsObject therefore we make the change.


sgueissa (approved):
Lgtm and my bad to not have seen this error at the last review

K1li4nL (approved):
Hopefully this time we got it right, lgtm !

PR #1813, opened 2024-04-17
Move federation messages from messageData level to JSON RPC level

1florentin:
As discussed in our last meeting, there are still some negative tests missing.


1florentin:
After a thorough reconsideration (including analyzing all implications and security/trust assumptions), we concluded that the protocol can be implemented on the messageData level, therefore we will continue on the messageData level.

PR #1820, opened 2024-04-20
Add karate UI tests rollcall create and join

4xiom5:
- Added tests for rollcall
- Updated MockClient


simone-kalbermatter (dismissed):
LGTM!
The CI shows that the coverage is not enough, you can try to add another test case.

simone-kalbermatter (approved):
LGTM!

PR #1822, opened 2024-04-21
Added paged catchup message to the protocol

osm-alt:
In an effort to reduce network traffic at catchup, a paged catchup message was added to the protocol with implementation planned to take place afterwards. For now, it is to be used to retrieve chirps on the social media channel by paging when a new client joins the LAO instead of getting all the chirps at once.


docs/protocol.md
```diff
@@ -1062,6 +1063,118 @@ RPC
...
+// ../protocol/examples/query/paged_catchup/paged_catchup.json
+
+{
+    "jsonrpc": "2.0",
+    "id": 3,
+    "method": "paged_catchup",
+    "params": {
+        "channel": "/root/p_EYbHyMv6sopI5QhEXBf40MO_eNoq7V_LygBd4c9RA=/social/chirps",
+        "number_of_messages": 10,
+        "everything_before_message_id": "DCBX48EuNO6q-Sr42ONqsj7opKiNeXyRzrjqTbZ_aMI="
```
Tyratox, line 1089: Are you sure about the naming here? `everything_before_message_id` seems a bit unclear to me? Maybe `before_message_id` or `cursor`?

docs/protocol.md
```diff
@@ -1062,6 +1063,118 @@ RPC
...
+
+RPC 
+
+```json5
+// ../protocol/examples/query/paged_catchup/paged_catchup.json
+
+{
+    "jsonrpc": "2.0",
+    "id": 3,
+    "method": "paged_catchup",
```
Tyratox, line 1085: Should it be a general catchup message? If so, how do you order non-chirp messages? Do you plan to add a timestamp to all messages?

docs/protocol.md
```diff
@@ -1062,6 +1063,118 @@ RPC
...
+### Catching up on past messages on a channel using paging
+
+🧭 **RPC Message** > **Query** > **Paged Catchup**
+
+By executing a paged catchup action, a client can ask the server to receive a specified number of
+past messages on a specific channel.
+
+For now, it is to be used to retrieve chirps on the social media channel by paging when a new client joins the LAO instead of getting all chirps at once. This is done in an effort to reduce network traffic at catchup.
+
+This may serve as a starting point for the paging of messages in other channels as a future optimization.
```
Tyratox, line 1075: I would try to clarify this further:
What does past messages mean here? (i.e. what is the order)
Why was the message designed this way? (basically what we discussed, this gives some background information)

What is "the social media channel"? I would try to be more precise here.
Also the docs say it is only used for chirps, what should happen if this is sent on some other channel? Is it a no-op for now? Or should the server try to find chirps on any channel and return them?
And which chirp message(-s) should this return? (I understand but the documentation should be precise in this regard)

Tyratox (changes requested):
Good starting point :)

Though I would try to clarify the documentation, I doubt that right now somebody unfamiliar with the application would understand.

Also don't forget about testing the schema ;) (https://github.com/dedis/popstellar/blob/master/protocol/test/main.test.js)

docs/protocol.md
osm-alt: before_message_id seems good to me

docs/protocol.md
osm-alt: The focus for this semester is on chirps, but if we can have one message that can then be used by future generations to extend paging to other types of messages, then that would be great.

docs/protocol.md
osm-alt: Noted
As for the "Also the docs say it is only used for chirps, what should happen if this is sent on some other channel? Is it a no-op for now? Or should the server try to find chirps on any channel and return them?" question, we decided to not implement paging for other channels this semester.

Tyratox: I assumed so but the message may be sent by (malicious) actors anyway. The docs should describe how to process the message even if it is sent in an unexpected way.

protocol/examples/query/paged_catchup/wrong_paged_catchup_missing_everything_after_message_id.json
Tyratox, line 1: Name of the file still has the old parameter name in it ;)

protocol/query/method/paged_catchup.json
```diff
@@ -0,0 +1,47 @@
...
+                    "description": "[String] name of the channel",
+                    "type": "string",
+                    "pattern": "^/root(/[^/]+)*$"
+                },
+                "number_of_messages": {
+                    "description": "[Integer] Number of messages requested",
+                    "type": "integer"
+                },
+                "before_message_id": {
+                    "description": "[Base64String] final message id on the client : HashLen(data, signature)",
```
Tyratox, line 27: Final? I don't think this is very clear but maybe that is just me?

protocol/examples/query/paged_catchup/wrong_paged_catchup_missing_everything_after_message_id.json
osm-alt: Oh yes. Thanks for pointing that out

protocol/query/method/paged_catchup.json
osm-alt: You have a point. I can try to make it clearer

docs/protocol.md
pierluca: I think @Tyratox here makes a very important point.

> we decided to not implement paging for other channels this semester.

That's perfectly okay, but what is the expected behavior when it *does* get called on a non-chirp channel ? Not leaving this kind of ambiguity in the specification is good for two reasons.

1. Unexpected stuff happens. Even if there are no malicious actors, a front-end could have bugs and make the wrong RPC call. What happens *then* ?
2. Without a clear specification, I can guarantee that different backends will spontaneously implement this (slightly) differently. It doesn't cost much to clarify and say "we return an error" or "we default to the normal catchup behavior". 
3. Conversely, if there's a bug and the points above are not addressed, you'll have to debug behaviors like "in some cases, with fe2 and be1 [for instance], the loading of chirps fails with some weird error". 

As such:
- if you pick to revert to a reasonable "default" behavior, such as behaving more or less like a catchup (without paging), document that !
- if you pick to return an error, define the error (what code ? what message ?), try and make it super unambiguous and consistent ! 

Best of luck :-)

docs/protocol.md
```diff
@@ -1062,6 +1063,118 @@ RPC
...
+Response (in case of success)
+
+
+```json5
+// ../protocol/examples/answer/general_message.json
+
+{
+    "jsonrpc": "2.0",
+    "id": 3,
+    "result": [
```
pierluca, line 1104: would it make sense to return an object, containing this list but also the next page that should be retrieved ? I believe I've encouraged you to do so before, but you might want to look into pagination API designs. 
E.g. 
- https://nordicapis.com/everything-you-need-to-know-about-api-pagination/
- https://www.getknit.dev/blog/api-pagination-best-practices
- https://dev.to/pragativerma18/unlocking-the-power-of-api-pagination-best-practices-and-strategies-4b49
- etc.

I don't see in the current design a fully thought-through pagination system.
It looks as though you're doing a variant of keyset pagination, but you might want to think about all the use cases here. You can start with a simple one: assume you were up-to-date until yesterday, stayed offline since, and come back online now. How do you retrieve the latest messages with the current protocol ?

docs/protocol.md
```diff
@@ -1062,6 +1063,118 @@ RPC
...
+                },
+                "before_message_id": {
+                    "description": "[Base64String] id of a message where the client requests messages that precede the specified message : HashLen(data, signature)",
+                    "type": "string",
+                    "contentEncoding": "base64",
+                    "$comment": "Note: the string is encoded in Base64"
+                }
+            },
+
+            "required": ["channel", "number_of_messages", "before_message_id"]
```
pierluca, line 1159: `before_message_id` here is marked as required. How do I know which `before_message_id` to provide if I want the latest messages, messages I've never heard of before ?

pierluca (changes requested):

docs/protocol.md
osm-alt: Noted
Thanks for the feedback

docs/protocol.md
osm-alt: I can remove it as a required parameter so that when it is not provided, it just return the latest messages on the server

docs/protocol.md
```diff
@@ -1062,6 +1063,118 @@ RPC
...
+{
+    "jsonrpc": "2.0",
+    "id": 3,
+    "result": [
+        {
+            "data": "eyJvYmplY3QiOiJsYW8iLCJhY3Rpb24iOiJjcmVhdGUiLCJuYW1lIjoiTEFPIiwiY3JlYXRpb24iOjE2MzMwMzU3MjEsIm9yZ2FuaXplciI6Iko5ZkJ6SlY3MEprNWMtaTMyNzdVcTRDbWVMNHQ1M1dEZlVnaGFLMEhwZU09Iiwid2l0bmVzc2VzIjpbXSwiaWQiOiJwX0VZYkh5TXY2c29wSTVRaEVYQmY0ME1PX2VOb3E3Vl9MeWdCZDRjOVJBPSJ9",
+            "sender": "J9fBzJV70Jk5c-i3277Uq4CmeL4t53WDfUghaK0HpeM=",
+            "signature": "ONylxgHA9cbsB_lwdfbn3iyzRd4aTpJhBMnvEKhmJF_niE_pUHdmjxDXjEwFyvo5WiH1NZXWyXG27SYEpkasCA==",
+            "message_id": "2mAAevx61TZJi4groVGqqkeLEQq0e-qM6PGmTWuShyY=",
+            "witness_signatures": []
```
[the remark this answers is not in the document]
osm-alt: I suppose having them sorted from first-to-last (oldest to newest before the message with the provided message id) is a viable option

docs/protocol.md
osm-alt: "would it make sense to return an object, containing this list but also the next page that should be retrieved ?": 
If I want to build upon the current design, it would involve introducing page-based pagination on top of the keyset pagination variant currently used. One way I can think of to add the next page is by taking in the `number_of_messages` parameter and using that as the size of each page, but it might get messy if the user changes the value of this parameter. An alternative is to just remove this parameter and have fixed-size pages, which makes implementing paged-based pagination simpler but removes the flexibility that comes with that parameter.

"How do you retrieve the latest messages with the current protocol ?":
The client can request the latest messages one page at a time until they get a page containing a message they already have.

Tyratox: > The client can request the latest messages one page at a time until they get a page containing a message they already have.
I think what @pierluca meant is that you should document how to retrieve the most recent set of messages using this protocol ;)

docs/protocol.md
```diff
@@ -1062,6 +1063,132 @@ RPC
...
+                    "description": "[String] name of the channel",
+                    "type": "string",
+                    "pattern": "^/root(/[^/]+)*$"
+                },
+                "number_of_messages": {
+                    "description": "[Integer] Number of messages requested",
+                    "type": "integer"
+                },
+                "before_message_id": {
+                    "description": "[Base64String] id of a message where the client requests messages that precede the specified message : HashLen(data, signature)",
```
Tyratox, line 1166: Maybe quickly mentioned that it is optional?

docs/protocol.md
```diff
@@ -1062,6 +1063,132 @@ RPC
...
+done in an effort to reduce network traffic at catchup.
+This message is also to be used to retrieve top chirps from a subchannel "/root/lao_id/social/top_chirps" and chirps
+of a specific user profile from a subchannel "/root/lao_id/social/profile/user_public_key/specific_user_public_key" 
+where "user_public_key" is the same as before and "specific_user_public_key" is the public key of the user whose messages
+the client wants to retrieve.
+
+This may serve as a starting point for the paging of messages in other channels as a future optimization.
+
+For now, if this message gets accidentally sent to a non-chirp channel, we return an error with a -1 code indicating that
+it is an invalid action and a description saying that paging is not supported on non-chirp channels.
```
Tyratox, line 1089: Great addition of detail! As we have already discussed this will be far from trivial for the top chirps. How will you go about it? (i.e. what is the ordering and how does paging work there?)

docs/protocol.md
```diff
@@ -1062,6 +1063,132 @@ RPC
...
 
+### Catching up on past messages on a channel using paging
+
+🧭 **RPC Message** > **Query** > **Paged Catchup**
+
+By executing a paged catchup action, a client can ask the server to receive a specified number of
+past messages sorted by timestamps from oldest to newest on a specific channel. When a message ID 
+is provided, the server returns the requested number of messages on the specified channel that precede
+this message. Otherwise, the server returns the latest messages it has on that channel taking into 
+account the requested number of messages. If the specified number of messages is greater than what
```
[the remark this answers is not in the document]
osm-alt: I mention earlier in line 1071 that the past messages "sorted by timestamps from oldest to newest"

docs/protocol.md
osm-alt: Changed the doc to mention that paging is not deemed necessary for top chirps for now

docs/protocol.md
```diff
@@ -1062,6 +1063,130 @@ RPC
...
+account the requested number of messages. If the specified number of messages is greater than what 
+is on the server in the current page, then the server only returns the messages it has on that page.
+
+For now, this message is to be used to retrieve chirps on the social media channel (/root/lao_id/social/user_public_key) 
+by paging when a new client joins the LAO instead of getting all the chirps at once. The user's public key is used
+to have a separate paging subchannel for each user to avoid sending irrelevant messages to other users. This paging is 
+done in an effort to reduce network traffic at catchup.
+
+This message is also to be used to retrieve chirps of a specific user profile from a subchannel "/root/lao_id/social/profile/user_public_key/specific_user_public_key"  where "user_public_key" is the same as before and "specific_user_public_key" is the public key of the user whose messages the client wants to retrieve. Paging is not deemed necessary for retrieving top chirps for now and can be done with the regular catchup message from a subchannel "/root/lao_id/social/top_chirps".
```
Kaz-ookid, line 1082: I agree that paging for top chirps is not necessary, because for now top chirps are only the 3 first. But we can imagine in the future to be able to scroll down in the same paging way as in the feed, but in top chirps.

What is the top chirps channel supposed to do? Should it be specified? IMO it should send the top 3 on catchup, then send the new top 3 whenever an update is received server side (then front ends should listen to this channel).

Also just a quick note for front ends implementation: I think it would be easier for now to have separate caches for feed/top chirps/profile. Later on, some optimizations could be done (front end wise) to link the caches and avoid duplicate data (chirps) throughout those caches, but it should do without for now as there shouldn't be so much data to handle. Also, I think that we should not cache chirps of profiles (other than ours) since we will very rarely click on those profile (and even less come back to it right after).

Kaz-ookid (dismissed):
Looks overall good to me! It sure is a tricky PR to merge, since it represents a big change that a lot of future work will depend on. Although, I might miss some things, but I still think that the current state looks very usable and accurate to what we want to have.  

Very keen to these changes! These will finally relieve front ends from having do compute everything, everytime, on their ow (and needing to query the WHOLE data to do so). I am talking about the fact that the chirps sorting, top chirps calculations and else will now be computed server side :+1: . This is sure a great step towards inclusion of low quality devices/bad connection areas of the world!

docs/protocol.md
osm-alt: "IMO it should send the top 3 on catchup, then send the new top 3 whenever an update is received server side (then front ends should listen to this channel)." -> Yes that's the plan
"have separate caches for feed/top chirps/profile" -> Makes sense
"we should not cache chirps of profiles (other than ours) since we will very rarely click on those profile" -> I agree

docs/protocol.md
```diff
@@ -1062,6 +1063,130 @@ RPC
...
+
+This may serve as a starting point for the paging of messages in other channels as a future optimization.
+
+For now, if this message gets accidentally sent to a non-chirp channel, we return an error with a -1 code indicating that
+it is an invalid action and a description saying that paging is not supported on non-chirp channels.
```
Tyratox, line 1087:
```suggestion
Paged catchup messages must not be sent on any other channel than the aforementioned chirp channels (global chirp timeline and profile chirp timelines).
If a paged catchup message is sent on another channel, the backend returns an error with a `-1` code indicating that this is an invalid action and a description saying that paging is not supported on this channel.
If at any point a frontend receives a paged catchup message, it must treat it as a no-op and is ignore it.
```

Non-Chirp channels is inaccurate if you plan to use the regular catchup message for the top chirps :)

protocol/query/method/paged_catchup.json
```diff
@@ -0,0 +1,47 @@
...
+        },
+
+        "params": {
+            "type": "object",
+            "additionalProperties": false,
+            "properties": {
+                "channel": {
+                    "description": "[String] name of the channel",
+                    "type": "string",
+                    "pattern": "^/root(/[^/]+)*$"
```
Tyratox, line 20: Since the set of channels this message can be sent on is now clearly defined, it may make sense to actually enforce it here?

Tyratox (approved):
Good work, the documentation and examples are improving a lot! :)

I added final set of comments. However I don't want to be blocking you from merging this indefinitely and therefore approve it.

However, I would still recommend to go through the comments and at least document the different behavior of the regular catchup message.

Kaz-ookid:
I think addressing the comments if @Tyratox would really set this PR from rushed to actually clean and maintainable, juste one last step @osm-alt ! :1st_place_medal:

protocol/query/method/paged_catchup.json
osm-alt: Good idea

Tyratox (dismissed):
Great work, approved! 🥳

Tyratox (approved):

pierluca (approved):
Approving blindly

PR #1823, opened 2024-04-21
[BE2] Fix multiple connectTo messages upon changing the server-peers-list.conf file.

onsriahi14:


t1b00:
On this fix (assumes knowledge of the code):
Problem: test started suddenly failing on some machines because an actor expected only one message yet received more (usually three).

Upon investigation, the following problems were found:
1. Synchronization issues: Monitor starts a FileMonitor in a Thread that sends messages. This thread is not properly stopped. 
2. Improper checking of modifications in the config file: the WatchService may detect too many modifications, even if there was only one.
3. Incorrectly written test: the test writes non-valid inputs to the config file, which shouldn't trigger an update
4. Impromper sending of messages

The following fixes were implemented:
1. The thread created by FileMonitor essentially consisted in a loop that would check if it was interrupted, however, not call to Thread.interrupt() was made. Even with a call to Thread.interrupt(), this logic fails for two reasons: Thread.interrupt() simply tells the thread to interrupt itself, it doesn't immediately stop the instruction, secondly the call to watchService.take() is blocking and a thread could thus stay in take(), then be unblocked at some point and execute its message sending logic before potentially terminating at the end of the loop, seeing that it was terminated.
A fix to this problem was to replace the interrupted thread logic by a status Atomic boolean (running), which enables graceful termination of the thread and prevents sending messages in the case where the thread was blocked in watchService.take().
2. As explained, the watch service can detect multiple modifications in the file, even though there was only one. This could be due to the file being opened, modified and then closed being detected as three modifications. A simple solution to this is to save the previous state of the file and compare it each time we have a modification, ensuring that there was an actual modification.
3. The test in its previous form would write strings to the file which would not match the regex. This wouldn't be a problem before since the modification detected that made the test pass was not the correct one. With the fixes above, this proved to be a problem since no modification (and rightfully so) would be detected. A simple solution was to replace with strings matching the regex. 
4. Improper sending of messages: this is technically not a bug, but simply bad practice. The FileMonitor thread would send messages to the actor passed to its constructor. This would pose a problem as the FileMonitor is not an actual actor, and the messages received would thus appear to come from the DeadLetters actor, instead of coming from the Monitor actor. I modified the code so that the FileMonitor would send a message to the Monitor and the Monitor would then forward it. This doesn't completely solve the problem as the FileMonitor is still sending messages despite not being an actor, but it limits it to the Monitor scope and doesn't expose it to the rest of the code.

be2-scala/src/main/scala/ch/epfl/pop/decentralized/Monitor.scala
```diff
@@ -82,9 +85,19 @@ final case class Monitor(
...
+      new Thread(fileMonitor).start()
 
-    case _ => /* DO NOTHING */
+    case msg: ConnectionMediator.ConnectTo =>
+      connectionMediatorRef ! msg
+  }
+
+  override def postStop(): Unit = {
+    if (fileMonitor != null) {
+      if (!fileMonitor.running.compareAndSet(true, false)) {
```
DanielTavaresA, line 97: Both boolean evaluation could be merged into one, using "&&" operator which short circuits the evaluation if first result is false. Meaning that if fileMonitor is null, fileMonitor.running.compareAndSet will not be evaluated.

be2-scala/src/test/scala/ch/epfl/pop/decentralized/MonitorSuite.scala
```diff
@@ -185,23 +181,21 @@ class MonitorSuite extends TestKit(ActorSystem("MonitorSuiteActorSystem")) with
     mockConnectionMediator.send(monitorRef, ConnectionMediator.Ping())
 
     // Expect first read of the server peers list
-    mockConnectionMediator.expectMsgType[ConnectionMediator.ConnectTo](timeout)
+    mockConnectionMediator.expectMsgType[ConnectionMediator.ConnectTo](20.seconds)
```
DanielTavaresA, line 184: Might want to create a new constant value or use an existing one, so that we can have consistency on the desired timeout for this case on future tests.

be2-scala/src/main/scala/ch/epfl/pop/decentralized/Monitor.scala
```diff
@@ -82,9 +85,19 @@ final case class Monitor(
     case ConnectionMediator.Ping() =>
       log.info("Received ConnectionMediator ping")
       connectionMediatorRef = sender()
-      new Thread(new FileMonitor(connectionMediatorRef)).start()
+      fileMonitor = new FileMonitor(this.self)
+      new Thread(fileMonitor).start()
 
-    case _ => /* DO NOTHING */
+    case msg: ConnectionMediator.ConnectTo =>
+      connectionMediatorRef ! msg
```
DanielTavaresA, line 92: This might deserve a comment to explain the logic behind it has it isn't trivial to understand why it is used by just looking at the code.

DanielTavaresA (changes requested):
Some suggestions to be addressed but other than that LGTM. Works on my machine running linux. Very good documentation of the issue.

be2-scala/src/main/scala/ch/epfl/pop/decentralized/Monitor.scala
t1b00: No, because if we merge it will log an error message when fileMonitor is null, which can happen if no ConnectionMedatior.Ping() is received, which is not wanted

K1li4nL (changes requested):
That wasn't an easy one, your fix looks good to me @t1b00 ! I will let you add the comments that @DanielTavaresA suggested and I think you need to update the comments at line 127-128, and the typo. After that it should be good to merge

K1li4nL (approved):
That changes was needed for a while now, good job !

DanielTavaresA (approved):
Comments have been addressed. LGTM !

PR #1824, opened 2024-04-22
Add Db handle of Rumors

DanielTavaresA:
Provides a way of read and write Rumors in the Db. Rumors are not kept in a Channel because their storage is only meant for local purposes. Instead we create two new prefixes : "Rumor:" and "RumorData:". 

"Rumor:" is meant for accessing rumors, based on key {serverPk:rumorId}
"RumorData:" is analogous to "ChannelData:". It keeps a correspondence of received rumorIds per serverPk. This way we can quickly have an idea of the state of the rumors. 

Provided interface is WriteRumor(rumor: Rumor), ReadRumors(desiredRumors: Map[String, List[Int]]) and ReadRumorData(senderPk: String).

For now Rumors store messages in full, which will create storage duplication. This might need to be addressed in the future if this solution is not optimal. Edge case would be : "Messages can be modified (for example with witnessing). Do we want rumors to keep messages as there were received, or as a pointer (=messageId) to the message.". Also as a tradeoff : "Is it better to have duplicates, but only needs one query to Db, or no duplicates but rumor needs to be created on the fly, fetching each message by their ID".


K1li4nL (changes requested):
First half of the review, I have not looked at the tests yet, overall its good. Some type choice and a question with the db read logic

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ParamsHandler.scala
```diff
@@ -85,4 +86,19 @@ object ParamsHandler extends AskPatternConstants {
...
+      jsonRpcMessage.method match {
+        case MethodType.rumor =>
+          val rumor: Rumor = jsonRpcMessage.getParams.asInstanceOf[Rumor]
+          val senderPk: PublicKey = rumor.senderPk
+          val rumorId: Int = rumor.rumorId
+          val messages: Map[Channel, List[Message]] = rumor.messages
+          Right(jsonRpcMessage)
+        case _ => Left(PipelineError(ErrorCodes.SERVER_ERROR.id, "RumorHandler received a non expected jsonRpcRequest", jsonRpcMessage.id))
+      }
+    case graphMessage @ _ => Left(PipelineError(ErrorCodes.SERVER_ERROR.id, "RumorHandler received an unexpected message:" + graphMessage, None))
```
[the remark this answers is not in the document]
DanielTavaresA: Yes, it was necessary for the buildGraph to work. But it will be done in another PR

be2-scala/src/main/scala/ch/epfl/pop/storage/DbActor.scala
```diff
@@ -384,6 +386,50 @@ final case class DbActor(
...
+
+  @throws[DbActorNAckException]
+  private def readRumors(desiredRumors: Map[String, List[Int]]): Map[String, Map[Int, Rumor]] = {
+    desiredRumors.flatMap { case (senderPk, rumorIds) =>
+      val rumorsForSender: Map[Int, Rumor] = rumorIds.flatMap { rumorId =>
+        val rumorKey = generateRumorKey(senderPk, rumorId)
+        storage.read(rumorKey).map(json => rumorId -> Rumor.buildFromJson(json))
+      }.toMap
+      if (rumorsForSender.isEmpty)
+        throw DbActorNAckException(ErrorCodes.SERVER_ERROR.id, s"No rumors found for sender $senderPk")
```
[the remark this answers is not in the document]
DanielTavaresA: It's true that it doesn't seem the best behaviour. I changed it (and also in readRumorData) so that if a queried rumor is missing, it is handled just by returning nothing. Which seem a more desired behaviour

be2-scala/src/main/scala/ch/epfl/pop/storage/DbActor.scala
```diff
@@ -384,6 +386,50 @@ final case class DbActor(
...
+  private def readRumorData(senderPk: PublicKey): RumorData = {
+    Try(storage.read(generateRumorDataKey(senderPk))) match {
+      case Success(Some(json)) => RumorData.buildFromJson(json)
+      case Success(None)       => RumorData(List.empty)
+      case Failure(ex)         => throw ex
+    }
+  }
+
+  @throws[DbActorNAckException]
+  private def readRumors(desiredRumors: Map[PublicKey, List[Int]]): Map[PublicKey, Map[Int, Rumor]] = {
```
K1li4nL, line 407: I am wondering, isn't the rumor id already in `Rumor` ? Do you need a map of map ? (if it makes your life easier down the line then sure, keep it that way) 
```suggestion
  private def readRumors(desiredRumors: Map[PublicKey, List[Int]]): Map[PublicKey, List[Rumor]] = {
```

K1li4nL (commented):
All good ! Just one question before we can merge:

be2-scala/src/main/scala/ch/epfl/pop/storage/DbActor.scala
DanielTavaresA: My idea was that with this type we have an easy access of which rumors_ids have been recovered from memory, without the need to parse each entry to know the corresponding rumor_id. This is oriented towards what I expect readRumors to be used for. I expect that when we query for the rumors, having their rumor_id easily accessible would be useful. Do you think it is better to have a more "intuitive" readRumors, that returns a list of rumors, and then the additionnal work would be done by the receivers ? Or to already have this type that already makes a small step of what we expect would be done after ?

K1li4nL: Following "You aren't gonna need it", I would say keep it as simple as possible unless you already know it will be useful.

If you think you will often need to check for the the presence of an ID then a Map could make sense, it will avoid the need of building the ID set at the cost of some small amount of memory and a more complex function signature.

I am fine with either, I won't block the pr for that and it can always be changed later anyway ;)

DanielTavaresA: I will follow your advice and go for a more intuitive function signature.

K1li4nL (approved):

PR #1827, opened 2024-04-24
add be1-go implemention of federation authentication part

arnauds5:
Add handler for federation message that do the following (not including verification of sender/etc):
- handleRequestChallenge: generate a new challenge, store it in db, broadcast it
- handleExpect: verify that the challenge exist in db, then store the expect
- handleInit: store the init, create a connection to the other server, then send the challenge to it
- handleChallenge: verify that a matching expect exist in db, remove the challenge from db to avoid reuse, create a result message, store it in db, send the result to the other server, broadcast the result locally
- handleResult: verify that a matching init exist in db, store the result, broadcast the result locally

The channel "/root/<loa_id>/federation" is used for all the current messages.

<br><br>
The following description was from before the go refactor for the code with last commit: [c6b307ff6ee4b4822514ac890e6ca2a2698e091d](https://github.com/dedis/popstellar/tree/c6b307ff6ee4b4822514ac890e6ca2a2698e091d)

<strike>
Implements some message handlers of the federation authentication (FederationResult is being done by Florentin)

- processChallengeRequest: generate a new challenge and send it back to the organizer
- processFederationExpect: Store informations about the other organization, with the expected challenge.
- processFederationInit: Store informations about the other organization, initiate a ws connection to the other server, then publish the FederationChallenge on the other server's federation channel
- processFederationChallenge: check that the challenge is valid, send federation result

A channel "/root/<lao_id>/federation" is created when a lao is created.

Informations about the other organization are stored in a map, such that it could be possible to handle multiple organizations. The public key of the organizer of the other lao is used as the key.

The challenge field of the json schema of FederationInit was changed from only a challenge to a general message containing a FederationChallenge. So that it's possible to send a FederationChallenge signed by the organizer (not the server)
</strike>


sgueissa (approved):
LGTM I have nothing to say you have perfectly fit your feature in the new backend with the same convention as we did. GGWP

emonnin-epfl (approved):
LGTM!

MariemBaccari (approved):
Approved blindly because of time constraints.

PR #1831, opened 2024-04-29
Fixes a test that randomly fails in be1-go

arnauds5:
`Test_Authenticate_User` was failling randomly some time, without any code change.
In this test, a websocket server is started, a popcha_authenticate message is published, and a message is expected to be received by the ws server.
The problem was that the message was not always received soon enough.
This PR use channels instead of expecting the messages to be received after some time delay.

### DIFF
--- a/be1-go/channel/authentication/authentication_test.go
+++ b/be1-go/channel/authentication/authentication_test.go
@@ -158,6 +158,9 @@ func Test_Authenticate_User(t *testing.T) {
 	err = authCha.Publish(msg, socket.ClientSocket{})
 	require.NoError(t, err)
 
+	// The websocket server may take some time to receive and process a message
+	time.Sleep(time.Millisecond * 100)
+
 	require.True(t, rb.get())
 }
 
--- a/be1-go/channel/authentication/authentication_test.go
+++ b/be1-go/channel/authentication/authentication_test.go
@@ -108,25 +108,17 @@ func Test_Authenticate_User(t *testing.T) {
 	_, found := fakeHub.channelByID[name]
 	require.True(t, found)
 
-	time.Sleep(time.Millisecond)
-
 	// Create the message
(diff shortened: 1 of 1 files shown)


1florentin (approved):
lgtm

be1-go/channel/authentication/authentication_test.go
```diff
@@ -158,6 +158,9 @@ func Test_Authenticate_User(t *testing.T) {
 	err = authCha.Publish(msg, socket.ClientSocket{})
 	require.NoError(t, err)
 
+	// The websocket server may take some time to receive and process a message
+	time.Sleep(time.Millisecond * 100)
```
MariemBaccari, line 162: Thanks for taking the initiative to fix this test ! Unfortunately, adding arbitrary delays in tests is not only unreliable (e.g. false negatives) but can also slow down the pipeline if it is done often (which is the case for our system).
You could fix this by using channels and checking the boolean once you know the value has been received (you can also add a timeout if you're not sure the check can be reached).  
I'm aware that using `time.Sleep` is common in our codebase but it's probably something we should refactor at some point. 
Let me know if you need clarifications from my part :)

arnauds5:
> Thanks for taking the initiative to fix this test ! Unfortunately, adding arbitrary delays in tests is not only unreliable (e.g. false negatives) but can also slow down the pipeline if it is done often (which is the case for our system). You could fix this by using channels and checking the boolean once you know the value has been received (you can also add a timeout if you're not sure the check can be reached). I'm aware that using `time.Sleep` is common in our codebase but it's probably something we should refactor at some point. Let me know if you need clarifications from my part :)

I removed all time.sleep and replaced the struct that stored if a message was received, by a buffered channel and a select that check if a message is received or throw an error if nothing was received after 1 seconds.

MariemBaccari (approved):
LGTM now, good job :)

Issue #1828, opened 2024-04-25
[BUG] Android Copy to Clipboard Buttons invisible in dark mode


Kaz-ookid:
### Description (Actual behavior)
The recently added Copy to Clipboard buttons in Invite tab are appearing to be black on black background when in darkmode, making them impossible to see. Should not be hard to fix.

### Expected behavior
They should turn white when phone settings are darkmode.

### How to reproduce
1. Go to a LAO, in Invite section. If your phone is in darkmode, you will not see the buttons even though they are indeed here.
Below is the example in white mode, where you can see the buttons we are talking about :
![image](https://github.com/dedis/popstellar/assets/100324323/ab8033c0-5ef5-45a1-8f22-b3f6842646ba)

### Impact
Minor, since the text is still available to manual selection so there is a way to copy the token. Also, even if not visible, the button can still be clicked.

PR #1833, opened 2024-04-29
fixed copy clipoboard button appearing black on darkmode

Kaz-ookid:
Addressing issue #1828 

newly added button was not turning white when phone is set to darkmode


quadcopterman (approved):
Looks good to me !

matteosz (approved):

PR #1837, opened 2024-05-01
Federation authentication protocol

quadcopterman:
This contains the first version of the authentication of the federation protocol. It will be merged when the implementation in BE1-2 and FE1-2 is done.


pierluca (approved):
As discussed, merging without review.

PR #1838, opened 2024-05-01
Protocol part of the federation authentication on FE2

quadcopterman:
This PR will contain the messageData for the federation authentication. There are also small UI changes to the invite and join pages :
image
image


Kaz-ookid (changes requested):
Looks good, tho I did not have an exhaustive view of all the code. I mainly focused for now on stuff I can talk about, since I worked on it very recently! QR fragments (especially manual inputs) are getting reworked in this #1850.

About the QR part, I think you tests lack testing manual inputs ! Let's meet someday and talk about how I can help you adapt to the upcoming QR fragments changes :+1:

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/qrcode/QrScannerFragment.kt
```diff
@@ -176,6 +184,26 @@ class QrScannerFragment : Fragment() {
 
   private fun onResult(data: String?) {
     scanningViewModel.handleData(data)
+
+    // Special case for federation
```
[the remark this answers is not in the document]
quadcopterman: Not sure if that's what you mean but as this opens a new fragment it has to be done from a fragment. The ViewModel cannot open a new fragment if I understood correctly. It doesn't look nice but I don't see any nicer way to do it, in all other uses of the QR code in the app it never has to open a new fragment in the end (the only kind of similar situation is in LAO join but here the view model just creates a whole new activity, which is not really adapted to this situation).

Kaz-ookid:
Hey @quadcopterman, everything seems to work fine after the merge. I adapted the QR part to fit the changes on master. I have setup manual inputs that may be wrong, just for you to see how it works and adapt it to your need! I also tried to pre-shot the strings resources you will need for that.

While testing it out, I saw that erroneous manual inputs (non-empty) are not handled correctly. They indeed pop a toas with `qr_code_not_federation_details` stringres, but it still lets you get to the next stage of Organizations linking. 

Don't forget to write some tests for the QR Scanning and Manual inputs! :+1:

quadcopterman:
> Hey @quadcopterman, everything seems to work fine after the merge. I adapted the QR part to fit the changes on master. I have setup manual inputs that may be wrong, just for you to see how it works and adapt it to your need! I also tried to pre-shot the strings resources you will need for that.
> 
> Also, the json format I put in `jsonFormatter` attribute of the `ScanningAction`s are most likely wrong, I'll let you fill those with the correct formats!
> 
> While testing it out, I saw that erroneous manual inputs (non-empty) are not handled correctly. They indeed pop a toas with `qr_code_not_federation_details` stringres, but it still lets you get to the next stage of Organizations linking.
> 
> Don't forget to write some tests for the QR Scanning and Manual inputs! 👍

Hey @Kaz-ookid, did you check that the app runs after your changes ? Because I still have the exact same problem I had when I tried to merge the code : the app keeps stopping when you run it. Maybe it's my Android Studio that has a problem... Just let me know if it was working for you

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/qrcode/QrScannerFragment.kt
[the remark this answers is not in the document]
quadcopterman: I resolved the conversation, as we discussed that during last week's meeting

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/Objects.kt
```diff
@@ -41,7 +42,7 @@ enum class Objects
       COIN.`object`,
       MESSAGE.`object` -> true
 
-      // Consensus and Popcha are for now the ones excluded from persistence
+      // Consensus, Popcha and Federation are for now the ones excluded from persistence
       else -> false
```
matteosz, line 46: Did you decide to not persist federation messages for a matter of time and it's gonna happen in a future PR or is there any other reason?

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/federation/FederationExpect.kt
```diff
@@ -0,0 +1,59 @@
...
+   * @param serverAddress public address of the remote organizer server
+   * @param publicKey public key of the remote organizer
+   * @param challenge challenge for the server
+   */
+  constructor(laoId: String, serverAddress: String, publicKey: String, challenge: MessageGeneral) {
+    this.laoId = laoId
+    this.serverAddress = serverAddress
+    this.publicKey = publicKey
+    this.challenge = challenge
+  }
```
matteosz, line 29:
```suggestion
class FederationExpect
/**
 * Constructor for a data Federation Expect
 *
 * @param laoId ID of the remote LAO
 * @param serverAddress public address of the remote organizer server
 * @param publicKey public key of the remote organizer
 * @param challenge challenge for the server
 */
(
    @SerializedName("lao_id") val laoId: String,
    @SerializedName("server_address") val serverAddress: String,
    @SerializedName("public_key") val publicKey: String,
    val challenge: MessageGeneral
) : Data {
```

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/federation/LinkedOrganizationsInviteFragment.kt
```diff
@@ -27,48 +35,93 @@ class LinkedOrganizationsInviteFragment : Fragment() {
...
+      savedInstanceState: Bundle?,
+  ): View {
     val binding = LinkedOrganizationsInviteFragmentBinding.inflate(inflater, container, false)
     laoViewModel = LaoActivity.obtainViewModel(requireActivity())
+    linkedOrganizationsViewModel =
+        obtainLinkedOrganizationsViewModel(requireActivity(), laoViewModel.laoId)
 
-    binding.linkedOrganizationsServerText.text = networkManager.currentUrl
+    if (CREATES_INVITATION) {
+      // When the user creates the invitation
```
matteosz, line 51: I guess a previous PR got blindly merge as I don't remember reviewing this. Why is `CREATES_INVITATION `defined as static constant when then it's not? You could have simply passed it in a custom fragment constructor

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/federation/LinkedOrganizationsViewModel.kt
```diff
@@ -0,0 +1,199 @@
...
+  private val connecting = AtomicBoolean(false)
+  private val disposables = CompositeDisposable()
+  var manager: FragmentManager? = null
+  override val nbScanned = MutableLiveData<Int>()
+
+  fun setLaoId(laoId: String?) {
+    if (laoId != null) {
+      this.laoId = laoId
+    }
+  }
```
matteosz, line 50: Opt.: Here you could force directly to accept `String` instead of `String?` (it would require some more cascade modifications until the source of that string though). And in that case you could simply remove the custom setter

matteosz (changes requested):
Good work! But there are some mistakes in the application of the MVVM pattern

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/Objects.kt
quadcopterman: Yes, for now it doesn't persist but it will happen in a future PR. We are still discussing about persistence and reconnection with the federation team.

fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/LinkedOrganizationsRepository.kt
```diff
@@ -0,0 +1,45 @@
...
+import javax.inject.Singleton
+
+/**
+ * This class is the repository of federation
+ *
+ * Its main purpose is to store received messages
+ */
+@Singleton
+class LinkedOrganizationsRepository @Inject constructor() {
+  private var challenge: Challenge? = null
```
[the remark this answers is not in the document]
quadcopterman: Actually the Challenge is only needed during the connection. To store the state of a connection, this is done via a new message, federationResult (that we receive at the end of the connection), that will come in a future PR.

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/federation/LinkedOrganizationsViewModel.kt
```diff
@@ -0,0 +1,199 @@
...
+        }
+
+    if (federationDetails.challenge == null) {
+      // The federationDetails object is sent by the invitation creator
+      disposables.add(
+          sendFederationExpect(
+                  federationDetails.laoId,
+                  federationDetails.serverAddress,
+                  federationDetails.publicKey,
+                  getMessageGeneralFromChallenge(linkedOrgRepo.getChallenge()!!))
```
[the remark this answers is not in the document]
quadcopterman: Yes, this is part of the protocol we defined for federation. Basically, the federationDetails message could have a null challenge in normal operation, but in this case the federationExpect message that follows will contain the challenge received by the server just before. When sending an Expect/Init the challenge cannot be null, it comes either from the federationDetails message or from the repository as it was sent by the server before. This is probably explained better (with drawings) in the doc (in issue #1800)

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/federation/LinkedOrganizationsInviteFragment.kt
```diff
@@ -141,20 +142,17 @@ class LinkedOrganizationsInviteFragment : Fragment() {
...
               .sendFederationInit(
-                  repo.otherLaoId!!,
-                  repo.otherServerAddr!!,
-                  repo.otherPublicKey!!,
-                  repo.getChallenge()!!,
+                  linkedOrganizationsViewModel.getOtherLao()!!,
+                  linkedOrganizationsViewModel.getOtherServerAddress()!!,
+                  linkedOrganizationsViewModel.getOtherPublicKey()!!,
+                  linkedOrganizationsViewModel.getChallenge()!!,
               )
```
matteosz: This seems very odd, you're calling a method in the view model (sendFederationInit and passing arguments by calling again methods of the view model? Why isn't it done entirely in the view model?

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/federation/LinkedOrganizationsInviteFragment.kt
```diff
@@ -179,11 +177,11 @@ class LinkedOrganizationsInviteFragment : Fragment() {
   companion object {
     private val TAG: String = LinkedOrganizationsInviteFragment::class.java.simpleName
     private const val QR_SIDE = 800
-    private var CREATES_INVITATION = false
+    private var createdInvitation = false
 
     @JvmStatic
     fun newInstance(createsInvitation: Boolean): LinkedOrganizationsInviteFragment {
-      CREATES_INVITATION = createsInvitation
+      createdInvitation = createsInvitation
```
matteosz, line 200: I didn't mean to simply switch the var name to lower case, but to make it not a static constant but a fragment variable. If you leave it as static constant then the upper case it's correct

quadcopterman: Oh yes now I understand, I corrected it :)

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/federation/LinkedOrganizationsInviteFragment.kt
quadcopterman: Yes doing it in the view model is actually way better, I can then also remove some getters that are only used for that

matteosz (approved):
LGTM! Great job 😄

PR #1841, opened 2024-05-01
[be1] add federation result handling

1florentin:
This PR adds handling of the federation result message.

Also it modifies the already existing handling of the federation init message, to forward the challenge to the remote server.


be1-go/channel/federation/federation.go
```diff
@@ -369,9 +376,57 @@ func (c *Channel) processFederationChallenge(msg message.Message,
...
+	dataBase64 := base64.URLEncoding.EncodeToString(dataBytes)
+	signatureBytes, err := c.hub.Sign(dataBytes)
+	if err != nil {
+		return xerrors.Errorf("failed to sign federation result message: %v", err)
+	}
+	signatureBase64 := base64.URLEncoding.EncodeToString(signatureBytes)
+
+	federationResultMsg := message.Message{
+		Data:              dataBase64,
+		Sender:            c.hub.GetPubKeyServ().String(),
```
arnauds5, line 402: c.hub.GetPubKeyServ().String() returns an hexadecimal representation of the key.
It should be the base64 representation of the key

be1-go/channel/federation/federation.go
```diff
@@ -369,9 +376,57 @@ func (c *Channel) processFederationChallenge(msg message.Message,
...
+			Channel: remoteOrg.fedChannel,
+			Message: federationResultMsg,
+		},
+	}
+	buf, err := json.Marshal(&rpcMessage)
+	if err != nil {
+		return xerrors.Errorf("failed to marshal publish query: %v", err)
+	}
+
+	remoteOrg.socket.Send(buf)
```
arnauds5, line 429: At this point remoteOrg.state is set to WaitResult, should we set it to Connected here ?

be1-go/channel/federation/federation.go
```diff
@@ -475,6 +525,97 @@ func (c *Channel) processChallengeRequest(msg message.Message,
...
+	challengeSignatureBytes, err := base64.URLEncoding.DecodeString(federationResult.ChallengeMsg.Signature)
+	if err != nil {
+		return xerrors.Errorf("failed to decode challenge signature in FederationResult: %v", err)
+
+	}
+
+	remoteOrg := c.getRemoteOrganization(federationChallenge.Value)
+	remoteOrg.Lock()
+	defer remoteOrg.Unlock()
+	pkBytes, err := base64.URLEncoding.DecodeString(remoteOrg.organizerPk)
```
arnauds5, line 578: We should check that the remoteOrg.state is WaitResult before trying to process the message

be1-go/channel/federation/federation.go
```diff
@@ -40,15 +42,17 @@ const (
 )
 
 type remoteOrganization struct {
-	laoId       string
-	fedChannel  string
-	organizerPk string
+	laoId             string
+	fedChannel        string
+	organizerPk       string
+	signedOrganizerPk string
```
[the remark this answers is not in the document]
1florentin: The original purpose of that field is that, in the result message the server must be able to verify the signature on the challenge (this is done because the challenge is a standalone-message signed by the remote organizer) and also verify a signature on the pk of the local organizer, this is however not happening right now---this is an oversight, i'll change asap.

be1-go/channel/federation/federation.go
```diff
@@ -475,6 +534,100 @@ func (c *Channel) processChallengeRequest(msg message.Message,
...
+	if !(bytes.Equal(resultPkBytes, localPkBytes)) {
+		return xerrors.Errorf("invalid public key contained in FederationResult message")
+
+	}
+
+	//err = schnorr.Verify(crypto.Suite, remotePk, localPkBinary, pkSignatureBytes)
+	//if err != nil {
+	//	return xerrors.Errorf("failed to verify remote signature on local organizer public key: %v", err)
+
+	//}
```
arnauds5, line 624: If the PublicKey is only the public key of Alice, then we could directly compare it to c.localOrganizerPk
```suggestion
	if c.localOrganizerPk != federationResult.PublicKey {
		return xerrors.Errorf("invalid public key contained in FederationResult message")
	}
```

PR #1842, opened 2024-05-02
Federation Authentication protocol documentation

onsriahi14:
This PR defines federation's authentication protocol in the messageData.md file


arnauds5 (approved):
The images about the protocol are still the old ones, for example the request challenge is not displayed in the illustrations.

Otherwise, the changes look good to me

quadcopterman (approved):
Looks good to me

docs/messageData.md
```diff
@@ -2920,8 +2947,107 @@ The two organizers meet in person to exchange the federation_details message (th
...
+    },
+    "timestamp": {
+      "type": "integer",
+      "description": "[Timestamp] of the request",
+      "minimum": 0
+    }
+  },
+  "additionalProperties": false,
+  "required": [
+    "timestamp"
```
K1li4nL, line 2992: I believe the `object` and `action` field are missing there.

docs/messageData.md
```diff
@@ -2920,8 +2947,107 @@ The two organizers meet in person to exchange the federation_details message (th
...
+    "valid_until": {
+      "type": "integer",
+      "description": "[Timestamp] of the expiration time",
+      "minimum": 0
+    }
+  },
+  "additionalProperties": false,
+  "required": [
+    "value",
+    "valid_until"
```
K1li4nL, line 3044: Same as above, I believe the `object` and `action` field are missing there.

docs/messageData.md
```diff
@@ -3024,79 +3139,171 @@ It contains the necessary connection details, and a challenge which the remote p
...
+}
+```
+</details>
+
+```json5
+// ../protocol/query/method/message/data/dataFederationResult.json
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://raw.githubusercontent.com/dedis/popstellar/master/protocol/query/method/message/data/dataFederationResult.json",
+  "description": "Sent by an server to a remote server, to inform them about the result of the authentication procedure",
```
K1li4nL, line 3238: typo: 
```suggestion
  "description": "Sent by a server to a remote server, to inform them about the result of the authentication procedure",
```

docs/messageData.md
```diff
@@ -3024,79 +3139,171 @@ It contains the necessary connection details, and a challenge which the remote p
...
+    "challenge"
+  ]
+}
 
+
+
+```
+
+## Establishing or Aborting the connection (federation#result)
+This message is used to know the status of the connection, whether it was a success and a connection between the two LAOs was established or it was a failure.
```
K1li4nL, line 3209: Which server send this message ?

K1li4nL (commented):
The doc looks good to me overall, some of these comments seem a bit late since they require changing the json themselves. One big question I have is: on which channel is this all happening ? Unless I missed it, its not mentioned in the doc

docs/messageData.md
```diff
@@ -2920,8 +2947,107 @@ The two organizers meet in person to exchange the federation_details message (th
...
+  "required": [
+    "value",
+    "valid_until"
+  ]
+}
+
+```
+
+## Expecting a connection (federation#expect)
 With this message the organizer lets the server know to expect a federation_init message from the remote party idetailed in the message body.
```
pierluca, line 3051:
```suggestion
With this message the organizer lets the server know to expect a federation_init message from the remote party detailed in the message body.
```

docs/messageData.md
```diff
@@ -2978,24 +3107,10 @@ With this message the organizer lets the server know to expect a federation_init
...
-                    "description": "[Timestamp] of the expiration time",
-                    "minimum": 0
-                }
-            },
-            "additionalProperties": false,
-            "required": ["value", "valid_until"]
-        }
-    },
+          "$ref": "../message.json",
+          "$comment": "message/message containing a FederationChallenge data"
```
pierluca, line 3111: what's a `message/message` ? :-)

docs/messageData.md
```diff
@@ -2978,24 +3107,10 @@ With this message the organizer lets the server know to expect a federation_init
             "$comment": "public key of the remote organizer"
```
pierluca, line 3107: here and in other places you refer to a "remote" organizer, it might be useful to define this once at the beginning of the federation protocol to clarify who is considered "local" and who's considered remote. You might also want to consider renaming this to "inviting" / "joining" , as that might be less ambiguous.

docs/messageData.md
```diff
@@ -2939,8 +3065,11 @@ With this message the organizer lets the server know to expect a federation_init
     "public_key": "UvViTxoKsB3XVP_ctkmOKCJpMWb7fCzrcb1XDmhNe7Q=",
     "server_address": "wss://ethz.ch:9000/server",
     "challenge": {
```
pierluca, line 3067: if the point here is to refer to the original challenge message, why not just refer to its message_id ? That would be entirely sufficient to someone having access to that message, seeing that it is a hash of the data and signature.

docs/messageData.md
```diff
@@ -2978,24 +3107,10 @@ With this message the organizer lets the server know to expect a federation_init
             "$comment": "public key of the remote organizer"
         },
         "challenge": {
```
pierluca, line 3109: same question as above, why not just refer to the challenge's `message_id`?

docs/messageData.md
```diff
@@ -3011,7 +3126,7 @@ With this message the organizer lets the server know to expect a federation_init
 
 
 
-## Federation init
+## Initiating a connection (federation#init)
 This message is used to let a server know to initiate a connection with a remote server.
 
 It contains the necessary connection details, and a challenge which the remote party will expect to receive and the server will use to authenticate himself.
```
pierluca, line 3132: `remote party` is ambiguous. Server or client ?

pierluca (commented):
Right now, the messages themselves are more or less clear, but how to use them isn't. For instance:
- Who should be sending which message ? 
- On which channel are these messages expected ? 
- If it's on a `/root/.../federation` channel, which LAO's ?

... can you spot other such issues ? :)
(I haven't tried to be exhaustive - some other issues may be latent, I just haven't readily spotted them)

docs/messageData.md
1florentin: We have to include the whole message here, because the 'original' challenge message (i.e. the one sent from the server to the organizer after a federation_challenge_request is not signed by the organizer). Upon receiving the message from the server, the organizer must extract the challenge data and create a new challenge message (now signed by him---the organizer) and pack that message into the federation_expect message s.t. the server can then again extract the challenge message from the federation_expect (which is now signed by the organizer) and send it back to the remote server as part of the federation result message.

Typing this out i realize this sounds very complicated but I think we discussed this and ultimately decided that  this is the best (and possibly only) way to implement the scheme with the trust boundaries wie defined.

docs/messageData.md
1florentin: I agree that the current wording can at some points be ambiguous, but I fear that if we define it at the start and subsequently remove these adjectives in the later definitions it could be even more ambiguous.

Maybe it would be good to sort of explain at the beginning that the documentation (and protocol definition) is written from the perspective of a specific party (i.e. an organizer Alice) who takes on defined role (i.e. Alice could be the organizer receiving the challenge from the other)?

docs/messageData.md
1florentin, line 3111:
```suggestion
          "$comment": "message containing a FederationChallenge"
```

docs/messageData.md
1florentin, line 3132:
```suggestion
It contains the necessary connection details, and a challenge which the remote server will expect to receive and the server will use to authenticate himself.
```

docs/messageData.md
1florentin: This message is sent by the server who verifies the message. The purpose of the message is to let the other server (i.e. the one that sent the challenge) know if the connection succeeded or not.

Maybe we could describe it more like this?

```suggestion
This message is sent by the server verifying a given challenge to the server who sent the specific challenge. The purpose of this message is to provide the non-verifying server a final status of the connection, whether it was a success and a connection between the two LAOs is now established or there was a failure. (In this case the `reason` field may be used to provide further details)
```

docs/messageData.md
```diff
@@ -3024,79 +3139,171 @@ It contains the necessary connection details, and a challenge which the remote p
...
+    "public_key",
+    "challenge"
+  ]
+}
 
+
+
+```
+
+## Establishing or Aborting the connection (federation#result)
```
1florentin, line 3208:
```suggestion
## Finalizing or Disregarding the connection attempt (federation#result)
```

1florentin:
> Right now, the messages themselves are more or less clear, but how to use them isn't. For instance:
> 
> * Who should be sending which message ?
> * On which channel are these messages expected ?
> * If it's on a `/root/.../federation` channel, which LAO's ?
> 
> ... can you spot other such issues ? :) (I haven't tried to be exhaustive - some other issues may be latent, I just haven't readily spotted them)

I agree that these facts remain a bit clouded. I don't know if it desired by the team but I think that a graphic showing the concrete steps and messages in a sequential format would make things clearer (like the one I designed here: https://github.com/dedis/popstellar/issues/1800#issue-2235464579 - of course this would have to be updated)?

docs/messageData.md
arnauds5: These fields are missing in the file ../protocol/query/method/message/data/dataFederationChallengeRequest.json

docs/messageData.md
arnauds5: These fields are also missing in the file ../protocol/query/method/message/data/dataFederationChallenge.json

docs/messageData.md
arnauds5: We are using the general message found at protocol/query/method/message/message.json, where the data field contains a dataFederationChallenge

1florentin (approved):

PR #1843, opened 2024-05-02
Add rumor handler

DanielTavaresA:
Adding handling for the rumors. If a server receives a rumor it should do the following : 
Checks if this rumor is new, if it is : 
- [ ] Processes the messages that it contains. => removed from this PR
- [x] Store it in the database.
- [x] Sends back a result message to the sender with the status.
- [x] Trigger a gossip protocol with this rumor

Removing the processing of rumors from this PR as it will probably require a separate PR. Tried to use already existing logic from getMsgByIdAnswer processing, but tests keep failing for some unknown reason.


be2-scala/src/main/scala/ch/epfl/pop/decentralized/GossipManager.scala
```diff
@@ -0,0 +1,123 @@
...
+    Await.result(randomPeer, duration) match {
+      // updates the list based on response
+      // if some peers are available we send
+      case ConnectionMediator.GetRandomPeerAck(serverRef, greetServer) =>
+        val alreadySent: List[ServerInfos] = activeGossip :+ (serverRef -> greetServer)
+        activeGossipProtocol += (rumorRpc -> alreadySent)
+        serverRef ! ClientAnswer(
+          Right(rumorRpc)
+        )
+      // else remove entry
```
t1b00, line 64: Maybe add a case _ so that match is complete

be2-scala/src/main/scala/ch/epfl/pop/decentralized/GossipManager.scala
```diff
@@ -0,0 +1,123 @@
...
+    val activeGossipPeers = activeGossipProtocol.filter((k, _) => k.id == response.id)
+    // response is expected because only one entry exists
+    if (activeGossipPeers.size == 1) {
+      activeGossipPeers.foreach { (rumorRpc, _) =>
+        if (response.result.isEmpty && Random.nextDouble() < stopProbability) {
+          activeGossipProtocol -= rumorRpc
+        } else {
+          sendRumorToRandomPeer(rumorRpc)
+        }
+      }
```
t1b00, line 80: Same as discussed in meeting, shoudln't we handle case where size != 1?

be2-scala/src/test/scala/ch/epfl/pop/decentralized/GossipManagerSuite.scala
```diff
@@ -0,0 +1,137 @@
...
+    val peerServer3 = TestProbe()
+    val peerServer4 = TestProbe()
+
+    val peers = List(peerServer1, peerServer2, peerServer3, peerServer4)
+
+    // register server
+    connectionMediatorRef ? ConnectionMediator.NewServerConnected(peerServer1.ref, GreetServer(PublicKey(Base64Data("")), "", ""))
+    connectionMediatorRef ? ConnectionMediator.NewServerConnected(peerServer2.ref, GreetServer(PublicKey(Base64Data("")), "", ""))
+    connectionMediatorRef ? ConnectionMediator.NewServerConnected(peerServer3.ref, GreetServer(PublicKey(Base64Data("")), "", ""))
+    connectionMediatorRef ? ConnectionMediator.NewServerConnected(peerServer4.ref, GreetServer(PublicKey(Base64Data("")), "", ""))
```
t1b00, line 84: You can iterate on the list here ;)

be2-scala/src/test/scala/ch/epfl/pop/decentralized/GossipManagerSuite.scala
```diff
@@ -0,0 +1,137 @@
...
```
t1b00, line 110: Same here, possible to iterate on a list

t1b00 (changes requested):
Good code, a few minor improvements can be done :)

be2-scala/src/main/scala/ch/epfl/pop/decentralized/GossipManager.scala
```diff
@@ -0,0 +1,123 @@
...
+    case Right(jsonRpcResponse: JsonRpcResponse) =>
+      gossipManager ? ProcessResponse(jsonRpcResponse)
+      Right(jsonRpcResponse)
+    case graphMessage @ _ => graphMessage
+  }
+
+  sealed trait Event
+  final case class MonitoredRumor(jsonRpcRumor: JsonRpcRequest)
+  final case class SendRumorToRandomPeer(jsonRpcRequest: JsonRpcRequest)
+  final case class ProcessResponse(jsonRpcResponse: JsonRpcResponse)
```
K1li4nL, line 118: I find that name a bit confusing, its going to clash with the code that process the messages in the rumors I think

K1li4nL (changes requested):
Here is a first half of my review, good job overall !

be2-scala/src/main/scala/ch/epfl/pop/decentralized/GossipManager.scala
DanielTavaresA: I am not sure what could be done in that case. For two reasons : 
- If there is 0 match => response is related to a gossip that isn't active, so nothing should be done related to this response
- If there is 2 matches it means that we have two rumor to which this response could be related to, and for now we don't have a way to distinguish them.

What do you see as the most suitable behaviour ? Select a random rumor from the pool of possible rumors and process ? or ignore  ? (or any other suggestion ?)

be2-scala/src/main/scala/ch/epfl/pop/decentralized/GossipManager.scala
```diff
@@ -0,0 +1,123 @@
...
+    dbActorRef: AskableActorRef,
+    monitorRef: ActorRef,
+    connectionMediator: AskableActorRef,
+    stopProbability: Double = 0.5
+) extends Actor with AskPatternConstants {
+
+  private type ServerInfos = (ActorRef, GreetServer)
+
+  monitorRef ! GossipManager.Ping()
+  connectionMediator ? ConnectionMediator.Ping()
```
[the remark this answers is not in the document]
DanielTavaresA: As explained during the meeting, the type of ConnectionMediator forces me to use "?". But in this case I do not expect any response back. This is the reason why there is no handle of the response.

be2-scala/src/test/scala/ch/epfl/pop/decentralized/GossipManagerSuite.scala
```diff
@@ -0,0 +1,135 @@
...
+
+    val output = Source.single(Right(rumorRequest)).via(gossipHandler).runWith(Sink.head)
+
+    Await.result(output, duration)
+
+    peers.map(_.receiveOne(duration)).count(_ != null) shouldBe 1
+
+  }
+
+  test("gossip handler should send rumor if there is an ongoing gossip protocol") {
```
K1li4nL, line 93: Not sure I understand what you are testing here ?  How is it different from the previous test ?

K1li4nL (changes requested):
Almost good to me, I just don't understand one of your tests.

be2-scala/src/test/scala/ch/epfl/pop/decentralized/GossipManagerSuite.scala
DanielTavaresA: It asserts that the gossip protocol goes on as expected (do not send to a peer that already received). But looking back it is true that it includes what previous tests is doing. Previous test can be removed imo.

K1li4nL: Okay, I got it now, you can keep the previous one, it will be easier to spot the break in it if that part fails. This last test is quite difficult to understand, I would love to see some comments for the major steps.

K1li4nL (changes requested):

K1li4nL (approved):
All good for me, good job !

t1b00 (approved):

PR #1844, opened 2024-05-02
change signature of readRumor for more intuitive

DanielTavaresA:
Changed from readRumors (db handle multiples queries at once) to readRumor that handles only one query at time. This shifts the responsibility of handling multiples query to the querier. It makes the use of the database more intuitive

### DIFF
--- a/be2-scala/src/main/scala/ch/epfl/pop/storage/DbActor.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/storage/DbActor.scala
@@ -613,10 +613,10 @@ final case class DbActor(
         case failure    => sender() ! failure.recover(Status.Failure(_))
       }
 
-    case ReadRumors(desiredRumor) =>
+    case ReadRumor(desiredRumor) =>
       log.info(s"Actor $self (db) received a ReadRumor request")
       Try(readRumor(desiredRumor)) match {
-        case Success(foundRumor) => sender() ! DbActorReadRumors(foundRumor)
+        case Success(foundRumor) => sender() ! DbActorReadRumor(foundRumor)
         case failure             => sender() ! failure.recover(Status.Failure(_))
       }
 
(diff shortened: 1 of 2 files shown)


t1b00 (dismissed):

be2-scala/src/main/scala/ch/epfl/pop/storage/DbActor.scala
```diff
@@ -951,7 +946,7 @@ object DbActor {
 
   /** Response for a [[ReadRumors]]
     */
-  final case class DbActorReadRumors(foundRumors: Map[PublicKey, List[Rumor]]) extends DbActorMessage
+  final case class DbActorReadRumors(foundRumor: Option[Rumor]) extends DbActorMessage
```
K1li4nL, line 949: If it returns only one rumor:
```suggestion
  final case class DbActorReadRumor(foundRumor: Option[Rumor]) extends DbActorMessage
```

be2-scala/src/main/scala/ch/epfl/pop/storage/DbActor.scala
```diff
@@ -857,10 +852,10 @@ object DbActor {
   final case class WriteRumor(rumor: Rumor) extends Event
 
   /** Requests the Db for rumors corresponding to keys {server public key:rumor id}
-    * @param desiredRumors
+    * @param desiredRumor
     *   Map of server public keys and list of desired rumor id for each
     */
-  final case class ReadRumors(desiredRumors: Map[PublicKey, List[Int]]) extends Event
+  final case class ReadRumors(desiredRumor: (PublicKey, Int)) extends Event
```
K1li4nL, line 858: If it returns only one rumor:
```suggestion
  final case class ReadRumor(desiredRumor: (PublicKey, Int)) extends Event
```

K1li4nL (changes requested):
Just a typo otherwise LGTM

K1li4nL (approved):
LGTM !

Issue #1848, opened 2024-05-06
[BUG] Spamming Warning Toast when Social without Token


Kaz-ookid:
### Description (Actual behavior)
A (legitimate) warning toasts pops up in Social when you don't have a token (not participated in any roll-calls). The warning is beeing spammed for infinity, damaging user experience. 

### Expected behavior
The warning toast should appear only once, upon opening the Social tab.

### How to reproduce
1. Create a LAO on device 1, open a roll call, close the roll call and add some chirps in the Social Tab
2. Join LAO with device 2
3. Go to Social Tab
4. See the warning toast `could not retrieve your own pop token`

![image](https://github.com/dedis/popstellar/assets/100324323/6faf7baf-d5d7-46eb-aca0-7cbc566d367b)

##### Front-ends:
- [ ] Fe1-Web (please include browser's names & version)
- [X] Fe2-Android (please specify if phone or emulation, and Android version)
- [ ] Not applicable

##### Back-ends:
- [ ] Be1-Go
- [ ] Be2-Scala
- [X] Not applicable

##### Environment (as applicable):
 - Device: [e.g. Pixel 5 emulator on Mac]
 - OS: [e.g. Android 11, Windows 10]
 - Browser: [e.g. Chrome 57]

### Workaround
Deal with getting spammed for ages

### Impact
Bad user experience

### Possible root cause
Bad logic for the toast displayIf you have any idea or hypothesis as to what is the root cause of this bug, feel free to elaborate here.

PR #1849, opened 2024-05-06
Fixed constant display of error message when Social tab without Token

Kaz-ookid:
addressing issue #1848 

Disabled error toast when the error is `InvalidPoPTokenException`, otherwise it would just pop multiple times for every single chirps. The same error is displayed once on opening the Social Tab in `isOwner` so the user is still aware that no token is available.

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/socialmedia/SocialMediaViewModel.kt
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/socialmedia/SocialMediaViewModel.kt
@@ -61,9 +61,9 @@ constructor(
   private val mNumberCharsLeft = MutableLiveData<Int>()
   val bottomNavigationTab = MutableLiveData(SocialMediaTab.HOME)
 
-    private val _hasValidToken = MutableLiveData<Boolean>()
+  private val _hasValidToken = MutableLiveData<Boolean>()
 
-    private val disposables: CompositeDisposable = CompositeDisposable()
+  private val disposables: CompositeDisposable = CompositeDisposable()
 
   override fun onCleared() {
     super.onCleared()
@@ -75,7 +75,6 @@ constructor(
      * Getters for MutableLiveData instances declared above
      */
     get() = mNumberCharsLeft
-
 
   /*
    * Methods that modify the state or post an Event to update the UI.
@@ -275,10 +274,10 @@ constructor(
 
       sender == token.publicKey.encoded
     } catch (e: KeyException) {
-        if (_hasValidToken.value == true) {
-            logAndShow(getApplication(), TAG, e, R.string.error_retrieve_own_token)
-        }
-        false
+      if (_hasValidToken.value == true) {
+        logAndShow(getApplication(), TAG, e, R.string.error_retrieve_own_token)
+      }
+      false
     }
   }
 
@@ -289,19 +288,17 @@ constructor(
    * @return true if we have sent such reaction, false otherwise
    */
   fun isReactionPresent(senders: Set<String?>): Boolean {
-      return try {
-          val token = validPoPToken
-          val toSearch = token.publicKey.encoded
-          senders.contains(toSearch)
-      } catch (e: KeyException) {
-          if (_hasValidToken.value == true) {
-              logAndShow(getApplication(), TAG, e, R.string.error_retrieve_own_token)
-          }
-          false
+    return try {
+      val token = validPoPToken
+      val toSearch = token.publicKey.encoded
+      senders.contains(toSearch)
+    } catch (e: KeyException) {
+      if (_hasValidToken.value == true) {
+        logAndShow(getApplication(), TAG, e, R.string.error_retrieve_own_token)
       }
+      false
+    }
   }
-
-
 
   fun setLaoId(laoId: String) {
     this.laoId = laoId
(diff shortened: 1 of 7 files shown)


fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/socialmedia/SocialMediaViewModel.kt
```diff
@@ -286,6 +288,11 @@ constructor(
...
       val toSearch = token.publicKey.encoded
       senders.contains(toSearch)
+    } catch (e: InvalidPoPTokenException) {
+      // InvalidPoPTokenException means the user did not attend any roll calls and had no token.
+      // This is not an issue worth displaying the error for every single chirps.
+      // This error will already show once in isOwner.
+      return false
     } catch (e: KeyException) {
       logAndShow(getApplication(), TAG, e, R.string.error_retrieve_own_token)
       false
```
matteosz, line 298: To simplify a bit the code clarity you can exploit the fact that InvalidPoPTokenException is a subclass of KeyException, so you can use the catch only once for the most generic KeyException and use an if to logAndShow

```suggestion
    } catch (e: KeyException) {
      if (e !is InvalidPoPTokenException) { 
        logAndShow(getApplication(), TAG, e, R.string.error_retrieve_own_token)
      }
      false
```

matteosz (commented):
Overall, I don't think that your solution solves the issue in the right way (but also that solves the issue at all, I tested the code and it seems that the message keeps spamming). 

If we want the message of the poptoken missing appearing only once, the right approach would be an ad-hoc check done in the fragment creation, which maybe sets a boolean constant that tells whether the user has a valid pop token. Then, all the current methods in the social media view model which catch and log a key exception must be kept to check errors (but simply made silent in case of no token). This is obviously a much longer fix to do and I'm also happy with a more provisional solution, as long as it works though.

I suggest you test this with 2 emulator instances so you can actually simulate this behaviour.

Kaz-ookid:
I've made some updates to the token check implementation. Here's a quick rundown:

- Launched the PoPToken validation in the `SocialMediaHomeFragment`'s `onCreateView` to kick off right when the view is created.
- Updated the ViewModel to maintain a LiveData that reflects the validity of the token.
- Adjusted the error handling in `isReactionPresent` and `isOwner` to suppress token-related toasts if we already know the token is not setup.

Now, when opening the Social tab, we are not even getting a single toast error, but we are only getting it when trying to send a chirps while not having the token, I think it makes more sense. Finally, the error should trigger if there is an issue with the retrieval of the token, other than it being uninitialized.

This should handle the spamming issue more gracefully by centralizing the token validation and making our error handling smarter. Let me know if this works well on your end too.

matteosz (dismissed):
LGTM! Nice job also with testing, now the error handling is much cleaner :)

matteosz:
> LGTM! Nice job also with testing, now the error handling is much cleaner :)

Tests are failing because it hasn't been formatted, so remember to format and then either squash the commit or push it as is

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/socialmedia/SocialMediaViewModel.kt
```diff
@@ -60,6 +61,8 @@ constructor(
   private val mNumberCharsLeft = MutableLiveData<Int>()
   val bottomNavigationTab = MutableLiveData(SocialMediaTab.HOME)
 
+  private val _hasValidToken = MutableLiveData<Boolean>()
```
matteosz, line 64: How come you decided to use this name convention here (I mean underscore to start the variable name)

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/socialmedia/SocialMediaViewModel.kt
```diff
@@ -109,7 +112,7 @@ constructor(
     val addChirp = AddChirp(text, parentId, timestamp)
 
     return Single.fromCallable { validPoPToken }
-        .doOnSuccess { token: PoPToken ->
+        .doOnSuccess { token: PoPToken? ->
```
matteosz, line 115: Why did you change that? Technically the token cannot be null there: if a KeyException error is thrown and then become null, the doOnSuccess won't be triggered anyway (as it won't be a success)

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/socialmedia/SocialMediaViewModel.kt
```diff
@@ -304,7 +312,26 @@ constructor(
...
 
+  fun checkValidPoPToken() {
+    disposables.add(
+        Single.fromCallable {
+              keyManager.getValidPoPToken(laoId, rollCallRepo.getLastClosedRollCall(laoId))
+            }
+            .subscribeOn(schedulerProvider.io())
+            .observeOn(schedulerProvider.mainThread())
+            .subscribe(
+                { _ -> _hasValidToken.postValue(true) },
```
matteosz, line 323: You can avoid the ` _ ->` here

matteosz (commented):
LGTM!

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/socialmedia/SocialMediaViewModel.kt
```diff
@@ -287,7 +293,9 @@ constructor(
       val toSearch = token.publicKey.encoded
       senders.contains(toSearch)
     } catch (e: KeyException) {
-      logAndShow(getApplication(), TAG, e, R.string.error_retrieve_own_token)
+      if (_hasValidToken.value == true) {
```
quadcopterman, line 296: Can't you remove the "== true" in the if ?

quadcopterman (approved):
Looks good to me, just a little comment. Nice to see that you added many tests for social media

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/socialmedia/SocialMediaViewModel.kt
Kaz-ookid: It is supposed to be the convention for mutable and private data. Copilot chose the name for me, and it felt normal since I've already come across this kind of naming. But i recon it can be misleading and there should be no issue not following this dark and unknown convention

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/socialmedia/SocialMediaViewModel.kt
Kaz-ookid: .... It helped raising the coverage 😅 (since the tests are having trouble creating valid data, some of them just could not get passed this without the "?"

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/socialmedia/SocialMediaViewModel.kt
Kaz-ookid: I am not sure I can since the value is `Boolean?` (could be null), but I'll check it out while addressing the other comments :thum

Kaz-ookid: In fact, I could not. I tried setting up the value to false in an early stage but did not change it, I guess it's really just a detail

matteosz (approved):
LGTM!

PR #1850, opened 2024-05-11
Improvements to Manual Input in Scanning Actions

Kaz-ookid:
# OUTDATED
this PR description is outdated (scroll down for the new one). In fact, I figured out that the Issue i was trying to solve was not only when manually joining a LAO, but for any scanning actions (add a witness, add attendee...). Then I had to find a modular solution for all of those actions. I let this description here, just to keep a trace of the event :+1: 

#### Disclaimer : tests are to come!

This PR introduces several enhancements to the manual entry process for joining a LAO, making it more user-friendly and efficient. Previously, users had to input a JSON string, which was not intuitive. Now, users can enter the server address and LAO ID directly through two separate text fields. User will also receive notifications if they fail to fill in both inputs, or if inputs are wrong.

### Key Changes:

1. **Separate Text Inputs**: Replaced the single JSON input with two distinct fields for the server address and the LAO ID. 

2. **Clipboard Integration**: Added "Paste from Clipboard" buttons next to the input fields. This quality-of-life feature allows users to quickly fill in fields with copied content.

3. **UX Enhancements**:
- **Keyboard Management**: The keyboard now automatically dismisses when exiting the Manual Add Menu (using the cross). This prevents the keyboard from obstructing the view or staying active unnecessarily.
- **Static Input Field Size**: Modified the text input fields to prevent them from scaling based on the length of the entered text. Users will now need to scroll through text within the input field if it exceeds the visible area, keeping the UI clean.

### Utility Updates:
- **ClipboardUtil Expansion**: `ClipboardUtil` now supports pasting functionality, could only handling copy actions until now.

### Screenshots:

image
image


Kaz-ookid:
Ok found out I fucked up, every QR scan in the app (add witness, add attendee...) Are using the same Manual Enter Menu that I modified. I need to find a more global solution that will modulate for the need...

Kaz-ookid:
## PR Summary
### Updates to QR Scanner for Flexible Input Handling

#### Note: Tests are pending!

This PR updates the QR Scanner Fragment to handle different types of manual inputs depending on the scanning action. It allows specific actions like joining a LAO or adding a witness to have customized input fields. This replaces the previous way of manual input : one and only input text fields that required a JSON format.

### Key Changes:

1. **Dynamic Input Fields**: Each scanning action now determines how many input fields it requires.

2. **Action-Specific JSON Formatting**: Custom JSON formatting logic is included for each action, ensuring the data is processed correctly based on the provided inputs.

3. **Enhanced Clipboard Integration**: Users can now paste content directly from the clipboard into the input fields, speeding up data entry.

4. **UX Improvements**:
   - **Automatic Keyboard Management**: The keyboard now automatically dismisses when exiting the Manual Add Menu (using the cross). This prevents the keyboard from obstructing the view or staying active unnecessarily.
   - **Fixed Input Field Size**: Modified the text input fields to prevent them from scaling based on the length of the entered text. Users will now need to scroll through text within the input field if it exceeds the visible area, keeping the UI clean.

### Utility Updates:
- **ClipboardUtil Expansion**: `ClipboardUtil` now supports pasting functionality, could only handling copy actions until now.

### screenshots:

image
image

matteosz (commented):
Very nice job! I like the pattern you used to solve the issue; also the code looks very clean (but don't worry I'll dive deeper to leave some picky comments once the PR is finished :) ), I just wanted to point out some general considerations.

I don't see the point of setting in the layout a pre-determined (4 in this case) number of entries. You could use some more general abstractions (e.g. a ListView), so you don't actually have to deal with the visibility of the remaining items (or also to avoid the case of a json having more than 4 fields). Then it would be make it much more extendible.

I don't understand the need of the paste button along each entry, as it something that the android keyboard should already provide (or am I missing something?).

Kaz-ookid:
Thanks for the help!

> I don't see the point of setting in the layout a pre-determined (4 in this case) number of entries. You could use some more general abstractions (e.g. a ListView), so you don't actually have to deal with the visibility of the remaining items (or also to avoid the case of a json having more than 4 fields). Then it would be make it much more extendible.
--> This is a solution I did not think about, that is why i went through tricky stuff. I will try and find a better solution, as a ListView :)


> I don't understand the need of the paste button along each entry, as it something that the android keyboard should already provide (or am I missing something?).
--> In fact Android has a quick paste button. But the button I provided enables to paste without even having to open the keyboard, which makes it faster. Android users could still be faster if they know how to use the clipboard history, by copying both fields and pasting them at once, but I still think it would be a good addition for most users, don't you think?

Kaz-ookid:
I tried out Recycler View, seems to be way more clean. I created `InputFieldsAdapter` and put it in GeneralUtils since i think it could serve any UI components that require variable number of inputs. For now it only works for String inputs, but it can be upgraded to accept more (such as Int) in the future.

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/GeneralUtils.kt
```diff
@@ -146,4 +169,122 @@ object GeneralUtils {
       emptyArray()
     }
   }
+
+  /**
```
matteosz, line 173: I'm not very sure the code you put in GeneralUtils fits the purpose of the file, perhaps a dedicated file for this dynamic list setup could be better, or even creating a UI Utils file.

matteosz (commented):
Very impressive work, especially with testing! LGTM but approving after the review from @quadcopterman as discussed on Slack

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/GeneralUtils.kt
Kaz-ookid: I'll check it out thanks

quadcopterman (dismissed):
It looks good to me, nice work !

matteosz (approved):
LGTM!

PR #1851, opened 2024-05-12
Add federation authentication protocol messages + tests + examples in Scala

onsriahi14:
This PR defines : 
- FederationInit message
- FederationExpect Message
- FederationChallengeRequest message
- FederationChallenge message
- FederationResult message
- Their respective JsonReader and JsonWriter
It also includes examples of each message and some unit tests


be2-scala/src/main/scala/ch/epfl/pop/json/MessageDataProtocol.scala
```diff
@@ -459,4 +461,149 @@ object MessageDataProtocol extends DefaultJsonProtocol {
       POPCHA_ADDRESS -> obj.popchaAddress.toJson
     )
   }
+
+  implicit object FederationChallengeFormat extends JsonFormat[FederationChallenge] {
+    final private val PARAM_CHALLENGE_VALUE: String = "value"
+    final private val PARAM_CHALLENGE_VALID_UNTIL: String = "validUntil"
```
DanielTavaresA, line 467: This field should match the one in the jsonRPC description : "valid_until". If not, when trying to parse a received message, an error will be thrown

be2-scala/src/main/scala/ch/epfl/pop/json/MessageDataProtocol.scala
```diff
@@ -459,4 +461,149 @@ object MessageDataProtocol extends DefaultJsonProtocol {
...
+      JsObject(jsObjectContent)
+    }
+  }
+
+  implicit object FederationExpectFormat extends JsonFormat[FederationExpect] {
+    final private val PARAM_ID: String = "laoId"
+    final private val PARAM_SERVER_ADDRESS: String = "serverAddress"
+    final private val PARAM_OTHER_ORGANIZER: String = "other_organizer"
+    final private val PARAM_CHALLENGE: String = "challenge"
+
```
DanielTavaresA, line 493: Same remark as previously. The PARAM should match some keys in the RPC message defined. But once you extracted them, you can use them as you want in your object "FederationExpect". "laoId" -> "lao_id", "serverAddress" -> "server_address", "other_organizer" -> "public_key"

be2-scala/src/main/scala/ch/epfl/pop/json/MessageDataProtocol.scala
```diff
@@ -459,4 +461,149 @@ object MessageDataProtocol extends DefaultJsonProtocol {
...
+      )
+      JsObject(jsObjectContent)
+    }
+  }
+
+  implicit object FederationInitFormat extends JsonFormat[FederationInit] {
+    final private val PARAM_ID: String = "laoId"
+    final private val PARAM_SERVER_ADDRESS: String = "serverAddress"
+    final private val PARAM_OTHER_ORGANIZER: String = "other_organizer"
+    final private val PARAM_CHALLENGE: String = "challenge"
```
DanielTavaresA, line 522: Same remark as federationExpect

be2-scala/src/main/scala/ch/epfl/pop/json/MessageDataProtocol.scala
```diff
@@ -459,4 +461,149 @@ object MessageDataProtocol extends DefaultJsonProtocol {
...
+      )
+      JsObject(jsObjectContent)
+    }
+  }
+
+  implicit object FederationResultFormat extends JsonFormat[FederationResult] {
+    final private val PARAM_STATUS: String = "status"
+    final private val PARAM_REASON: String = "reason"
+    final private val PARAM_ORGANIZER: String = "organizer"
+    final private val PARAM_CHALLENGE_MESSAGE = "challenge_message"
```
DanielTavaresA, line 572: Same remark as before.

be2-scala/src/main/scala/ch/epfl/pop/json/MessageDataProtocol.scala
```diff
@@ -459,4 +461,149 @@ object MessageDataProtocol extends DefaultJsonProtocol {
...
+      val jsObj = json.asJsObject
+      val status = jsObj.fields.get(PARAM_STATUS).collect {
+        case JsString(s) => s
+      }.getOrElse(throw new IllegalArgumentException(s"Can't parse json value $json to a FederationResult object"))
+
+      val reason = jsObj.fields.get(PARAM_REASON).collect {
+        case JsString(r) => r
+      }
+
+      val organizer = jsObj.fields.get(PARAM_ORGANIZER).flatMap(_.convertTo[Option[PublicKey]])
```
DanielTavaresA, line 584: nit : can't the same logic as reason be applied here ? 
val organizer = jsObj.fields.get(PARAM_ORGANIZER).collect {
        case key @ JsString(_) => key.convertTo[PublicKey]
    }

be2-scala/src/main/scala/ch/epfl/pop/model/network/method/message/data/ActionType.scala
```diff
@@ -25,3 +25,9 @@ enum ActionType(val action: String):
   case post_transaction extends ActionType("post_transaction")
   // popcha
   case authenticate extends ActionType("authenticate")
+  // federation actions
+  case init extends ActionType("init")
+  case expect extends ActionType("expect")
+  case challenge_request extends ActionType("challenge_request")
+  case challenge extends ActionType("challenge")
+  case federation_result extends ActionType("federation_result")
```
DanielTavaresA, line 33: If I am not mistaken, string should match actions from jsonRPC format. "federation_result" -> "result"

be2-scala/src/main/scala/ch/epfl/pop/model/network/method/message/data/federation/FederationExpect.scala
```diff
@@ -0,0 +1,27 @@
...
+import ch.epfl.pop.model.objects.{Hash, PublicKey}
+import spray.json.*
+
+final case class FederationExpect(
+    laoId: Hash,
+    serverAddress: String,
+    other_organizer: PublicKey,
+    challenge: Message
+) extends MessageData {
+  require(serverAddress.matches("^(ws|wss):\\/\\/.*(:\\\\d{0,5})?\\/.*$"), s"This is an invalid server address")
```
DanielTavaresA, line 16: As mentionned during meeting, using "require" doesn't allow to handle errors gracefully. As you have this pattern on your jsonRPC, you could assume that if a message reaches here, it has the right format. Or you could try to handle it gracefully and return an error/default value in case of wrong format.

DanielTavaresA (changes requested):
First part of my review. Overall good code, just small inconsistencies between jsonRPC and scala.

be2-scala/src/test/scala/ch/epfl/pop/model/network/method/message/data/federation/ChallengeRequestSuite.scala
```diff
@@ -0,0 +1,24 @@
...
+    CHALLENGE_REQUEST.timestamp should equal(TIMESTAMP)
+    CHALLENGE_REQUEST._object should equal(ObjectType.federation)
+    CHALLENGE_REQUEST.action should equal(ActionType.challenge_request)
+  }
+
+  test("BuildFromJson for RequestChallenge works as expected") {
+    val request_challenge_ = FederationChallengeRequest.buildFromJson(CHALLENGE_REQUEST.toJson.toString)
+
+    request_challenge_ should equal(CHALLENGE_REQUEST)
+  }
```
DanielTavaresA, line 23: To test that your implementation works with the json examples you could add the json in the /util/examples/json and use readJsonFromPath. That way you can be sure that your implementation is correct with the messages defined in the protocol. Because a subtle bug could be that you are building from a json that you created, so your implementation work to read and write with object within the app, but not for the outside world. You can find some examples in other tests that use readJsonFromPath

DanielTavaresA (changes requested):
Just one suggestion about tests to avoid finding bugs latter on. This suggestion applies to all tests about messages. After that LGTM

be2-scala/src/main/scala/ch/epfl/pop/json/MessageDataProtocol.scala
```diff
@@ -459,4 +461,150 @@ object MessageDataProtocol extends DefaultJsonProtocol {
...
+        case key @ JsString(_) => key.convertTo[PublicKey]
+      }
+
+      val challengeMessage = jsObj.fields.get(PARAM_CHALLENGE).collect {
+        case jsObj: JsObject => jsObj.convertTo[Message]
+      }.getOrElse(throw new IllegalArgumentException(s"Can't parse json value $json to a FederationResult object"))
+
+      (reason, publicKey) match {
+        case (None, Some(key)) => FederationResult(status, key, challengeMessage)
+        case (Some(r), None)   => FederationResult(status, r, challengeMessage)
```
K1li4nL, line 594: I believe the public key is a mandatory parameter, is this case really reachable ?

be2-scala/src/main/scala/ch/epfl/pop/model/network/method/message/data/ActionType.scala
```diff
@@ -25,3 +25,8 @@ enum ActionType(val action: String):
   case post_transaction extends ActionType("post_transaction")
   // popcha
   case authenticate extends ActionType("authenticate")
+  // federation actions
+  case init extends ActionType("init")
+  case expect extends ActionType("expect")
+  case challenge_request extends ActionType("challenge_request")
```
K1li4nL, line 31: snake case are a bit surprising here ?

be2-scala/src/main/scala/ch/epfl/pop/model/network/method/message/data/federation/FederationResult.scala
```diff
@@ -0,0 +1,30 @@
...
+    challenge: Message
+) extends MessageData {
+  def this(status: String, publicKey: PublicKey, challenge: Message) = this(status, None, Some(publicKey), challenge)
+  def this(status: String, reason: String, challenge: Message) = this(status, Some(reason), None, challenge)
+  override val _object: ObjectType = ObjectType.federation
+  override val action: ActionType = ActionType.result
+}
+
+object FederationResult extends Parsable {
+  def apply(status: String, reason: Option[String], public_key: Option[PublicKey], challenge: Message): FederationResult = {
```
K1li4nL, line 23: snake case are a bit surprising here ?

be2-scala/src/main/scala/ch/epfl/pop/model/objects/Base16Data.scala
```diff
@@ -0,0 +1,52 @@
...
+  /** Create a Base16Data instance from a string, ensuring it represents exactly 32 bytes when decoded. */
+  def encode(data: String): Base16Data = {
+    val hexString = stringToHexString(data)
+    if (hexString.length != 64) throw new IllegalArgumentException("Encoded data must represent exactly 64 hexadecimal characters (32 bytes).")
+    Base16Data(hexString)
+  }
+
+  /** Create a Base16Data instance from a byte array, ensuring it is exactly 32 bytes long. */
+  def encode(data: Array[Byte]): Base16Data = {
+    if (data.length != 32) throw new IllegalArgumentException("Byte array must be exactly 32 bytes long.")
```
K1li4nL, line 44: Why is that necessary ? Is this really the place to do this check ? We may get unhandled exceptions in the middle of the code, if not handled it could crash the entire backend

K1li4nL (changes requested):
Looks good to me overall, two issues I believe could be important otherwise a couple nitpicks

be2-scala/src/main/scala/ch/epfl/pop/json/MessageDataProtocol.scala
onsriahi14: the public key is only mandatory when we have success as a result

K1li4nL: Yes indeed, I misread the doc, perfect !

be2-scala/src/main/scala/ch/epfl/pop/model/network/method/message/data/ActionType.scala
K1li4nL: actually, we have a couple there already, so we may be stay coherent

K1li4nL (approved):
LGTM !

DanielTavaresA (approved):
Great work ! LGTM

PR #1859, opened 2024-05-16
[be1] Fix connexion problem

sgueissa:
fix for the pre-release [be1-v1.15.2.1](https://github.com/dedis/popstellar/releases/tag/be1-v1.15.2.1).

1. Remove the return inside the go routine handling the hub
2. Unsubscribe from all channels when sockets disconnect

### DIFF
--- a/be1-go/internal/popserver/hub.go
+++ b/be1-go/internal/popserver/hub.go
@@ -56,6 +56,7 @@ func (h *Hub) Start() {
 	go func() {
 		utils.LogInfo("start the Hub")
 		for {
+			utils.LogInfo("waiting for a new message")
 			select {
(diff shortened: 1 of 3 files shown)


be1-go/internal/popserver/hub.go
```diff
@@ -56,6 +56,7 @@ func (h *Hub) Start() {
 	go func() {
 		utils.LogInfo("start the Hub")
 		for {
+			utils.LogInfo("waiting for a new message")
```
pierluca, line 59: When you clean up logs, this would have to be a `Debug`. This will spam like crazy when the server is active

pierluca (approved):

PR #1871, opened 2024-05-17
[BE2] Work on more rules for linter

t1b00:
Newly implemented rules:
* IllegalFormatString
* IncorrectNumberOfArgs
* IncorrectlyNamedExceptions
* LonelySealedTrait
* MapGetAndGetOrElse
* NanComparison
* OptionSize
* OptionGet
* StripMarginOnregex
* TryGet

Improved rules
* EmptyInterpolatedString now supports literal.format() warning

Added Util file for utility methods (finding type, finding declarations...)

Task list: https://docs.google.com/spreadsheets/d/1LjsvOvH56TTYzyhhRpGPTqyeaSzew-pMDEPzRPVO3tE/edit?usp=sharing


be2-scala/linter/rules/src/main/scala/fix/NanComparison.scala
```diff
@@ -0,0 +1,46 @@
...
+      // => matcher function avoids repetition. Type of arguments is any since findDefinition will return Any,
+      // but it doesn't matter since we do pattern matching
+      (l, r) match {
+        case (Term.Select(Term.Name("Double"), Term.Name("NaN")), _) | (_, Term.Select(Term.Name("Double"), Term.Name("NaN"))) => Patch.lint(NanComparisonDiag(t))
+        case _ => Patch.empty
+      }
+    }
+
+    doc.tree.collect {
+      case t @ Term.ApplyInfix.After_4_6_0(lhs, Term.Name("=="), _, Term.ArgClause(List(rhs), _)) =>
```
DanielTavaresA, line 35: Like last time, shouldn't .equals be also handled ?

be2-scala/linter/input/src/main/scala/fix/StripMarginOnRegex.scala
```diff
@@ -0,0 +1,13 @@
+/*
+rule = StripMarginOnRegex
+ */
+package fix
+
+object StripMarginOnRegex {
+  def test() = {
+//    val regex = "match|this".stripMargin.r // assert: StripMarginOnRegex
+//    "match|that".stripMargin.r // assert: StripMarginOnRegex
```
DanielTavaresA, line 9: If not necessary anymore should be removed. In addition, your test should have at least a match checking that your rule is well triggered

DanielTavaresA (changes requested):
Overall good code. Couple of comments. If I could suggest an improvement it would be to add more documentation/comment in your code. Lint code is not trivial for persons for the outside and it can be hard to understand what is being done. In particular I think having a small comment on `case` saying what it is matching would improve greatly readability.

be2-scala/linter/rules/src/main/scala/fix/NanComparison.scala
t1b00: Yes! Added it with my last commit.
You'll notice I didn't test for .equals between Double and Float and Float and Double because it already throws a warning upon compilation (comparing values of types Float and Double using `equals` unsafely bypasses cooperative equality; use `==` instead).

K1li4nL (approved):
Really a non-expert eye here, your code looks good to me, the tests make sense so I will trust that, good job !

DanielTavaresA (approved):
Most important comments addressed. LGTM

PR #1877, opened 2024-05-17
New messages for pull rumors

DanielTavaresA:
To be able to use pull gossiping, we need new messages that will share the state of the received rumors per server.

### DIFF
--- a/protocol/query/method/rumor_state.json
+++ b/protocol/query/method/rumor_state.json
@@ -14,11 +14,11 @@
             "type": "object",
             "additionalProperties": false,
             "properties": {
-                "rumor_state": {
+                "state": {
                     "$ref": "object/rumor_state.json"
                 }
             },
-            "required": ["rumor_state"]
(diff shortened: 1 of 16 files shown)


protocol/query/method/rumor_state.json
```diff
@@ -0,0 +1,34 @@
...
+        "method": {
+            "description": "[String] operation to be performed by the query",
+            "const": "rumor_state"
+        },
+
+        "params": {
+            "type": "object",
+            "additionalProperties": false,
+            "properties": {
+                "rumor_state": {
```
sgueissa, line 17: you could just write state, we already know it is a rumor_state with the method

sgueissa (commented):
great I would just suggest changing the name of the param to just state

sgueissa (approved):
all good for me now with the added answer

K1li4nL (approved):
Approving blindly

PR #1887, opened 2024-05-25
[BE2] Add parsing of RumorState messages

DanielTavaresA:
Pending on #1877 .

Handles parsing from jsonRpc to usable object for new message RumorState


K1li4nL (approved):
LGTM !

PR #1889, opened 2024-05-27
[BE2] Added RumorStateAns to possible values of response

DanielTavaresA:
Pending on #1877

Added the possibility to have a list of rumors in the response. Some refactoring was needed because Scala do not see two List[A] and List[B] as different types. For that I added a new sealed trait ResultTypes that represents all possible ResultTypes. With this structure it is more manageable for future additions is more intuitive with a ResultObject having only one result and not one result for each possible type. ResultEmptyList is a "dirtyfix". It is necessary because for jsonRPC coming in, if list is empty it is impossible to differentiate a List[Rumor] or List[Message]. ResultEmptyList is a handled in a way that it returns a emptyList for both List[Rumor] and List[Message]. Such behaviour si not a real problem because there using jsonRPC id should avoid any conflict.


K1li4nL (approved):
Went quickly over it, looks good to me

PR #1892, opened 2024-05-29
[BE2] Fix be2 test db ref injection

DanielTavaresA:
fixes the issue of using the wrong refs for the database. 

Now all refs from code come from the PublishSubscribe graph, and if something should be injected, it is being done through PublishSubscribe.buildgraph(...)


be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/MessageHandler.scala
```diff
@@ -70,6 +70,7 @@ trait MessageHandler extends AskPatternConstants {
   def dbAskWritePropagate(rpcRequest: JsonRpcRequest): Future[GraphMessage] = {
     rpcRequest.getParamsMessage match
       case Some(m) =>
+        println(s"dbActorWriteAndPropagate : ${dbActor.actorRef}")
```
K1li4nL, line 73: This is might log a bit too much

be2-scala/src/test/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandlerTest.scala
```diff
@@ -214,142 +218,144 @@ class ElectionHandlerTest extends TestKit(ActorSystem("Election-DB-System")) wit
     system.actorOf(dbActorMock)
   }
 
+  private def injectDb(dbRef: AskableActorRef) = PublishSubscribe.buildGraph(Actor.noSender, dbRef, Actor.noSender, MessageRegistry(), Actor.noSender, Actor.noSender, Actor.noSender, false)
```
K1li4nL, line 221: Can't we use something like: https://www.scalatest.org/scaladoc/1.8/org/scalatest/BeforeAndAfterEach.html ?

K1li4nL (changes requested):
I am not entirely sure what these changes means for the overall design, it reduces a level of indirection in the handlers which is nice but means PublishSubscribe is a hard dependency for the entire system. I don't see any other way to fix the problem you are trying to fix without a larger changes right now so we might as well go that way.

be2-scala/src/test/scala/ch/epfl/pop/pubsub/graph/handlers/ElectionHandlerTest.scala
DanielTavaresA: in this case not, because different types of mocked db are injected. Using BeforeAndAfterEach would force to inject the same db for all test.

DanielTavaresA:
> I am not entirely sure what these changes means for the overall design, it reduces a level of indirection in the handlers which is nice but means PublishSubscribe is a hard dependency for the entire system. I don't see any other way to fix the problem you are trying to fix without a larger changes right now so we might as well go that way.

Agree, buildGraph is tight to adding references to the system, which doesn't seem a natural behaviour. Overall, a lot of references are passed around, eventhought they are unique in the system. We could try to centralize those references using maybe https://doc.akka.io/docs/akka/current/typed/actor-discovery.html as you recommended.

K1li4nL (approved):
LGTM

PR #1893, opened 2024-05-29
 [BE2] Federation Implementation

onsriahi14:
This PR validates and handles the federation messages. Upon receiving :
- FederationChallengeRequest, the server generates the challenge and sends it to its own organizer
- FederationExpect, the server validates the message and stores it in the database
- FederationInit, the server stores the message in the database, extracts the challenge from it and sends it to the server of the other LAO
- FederationChallenge, the server checks if the challenge is valid and prepares the result, sends it to the server of the other LAO and to its own organizer to let him know whether the federation is successful or not 
- FederationResult, the server sends the result to its own organizer to let him know whether the federation is successful or not  

Some examples and unit tests are provided to test the code


K1li4nL (approved):
Not much time to go over it unfortunately, so here is a blind approval

K1li4nL (approved):

PR #1902, opened 2024-06-03
[BE1] Better error handling

sgueissa:
This PR has for purpose to rework all the errors to  prepare the code for having better logs.

We removed the bad return type *answer.Error from everywhere and replace it with error, for that we created an error type with a stacktrace for this project named PopError.


K1li4nL (approved):
blind approval

PR #1903, opened 2024-06-03
[BE2] Added rumor state ans handler

DanielTavaresA:
When receiving a rumor state answer, which is a list of rumor. We now merge all messages into one processing pipeline, and try to process them.

### DIFF
--- a/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ProcessMessagesHandler.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ProcessMessagesHandler.scala
@@ -50,7 +50,7 @@ object ProcessMessagesHandler extends AskPatternConstants {
   }
 
   def rumorStateAnsHandler(messageRegistry: MessageRegistry)(implicit system: ActorSystem): Flow[GraphMessage, GraphMessage, NotUsed] = Flow[GraphMessage].map {
-    case msg @ Right(JsonRpcResponse(_, Some(resultObject), None, _)) =>
+    case msg @ Right(JsonRpcResponse(_, Some(resultObject), None, jsonId)) =>
       resultObject.resultRumor match
         case Some(rumorList) =>
           val mergedMsg = rumorList.flatMap(rumor => rumor.messages)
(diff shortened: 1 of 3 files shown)


be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ProcessMessagesHandler.scala
```diff
@@ -49,6 +49,29 @@ object ProcessMessagesHandler extends AskPatternConstants {
...
+  def rumorStateAnsHandler(messageRegistry: MessageRegistry)(implicit system: ActorSystem): Flow[GraphMessage, GraphMessage, NotUsed] = Flow[GraphMessage].map {
+    case msg @ Right(JsonRpcResponse(_, Some(resultObject), None, jsonId)) =>
+      resultObject.resultRumor match
+        case Some(rumorList) =>
+          val mergedMsg = rumorList.flatMap(rumor => rumor.messages)
+            .map((channel, msgList) => (channel, msgList.toSet))
+            .foldLeft(Map.empty[Channel, Set[Message]]) { case (acc, (key, values)) =>
+              acc.updated(key, acc.getOrElse(key, Set.empty[Message]) ++ values)
+            }
+          processMsgMap(mergedMsg, messageRegistry)
```
onsriahi14, line 61: maybe it's better to simplify it like that, it should lead to the same result 
val mergedMsg = rumorList
  .flatMap(_.messages)
  .groupBy(_._1)
  .mapValues(_.flatMap(_._2).toSet)

onsriahi14 (commented):
Looks good to me, just left a small comment

be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ProcessMessagesHandler.scala
DanielTavaresA: I updated with a small difference because mapValues is deprecated in scala3.

onsriahi14 (approved):

K1li4nL (approved):
LGTM !

PR #1905, opened 2024-06-03
[BE2] Add rumor state creation

DanielTavaresA:
Pending on #1900

Periodically, we create rumor state with the map of the last rumor id viewed for each public key known to the server


be2-scala/src/main/scala/ch/epfl/pop/decentralized/ConnectionMediator.scala
```diff
@@ -69,6 +70,7 @@ final case class ConnectionMediator(
     case ConnectionMediator.NewServerConnected(serverRef, greetServer) =>
       if (serverMap.isEmpty) {
         monitorRef ! Monitor.AtLeastOneServerConnected
+        gossipManagerRef ? Monitor.AtLeastOneServerConnected
       }
```
onsriahi14, line 74: same as above

be2-scala/src/main/scala/ch/epfl/pop/decentralized/GossipManager.scala
```diff
@@ -129,6 +148,27 @@ final case class GossipManager(dbActorRef: AskableActorRef, stopProbability: Dou
...
+          case DbActorGetRumorStateAck(rumorState) =>
+            serverRef ! ClientAnswer(
+              Right(JsonRpcRequest(
+                RpcValidator.JSON_RPC_VERSION,
+                rumor_state,
+                rumorState,
+                Some(jsonId)
+              ))
+            )
+            jsonId += 1
```
onsriahi14, line 166: the pattern matching is not exhaustive here

onsriahi14 (approved):

K1li4nL (approved):

PR #1909, opened 2024-06-04
[BE1] Fix a segmentation fault

arnauds5:
Add a null check to avoid a segmentation fault when an answer id is null.

SIGSEGV


emonnin-epfl:
Would you mind applying the fix on the be1-stuart-etienne-top-down-injection branch instead? Thank you!

MariemBaccari (approved):
LGTM 👍 I'll let you change the target branch according to @emonnin-epfl's comment.

PR #1914, opened 2024-06-06
[BE2] Integration of linter in the main code

t1b00:
PR makes the linter publishable, with modifications to the build.sbt of the linter and provides documentation in the README.md.
Small modification at the end of build.sbt of the main project (most edits were done by automatic cleanup), that adds custom rules 
Rules use a diag() function instead of a class, which would cause compilation errors in Scala 3


be2-scala/linter/project/plugins.sbt
```diff
@@ -1,3 +1,4 @@
 resolvers += Resolver.sonatypeRepo("releases")
-addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.12.0")
+addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.12.1")
 addSbtPlugin("com.eed3si9n" % "sbt-projectmatrix" % "0.9.2")
+addSbtPlugin("com.geirsson" % "sbt-ci-release" % "1.5.7")
```
K1li4nL, line 4: Why do we need that ?

be2-scala/linter/rules/src/main/scala/fix/Util.scala
```diff
@@ -31,18 +53,38 @@ object Util {
...
-  // Finds multiple definitions in one tree traversal
+  /**
+   * Finds multiple definitions in one tree traversal, returning a map of the definitions
+   * @param tree The tree to search in
+   * @param nameSet The set of names to search for
+   * @return A map of the definitions
+   */
   def findDefinitionsMap(tree: Tree, nameSet: Set[Term]): Map[Term, Any] = {
     findDefinitions(tree, nameSet).toMap
   }
```
K1li4nL, line 77: Should have seen it sooner but, do we need both of them ? especially since the the latter is just a `.toMap`

K1li4nL (commented):
Good job adding documentation ! Only minor comment / question

be2-scala/linter/project/plugins.sbt
t1b00: It's to be able to publish the rules to Maven using sbt publish

K1li4nL (dismissed):
lgtm

be2-scala/linter/rules/src/main/scala/fix/Util.scala
K1li4nL: I believe it should be fine given the one-liner it is today, I approved a bit fast so you are free to keep it or remove it.

t1b00: Removed it :)

K1li4nL (approved):

Issue #1915, opened 2024-06-06
Displaying Deterministic Username for a given Token


Kaz-ookid:
placeholder to link in the global report, will fill soon

PR #1918, opened 2024-06-11
Work fe1 ljankoschek poptoken mnemonic

ljankoschek:
Implementation of the fe-1 part for the deterministic username generator for a pop token to improve usability. In the roll call attendee list, and the user profile we display both the pop token and the username. In the chirps , the user list and the beneficiary selector we display only the user name now. Issue: #1915


4xiom5 (approved):
LGTM

K1li4nL (approved):
Blind approval

PR #1919, opened 2024-06-12
[BE1] better logging

sgueissa:
In this PR, I upgraded the log system with the following features:
1. Log for every sent/received message
2. Error log with stacktrace inside the Hub

image

In a next PR we should refine the error messages from the database


be1-go/internal/hub/hub.go
```diff
@@ -242,23 +246,35 @@ func (h *Hub) SendGreetServer(socket socket.Socket) error {
...
 		case incomingMessage := <-h.messageChan:
-			logger.Logger.Debug().Msg("start handling a message")
+			h.log.Debug().Msg("start handling a message")
 			err := h.jsonRpcHandler.Handle(incomingMessage.Socket, incomingMessage.Message)
 			if err != nil {
-				logger.Logger.Error().Err(err)
+				popError := &poperrors.PopError{}
+				if !errors.As(err, &popError) {
+					popError = poperrors.NewPopError(-6, err.Error())
+				}
```
arnauds5, line 263: Maybe changing the `-6` to `poperrors.InternalServerErrorCode` could be easier to remember what code it is for.

Using err.Error() directly as the format string could potentially result in a format string attack ?

sgueissa: done

K1li4nL (approved):
blind approval

PR #1920, opened 2024-06-12
[BE1] Add a simple buffer to rumor

sgueissa:
this PR add the first version of the buffer for the simple version of the rumors without the state inside the rumor.

It will be upgraded once the new rumor message is created.


be1-go/internal/handler/method/rumor/hrumor/buffer.go
```diff
@@ -0,0 +1,66 @@
...
+
+func (b *buffer) deleteWithDelay(ID string) {
+	defer b.Unlock()
+
+	timer := time.NewTimer(bufferEntryLifeTime)
+
+	<-timer.C
+
+	b.Lock()
+	delete(b.values, ID)
```
arnauds5, line 50: Having the defer unlock at the beginning works fine, but maybe having the (defer) unlock near/after the lock could be more readable.

sgueissa: fixed

K1li4nL (approved):
blind approval

PR #1928, opened 2024-06-13
[BE1] Increase usability of module with handlers

sgueissa:
In this PR, I changed the switch inside the modules query and message to allow future users to be able to add new methods and channels without having to directly change those modules.

Maybe we should return an error in each New() function if one of the argument is nil in a future PR.


arnauds5 (approved):
lgtm

emonnin-epfl (approved):
LGTM

PR #1929, opened 2024-06-13
[BE1] Add publish test

sgueissa:
In this PR, I added one unit test for the publish method.


be1-go/internal/handler/method/publish/hpublish/publish_test.go
```diff
@@ -0,0 +1,57 @@
...
+	hub.On("NotifyResetRumorSender").Return(nil).Once()
+
+	id, err := publishHandler.Handle(socket, publishBuf)
+	require.Nil(t, id)
+	require.NoError(t, err)
+
+	// succeed handling the publish but no notify rumor
+
+	messageHandler.On("Handle", channelPath, msg, false).Return(nil).Once()
+	db.On("AddMessageToMyRumor", msg.MessageID).Return(thresholdMessagesByRumor-1, nil).Once()
```
emonnin-epfl, line 41: By default, doesn't Mockery expect only one call?

sgueissa: I got some problem using mockery without the .Once() so now I'm always using it

PR #1933, opened 2024-06-14
fix embedme not working

DanielTavaresA:
Characters " \` " were escaped. But for some reason, matching regex didn't trigger. When those characters are not escaped, matching regex for `json` code is triggered.


K1li4nL (approved):
LGTM

PR #1936, opened 2024-06-14
Data exchange protocol message definition

onsriahi14:
After a successful establishment of the federation, now participants can start sharing data between each other. For now, data exchange concerns the social media channel.
In this PR, we define the json schema for the data exchange message and we update the documentation accordingly.


docs/messageData.md
```diff
@@ -3311,4 +3312,77 @@ This message is sent by Bob's server. The purpose of this message is to provide
   ]
 }
 ```
+## Exchanging tokens to share data between LAOs (federation#tokens_exchange)
+For now, data exchange is only implemented for social media channel.
 
+This message is sent by both organizers to their servers to be broadcast. The message sent from Alice's server to Bob's server contains the pop tokens of Alice's LAO and vice versa. Upon receiving this message, participants will be able to subscribe to each channel /root/lao_id/social/token.
```
arnauds5, line 3318: Since participants of one LAO only have trust to their respective organizer, isn't this message only needed to be sent by alice to alice server and broadcast to all participants of alice LAO, (and the same on bob side) ?
If that is the case, then the message shouldn't be send from Alice server to Bob server ?

quadcopterman (approved):

arnauds5 (approved):

quadcopterman (approved):

matteosz (approved):

PR #1938, opened 2024-06-14
[BE1] Upgrade gossip rumor with vector clock

sgueissa:
In this PR we did:
1. Rumor Push protocol with vector clock
2. Buffer inside Rumor handler to handle cases where a newer rumor is arriving few second before an older rumor
3. Rumor Pull protocol with vector clock
4. Fix order of the answer to catchup query


emonnin-epfl (approved):
LVGTM!

sgueissa:
> LVGTM!

YOU TOO!!!

K1li4nL (approved):
Blind approval

PR #1943, opened 2024-06-23
Federation Data Exchange for FE2

quadcopterman:
Here is the implementation of the data exchange protocol for FE2 Android. The data exchange protocol is defined in issue #1802

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/federation/FederationResult.kt
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/federation/FederationResult.kt
@@ -25,18 +25,7 @@ class FederationResult
 ) : Data {
 
   init {
-    when (status) {
-      FAILURE -> {
-        require(reason != null) { "Reason must be provided for $FAILURE status." }
-        require(publicKey == null) { "Public key must be null for $FAILURE status." }
-      }
-      SUCCESS -> {
-        require(publicKey != null) { "Public key must be provided for $SUCCESS status." }
-        require(reason == null) { "Reason must be null for $SUCCESS status." }
-      }
-      else -> throw IllegalArgumentException("Status must be either '$FAILURE' or '$SUCCESS'.")
-    }
-    MessageValidator.verify().validMessage(challenge)
+    MessageValidator.verify().isValidFederationResult(status, reason, publicKey, challenge)
   }
 
   override val `object`: String
@@ -78,7 +67,7 @@ class FederationResult
   }
 
   companion object {
-    private const val SUCCESS = "success"
-    private const val FAILURE = "failure"
+    const val SUCCESS = "success"
+    const val FAILURE = "failure"
   }
 }
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/federation/FederationResult.kt
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/federation/FederationResult.kt
@@ -62,9 +62,9 @@ class FederationResult
 
   override fun toString(): String {
     if (status == "failure") {
-      return "FederationResult{status='$status', reason='$reason'," + "challenge='$challenge'}"
+      return "FederationResult{status='$status', reason='$reason', challenge='$challenge'}"
     } else if (status == "success") {
-      return "FederationResult{status='$status', public_key='$publicKey'," +
+      return "FederationResult{status='$status', public_key='$publicKey', " +
           "challenge='$challenge'}"
     }
     return "FederationResult{ERROR}"
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/federation/FederationResult.kt
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/federation/FederationResult.kt
@@ -1,0 +1,76 @@
+package com.github.dedis.popstellar.model.network.method.message.data.federation
+
+import com.github.dedis.popstellar.model.network.method.message.data.Action
+import com.github.dedis.popstellar.model.network.method.message.data.Data
+import com.github.dedis.popstellar.model.network.method.message.data.Objects
+import com.google.gson.annotations.SerializedName
+
+/** Informs about the result of the authentication procedure */
+class FederationResult
+/**
+ * Constructor for a data Federation Result
+ *
+ * @param status status of the result (either success or failure)
+ * @param reason reason of the failure
+ * @param publicKey public key of the other LAO organizer
+ * @param challenge challenge used to connect the LAOs
+ */
+(
+    val status: String,
+    val reason: String? = null,
+    @SerializedName("public_key") val publicKey: String? = null,
+    val challenge: Challenge,
+) : Data {
+
+  init {
+    when (status) {
+      "failure" -> {
+        require(reason != null) { "Reason must be provided for failure status." }
+        require(publicKey == null) { "Public key must be null for failure status." }
+      }
+      "success" -> {
+        require(publicKey != null) { "Public key must be provided for success status." }
+        require(reason == null) { "Reason must be null for success status." }
+      }
+      else -> throw IllegalArgumentException("Status must be either 'failure' or 'success'.")
+    }
+  }
+
+  override val `object`: String
+    get() = Objects.FEDERATION.`object`
+
+  override val action: String
+    get() = Action.RESULT.action
+
+  override fun equals(other: Any?): Boolean {
+    if (this === other) {
+      return true
+    }
+    if (other == null || javaClass != other.javaClass) {
+      return false
+    }
+    val that = other as FederationResult
+    return status == that.status &&
+        reason == that.reason &&
+        publicKey == that.publicKey &&
+        challenge == that.challenge
+  }
+
+  override fun hashCode(): Int {
+    return java.util.Objects.hash(status, reason, publicKey, challenge)
+  }
+
+  override fun toString(): String {
+    if (status == "failure") {
+      return "FederationResult{status='$status', reason='$reason'," + "challenge='$challenge'}"
+    } else if (status == "success") {
+      return "FederationResult{status='$status', public_key='$publicKey'," +
+          "challenge='$challenge'}"
+    }
+    return "FederationResult{ERROR}"
+  }
+
+  fun isSuccess(): Boolean {
+    return status == "success"
+  }
+}
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/federation/FederationResult.kt
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/federation/FederationResult.kt
@@ -4,6 +4,7 @@ import com.github.dedis.popstellar.model.network.method.message.MessageGeneral
 import com.github.dedis.popstellar.model.network.method.message.data.Action
 import com.github.dedis.popstellar.model.network.method.message.data.Data
 import com.github.dedis.popstellar.model.network.method.message.data.Objects
+import com.github.dedis.popstellar.utility.MessageValidator
 import com.google.gson.annotations.SerializedName
 
 /** Informs about the result of the authentication procedure */
@@ -35,6 +36,7 @@ class FederationResult
       }
       else -> throw IllegalArgumentException("Status must be either '$FAILURE' or '$SUCCESS'.")
     }
+    MessageValidator.verify().validMessage(challenge)
   }
 
   override val `object`: String
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/federation/FederationResult.kt
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/federation/FederationResult.kt
@@ -1,5 +1,6 @@
 package com.github.dedis.popstellar.model.network.method.message.data.federation
 
+import com.github.dedis.popstellar.model.network.method.message.MessageGeneral
 import com.github.dedis.popstellar.model.network.method.message.data.Action
 import com.github.dedis.popstellar.model.network.method.message.data.Data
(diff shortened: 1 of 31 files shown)


fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/federation/FederationResult.kt
```diff
@@ -0,0 +1,77 @@
...
+    val challenge: MessageGeneral,
+) : Data {
+
+  init {
+    when (status) {
+      "failure" -> {
+        require(reason != null) { "Reason must be provided for failure status." }
+        require(publicKey == null) { "Public key must be null for failure status." }
+      }
+      "success" -> {
```
matteosz, line 32: Perhaps it would be better to define these ("failure" and "success") as static constants, so in case we want to change them in future it's just 1 place to modify

fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/LinkedOrganizationsRepository.kt
```diff
@@ -35,6 +34,36 @@ class LinkedOrganizationsRepository @Inject constructor() {
...
+  }
+
+  fun updateAndNotifyLinkedLao(
+      laoId: String,
+      otherLaoId: String,
+      tokens: Array<String>,
+      rollCallId: String
+  ) {
+    addLinkedLao(laoId, otherLaoId, tokens)
+    newTokensNotifyFunction?.invoke(laoId, otherLaoId, rollCallId, tokens)
```
matteosz, line 50: If you take a look at other repos (any one, e.g. rollcallrepo) you'd see that in general this is not the preferred way to notify modifications on a given data structure, but that we use java rx constructs (subject, observables and so on)

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/federation/LinkedOrganizationsFragment.kt
```diff
@@ -59,6 +59,19 @@ class LinkedOrganizationsFragment : Fragment() {
...
+    linkedOrganizationsViewModel.doWhenLinkedLaosIsUpdated { laoId, laoMap ->
+      if (laoId == laoViewModel.laoId) {
+        CoroutineScope(Dispatchers.Main).launch {
+          val currentLaos = laoMap.keys
+          displayLinkedOrganizations(currentLaos)
+        }
+      }
+    }
+
+    linkedOrganizationsViewModel.setLinkedLaosNotifyFunction()
```
matteosz, line 74: Why calling display first with the current list and then putting a custom and complex observer that uses coroutines and not directly using a simple observer? This relates to my previous comment on how to deal with modifications using observables and subjects

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/socialmedia/SocialMediaViewModel.kt
```diff
@@ -225,36 +227,50 @@ constructor(
...
+                .sorted(
+                    Comparator.comparing { chirp: Chirp? ->
+                      if (chirp != null) -chirp.timestamp else 0
+                    })
+                .collect(Collectors.toList())
+          }
+      // We want to observe these changes on the main thread such that any modification done
+      // to the view are done on the thread. Otherwise, the app might crash
+      return combinedObservables.observeOn(schedulerProvider.mainThread())
+    }
```
matteosz, line 270: This getter will be executed every time we access that field, which looks very heavy as getter method. You could precompute this variable in initialization, such that this function is only called once

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/data/LinkedOrganizationsHandler.kt
```diff
@@ -24,4 +31,84 @@ constructor(
...
+   */
+  @Throws(UnknownLaoException::class)
+  fun handleResult(context: HandlerContext, result: FederationResult) {
+    if (result.isSuccess()) {
+      if (result.challenge.data == linkedOrgRepo.getChallenge() &&
+          linkedOrgRepo.otherLaoId != null) {
+        val laoId = context.channel.extractLaoId()
+        linkedOrgRepo.addLinkedLao(laoId, linkedOrgRepo.otherLaoId!!, arrayOf())
+        laoRepo.addDisposable(
+            context.messageSender
```
matteosz, line 49: Maybe for better separability it would be nice to have an ad-hoc disposable in the linked org repo, without using the lao one

fe2-android/app/src/test/ui/robolectric/com/github/dedis/popstellar/ui/lao/federation/LinkedOrganizationsFragmentOrganizerTest.kt
matteosz, line 1: Could you add a test when also there are some organizations (so not just the empty case)?

matteosz (changes requested):
Great work!

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/federation/FederationResult.kt
quadcopterman: Ok, I'll change that

fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/LinkedOrganizationsRepository.kt
quadcopterman: Well the thing is that I tried using observables and for some reason I had many problems with the UI not updating or weird other issues... That's why in the end I changed to using callbacks, I know it's a bit more complex but at least it's working

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/data/LinkedOrganizationsHandler.kt
quadcopterman: Yes, I'll do that

fe2-android/app/src/main/java/com/github/dedis/popstellar/repository/LinkedOrganizationsRepository.kt
matteosz: Sure, we can keep it as is then

fe2-android/app/src/main/java/com/github/dedis/popstellar/ui/lao/socialmedia/SocialMediaViewModel.kt
matteosz: Yeah sorry I was missing the context of execution of that function, now I looked it up better and it's how it's supposed to work, even if it's not very efficient and perhaps in the future we could change it

Kaz-ookid (changes requested):
Matteo's comments cover all the important parts but I have one last request:
FederationResult and TokensExchange are new message data, that require validation of inputs.

You should add the validation using MessageValidation.validate() to avoid having technical debt, as total validation is a goal documented in #1751. You should also update this Issue and add the new data in the expandable columns :)

You can check in #1952 how to use validation, and how to create simple tests to check the constructor.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/federation/FederationResult.kt
```diff
@@ -0,0 +1,84 @@
...
+ */
+(
+    val status: String,
+    val reason: String? = null,
+    @SerializedName("public_key") val publicKey: String? = null,
+    val challenge: MessageGeneral,
+) : Data {
+
+  init {
+    when (status) {
```
Kaz-ookid, line 28: This looks like input validation, maybe you should complete MessageValidator class to check a federation status? Validation would then just be a clear oneliner 👍

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/federation/TokensExchange.kt
```diff
@@ -0,0 +1,59 @@
...
+ */
+(
+    @SerializedName("lao_id") val laoId: String,
+    @SerializedName("roll_call_id") val rollCallId: String,
+    val tokens: Array<String>,
+    val timestamp: Long
+) : Data {
+
+  init {
+    MessageValidator.verify().isBase64(rollCallId, "rollCallId")
```
Kaz-ookid, line 27: isBase64 (as every other functions in MessageValidator) returns the builder so you can write : 

`MessageValidator.verify().isBase64(rollCallId, "rollCallId").validPastTimes(timestamp)`

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/data/LinkedOrganizationsHandler.kt
```diff
@@ -76,15 +77,15 @@ constructor(
     // LAO. This might be changed in the future (making a pop-up asking the user if he/she wants
     // to subscribe to that)
     tokenExchange.tokens.forEach { t ->
-      laoRepo.addDisposable(
+      disposables.add(
```
matteosz: Using the disposable from the handler doesn't give you the opportunity to clear the added disposables when closing a lao or the application. Take a look at the usage of disposables in the RollCallRepository or any other repository.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/federation/TokensExchange.kt
Kaz-ookid: given the documentation of the protocol :

<details>
<summary><code>consensus</code></summary>

```
// ../protocol/query/method/message/data/dataFederationTokensExchange.json
{
  "$id": "https://raw.githubusercontent.com/dedis/popstellar/master/protocol/query/method/message/data/dataFederationTokensExchange.json",
  "description": "Sent by an organizer client to its server, to broadcast the Pop tokens",
  "type": "object",
  "properties": {
    "object": {
      "const": "federation"
    },
    "action": {
      "const": "tokens_exchange"
    },
    "lao_id": {
      "type": "string",
      "contentEncoding": "base64",
      "$comment": "Hash : HashLen(organizer, creation, name)"
    },
    "roll_call_id": {
      "type": "string",
      "contentEncoding": "base64",
      "$comment": "last roll call id"
    },
    "tokens": {
      "description": "[Array[Base64String]] list of Pop tokens",
      "type": "array",
      "uniqueItems": true,
      "items": {
        "type": "string",
        "contentEncoding": "base64"
      },
      "$comment": "List must be sorted according to byte encoding: -,0...9,A...Z,_,a...z"
    },
    "timestamp": {
      "type": "integer",
      "description": "[Timestamp] of the tokens' exchange",
      "minimum": 0
    }
  },
  "required": [
    "object",
    "action",
    "lao_id",
    "roll_call_id",
    "tokens",
    "timestamp"
  ]
}
```


</details>

- It looks like that laoId should be checked for `notEmptybase64`.
- I am not sure if an empty rollCallId makes sense, if not then switch to `notEmptybase64 `as well.
- Tokens look like they individually need to be checked as well (they should be base64, maybe we wan't them not empty as well?). Also they should be checked for uniqueness. Is a null or empty array acceptable? This maybe needs to be defined.

fe2-android/app/src/test/unit/java/com/github/dedis/popstellar/model/network/method/message/data/federation/FederationResultTest.kt
```diff
@@ -0,0 +1,137 @@
...
+        }
+        assert(exception.message == "Public key must be provided for success status.")
+    }
+
+    @Test
+    fun invalidSuccessTest2() {
+        val exception = assertThrows(IllegalArgumentException::class.java) {
+            FederationResult(SUCCESS, publicKey = PK.encoded, reason = "reason", challenge = MG_CHALLENGE)
+        }
+        assert(exception.message == "Reason must be null for success status.")
```
Kaz-ookid, line 105: maybe we want to add those strings in Constants to avoid failing if the message comes to change

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/data/LinkedOrganizationsHandler.kt
quadcopterman: Oh ok, then should I create the disposables in LinkedOrganizationsRepository (as done in RollCallRepository) and then call it from the LinkedOrganizationsHandler ? Or should I do it directly in the handler ?

fe2-android/app/src/test/unit/java/com/github/dedis/popstellar/model/network/method/message/data/federation/TokensExchangeTest.kt
```diff
@@ -0,0 +1,77 @@
...
+
+import com.github.dedis.popstellar.model.network.method.message.data.Action
+import com.github.dedis.popstellar.model.network.method.message.data.Objects
+import com.github.dedis.popstellar.model.objects.Lao
+import com.github.dedis.popstellar.testutils.Base64DataUtils
+import org.junit.Assert
+import org.junit.Test
+import java.time.Instant
+
+class TokensExchangeTest {
```
Kaz-ookid, line 11: don't forget to add constructors succeed and failing tests as in the other class

Kaz-ookid (changes requested):
Great! Just a few last steps in message validation 👍

fe2-android/app/src/main/java/com/github/dedis/popstellar/utility/handler/data/LinkedOrganizationsHandler.kt
matteosz: Yep you can do it from the repository as for roll calls and keep it consistent

matteosz (approved):

PR #1945, opened 2024-06-25
[BE2] Added vector clock 

DanielTavaresA:
Now, rumors should be able to be sorted based on their timestamp which is a vector clock. With this PR we add the possibility to sort on a RumorState and process rumors in sorted order

### DIFF
--- a/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ProcessMessagesHandler.scala
+++ b/be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ProcessMessagesHandler.scala
@@ -56,7 +56,6 @@ object ProcessMessagesHandler extends AskPatternConstants {
       resultObject.resultRumor match
         case Some(rumorList) =>
           val orderedRumors = rumorList.sortBy(_.timestamp)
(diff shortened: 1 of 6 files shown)


be2-scala/src/main/scala/ch/epfl/pop/pubsub/graph/handlers/ProcessMessagesHandler.scala
```diff
@@ -55,14 +55,23 @@ object ProcessMessagesHandler extends AskPatternConstants {
...
       resultObject.resultRumor match
         case Some(rumorList) =>
-          val mergedMsg = rumorList
-            .flatMap(_.messages)
-            .groupBy(_._1)
-            .view.mapValues(_.flatMap(_._2).toSet).toMap
-          if (processMsgMap(mergedMsg, messageRegistry) && writeRumorsInDb(dbActorRef, rumorList))
-            msg
+          val orderedRumors = rumorList.sortBy(_.timestamp)
+          println(s"orderedRumors $orderedRumors")
```
K1li4nL, line 59: Might clog a bit the logs ?

K1li4nL (approved):
LGTM

PR #1946, opened 2024-06-25
[BE1] Federation Data Exchange

arnauds5:
- Add federation tokens exchange handler
- Avoid duplicate websocket connections  in federation of multiple LAOs within one server
- Add the server socket to the list of server sockets and use NewServerSocket instead of NewClientSocket
- Fix missing federation result message if the same server is used
- handle challenge msg of a federation init directly instead of using a websocket if only one server


be1-go/internal/state/hub_params.go
```diff
@@ -17,7 +17,7 @@ type HubParams struct {
 
 func NewHubParams(log zerolog.Logger) *HubParams {
 	return &HubParams{
-		messageChan:      make(chan socket.IncomingMessage),
+		messageChan:      make(chan socket.IncomingMessage, 100),
```
sgueissa, line 20: If you want to buffer a message because the arriving order is random then you can do it with a temporary buffer that flush after x time like we did for the rumors directly inside the federation package. And you should not use buffered channel like that because it will just fill as we are handling messages sequentialy.

sgueissa (changes requested):

be1-go/internal/state/hub_params.go
sgueissa: And imagine that 100 messages arrive on the server while you are handling your message then you will try to send your new message on the incoming message channel but it will result to the server stalling forever.

sgueissa (approved):

sgueissa:
LGTM

K1li4nL (approved):
Blind approval

PR #1948, opened 2024-06-26
[BE2] Clean Log

DanielTavaresA:
Previously logs were crowded with unintuitive messages and inconsistent over all code. Now logs should be simpler, more readable and contain only useful information


K1li4nL (approved):
Nicely done !

PR #1958, opened 2024-06-27
Send rumors state when connecting federaction

arnauds5:
- Send rumors state when connecting federaction
- Send rumors state when receiving a connection from federation
- Send greet server when connecting federaction


be1-go/internal/handler/channel/federation/hfederation/federation.go
```diff
@@ -366,19 +378,28 @@ func (h *Handler) handleChallenge(msg mmessage.Message, channelPath string,
...
+			federationExpect.LaoId)
 	} else {
+		// Add the socket to the list of server sockets
+		h.sockets.Upsert(socket)
+
+		// Send the rumor state directly to avoid delay while syncing
+		err = h.rumors.SendRumorStateTo(socket)
+		if err != nil {
+			return err
+		}
```
sgueissa, line 392: When a server connect to an other server, the receiving server already notify the hub of a new socket and will already upsert the socket and send the rumor state message so you don't need to do it here

sgueissa (changes requested):

be1-go/internal/handler/channel/federation/hfederation/federation.go
sgueissa: as for now, the federation protocol is using the client socket, therefore we need to do that

sgueissa (approved):

sgueissa:
LGTM

K1li4nL (approved):
Blind approval

PR #1959, opened 2024-06-27
Message validation for the rest of Federation messages

quadcopterman:
Added message validation on all federation messages that hadn't validation before

### DIFF
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/federation/Challenge.kt
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/federation/Challenge.kt
@@ -3,6 +3,7 @@ package com.github.dedis.popstellar.model.network.method.message.data.federation
 import com.github.dedis.popstellar.model.network.method.message.data.Action
 import com.github.dedis.popstellar.model.network.method.message.data.Data
 import com.github.dedis.popstellar.model.network.method.message.data.Objects
+import com.github.dedis.popstellar.utility.MessageValidator
 import com.google.gson.annotations.SerializedName
 import kotlin.math.max
 
@@ -19,6 +20,7 @@ class Challenge
 
   init {
     this.validUntil = max(0L, validUntil)
+    MessageValidator.verify().isNotEmptyBase64(value, "challenge").validFutureTimes(this.validUntil)
   }
 
   override val `object`: String
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/federation/Challenge.kt
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/federation/Challenge.kt
@@ -19,8 +19,8 @@ class Challenge
   @SerializedName("valid_until") val validUntil: Long
 
   init {
+    MessageValidator.verify().stringNotEmpty(value, "challenge").validFutureTimes(validUntil)
     this.validUntil = max(0L, validUntil)
-    MessageValidator.verify().isNotEmptyBase64(value, "challenge").validFutureTimes(this.validUntil)
   }
 
   override val `object`: String
--- a/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/federation/ChallengeRequest.kt
+++ b/fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/federation/ChallengeRequest.kt
@@ -4,6 +4,7 @@ import com.github.dedis.popstellar.model.Immutable
 import com.github.dedis.popstellar.model.network.method.message.data.Action
 import com.github.dedis.popstellar.model.network.method.message.data.Data
 import com.github.dedis.popstellar.model.network.method.message.data.Objects
+import com.github.dedis.popstellar.utility.MessageValidator
 
 /** Data sent to get a challenge */
 @Immutable
@@ -14,6 +15,10 @@ class ChallengeRequest
  * @param timestamp time of the Challenge Request
  */
 (val timestamp: Long) : Data {
+
+  init {
+    MessageValidator.verify().validPastTimes(timestamp)
+  }
 
   override val `object`: String
     get() = Objects.FEDERATION.`object`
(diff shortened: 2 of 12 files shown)


fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/federation/Challenge.kt
```diff
@@ -19,6 +20,7 @@ class Challenge
 
   init {
     this.validUntil = max(0L, validUntil)
```
Kaz-ookid, line 22: validation should ideally be before instantiating attributes, since if the input is wrong and it is meant to crash, no need to fill the attribute beforehand.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/federation/Challenge.kt
```diff
@@ -19,6 +20,7 @@ class Challenge
...
+    MessageValidator.verify().isNotEmptyBase64(value, "challenge").validFutureTimes(this.validUntil)
```
Kaz-ookid, line 23: By checking the protocol, it seems that the `value ` argument has some guidelines that should probably be checked (the regex pattern and encoding, the 32 byte length). Also the encoding seems to be Hex and not base64, I am not sure it this changes something about it.

<details>
<summary> json protocol </summary>

```
// ../protocol/query/method/message/data/dataFederationChallenge.json

{
    "$id": "https://raw.githubusercontent.com/dedis/popstellar/master/protocol/query/method/message/data/dataFederationChallenge.json",
    "description": "Challenge object in the context of federation authentication",
    "type": "object",
    "properties": {
        "object": {
            "const": "federation"
        },
        "action": {
            "const": "challenge"
        },
        "value": {
            "type": "string",
            "contentEncoding": "hex",
            "pattern": "^[0-9a-fA-F]{64}$",
            "$comment": "A 32 bytes array encoded in hexadecimal"
        },
        "valid_until": {
            "type": "integer",
            "description": "[Timestamp] of the expiration time",
            "minimum": 0
        }
    },
    "required": [
        "object",
        "action",
        "value",
        "valid_until"
    ]
}
```


</details


Also, why is `this.value` never assigned? I might just not have catch the thing

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/federation/ChallengeRequest.kt
```diff
@@ -15,6 +16,10 @@ class ChallengeRequest
  */
 (val timestamp: Long) : Data {
 
+  init {
+    MessageValidator.verify().validPastTimes(timestamp)
```
Kaz-ookid, line 20: given the protocol : 

```
        "timestamp": {
            "type": "integer",
            "description": "[Timestamp] of the request",
            "minimum": 0
        }
```

timestamp should be >= 0, however `validPastTime `does not check this. I think that this check should be preceded by `greaterOrEqualThan`

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/federation/FederationExpect.kt
```diff
@@ -23,6 +24,14 @@ class FederationExpect
     val challenge: MessageGeneral
 ) : Data {
 
+  init {
+    MessageValidator.verify()
+        .isNotEmptyBase64(laoId, "lao_id")
+        .stringNotEmpty(serverAddress, "server_address")
```
Kaz-ookid, line 30: I am wondering if it would be a good idea to create a function in MessageValidator to check for server addresses pattern. It is specified in the protocol :

```
"server_address": {
            "type": "string",
            "pattern": "^(ws|wss):\/\/.*(:\\d{0,5})?\/.*$",
            "$comment": "public address of the remote organizer server"
        },
```

Kaz-ookid (changes requested):
Great job !

I have multiple comments : 
- some picky comments in the review. Good and complete validation enhances security, but also robustness of the code : others tests using some federation data will not be able to instantiate wrong data, therefore avoiding any case of uncatched bugs and issues. From my experience, as I was adding validations a lot of tests aside were failing due to bad data instantiation. This showed how important it is to give clear and strict guidelines.
- you should edit the Issue #1751 with the new classes and check the classes done :)
- I feel like there is a lack of tests. There should be one tests for every construction scenario, checking if the constructor fails or succeeds accordingly. A great example can be found [here](https://github.com/dedis/popstellar/blob/6c8076d96dd0316796aa6b52939754099429ab25/fe2-android/app/src/test/unit/java/com/github/dedis/popstellar/model/network/method/message/data/rollcall/CloseRollCallTest.kt#L90). I know it looks big but it really is quick, and enforces the well definitions of the validations.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/federation/FederationExpect.kt
quadcopterman: For that point, the pattern might change as (if I remember well) some addresses didn't fit in the pattern. It will be added once the protocol is updated.

fe2-android/app/src/main/java/com/github/dedis/popstellar/model/network/method/message/data/federation/Challenge.kt
quadcopterman: Ooh true, good catch ! I thought it was base64. However the challenges are not always hex, I need to check that with the team. For now, I'll check that it is not empty, otherwise it would break everything. Oh and this.value is set when constructing the challenge

Kaz-ookid (approved):
Looks good to me. 

Maybe you should just write self-explanatory TODOs for the value validation (encoding, regex), and for the server address. The thing is, by the time these points are clearer in the protocol, nobody will still be there to remember that this was to do to finalize a clean validation of the data.

I think very clear TODOs with your name + date would be a good addition 👍

quadcopterman:
Yes it's a good idea to write that somewhere. However we might forget a TODO in the code (nobody will see it unless working on these specific lines of the project), so I think it's better if I mention that in issue #1751 so we can see what still needs validation more globally :)

matteosz (approved):
LGTM

PR #1965, opened 2024-06-28
added commands to be able to run new server config from files

DanielTavaresA:
Added parameters to be able to run multiples configs of the server by providing files from cmd


DanielTavaresA:
After discussion with @K1li4nL . Found out there is already a way to do it. Closing

PR #1969, opened 2024-06-29
fix sqlite transactions

arnauds5:
- Add a missing defer tx.Rollback
- Change some s.database.Exec(...) into tx.Exec(...) if it there is a transaction
- Change unprocessed message sql query from "insert ..." to "insert or ignore ..." so that duplicate message doesn't block all the others messages to be handled


emonnin-epfl (approved):
LGTM

K1li4nL (approved):

PR #1970, opened 2024-06-30
Add some delay before servers are sync

arnauds5:
When a federation challenge is successfully handled in go, a federation result is broadcast and the organizers frontends will try to subscribe/catchup to the other lao channel, but if the server isn't synced yet, then they will failed.
A goroutine is started to avoid blocking the messages handlers (single thread), and will wait until the remote channel is available locally.


sgueissa (approved):
LGTM

MariemBaccari (approved):

PR #1974, opened 2024-06-30
Revert "Add some delay before servers are sync"

arnauds5:
Reverts dedis/popstellar#1970

delay on backend is not reliable enough, it should maybe be done on frontends instead


sgueissa (approved):
LGTM

K1li4nL (approved):
Blind approval

